Issue #3, opened 2021-12-23
Service with no operations gen


lewisjkl:
When a service has no operations in it, the generated code has a compilation error. For example:

```smithy
namespace test

use smithy4s.api#simpleRestJson

@simpleRestJson
service MyService {
  version: "1",
  operations: []
}
```

would generate:

```scala
sealed trait MyServiceOperation[Input, Err, Output, StreamedInput, StreamedOutput]

def asTransformationGen[P[_, _, _, _, _]](impl : MyServiceGen[P]): smithy4s.Transformation[MyServiceOperation, P] = new smithy4s.Transformation[MyServiceOperation, P] {
    def apply[I, E, O, SI, SO](op : MyServiceOperation[I, E, O, SI, SO]) : P[I, E, O, SI, SO] = op match  { // error empty match
    }
  }

def endpoint[I, E, O, SI, SO](op : MyServiceOperation[I, E, O, SI, SO]) = op match { // error empty match
  }
```

Of course, why would one want to generate code for a service with no operations? They probably wouldn't.. but maybe I am not thinking of other use cases.
 
I am thinking we should make it so that we automatically do not generate any code for services with no operations.

For those wondering why one would even have a service with no operations in the first place, it can be helpful for if you have a prior version of a service that you want to represent/annotate with traits in smithy without documenting it fully.

kubukoz:
Sounds like this could be a validator.

lewisjkl:
> Sounds like this could be a validator.

You mean a validator that will prevent services from having an empty list of operations?

kubukoz:
Yeah, seems useless :) it should be part of the default smithy validators, really (just like you can't have an enum without values)

Baccata:
Actually internally we have a use-case for empty services. 

In any case, it's gonna be an easy fix. I'll get it before 0.9.1

PR #5, opened 2021-12-23
Make generated code from empty services compile

Baccata:
Fixes https://github.com/disneystreaming/smithy4s/issues/3

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/Renderer.scala
@@ -45,8 +45,7 @@ object Renderer {
       val p = s"package ${unit.namespace}"
 
       val allImports =
-        (renderResult.imports + "smithy4s.syntax._")
-          .filterNot(_.startsWith(unit.namespace))
+        renderResult.imports.filterNot(_.startsWith(unit.namespace))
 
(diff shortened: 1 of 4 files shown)


modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -165,14 +164,20 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
...
-          s"def endpoint[I, E, O, SI, SO](op : $opTraitName[I, E, O, SI, SO]) = op match"
-        ) {
-          ops.map {
-            case op if op.input != Type.unit =>
-              s"case ${op.name}(input) => (input, ${op.name})"
-            case op =>
-              s"case ${op.name}() => ((), ${op.name})"
+        if (ops.isEmpty) {
+          line(
+            s"""def endpoint[I, E, O, SI, SO](op : $opTraitName[I, E, O, SI, SO]) = sys.error("impossible")"""
```
Baccata, line 169: Because the sealed trait is unpopulated, it is effectively a phantom type, therefore it is literally impossible under normal circumstances (unless runtime reflection is involved, or black magic) for this method to be called.

kubukoz (approved):

PR #12, opened 2021-12-24
ci(Mergify): configuration update

kubukoz:
This change has been made by @kubukoz from https://mergify.com config editor.


kubukoz:
@Baccata can we merge this for now, and do something smarter like what you described in #13 later?

Baccata (approved):

PR #17, opened 2021-12-27
[POC] Adds a "dynamic" module to dynamically load services from smithy models

Baccata:
This relies on a one-pass + suspension to build a map of schemas that work against dynamic data structures, backing product types with `Array[Any]`, and union-types with `(Int, Any)`

The goal is notably to get to a state where services could be proxied dynamically, whether for actual proxyfication purposes, or to create CLIs that do not need recompilation, or create UIs/playgrounds dynamically.


modules/dynamic/src/smithy4s/dynamic/DynamicEndpoint.scala
```diff
@@ -0,0 +1,21 @@
+package smithy4s
+package dynamic
+
+case class DynamicEndpoint(
+    namespace: String,
+    name: String,
+    input: Schema[DynData],
```
Baccata, line 7: need to add the `Input` / `Output` hint there.

build.sbt
```diff
@@ -391,6 +391,27 @@ lazy val protocol = projectMatrix
...
+ * Model instances.
+ */
+lazy val dynamic = projectMatrix
+  .in(file("modules/dynamic"))
+  .dependsOn(core)
+  .settings(
+    isCE3 := true,
+    libraryDependencies ++= Seq(
+      Dependencies.Smithy.model,
+      "org.scala-lang.modules" %% "scala-collection-compat" % "2.6.0",
```
kubukoz, line 406: Should this be %%%?

modules/dynamic/src/smithy4s/dynamic/DynamicModelCompiler.scala
```diff
@@ -0,0 +1,171 @@
...
+        shape.getId.getName(),
+        maybeSchema(shape.getInput()),
+        maybeSchema(shape.getOutput()),
+        Hints(allHints(shape): _*)
+      )
+      endpointMap += shape.getId() -> ep
+    }
+
+    override def serviceShape(shape: ServiceShape): Unit = {
+      val getEndpoints = () =>
```
kubukoz, line 120: Soooo why is this lazy?

build.sbt
Baccata: Possibly

modules/dynamic/src/smithy4s/dynamic/DynamicModelCompiler.scala
Baccata: For simplicity of implementation : we traverse the shapes only once. When we encounter a service, it's endpoints might not have been created yet.

That being said I could force evaluation at the end of the traversal, but in practice it shouldn't make a difference since the interpreters often precompile stuff on their end

kubukoz: fair nuff

Baccata:
Don't trust the CI, the new module is not currently wired (nor do I intend to do it right away)

kubukoz:
I can now confirm this works on JS :)

modules/dynamic/src/smithy4s/dynamic/DynamicEndpoint.scala
kubukoz: Added in https://github.com/disneystreaming/smithy4s/pull/17/commits/41e5d1fea6e7fc825a397828578c0cc727d8ca28

kubukoz:
@baccata, [`eeef6b8` (#17)](https://github.com/disneystreaming/smithy4s/pull/17/commits/eeef6b8760d4d458c8f36604c38b8f7f3b98362d) adds a failing test for something I've experienced when trying to compile dynamic schemas in the other CLI module. The stack trace in this test is as follows:

```
[E]   NoSuchElementException: key not found: smithy.api#String
[E] 
[E]   Map.scala:274                           scala.collection.MapOps#default
[E]   Map.scala:273                           scala.collection.MapOps#default$
[E]   Map.scala:405                           scala.collection.AbstractMap#default
[E]   HashMap.scala:425                       scala.collection.mutable.HashMap#apply
[E]   DynamicModelCompiler.scala:82           smithy4s.dynamic.Compiler$CompileVisitor#schema
[E]   DynamicModelCompiler.scala:180          smithy4s.dynamic.Compiler$CompileVisitor#$anonfun$structureShape$3
[E]   suspended.scala:32                      schematic.suspended$Schema#deferredSchema$lzycompute
[E]   suspended.scala:32                      schematic.suspended$Schema#deferredSchema
[E]   suspended.scala:35                      schematic.suspended$Schema#$anonfun$compile$1
[E]   DynamicModelSpec.scala:131              smithy4s.dynamic.model.DynamicModelSpec$Interpreter$GetFieldNames$#$anonfun$suspend$1
[E]   DynamicModelSpec.scala:127              smithy4s.dynamic.model.DynamicModelSpec$Interpreter$GetFieldNames$#$anonfun$genericStruct$2
[E]   StrictOptimizedIterableOps.scala:118    scala.collection.StrictOptimizedIterableOps#flatMap
[E]   StrictOptimizedIterableOps.scala:105    scala.collection.StrictOptimizedIterableOps#flatMap$
[E]   Vector.scala:113                        scala.collection.immutable.Vector#flatMap
[E]   DynamicModelSpec.scala:126              smithy4s.dynamic.model.DynamicModelSpec$Interpreter$GetFieldNames$#$anonfun$genericStruct$1
[E]   DynamicModelSpec.scala:131              smithy4s.dynamic.model.DynamicModelSpec$Interpreter$GetFieldNames$#$anonfun$suspend$1
[E]   DynamicModelSpec.scala:163              smithy4s.dynamic.model.DynamicModelSpec$Interpreter$#$anonfun$toFieldNames$1
[E]   List.scala:293                          scala.collection.immutable.List#flatMap
[E]   DynamicModelSpec.scala:162              smithy4s.dynamic.model.DynamicModelSpec$Interpreter$#toFieldNames
[E]   DynamicModelSpec.scala:174              smithy4s.dynamic.model.DynamicModelSpec$#$anonfun$new$8
[E]   suites.scala:82                         delay @ weaver.MutableFSuite#$anonfun$pureTest$1
// only weaver stuff below
```

modules/dynamic/test/src-jvm/smithy4s/dynamic/model/DynamicModelSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+      svc.endpoints.flatMap { endpoint =>
+        endpoint.input.compile(GetFieldNames)() ++
+          endpoint.output.compile(GetFieldNames)()
+      }
+  }
+
+  pureTest(
+    "Extract field names from all structures in a service's endpoints"
+  ) {
+    val svc = dynamic.Compiler.compile(expected).allServices.head
```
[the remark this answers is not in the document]
Baccata: The test fails because the smithy prelude is not dumped in the json representation. We would prepopulate the visitor by running it on a handcrafted model containing just the primitives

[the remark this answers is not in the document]
kubukoz: makes sense. I'll let you know if I get to working on it first :)

kubukoz:
also, can we include this in the root project? It makes local publishing much easier.

Baccata:
> also, can we include this in the root project? It makes local publishing much easier.

@kubukoz sure yeah, go for it :)

modules/dynamic/test/src-jvm/smithy4s/dynamic/model/DynamicModelSpec.scala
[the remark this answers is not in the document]
kubukoz: Tried that, turns out the `floatShape` visitor was not doing anything in particular - still the case for other cases. E.g. when I tried to compile PizzaSpec's model, it's doesn't recognize `Food` (a union). I can look at this again after the weekend.

kubukoz:
Added new spec, trying to dump the model of PizzaSpec to find more issues - currently it's failing because it doesn't recognize the `@simpleRestJson` trait, so we probably need to change how the model assembler is used in that spec. Then there's the union issue: https://github.com/disneystreaming/smithy4s/pull/17#discussion_r785381131 - some other cases are also probably not being visited

modules/dynamic/test/src-jvm/smithy4s/dynamic/model/DynamicModelSpec.scala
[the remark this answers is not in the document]
Baccata: > Tried that, turns out the floatShape visitor was not doing anything in particular - still the case for other cases. E.g. when I tried to compile PizzaSpec's model, it's doesn't recognize Food (a union). I can look at this again after the weekend.

Yeah, I had left a fair amount there as TODOs

Baccata:
> currently it's failing because it doesn't recognize the @simpleRestJson trait, so we probably need to change how the model assembler is used in that spec

See 32bfcae. Not ideal, but I can live with it. 

> Then there's the union issue: #17 (comment) - some other cases are also probably not being visited

Covered more of the metamodel in eb05278 (totally untested, as usual 😄 )

modules/dynamic/src/smithy4s/dynamic/DynamicModel.scala
```diff
@@ -0,0 +1,18 @@
+package smithy4s
+package dynamic
+
+class DynamicModel(
```
Baccata, line 4: TODO : we'll need to 

* have the public methods of this construct return existential versions (honestly the allServices is gonna be a pita on Scala 2 ...) 
* move most things that leads to the construction in an `internals` package

modules/dynamic/test/src-jvm/smithy4s/dynamic/model/DynamicModelSpec.scala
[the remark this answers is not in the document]
Baccata: Most of it should be implemented in eb05278

modules/dynamic/test/src-jvm/smithy4s/dynamic/model/JsonIOProtocol.scala
```diff
@@ -0,0 +1,83 @@
...
+import cats.effect.IO
+import cats.syntax.all._
+
+/**
+  * These are toy interpreters that turn services into json-in/json-out
+  * functions, and vice versa.
+  *
+  * Created for testing purposes.
+  */
+object JsonIOProtocol {
```
Baccata, line 13: @kubukoz, I've created this to facilitate testing. Given a service given by the dynamic model, we can create a "proxy" function, that round trips a `Document => IO[Document]` kleisli through the `fromJsonIO` and `toJsonIO` functions. 

Then, if we take that proxy and create a corresponding statically generated stub from it, in theory the data should go through the proxy back and forth, in an untampered fashion. 

We can add some middleware to intercept and store the incoming and outgoing json in a CE-Ref, and either perform assertions on the json itself, or use it for logging purposes when we see that the data has not gone through the dynamic proxy as expected.

modules/dynamic/src/smithy4s/dynamic/DynamicModel.scala
Baccata: First point is done : I used abstract type members to circumvent the lack of higher-kinded existentials in Scala 2. That hides the `Any` from the end user, forcing them to only use dynamic services as "proxies"

modules/dynamic/test/src-jvm/smithy4s/dynamic/model/JsonIOProtocol.scala
```diff
@@ -0,0 +1,83 @@
+package smithy4s
```
kubukoz, line 1: should this go to shared sources instead of jvm-specific?

Baccata: Yup, you are correct

modules/dynamic/test/src/smithy4s/dynamic/DefaultSchematic.scala
```diff
@@ -0,0 +1,96 @@
...
+package smithy4s
+package dynamic
+
+import cats.Id
+import java.util.UUID
+import schematic.ByteArray
+import schematic.Field
+import schematic.Alt
+
+object DefaultSchematic
```
kubukoz, line 26: This seems useful for working with schemas, maybe it could be exposed to the users in some module closer to core?

Baccata: I thought about, but decided against it because of a lack for a better motivation, at this time. When the time arise, we'll move it

kubukoz:
Sooo this is currently working in all capacity that I'd need it to, is there anything major that needs to happen before we merge?

Baccata:
@kubukoz sorry I had missed that. I need to do another pass, clean some things up (make most of them private) and move stuff to an `internals`package. I'll try to do it next week.

Baccata:
@kubukoz I've cleaned things up, renamed some things, etc. Time for some bikeshedding :)

kubukoz (approved):

PR #25, opened 2022-01-04
Make ShapeId a datatype

Baccata:
And use it as a hint everywhere. The rationale is as follows :

1. would be useful for error messages
2. will be required for implementing some protocols
3. `ShapeId` can be used as a map key for dynamic purposes


lewisjkl (dismissed):

kubukoz (dismissed):

modules/core/test/src/smithy4s/http/internals/MetadataSpec.scala
```diff
@@ -211,9 +212,9 @@ object MetadataSpec extends FunSuite {
 
   test("map of list of strings query param") {
```
kubukoz, line 213: The name seems not to be accurate anymore, same in the other test

kubukoz (approved):

PR #26, opened 2022-01-04
add js.Any => Document converter

### LINKED ISSUE #21: Create `js.Any => Document` conversion function
Basically recreate [this](https://github.com/circe/circe/blob/e6685317bc9e6e65539bb6194900e0e754fbcb7c/modules/scalajs/src/main/scala/io/circe/scalajs/package.scala#L12-L24) in the core module, so that we can easily convert native js data to Document to benefit from the Document decoders

lewisjkl:
Closes #21.

### DIFF
--- a/modules/core/test/src-js/smithy4s/JsConvertersSpec.scala
+++ b/modules/core/test/src-js/smithy4s/JsConvertersSpec.scala
@@ -1,0 +1,54 @@
+package smithy4s
+
+import weaver._
+import scala.scalajs.js
+
+object JsConvertersSpec extends FunSuite {
+
+  test("object - strings") {
+    val input = js.Dictionary("one" -> "1")
+    val result = JsConverters.convertJsToDocument(input)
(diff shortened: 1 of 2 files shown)


modules/core/test/src-js/smithy4s/JsConvertersSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  test("integer") {
+    val input = 1
+    val result = JsConverters.convertJsToDocument(input)
+    val expected = Right(Document.fromInt(1))
+    expect(result == expected)
+  }
+
+  test("null") {
```
kubukoz, line 31: Another one for undefined would be great here :)

lewisjkl: Good call, just added 🙂

kubukoz (approved):

PR #29, opened 2022-01-05
add header validator

### LINKED ISSUE #28: Ensure headers referenced in smithy specs are actually settable
We have an API where we have required that the "content-type" header be set as part of the smithy spec, eg

    @required
    @httpHeader("content-type")
    contentType: CustomContentType

Unfortunately due to how http4s works, if you derive a client for this API, the trait will require the parameter, but the actual request generated will have the "static" header value ie "Content-Type: application/json; charset=UTF-8". 

Smithy4s should disallow headers that it knows it cannot reliably set.

lewisjkl:
Adds validator to make sure headers cannot be set if they are automatically set down the line.

Closes #28.

### DIFF
--- a/modules/protocol/src/smithy4s/api/validation/HttpHeaderValidator.java
+++ b/modules/protocol/src/smithy4s/api/validation/HttpHeaderValidator.java
@@ -33,8 +33,8 @@ public final class HttpHeaderValidator extends AbstractValidator {
     public List<ValidationEvent> validate(Model model) {
         return model.getShapesWithTrait(HttpHeaderTrait.class).stream().flatMap(headerShape -> {
             String value = headerShape.getTrait(HttpHeaderTrait.class).get().getValue();
-            if (disallowedHeaderNames.contains(value)) {
-                return Stream.of(error(headerShape, String.format("Header named `%s` should not be present since it is automatically filled", value)));
+            if (disallowedHeaderNames.contains(value.toLowerCase())) {
+                return Stream.of(warning(headerShape, String.format("Header named `%s` should not be present since it is automatically filled", value)));
             } else {
                 return Stream.empty();
             }
--- a/modules/protocol/src/smithy4s/api/validation/HttpHeaderValidator.java
+++ b/modules/protocol/src/smithy4s/api/validation/HttpHeaderValidator.java
@@ -1,0 +1,43 @@
+/*
+ *  Copyright 2021 Disney Streaming
+ *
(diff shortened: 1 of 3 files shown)


modules/protocol/src/smithy4s/api/validation/HttpHeaderValidator.java
```diff
@@ -0,0 +1,43 @@
...
+public final class HttpHeaderValidator extends AbstractValidator {
+
+    List<String> disallowedHeaderNames = java.util.Arrays.asList("content-type");
+
+    @Override
+    public List<ValidationEvent> validate(Model model) {
+        return model.getShapesWithTrait(HttpHeaderTrait.class).stream().flatMap(headerShape -> {
+            String value = headerShape.getTrait(HttpHeaderTrait.class).get().getValue();
+            if (disallowedHeaderNames.contains(value)) {
+                return Stream.of(error(headerShape, String.format("Header named `%s` should not be present since it is automatically filled", value)));
```
kubukoz, line 37: Maybe we should mention the @mediaType trait here

modules/protocol/src/smithy4s/api/validation/HttpHeaderValidator.java
```diff
@@ -0,0 +1,43 @@
...
+
+public final class HttpHeaderValidator extends AbstractValidator {
+
+    List<String> disallowedHeaderNames = java.util.Arrays.asList("content-type");
+
+    @Override
+    public List<ValidationEvent> validate(Model model) {
+        return model.getShapesWithTrait(HttpHeaderTrait.class).stream().flatMap(headerShape -> {
+            String value = headerShape.getTrait(HttpHeaderTrait.class).get().getValue();
+            if (disallowedHeaderNames.contains(value)) {
```
Baccata, line 36: Headers are case insensitive, so this doesn't quite cut it

modules/protocol/src/smithy4s/api/validation/HttpHeaderValidator.java
Baccata: Not sure, tbh. Like, the simpleRestJson protocol doesn't take it into consideration

Baccata: Also, error is too hard a severity. It might be that someone provides an implementation that allows for it, so
the error severity should only be used if we check that the header is in the closure of a simpleRestJson service.

If we don't perform that check, it should be a warning, not an error

Baccata (approved):

PR #30, opened 2022-01-06
add discriminated union trait

lewisjkl:


modules/protocol/resources/META-INF/smithy/smithy4s.smithy
```diff
@@ -41,3 +41,7 @@ structure UncheckedExample {
 @trait(selector: "string")
 structure uuidFormat {
 }
+
+@trait(selector: "union")
+structure discriminated {
```
kubukoz, line 46: I think this should be parameterized by the field name. Personally I've seen "@type", "type", "kind", "node_kind" and so on, so it should probably be up to the API drafter :)

kubukoz (changes requested):

modules/protocol/resources/META-INF/smithy/smithy4s.smithy
Baccata: That is indeed correct. See "propertyName" in openapi : https://spec.openapis.org/oas/latest.html#discriminator-object. 

I think I've seen `type` more than anything else, but making it explicit is probably a good call.

modules/core/src/smithy4s/internals/SchematicDocumentDecoder.scala
```diff
@@ -344,49 +345,103 @@ object SchematicDocumentDecoder
...
+  )(
+      total: S => Alt.WithValue[DocumentDecoderMake, S, _]
+  ): DocumentDecoderMake[S] = {
+    def jsonLabel[A](alt: Alt[DocumentDecoderMake, S, A]): String =
+      alt.instance.hints.get(JsonName).map(_.value).getOrElse(alt.label)
+
+    val decoders: DecoderMap[S] =
+      (first +: rest).map { case alt @ Alt(_, instance, inject) =>
+        val encoder = { (pp: List[PayloadPath.Segment], doc: Document) =>
+          inject(instance.get.apply(jsonLabel(alt) :: pp, doc))
```
Baccata, line 435: extract the `jsonLabel` call from here so that it ain't on the hot path.

modules/core/src/smithy4s/internals/SchematicDocumentDecoder.scala
```diff
@@ -344,49 +345,103 @@ object SchematicDocumentDecoder
...
+        }
+        jsonLabel(alt) -> encoder
+      }.toMap
+
+    Hinted[DocumentDecoder].onHintOpt[Discriminated, S] {
```
Baccata, line 440: 👍

modules/core/src/smithy4s/internals/SchematicDocumentDecoder.scala
```diff
@@ -344,49 +345,103 @@ object SchematicDocumentDecoder
...
+                s"Unable to locate discriminator under property '${discriminated.propertyName}'"
+              )
           }
-          jsonLabel(alt) -> encoder
-        }.toMap
+        case _ =>
+          throw new PayloadError(
+            PayloadPath(pp.reverse),
+            "Union",
+            "Expected more than single-key Json object"
```
Baccata, line 391: Not necessarily : the alternative could be an empty structure

modules/core/src/smithy4s/internals/SchematicDocumentDecoder.scala
```diff
@@ -344,49 +345,103 @@ object SchematicDocumentDecoder
...
+  private def discriminatedUnion[S](
+      discriminated: Discriminated,
+      decoders: DecoderMap[S]
+  ): DocumentDecoder[S] = handleUnion {
+    (pp: List[PayloadPath.Segment], document: Document) =>
+      document match {
+        case DObject(map) if (map.size > 1) =>
+          map
+            .get(discriminated.propertyName) match {
+            case Some(value: Document.DString) =>
```
Baccata, line 370: 👍

modules/core/src/smithy4s/internals/SchematicDocumentDecoder.scala
lewisjkl: Ah yeah good call, I will update and add a test case to make sure that is handled

modules/json/src/smithy4s/http/json/SchematicJCodec.scala
```diff
@@ -464,76 +466,163 @@ private[smithy4s] class SchematicJCodec(constraints: Constraints, maxArity: Int)
...
-        def encodeKey(x: Z, out: JsonWriter): Unit =
-          out.encodeError("Cannot use coproducts as keys")
+      def decodeValue(cursor: Cursor, in: JsonReader): Z = {
+        if (in.isNextToken('{')) {
+          if (in.isNextToken('}'))
+            in.decodeError("Expected at least a one key/value pair")
+          else {
+            in.rollbackToken()
+            in.setMark()
+            val key = if (in.skipToKey(discriminated.propertyName)) {
```
Baccata, line 573: oh really cool ! I thought we were gonna have to do that by hand, happy it's not the case

modules/json/src/smithy4s/http/json/SchematicJCodec.scala
```diff
@@ -464,76 +466,163 @@ private[smithy4s] class SchematicJCodec(constraints: Constraints, maxArity: Int)
...
+            } match {
+              case Some(value) => value
+              case None        => in.discriminatorValueError(key)
+            }
+            result
+          }
+        } else in.decodeError("Expected JSON object")
       }
+
+      val altCache = new PolyFunction[Alt[JCodecMake, Z, *], JCodec] {
```
Baccata, line 594: 👍

modules/json/src/smithy4s/http/json/SchematicJCodec.scala
lewisjkl: Yeah I was super happy to learn of `setMark` + `skipToKey` + `rollbackToMark`.. made the backtracking a lot easier

Baccata: Yeah I was aware of `setMark` + `rollbackToMark`. Kudos to @plokhotnyuk, I **really** appreciate jsoniter's API.

Baccata:
The implementation of the schematics look good to me. Maybe we could do with some property tests generating random schemas that have the discriminated trait, but I'm okay with us doing that later. 

There's two things that I'd like to have before merging : 

1. Some smithy validator that checks that a `union` with the `discriminated` trait only has alternatives that are structures. Alternatively, a selector on the `discriminated` trait that achieves the same thing. 
2. Should we make `discriminated` a string instead of a struct ? I'm thinking maybe making it a struct is overkill, because besides the name of the discriminator, I don't think any other property is relevant.

lewisjkl:
> Some smithy validator that checks that a union with the discriminated trait only has alternatives that are structures. Alternatively, a selector on the discriminated trait that achieves the same thing.

Agreed, thanks for the reminder.

> Should we make discriminated a string instead of a struct ? I'm thinking maybe making it a struct is overkill, because besides the name of the discriminator, I don't think any other property is relevant.

Yeah, I can't think of anything else that would ever be specified in there. I will update it to be a string.

Baccata (approved):
👍  pre-approving (conditioned by the addition of a validator for the discriminator trait)

modules/protocol/src/smithy4s/api/validation/DiscriminatedUnionValidator.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public final class DiscriminatedUnionValidator extends AbstractValidator {
+
+    @Override
+    public List<ValidationEvent> validate(Model model) {
+        return model.getShapesWithTrait(DiscriminatedUnionTrait.class).stream().flatMap(unionShape -> {
+           return unionShape.asUnionShape().get().getAllMembers().entrySet().stream().flatMap(entry -> {
+              System.out.println(model.getShape(entry.getValue().getTarget()));
```
Baccata, line 36: cleanup

modules/protocol/src/smithy4s/api/validation/DiscriminatedUnionValidator.java
```diff
@@ -0,0 +1,46 @@
...
+              Optional<Shape> maybeTarget = model.getShape(entry.getValue().getTarget());
+              if (maybeTarget.isPresent() && maybeTarget.get().isStructureShape()) { // if not defined then shape won't be structure
+                  return Stream.empty();
+              } else {
+                  return Stream.of(error(entry.getValue(), String.format("Member shape corresponding to key '%s' is not a structure shape", entry.getKey())));
```
Baccata, line 41:
```suggestion
                  return Stream.of(error(entry.getValue(), String.format("Target of member '%s' is not a structure shape", entry.getKey())));
```

Baccata (approved):
re-approving

kubukoz (approved):
The only thing that I'm kind of missing is an example in a `.smithy` file, but otherwise it looks good

lewisjkl:
> The only thing that I'm kind of missing is an example in a `.smithy` file, but otherwise it looks good

Good call, I will add that

lewisjkl:
One thing that we will want to look at (possibly in different PR) is the OpenAPI conversion respecting the discriminated union trait. I think there should be a `discriminator` field in the OpenAPI representation. (https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/)

Baccata:
Happy for the openapi conversion to happen in a subsequent PR

kubukoz:
🎉

PR #36, opened 2022-01-10
add id hints onto more generated types

### LINKED ISSUE #35: Struct/union name hints
It would be great if the `Schematic` instances for structs generated by smithy4s had a hint for their types' names, or some other way to get that value at any level of the structure.

Currently there's not a (clear?) way to get any of those names, which would be very useful for building tooling / documentation on top of the generated code.

lewisjkl:
This is an attempt to close #35. There are a few areas I am unsure of (will add comments below inline). Additionally, I wasn't sure if we want to be adding the hints onto primitive types within structures and such. Thoughts?

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/Renderer.scala
@@ -466,7 +466,7 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
         newline, {
           val union =
             if (error)
-              s"val schema: $errorUnion_.Schema[$name] = errors"
+              s"val schema: $errorUnion_.Schema[$name] = (errors"
             else if (recursive)
               s"val schema: $Schema_[$name] = recursive(union"
             else
@@ -480,7 +480,9 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
                 s"case c : $caseName => $caseName.alt(c)"
               }
             }
-            .appendToLast(if (hints.nonEmpty) ".withHints(hints)" else "")
+            .appendToLast(
+              if (error) ").withHints(hints)" else ".withHints(hints)"
+            )
             .appendToLast(if (recursive) ")" else "")
         },
         s"implicit val staticSchema : $Static_[$Schema_[$name]] = $Static_(schema)"
@@ -505,6 +507,8 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
     obj(name, ext = s"$Enumeration_[$name]", w = hintKey(name, hints))(
       renderId(name),
       newline,
+      renderHintsValWithId(hints),
+      newline,
       values.zipWithIndex.map { case (e @ EnumValue(value, _, _), index) =>
         line(
           s"""case object ${e.className} extends $name("$value", $index)"""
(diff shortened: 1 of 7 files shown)


modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -517,7 +521,7 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
       newline,
       line(s"def to(e: $name) : (String, Int) = (e.value, e.ordinal)"),
       lines(
-        s"val schema: $Schema_[$name] = enumeration(to, valueMap, ordinalMap)",
+        s"val schema: $Schema_[$name] = (enumeration(to, valueMap, ordinalMap): $Schema_[$name]).withHints(hints)",
```
lewisjkl, line 524: Doesn't compile without the type hint here, not sure of another way around this atm

modules/example/src/smithy4s/example/BucketName.scala
```diff
@@ -5,7 +5,10 @@ import smithy4s.syntax._
 
 object BucketName extends Newtype[String] {
   val id: smithy4s.ShapeId = smithy4s.ShapeId("smithy4s.example", "BucketName")
-  val underlyingSchema : smithy4s.Schema[String] = string
+  val hints : smithy4s.Hints = smithy4s.Hints(
+    id,
+  )
+  val underlyingSchema : smithy4s.Schema[String] = string.withHints(hints)
   val schema : smithy4s.Schema[BucketName] = bijection(underlyingSchema, BucketName(_), (_ : BucketName).value)
```
lewisjkl, line 12: Wasn't sure if this was correct or not to have the hints on `underlyingSchema`.. I am thinking they should actually be on the `schema` following the `bijection`?

kubukoz: I think on the whole thing, yeah (not `underlyingSchema`)

Baccata: It shouldn't make a difference :)

modules/schematic-core/src/schematic/enumeration.scala
```diff
@@ -39,12 +39,12 @@ object enumeration {
   }
 
   trait Syntax {
-    def enumeration[H, S[x[_]] <: Schematic[x], A](
```
Baccata, line 42: This was un-necessary, and preventing type inference from working properly

modules/core/src/smithy4s/syntax/package.scala
```diff
@@ -45,9 +31,53 @@ package object syntax
     with schematic.suspended.ClosedSyntax[Schematic]
     with schematic.bijection.ClosedSyntax[Schematic]
     with Hints.ClosedSyntax[Schematic]
-    with Document.ClosedSyntax[Schematic]
     with errorUnion.Syntax {
 
+  val short: Schema[Short] =
```
Baccata, line 36: the changes bellow should address the concern of not having shape ids for primitive types

modules/example/src/smithy4s/example/BucketName.scala
Baccata: Actually,@kubukoz is correct. On the `schema`, we have a chance of writing interpreters that retain the information of the id of the underlying schema;

Baccata: Actually maybe not. I think it's fine as it is : the hints should be applied on the "underlyingSchema". That is because we do not use the newtypes for collections, except when they are used as hints themselves. However, we still want the underlying schema of collections to contain the hints.

Baccata: (We should really have some smoke tests for this kind of thing 😓 )

kubukoz: I thought to have this on the big thing so that the implementation of the Schematic would carry it over to the wrapped schema in a `Hinted` or something

Baccata: > I thought to have this on the big thing so that the implementation of the Schematic would carry it over to the wrapped schema in a Hinted or something

Yeah, that is not incorrect, but having it on the small thing makes more sense, because otherwise you wouldn't have the `id` for collection shapes.

kubukoz: ahh I see now

Baccata (approved):

PR #38, opened 2022-01-11
remove core dependency auto-add from sbt plugin

### LINKED ISSUE #34: schematic.StubSchematic can't be used in Scala.JS
It breaks linking because of extending `javatime.StubSchematic`:

```scala
object CompletionSchematic {
  type Result[A] = List[String]
}

class CompletionSchematic extends smithy4s.StubSchematic[CompletionSchematic.Result] {
  def default[A]: List[String] = Nil
}
```

```
[error] Referring to non-existent method schematic.javatime$StubSchematic.$init$()void
[error]   called from constructor playground.CompletionSchematic.<init>()void
[error]   called from private playground.completions$.$anonfun$complete$10(scala.collection.immutable.List,smithy4s.Service,playground.smithyql.Query,playground.smithyql.WithSource$Thing)scala.collection.IterableOnce
[error]   called from private playground.completions$.$anonfun$complete$9(scala.collection.immutable.List,smithy4s.Service,typings.vscode.mod.TextDocument,typings.vscode.mod.Position)scala.collection.immutable.List
[error]   called from playground.completions$.complete(smithy4s.Service)scala.Function2
[error]   called from playground.extension$.activate(typings.vscode.mod.ExtensionContext)void
[error]   exported to JavaScript with @JSExport
[error] involving instantiated classes:
[error]   playground.completions$
[error]   playground.extension$
[error] There were linking errors
[error] (vscodeJS / Compile / fastLinkJS) There were linking errors
```

lewisjkl:
Closes #34.

The issue was that the sbt plugin was auto-adding the `smithy4s-core` dependency. This worked in all cases except for Scala JS since that would require the use of `%%%`. We could have updated the plugin to use `%%%`, but that would require the Scala JS SBT plugin be present on all projects using it, which we do not want to require.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,11 @@ import smithy4s.codegen.Smithy4sCodegenPlugin
 val myModule = project
   .in(file("modules/my-module"))
   .enablePlugins(Smithy4sCodegenPlugin)
+  // version for smithy4s-core is sourced from Smithy4sCodegenPlugin
+  .settings(libraryDependencies += "com.disneystreaming.smithy4s" %%% "smithy4s-core" % smithy4sVersion.value)
 ```
 
-This will add the `smithy-core` dependency to `myModule`, since it is needed
+This will enable the plugin on `myModule`. We also need to add `smithy4s-core here since it is needed
 for compiling the generated code.
 
 This will look for the smithy specs in the folder `$MY_MODULE/src/main/smithy` and will write scala code in `$MY_MODULE/target/scala-<version>/src_managed/` when invoking `compile`. The paths are configurable via the `smithy4sInputDir` and `smithy4sOutputDir` settings keys.
@@ -46,7 +48,8 @@ val myModule = project
   .settings(
     scalaVersion := "2.13.6",
     smithy4sInputDir in Compile  := (baseDirectory in ThisBuild).value / "smithy_input",
-    smithy4sOutputDir in Compile := (baseDirectory in ThisBuild).value / "smithy_output"
+    smithy4sOutputDir in Compile := (baseDirectory in ThisBuild).value / "smithy_output",
+    libraryDependencies += "com.disneystreaming.smithy4s" %%% "smithy4s-core" % smithy4sVersion.value
   )
 
 ```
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -30,9 +30,11 @@ import smithy4s.codegen.Smithy4sCodegenPlugin
 val myModule = project
   .in(file("modules/my-module"))
   .enablePlugins(Smithy4sCodegenPlugin)
+  // version for smithy4s-core is sourced from Smithy4sCodegenPlugin
+  .settings(libraryDependencies += "com.disneystreaming.smithy4s" %%% "smithy4s-core" % smithy4sVersion.value)
```
kubukoz, line 34: %%% won't work on builds without scalajs. In general, I think this approach is non-ideal (asking people to add the dependency themselves) but I don't have a better idea for now :/ hoping @Baccata has something better

kubukoz: ah, I now see the comment in #34 saying it's already been discussed

kubukoz: still, the recommendation should be `%%` I think

lewisjkl: Fair enough, I was trying to signal that it worked for ScalaJS as well, but maybe that is implied enough to change. And it is likely that ScalaJS folks will figure that out where a newcomer who isn't familiar with '%%%' might get confused

kubukoz: 💯

kubukoz (approved):
This will probably need an entry in the release notes :)

armanbilge:
> We could have updated the plugin to use `%%%`, but that would require the Scala JS SBT plugin be present on all projects using it, which we do not want to require.

This ship has probably sailed, but FYI `%%%` is a separate plugin.
https://github.com/portable-scala/sbt-platform-deps

Baccata:
> https://github.com/disneystreaming/smithy4s/pull/38#issuecomment-1092307103 

https://www.nicepng.com/png/detail/41-416284_doh-homer-simpson-quotes-on-quotestopics-homer-simpson.png

I think we might be able to revert then.

PR #40, opened 2022-01-11
Add smoke tests to check for hints on schemas

lewisjkl:
This is still a WIP, but the idea is adding a few test cases to ensure the fix in #36 is working.

It seems like there may be an easier way than the approach I am taking here, so please let me know if there is. I couldn't find a great way to get the hints from a schema without compiling the schema.

The union test is not working as of right now, but the other two are.

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/Renderer.scala
@@ -362,7 +362,7 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
               }
             )
             .block(s"$name.apply")
-            .appendToLast(if (hints.nonEmpty) ".withHints(hints)" else "")
+            .appendToLast(".withHints(hints)")
             .appendToLast(if (recursive) ")" else "")
         } else {
           line(
--- a/modules/core/test/src/smithy4s/ShapeIdHintsSmokeSpec.scala
+++ b/modules/core/test/src/smithy4s/ShapeIdHintsSmokeSpec.scala
@@ -1,0 +1,152 @@
+package smithy4s
+
+import java.util.UUID
+import schematic.ByteArray
+import schematic.Field
+import schematic.Alt
+import cats.Id
+import schematic.struct.GenericAritySchematic
+import smithy4s.capability.Invariant
+
+object ShapeIdHintsSmokeSpec extends weaver.FunSuite {
+
+  type HintTracker[A] = TestHinted[Id, A]
+
+  case class TestHinted[F[_], A](hints: List[Hint], make: List[Hint] => F[A]) {
+    def addHints(h: List[Hint]): TestHinted[F, A] = copy(hints = hints ++ h)
+    def imap[B](to: A => B, from: B => A)(implicit
+        I: Invariant[F]
+    ): TestHinted[F, B] =
+      TestHinted(hints, h => I.imap(make(h))(to, from))
+  }
+
+  object TestHinted {
+    def apply[F[_]]: PartiallyAppliedHintedTest[F] =
+      new PartiallyAppliedHintedTest[F]
+
+    class PartiallyAppliedHintedTest[F[_]] {
+      def from[A](f: List[Hint] => F[A]): TestHinted[F, A] =
+        TestHinted(List.empty, f(_))
+    }
+  }
+
+  implicit def invariantFromCats[F[_]: cats.Invariant]: Invariant[F] =
+    new Invariant[F] {
+      def imap[A, B](fa: F[A])(to: A => B, from: B => A): F[B] =
+        cats.Invariant[F].imap(fa)(to)(from)
+    }
+
+  private object TestCompiler
+      extends Schematic[HintTracker]
+      with GenericAritySchematic[HintTracker] {
+    def short: HintTracker[Short] = ???
+
+    def int: HintTracker[Int] = ???
+
+    def long: HintTracker[Long] = ???
+
+    def double: HintTracker[Double] = ???
+
+    def float: HintTracker[Float] = TestHinted[Id].from(_ => 0f)
+
+    def bigint: HintTracker[BigInt] = ???
+
+    def bigdecimal: HintTracker[BigDecimal] = ???
+
+    def string: HintTracker[String] = TestHinted[Id].from(_ => "")
+
+    def boolean: HintTracker[Boolean] = ???
+
+    def uuid: HintTracker[UUID] = ???
+
+    def byte: HintTracker[Byte] = ???
+
+    def bytes: HintTracker[ByteArray] = ???
+
+    def unit: HintTracker[Unit] = ???
+
+    def list[S](fs: HintTracker[S]): HintTracker[List[S]] = ???
+
+    def set[S](fs: HintTracker[S]): HintTracker[Set[S]] = ???
+
+    def vector[S](fs: HintTracker[S]): HintTracker[Vector[S]] = ???
+
+    def map[K, V](
+        fk: HintTracker[K],
+        fv: HintTracker[V]
+    ): HintTracker[Map[K, V]] = ???
+
+    def genericStruct[S](fields: Vector[Field[HintTracker, S, _]])(
+        const: Vector[Any] => S
+    ): HintTracker[S] = {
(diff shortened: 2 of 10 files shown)


modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -362,7 +362,7 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
               }
             )
             .block(s"$name.apply")
-            .appendToLast(if (hints.nonEmpty) ".withHints(hints)" else "")
```
lewisjkl, line 365: Structures should have hints on them containing the ShapeId, right?

kubukoz: think so

modules/core/test/src/smithy4s/ShapeIdHintsSmokeSpec.scala
```diff
@@ -0,0 +1,152 @@
...
+import schematic.Alt
+import cats.Id
+import schematic.struct.GenericAritySchematic
+import smithy4s.capability.Invariant
+
+object ShapeIdHintsSmokeSpec extends weaver.FunSuite {
+
+  type HintTracker[A] = TestHinted[Id, A]
+
+  case class TestHinted[F[_], A](hints: List[Hint], make: List[Hint] => F[A]) {
```
kubukoz, line 15: Is there any particular reason to have `make` at all?

```diff
diff --git a/modules/core/test/src/smithy4s/ShapeIdHintsSmokeSpec.scala b/modules/core/test/src/smithy4s/ShapeIdHintsSmokeSpec.scala
index eb2f9b7..0baf4f5 100644
--- a/modules/core/test/src/smithy4s/ShapeIdHintsSmokeSpec.scala
+++ b/modules/core/test/src/smithy4s/ShapeIdHintsSmokeSpec.scala
@@ -4,38 +4,18 @@ import java.util.UUID
 import schematic.ByteArray
 import schematic.Field
 import schematic.Alt
-import cats.Id
 import schematic.struct.GenericAritySchematic
-import smithy4s.capability.Invariant
 
 object ShapeIdHintsSmokeSpec extends weaver.FunSuite {
 
-  type HintTracker[A] = TestHinted[Id, A]
+  // todo remove
+  type HintTracker[A] = TestHinted[A]
 
-  case class TestHinted[F[_], A](hints: List[Hint], make: List[Hint] => F[A]) {
-    def addHints(h: List[Hint]): TestHinted[F, A] = copy(hints = hints ++ h)
-    def imap[B](to: A => B, from: B => A)(implicit
-        I: Invariant[F]
-    ): TestHinted[F, B] =
-      TestHinted(hints, h => I.imap(make(h))(to, from))
+  case class TestHinted[A](hints: List[Hint]) {
+    def addHints(h: List[Hint]): TestHinted[A] = copy(hints = hints ++ h)
+    def retag[B]: TestHinted[B] = copy()
   }
 
-  object TestHinted {
-    def apply[F[_]]: PartiallyAppliedHintedTest[F] =
-      new PartiallyAppliedHintedTest[F]
-
-    class PartiallyAppliedHintedTest[F[_]] {
-      def from[A](f: List[Hint] => F[A]): TestHinted[F, A] =
-        TestHinted(List.empty, f(_))
-    }
-  }
-
-  implicit def invariantFromCats[F[_]: cats.Invariant]: Invariant[F] =
-    new Invariant[F] {
-      def imap[A, B](fa: F[A])(to: A => B, from: B => A): F[B] =
-        cats.Invariant[F].imap(fa)(to)(from)
-    }
-
   private object TestCompiler
       extends Schematic[HintTracker]
       with GenericAritySchematic[HintTracker] {
@@ -47,13 +27,13 @@ object ShapeIdHintsSmokeSpec extends weaver.FunSuite {
 
     def double: HintTracker[Double] = ???
 
-    def float: HintTracker[Float] = TestHinted[Id].from(_ => 0f)
+    def float: HintTracker[Float] = TestHinted(Nil)
 
     def bigint: HintTracker[BigInt] = ???
 
     def bigdecimal: HintTracker[BigDecimal] = ???
 
-    def string: HintTracker[String] = TestHinted[Id].from(_ => "")
+    def string: HintTracker[String] = TestHinted(Nil)
 
     def boolean: HintTracker[Boolean] = ???
 
@@ -79,7 +59,7 @@ object ShapeIdHintsSmokeSpec extends weaver.FunSuite {
     def genericStruct[S](fields: Vector[Field[HintTracker, S, _]])(
         const: Vector[Any] => S
     ): HintTracker[S] = {
-      val result = TestHinted[Id].from(_ => const(fields))
+      val result = TestHinted[S](Nil)
       val allHints = fields
         .flatMap(_.instance.hints)
       result.addHints(allHints.toList)
@@ -102,7 +82,7 @@ object ShapeIdHintsSmokeSpec extends weaver.FunSuite {
         f: HintTracker[A],
         to: A => B,
         from: B => A
-    ): HintTracker[B] = f.imap(to, from)
+    ): HintTracker[B] = f.retag
 
     def timestamp: HintTracker[Timestamp] = ???
 
```

modules/core/test/src/smithy4s/ShapeIdHintsSmokeSpec.scala
```diff
@@ -0,0 +1,152 @@
...
+    }
+  }
+
+  implicit def invariantFromCats[F[_]: cats.Invariant]: Invariant[F] =
+    new Invariant[F] {
+      def imap[A, B](fa: F[A])(to: A => B, from: B => A): F[B] =
+        cats.Invariant[F].imap(fa)(to)(from)
+    }
+
+  private object TestCompiler
```
kubukoz, line 39: First thing I think seeing this...

image


I feel like this could be much simpler, a `StubSchematic` that compiles everything to its members' shape IDs. Unless we need the actual values provided by schematic 😅 

This seems to pass the two tests you already had passing:

```scala
type ToShapeIds[A] = List[ShapeId]

object SimpleTestCompiler extends StubSchematic[ToShapeIds] {
  def default[A]: List[ShapeId] = Nil
  override def bijection[A, B](
      f: ToShapeIds[A],
      to: A => B,
      from: B => A
  ): ToShapeIds[B] = f

  override def genericStruct[S](fields: Vector[Field[ToShapeIds, S, _]])(
      const: Vector[Any] => S
  ): ToShapeIds[S] =
    fields.flatMap(_.instance).toList

  override def withHints[A](fa: ToShapeIds[A], hints: Hints): ToShapeIds[A] =
    fa ++ hints.get(ShapeId)
}
```

kubukoz: We could just as well make this collect all the hints if we want to test for that in the future. The only concern I have right now is whether we want to check which field has which ID, which would be a bit more complicated.

Baccata: > I have right now is whether we want to check which field has which ID, which would be a bit more complicated. 

nah, I'm fine not doing it

Baccata: I like @kubukoz's idea. Add a method for unions (and corresponding shapes in the model for the test to retrieve), and I think we're good.

lewisjkl: Thanks for saving me from myself @kubukoz 😅 haha this looks great!

modules/core/test/src/smithy4s/ShapeIdHintsSmokeSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+        to: A => B,
+        from: B => A
+    ): ToShapeIds[B] = f
+
+    override def genericStruct[S](fields: Vector[Field[ToShapeIds, S, _]])(
+        const: Vector[Any] => S
+    ): ToShapeIds[S] =
+      fields.flatMap(_.instance).toList
+
+    override def list[S](fs: ToShapeIds[S]): ToShapeIds[List[S]] = fs
```
kubukoz, line 24: Should there be cases for sets, maps etc?

kubukoz (approved):

Baccata (approved):

PR #44, opened 2022-01-13
add support for transforms in codegen

lewisjkl:
Adds support for applying transformers from artifacts in codegen.

The argument `transformers` in the codegen args is for specifying the names of each transformer that should be applied.

### DIFF
--- a/modules/codegen-cli/src/smithy4s/codegen/cli/DumpModelCommand.scala
+++ b/modules/codegen-cli/src/smithy4s/codegen/cli/DumpModelCommand.scala
@@ -27,7 +27,8 @@ object DumpModelCommand {
   val options = (
     specsArgs,
     repositoriesOpt.map(_.getOrElse(Nil)),
-    dependenciesOpt.map(_.getOrElse(Nil))
+    dependenciesOpt.map(_.getOrElse(Nil)),
+    transformersOpt.map(_.getOrElse(Nil))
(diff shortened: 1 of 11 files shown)


modules/codegen-cli/src/smithy4s/codegen/cli/DumpModelCommand.scala
```diff
@@ -27,7 +27,8 @@ object DumpModelCommand {
   val options = (
     specsArgs,
     repositoriesOpt.map(_.getOrElse(Nil)),
-    dependenciesOpt.map(_.getOrElse(Nil))
+    dependenciesOpt.map(_.getOrElse(Nil)),
+    transformersOpt.map(_.getOrElse(Nil))
```
kubukoz, line 31: :+1:

kubukoz (approved):

lewisjkl:
> Curious, what's the usecase for this?

Though I imagine there are a multitude of uses, the reason for doing this now is to make Smithy4s capable of recognizing shapes/traits defined in a separate protocol. For example, if another protocol (defined in a different repo) also defines a discriminated union trait and we want it to be treated the same was as the smithy4s.api#discriminated trait. With this feature, we can implement a transformer which will change all instances of that discriminated to the smithy4s one. We have this situation more or less in an internal use case.

Baccata (approved):

PR #56, opened 2022-01-18
create documentation microsite

lewisjkl:
Moves the current README into a docs project with mdoc setup. From there adds a docusaurus microsite. Colors/copy/etc can all be changed of course, I just went with what made sense off the top of my head.

This will make it so the microsite is pushed to a new branch `gh-pages`. We will need to set this branch to be the github pages page in the settings for this repo.

PR #59, opened 2022-01-20
Omitted that comma

arinal:


kubukoz (approved):

Baccata:
Good catch ! 

We're moving the contents of the readme to a microsite : https://github.com/disneystreaming/smithy4s/pull/56, which we'll release today. Therefore I'm gonna close this PR to avoid having to sync, and will port that fix over to the microsite PR

PR #60, opened 2022-01-20
Tentative fix of the signature implementation

Baccata:
The sigv4 trait contains what seems to be the service-specific name
that should be used during the signature of the request.


kubukoz:
Think we've got this implemented already? AWS seems to work fine

Baccata:
I think it's still needed, in the sense that it's probably working fine by accident, because the sigv4 name is coincidentally the same as the endpoint prefix for most services

Baccata:
Anyhow, we need parity tests with existing SDKs

PR #62, opened 2022-01-20
add quickstart guide to docs

lewisjkl:
What I added is based off a simple project I created here: https://github.com/lewisjkl/smithy4s-minimal-example

Happy to move that project into this repo as an example if we want to.


modules/docs/src/01-overview/02-quickstart.md
```diff
@@ -0,0 +1,137 @@
...
+
+ThisBuild / scalaVersion := "@SCALA_VERSION@"
+
+val example = project
+  .in(file("modules/example"))
+  .enablePlugins(Smithy4sCodegenPlugin)
+  .settings(
+    libraryDependencies ++= Seq(
+      "com.disneystreaming.smithy4s" %% "smithy4s-http4s" % smithy4sVersion.value,
+      "com.disneystreaming.smithy4s" %% "smithy4s-http4s-swagger" % smithy4sVersion.value,
```
lewisjkl, line 31: So technically, `smithy4s-http4s-swagger` brings in `smithy4s-http4s` transitively, should we still include both here? My gut says we should leave both.. but wanted to get other thoughts

kubukoz: Personally I don't care much about explicit dependencies, transitive is fine for me - I'd leave just the one that takes in all the others

kubukoz:
LGTM, what do you think about also showing a client? Maybe a simple main class (IOApp) that hits a single endpoint?

lewisjkl:
> LGTM, what do you think about also showing a client? Maybe a simple main class (IOApp) that hits a single endpoint?

Yeah that's a good idea, I think it'd be nice to call out that it can do clients right up front like that.

modules/docs/src/01-overview/02-quickstart.md
Baccata: For the sake of this doc, I'd rather keep both.

Baccata (approved):

kubukoz (approved):

PR #63, opened 2022-01-21
Use FQNs in rendered `Product` and `Serializable` inheritance

Baccata:
Using full-qualified-names because users encounter issues when they use `Product` as a shape name.

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/Renderer.scala
@@ -401,7 +401,7 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
     val caseNames = alts.map(_.name).map(caseName)
     val imports = alts.foldMap(_.tpe.imports) ++ syntaxImport
     lines(
-      s"sealed trait $name",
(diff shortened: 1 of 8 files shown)


modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -460,7 +460,7 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
       values: List[EnumValue],
       hints: List[Hint]
   ): RenderResult = lines(
-    s"sealed abstract class $name(val value: String, val ordinal: Int) extends Product with Serializable",
+    s"sealed abstract class $name(val value: String, val ordinal: Int) extends scala.Product with scala.Serializable",
```
kubukoz, line 463: general thought, should we do the same for collections? I imagine a type like `Vector` might appear in users' smithy files

kubukoz (approved):

modules/codegen/src/smithy4s/codegen/Renderer.scala
Baccata: I don't think that should be default behaviour, but maybe we'll add that as an option to the rendering

PR #66, opened 2022-01-21
fix versioning for microsite snapshots

lewisjkl:


.github/workflows/ci.yml
```diff
@@ -46,6 +46,8 @@ jobs:
     steps:
       - name: Checkout current branch
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
lewisjkl, line 50: This will make it so it checks all all history from git: https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches

kubukoz: tbh just fetching the tag list would be nice, not sure if that's possible (in a way that doesn't download the whoooole repo)

lewisjkl: This is needed because we will never know how many commits back a previous tag will be

kubukoz: it's probably fine because we have FD: 0 in the release job too

kubukoz (approved):
Happy to merge on green :)

.github/workflows/ci.yml
lewisjkl: > tbh just fetching the tag list would be nice, not sure if that's possible (in a way that doesn't download the whoooole repo)

Yeah I would love it if there was a way.. I don't see a way using this github actions plugin though

PR #70, opened 2022-01-21
add hint mask construct

lewisjkl:
Introduces the concept of `HintMask` that is derived from smithy protocolDefinitions


modules/core/src/smithy4s/HintMask.scala
```diff
@@ -0,0 +1,36 @@
+package smithy4s
+
+abstract class HintMask {
```
kubukoz, line 3: Could this just be a newtype over Hints?

modules/core/src/smithy4s/syntax/package.scala
```diff
@@ -113,4 +113,8 @@ package object syntax
       OneOf(oneOf.label, oneOf.schema.withHints(hints: _*), oneOf.inject)
   }
 
+  implicit class withMaskSyntax[F[_]](val s: Schematic[F]) extends AnyVal {
```
kubukoz, line 116: s should be private, otherwise you can do schematic.s.s.s.s.s I think

modules/protocol/resources/META-INF/smithy/smithy4s.smithy
```diff
@@ -14,7 +14,22 @@ namespace smithy4s.api
 @uuidFormat
 string UUID
 
-@protocolDefinition
+@protocolDefinition(traits: [
+    "smithy.api#cors",
```
kubukoz, line 18: Does @discriminated belong here?

modules/protocol/resources/META-INF/smithy/smithy4s.smithy
```diff
@@ -14,7 +14,22 @@ namespace smithy4s.api
...
+    "smithy.api#endpoint",
+    "smithy.api#hostLabel",
+    "smithy.api#http",
+    "smithy.api#httpError",
+    "smithy.api#httpHeader",
+    "smithy.api#httpLabel",
+    "smithy.api#httpPayload",
+    "smithy.api#httpPrefixHeaders",
```
kubukoz, line 26: I gotta learn what this is :D

modules/protocol/resources/META-INF/smithy/smithy4s.smithy
kubukoz: Also @ error, if it's a thing (can't remember, on phone, just spitballing)

modules/json/src/smithy4s/http/json/codecs.scala
```diff
@@ -0,0 +1,15 @@
+package smithy4s
+package http.json
+
+import smithy4s.syntax._
+
+final case class codecs(mask: HintMask)
```
kubukoz, line 6: Interesting API design here (the parameterless and parameterized version of codecs) :+1:

modules/protocol/resources/META-INF/smithy/smithy4s.smithy
kubukoz: didn't see the previous comments when I was writing this :D

lewisjkl: No worries :) appreciate the comments!

modules/core/src/smithy4s/HintMask.scala
Baccata: nope, Hints contain keys and values, HintMask just contains keys

modules/core/src/smithy4s/syntax/package.scala
Baccata: Mmm not sure it should leave there tbh, nor that it should be a extension, but 🤷

modules/core/src/smithy4s/HintMask.scala
```diff
@@ -0,0 +1,36 @@
...
+    def apply(hints: Hints): Hints = {
+      val hintKeysToRemove = hints.toMap.keySet.diff(toSet)
+      hintKeysToRemove.foldLeft(hints)((all, key) => all.remove(key))
+    }
+  }
+
+  private[this] final class MaskSchematic[F[_]](
+      schematic: Schematic[F],
+      mask: HintMask
+  ) extends PassthroughSchematic[F](schematic) {
```
Baccata, line 29: good idea !

modules/http4s/src/smithy4s/http4s/SimpleRestJsonBuilder.scala
```diff
@@ -19,5 +19,7 @@ package http4s
 
 object SimpleRestJsonBuilder
     extends SimpleProtocolBuilder[smithy4s.api.SimpleRestJson](
-      smithy4s.http.json.codecs
+      smithy4s.http.json.codecs(
+        HintMask()
+      ) // TODO smithy4s.api.SimpleRestJson.hintMask??
```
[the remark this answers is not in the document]
Baccata: > It would be great it we could apply the hint mask across the codecs from the SimpleProtocolBuilder 

We could add a method to the CodecAPI interface tbh.

> I tried looking at the renderer, but couldn't confidently figure out how to begin approaching this 

that's cool, we'll have a brainstorm next week, but the progress at this point in time is already very satisfying.

modules/core/src/smithy4s/PassthroughSchematic.scala
```diff
@@ -0,0 +1,70 @@
+package smithy4s
+
+import schematic._
+import schematic.struct.GenericAritySchematic
+import java.util.UUID
+
+class PassthroughSchematic[F[_]](schematic: Schematic[F])
+    extends Schematic[F]
+    with GenericAritySchematic[F] {
```
Baccata, line 9: So unfortunately, this `Passthrough` construct should not extend this. We'll have to use the build-time Boilerplate.

modules/core/src/smithy4s/HintMask.scala
kubukoz: Ah true 😂

modules/core/src/smithy4s/PassthroughSchematic.scala
lewisjkl: Is that because the schematics passed in may not be using the generic arity trait so they would rely on the other functions directly?

modules/core/src/smithy4s/syntax/package.scala
lewisjkl: Fair enough, I think it has proved less useful than I thought it would initially anyway

modules/core/src/smithy4s/PassthroughSchematic.scala
Baccata: Yes indeed

modules/core/src/smithy4s/HintMask.scala
```diff
@@ -0,0 +1,52 @@
...
+    new Impl(hintKeys.toSet)
+  }
+
+  private[this] final class Impl(val toSet: Set[Hints.Key[_]])
+      extends HintMask {
+    def ++(other: HintMask): HintMask =
+      new Impl(toSet ++ other.toSet)
+    def apply(hints: Hints): Hints = {
+      val hintKeysToRemove = hints.toMap.keySet.diff(toSet)
+      hintKeysToRemove.foldLeft(hints)((all, key) => all.remove(key))
```
Baccata, line 38: Could we solve it without adding a `remove` method to Hints ?

modules/core/src/smithy4s/PassthroughSchematic.scala
Baccata: We could just delete the non-generic arity methods in the schematic interface. I think experience shows that we always use the generic one anyway. 

We would however keep the arity-based methods in the syntax package, to create schemas. They would just all delegate to the generic-arity one.

modules/core/src/smithy4s/syntax/package.scala
```diff
@@ -113,4 +113,10 @@ package object syntax
...
+    def mask(hintMask: HintMask): Schematic[F] = HintMask.mask(s, hintMask)
+  }
+
+  implicit def newTypeToHintKey[A](a: Newtype[A]): Hints.Key[_] = a.key
```
Baccata, line 116: What happens if you return `Hints.Key[A]` here ? Does the generated code still compile ? If it does, we should do that, and maybe move this implicit def into the companion object of Hints.Key

modules/core/src/smithy4s/HintMask.scala
lewisjkl: I guess we could instead construct a new `Hints` using the contents of the old one

modules/core/src/smithy4s/syntax/package.scala
lewisjkl: Cool, I will give that a try

modules/http4s/src/smithy4s/http4s/SimpleProtocolBuilder.scala
```diff
@@ -67,7 +68,8 @@ abstract class SimpleProtocolBuilder[P](codecs: CodecAPI)(implicit
             baseUri,
             service,
             Left(http4sClient),
-            EntityCompiler.fromCodecAPI[F](codecs)
+            EntityCompiler
+              .fromCodecAPI[F](codecs, protocol.hintMask)
```
Baccata, line 72: we need to append the `InputOutput` hint to the mask there, which we create manually

modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -308,9 +326,7 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
     val decl = s"case class $name(${renderArgs(fields)})"
     val imports = fields.foldMap(_.tpe.imports) ++ syntaxImport
     lines(
-      if (
-        hints.contains(Hint.ClientError) || hints.contains(Hint.ServerError)
-      ) {
+      if (hints.contains(Hint.Error)) {
```
Baccata, line 329: :+1:

modules/core/src/smithy4s/PassthroughSchematic.scala
lewisjkl: Yeah, I am fine with that. I just implemented them all in case we decide to go that route.. I agree though that the generic route seems to be sufficient.

modules/core/src/smithy4s/syntax/package.scala
lewisjkl: Doesn't compile like that:
```
type mismatch;
 found   : smithy4s.Hints.Key[a.Type]
 required: smithy4s.Hints.Key[A]
```

lewisjkl: Moved to the companion object though as I think that is better

modules/core/src/smithy4s/http/CodecAPI.scala
```diff
@@ -40,7 +40,7 @@ trait CodecAPI {
     * @param schema the value's schema
     * @return the codec associated to the A value.
     */
-  def compileCodec[A](schema: Schema[A]): Codec[A]
+  def compileCodec[A](schema: Schema[A], hintMask: HintMask): Codec[A]
```
Baccata, line 43: I don't really like this, the application of the hintMask should be an implementation detail of the `CodecAPI` instances I think

Baccata (approved):

modules/http4s/src/smithy4s/http4s/SimpleRestJsonBuilder.scala
```diff
@@ -17,7 +17,11 @@
 package smithy4s
 package http4s
 
+import smithy4s.internals.InputOutput
+
 object SimpleRestJsonBuilder
     extends SimpleProtocolBuilder[smithy4s.api.SimpleRestJson](
-      smithy4s.http.json.codecs
+      smithy4s.http.json.codecs(
+        smithy4s.api.SimpleRestJson.protocol.hintMask ++ HintMask(InputOutput)
```
kubukoz, line 25: **nitpick (if-minor):** Kind of makes me think the standard hints (smithy.api) should always be forwarded no matter what.

kubukoz (approved):

modules/http4s/src/smithy4s/http4s/SimpleRestJsonBuilder.scala
Baccata: Nah : imagine you want to have a json RPC API and also serve it over http/restJson. You need a mechanism to ignore the http binding traits when using the RPC protocol (or re-implement a bespoke schematic)

kubukoz: seems like `smithy.api` is too generic a package then 😅

PR #85, opened 2022-01-28
smithy4s.Hints.Key => smithy4s.ShapeTag

Baccata:
* Extracts the hint key into its own called ShapeTag : it may not
be used just for its hints anymore.
* Creates a `SchemaIndex` that is a map of `ShapeTag` to `Schema`.
* Updates the ProtocolInterface to be constructed with the SchemaIndex
instead, which the HintMask is derived

This is gonna be tremendously useful for the Dynamic work


modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -309,7 +309,7 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
         block(
           s"implicit val protocol: smithy4s.Protocol[$name] = new smithy4s.Protocol[$name]"
         ) {
-          s"def hintMask: smithy4s.HintMask = smithy4s.HintMask($protocolTraits)"
+          s"def schemas: smithy4s.SchemaIndex = smithy4s.SchemaIndex($protocolTraits)"
```
kubukoz, line 312: this looks very promising in general

modules/core/src/smithy4s/SchemaIndex.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  def apply[S](bindings: Binding[_]*): SchemaIndex = {
+    new Impl(bindings.map(b => b.tuple: (ShapeTag[_], Schema[_])).toMap)
+  }
+
+  trait Schematic[F[_]] {
+    def withHints[A](fa: F[A], hints: Hints): F[A]
+  }
+
+  private[smithy4s] class Impl(
```
kubukoz, line 44:
```suggestion
  private[smithy4s] final class Impl(
```

modules/core/src/smithy4s/Hints.scala
```diff
@@ -97,12 +78,12 @@ object Hints {
       s"Hints(${all.map(_.value).mkString(", ")})"
   }
 
-  case class Binding[A](key: Key[A], value: A) {
-    def tuple: (Key[A], Binding[A]) = key -> this
+  case class Binding[A](key: ShapeTag[A], value: A) {
+    def tuple: (ShapeTag[A], Binding[A]) = key -> this
```
kubukoz, line 82: if you wanna be fancy... does this work?

```suggestion
    def tuple: (ShapeTag[A], this.type) = key -> this
```

kubukoz (approved):

modules/core/src/smithy4s/Hints.scala
Baccata: 608c0c1

modules/core/src/smithy4s/SchemaIndex.scala
Baccata: f1d224c

lewisjkl (approved):

PR #98, opened 2022-02-07
fix for imports from sub-packages

lewisjkl:
Fixes an issue where imports were being filtered out if they were for a sub-package.

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/Renderer.scala
@@ -45,7 +45,11 @@ object Renderer {
       val p = s"package ${unit.namespace}"
 
       val allImports =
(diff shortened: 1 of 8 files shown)


modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -57,7 +61,13 @@ object Renderer {
...
     }
 
-    pack :: classes
+    val packageApplicableDecls = unit.declarations.filter {
+      case _: TypeAlias | _: Service => true
+      case _                         => false
+    }
+
+    if (packageApplicableDecls.isEmpty) classes
+    else pack :: classes
```
lewisjkl, line 70: This may not be the preferred way to do this, but without something like this an empty package object is created for the `import_test` package

Baccata (approved):

PR #102, opened 2022-02-08
remove generation of redundant package imports

### LINKED ISSUE #101: Redundant package import
Using Smithy `0.12.1`. Found this as the package import being generated is causing compilation errors because of our project structure. Can work around it but I don't think there's a reason for the package import to be added? Doesn't happen in all cases, seems to be if the namespace being imported entirely contains the other it works fine, otherwise the package import is added.

Example:

`errors.smithy`
```
namespace com.example.smithy.errors

@error("client")
@httpError(404)
structure NotFoundError {
  error: String,
}
```
`example.smithy`
```
namespace com.example.smithy.api

service ExampleService {
  operations: [ExampleOperation],
  errors: [com.example.smithy.errors#NotFoundError],
}

@http(method: "GET", uri: "/example", code: 200)
operation ExampleOperation {
  input: ExampleInput,
  output: ExampleOutput,
  errors: [com.example.smithy.errors#NotFoundError],
}

@input
structure ExampleInput {
  @httpQuery("id")
  id: String
}

@output
structure ExampleOutput {
  id: String
}
```

The generated `ExampleService.scala` has a package import for the namespace with the error models:
```
import ExampleServiceGen.ExampleOperationError
import com.example.smithy.errors
import com.example.smithy.errors.NotFoundError
import smithy4s.syntax._

trait ExampleServiceGen[F[_, _, _, _, _]] {
```

lewisjkl:
Closes #101.

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/Renderable.scala
+++ b/modules/codegen/src/smithy4s/codegen/Renderable.scala
@@ -43,8 +43,7 @@ object Renderable {
 
   implicit def tupleRenderable[A](implicit
       A: Renderable[A]
-  ): Renderable[(String, A)] = (t: (String, A)) =>
-    A.render(t._2).addImport(t._1)
+  ): Renderable[(String, A)] = (t: (String, A)) => A.render(t._2)
 
   implicit def nelRenderable[A](implicit
       A: Renderable[A]
--- a/modules/codegen/src/smithy4s/codegen/Renderable.scala
+++ b/modules/codegen/src/smithy4s/codegen/Renderable.scala
@@ -43,7 +43,8 @@ object Renderable {
 
   implicit def tupleRenderable[A](implicit
       A: Renderable[A]
-  ): Renderable[(String, A)] = (t: (String, A)) => A.render(t._2)
+  ): Renderable[(String, A)] = (t: (String, A)) =>
+    A.render(t._2).addImport(t._1)
 
   implicit def nelRenderable[A](implicit
       A: Renderable[A]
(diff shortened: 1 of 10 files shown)


modules/codegen/src/smithy4s/codegen/Renderable.scala
```diff
@@ -43,8 +43,7 @@ object Renderable {
 
   implicit def tupleRenderable[A](implicit
       A: Renderable[A]
-  ): Renderable[(String, A)] = (t: (String, A)) =>
-    A.render(t._2).addImport(t._1)
+  ): Renderable[(String, A)] = (t: (String, A)) => A.render(t._2)
```
Baccata, line 46: I'm a bit suspicious of this. The LHS of the tuple isn't used anymore, which means that this week's implicit is probably useless and its current usesites should be amended

lewisjkl: That's fair, I will go through and cleanup the usesites

kubukoz: > this week's implicit

:joy:

kubukoz (approved):

modules/codegen/src/smithy4s/codegen/Renderable.scala
Baccata: Lol, that's what happens when I look at github on the phone whilst being sleep deprived 😄

Baccata (approved):

PR #110, opened 2022-02-14
Fix typo in documentation

daddykotex:
Again, a small typo.

I'm not done reading, so there is a chance I find other typos as I read, so if we wait a bit I might be able to add commits to the PR.

### DIFF
--- a/modules/docs/src/02-the-smithy-idl/01-smithy-idl.md
+++ b/modules/docs/src/02-the-smithy-idl/01-smithy-idl.md
@@ -186,7 +186,7 @@ trait HelloServiceGen[P[_, _, _, _, _]]{
 ```
 
 P represents an abstract context against which operations are going
-to run against. The abstact context has 5 type parameters:
+to run. The abstact context has 5 type parameters:
 * input,
 * error,
 * output,
--- a/modules/docs/src/02-the-smithy-idl/01-smithy-idl.md
+++ b/modules/docs/src/02-the-smithy-idl/01-smithy-idl.md
@@ -46,7 +46,7 @@ integer Age
 long Identifier
 ```
 
-These get translated as unboxed type wrappers, or `newtypes`, that look like a case class but do not induce any boxing at runtiime.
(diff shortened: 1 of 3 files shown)


daddykotex:
I've also found that in the installation page for sbt, we suggest: `.settings(libraryDependencies += "com.disneystreaming.smithy4s" %%% "smithy4s-core" % smithy4sVersion.value)` but the triple `%%%` is only available when using project matrix I believe

should I fix that?

Baccata:
> should I fix that?

Yes please

daddykotex:
> > should I fix that?
> 
> Yes please

Done

I've also taken the initiative to do a small rewrite in https://github.com/disneystreaming/smithy4s/pull/110/commits/e6154af8973bb86055b953925e6a0f6c4f305713

What do you think?

modules/docs/src/02-the-smithy-idl/01-smithy-idl.md
```diff
@@ -186,7 +186,7 @@ trait HelloServiceGen[P[_, _, _, _, _]]{
 ```
 
 P represents an abstract context against which operations are going
-to run against. The abstact context has 5 type parameters:
+to run. The abstact context has 5 type parameters:
```
kubukoz, line 189:
```suggestion
to run. The abstract context has 5 type parameters:
```

kubukoz:
> > > should I fix that?
> > 
> > 
> > Yes please
> 
> Done
> 
> I've also taken the initiative to do a small rewrite in [e6154af](https://github.com/disneystreaming/smithy4s/commit/e6154af8973bb86055b953925e6a0f6c4f305713)
> 
> What do you think?

that looks good to me @daddykotex :+1:

kubukoz (approved):

Baccata:
@daddykotex you accidentally broke the docs I'm afraid

daddykotex:
From a comment in : https://github.com/scalameta/mdoc/issues/324#issuecomment-619996779

> This design allows individual compile-only blocks to access non-compile-only variables but not other compile-only blocks

Which would explain what happens, I'll just drop this commit from my PR.

lewisjkl (approved):

Issue #114, opened 2022-02-16
Take in request model directly in server function stubs


vangogh500:
Is there a way for the server function stubs generated by Smithy4s to take in the request model instead of all the parameters individually? For a request structure with many fields it can be a bit unruly and to pass it around we will just end up constructing the request object anyways. I guess it wouldn't be too much of a pain to manually create a trait to do this for me but wondering if there is a feature out of the box.

I think it would be a nice feature to include some flag to enable this. I see it generates the request case class anyways (just doesn't use it) so seems like the tool has all the required pieces already.

PR #122, opened 2022-02-22
Adds a way to get packed inputs in operations

Baccata:
Closes https://github.com/disneystreaming/smithy4s/issues/114

* Adds a way for users to tweak rendering of services methods, and substitute a list of parameters in methods for a single case-class parameter (which smithy4s uses under the hood anyway), corresponding to the smithy shape of the operation's input (when present). This might offer better UX in some occurences.
* Adds documentation

### DIFF
--- a/modules/docs/src/04-codegen/01-customisation.md
+++ b/modules/docs/src/04-codegen/01-customisation.md
@@ -1,0 +1,64 @@
+---
+sidebar_label: Protocols and smithy4s
+title: Protocols and smithy4s
+---
+
+Smithy4s is opinionated in what the generated code look like, there are a few things that can be tweaked.
+
+#### Packed inputs
+
+By default, smithy4s generates methods the parameters of which map to the fields of the input structure of the corresponding operation.
+
+For instance :
+
+```kotlin
+service PackedInputsService {
+  version: "1.0.0",
+  operations: [PackedInputOperation]
+}
+
+operation PackedInputOperation {
+  input: PackedInput,
+}
+
+structure PackedInput {
+    @required
+    a: String,
+    @required
+    b: String
+}
+```
+
+leads to something conceptually equivalent to :
+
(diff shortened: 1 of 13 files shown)


modules/docs/src/04-codegen/01-customisation.md
```diff
@@ -0,0 +1,64 @@
...
+
+```scala
+trait PackedInputServiceGen[F[_]] {
+
+  def packedInputOperation(a: String, b: String) : F[Unit]
+
+}
+```
+
+It is however possible to annotate the service (or operation) definition with the `smithy.meta#packedInputs` trait, in order for the rendered method to contain a single parameter, typed with actual input case class of the operation.
```
lewisjkl, line 42:
```suggestion
It is however possible to annotate the service (or operation) definition with the `smithy4s.meta#packedInputs` trait, in order for the rendered method to contain a single parameter, typed with actual input case class of the operation.
```

lewisjkl (approved):

Baccata:
@lewisjkl, going to bed. Might merging and cutting a patch release ?

lewisjkl:
> @lewisjkl, going to bed. Mind merging and cutting a patch release ?

Sounds good, will do!

modules/docs/src/04-codegen/01-customisation.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+It is however possible to annotate the service (or operation) definition with the `smithy4s.meta#packedInputs` trait, in order for the rendered method to contain a single parameter, typed with actual input case class of the operation.
+
+For instance :
+
+```scala
+use smithy4s.meta#packedInputs
+
+@packedInputs
```
kubukoz, line 49: **praise:** I really like this, and how it doesn't require configuration of the codegen.

Issue #129, opened 2022-03-04
more then 22 fields for function Apply


yisraelU:
a case class with 23 fields was generated for  a smithy structure resulting in an error 
```
functions may not have more than 22 parameters, but 23 given
[error]     XXX.apply
```

Issue #128, opened 2022-03-04
Invalid class name generated


yisraelU:
Need to research if this problem is more generalized.
The current case is a Smithy enum with a colon in value generates a class name without escaping the colon either via backticks or replacement
@enum([
    {
        value: "test:nested:value"
    }])
this generated
case object Test:nested:value extends ...

PR #130, opened 2022-03-07
Rendering fixes

Baccata:
Fixes https://github.com/disneystreaming/smithy4s/issues/129 
Fixes https://github.com/disneystreaming/smithy4s/issues/128

* Renders correct schemas for structures of arity > 22 
* Renders correct names for enum values, even in worst case scenario (purely-symbolic enum value)
* Adds smoke tests to check both fixes

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -130,7 +130,7 @@ lazy val core = projectMatrix
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "metadata.smithy",
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "recursive.smithy",
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "bodies.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "empty.smithy",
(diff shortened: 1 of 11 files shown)


daddykotex (approved):
Interesting. it gives good insight with regards to the Renderer

I was convinced the implementation was more "type safe" in the sense that it wasn't just string concatenating all the way!

Baccata:
> I was convinced the implementation was more "type safe" in the sense that it wasn't just string concatenating all the way!

To an extent, it is, there's an intermediate representation between the smithy model and the string representation that does a pretty decent job at prepping the data for rendering. Then the dsl produces a `List[String]`, which in turn are concatenated at the very end. 

Now the thing is, from experience, using something like Scalameta to get type-checked syntax is extremely counter productive (I have built code generators with Scalameta and it ended up being less maintainable than a string-concatenation DSL). 

Then there's things like [treehugger](https://eed3si9n.com/treehugger), but I find that the code becomes quickly less readable, especially when there's polymorphism involved.

lewisjkl (approved):

PR #133, opened 2022-03-09
Exclude sbt-test files from compile configuration

daddykotex:


daddykotex:
Rewrote the commit so that my user matches whatever CLA enforces

Baccata (approved):

Issue #42, opened 2022-01-12
Use SBT scopes to define dependencies for the codegen task 


Baccata:
I’d like us to have a similar UX to ScalaPB when it comes to defining dependencies for the codegen task in the SBT plugin : https://scalapb.github.io/docs/third-party-protos#there-is-a-library-on-maven-with-the-protos-and-possibly-generated-java-code. 

Right now, we define the dependencies as [maven strings](https://github.com/disneystreaming/smithy4s/blob/main/modules/codegen-plugin/src/smithy4s/codegen/CodegenPlugin.scala#L50-L53) that coursier can just parse. It works great, but it's not really sbt-idiomatic.  

Switching to scoped library dependencies might be better. I'm not sure whether it'd be picked up by Scala-steward, but we should definitely aim for it to be the case.

PR #136, opened 2022-03-11
Add a smithy4s config to import codegen dependencies

daddykotex:
The goal is the following: having codegen dependencies as
regular sbt libraryDependencies will allow tools (like scala-
steward) to work with them

### DIFF
--- a/modules/codegen-plugin/src/smithy4s/codegen/CodegenPlugin.scala
+++ b/modules/codegen-plugin/src/smithy4s/codegen/CodegenPlugin.scala
@@ -75,7 +75,7 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
       Compile / smithy4sOutputDir := (Compile / sourceManaged).value,
       Compile / smithy4sOpenapiDir := (Compile / resourceManaged).value,
       Compile / smithy4sCodegen := cachedSmithyCodegen(Compile).value,
-      Compile / smithy4sCodegenDependencies := Nil: @annotation.nowarn,
+      Compile / smithy4sCodegenDependencies := List.empty: @annotation.nowarn,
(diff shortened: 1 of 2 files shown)


modules/codegen-plugin/src/smithy4s/codegen/CodegenPlugin.scala
```diff
@@ -64,13 +67,15 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
...
+  override def projectConfigurations: Seq[Configuration] = Seq(Smithy4s)
+
   override lazy val projectSettings =
     Seq(
       Compile / smithy4sInputDir := (Compile / sourceDirectory).value / "smithy",
       Compile / smithy4sOutputDir := (Compile / sourceManaged).value,
       Compile / smithy4sOpenapiDir := (Compile / resourceManaged).value,
       Compile / smithy4sCodegen := cachedSmithyCodegen(Compile).value,
-      Compile / smithy4sCodegenDependencies := List.empty,
+      Compile / smithy4sCodegenDependencies := Nil: @annotation.nowarn,
```
lewisjkl, line 78: For my own education, what does using `Nil` rather than `List.empty` get us here?

lewisjkl (approved):

lewisjkl:
Also, this PR closes #42 I believe.

modules/codegen-plugin/src/smithy4s/codegen/CodegenPlugin.scala
daddykotex: absolutely nothing, I'll align with the line below

Baccata (approved):

PR #139, opened 2022-03-14
openapi datetime mods

### LINKED ISSUE #43: Default timestamp format inconsistency between OpenAPI and json codec
The default timestamp format in `SchematicJCodec` [defaults to `DATE_TIME`](https://github.com/disneystreaming/smithy4s/blob/38d51ec3764aa9964b018e14cf4f9152fa793aac/modules/json/src/smithy4s/http/json/SchematicJCodec.scala#L290), but the OpenAPI generator [defaults to EPOCH_SECONDS](https://github.com/disneystreaming/smithy4s/blob/38d51ec3764aa9964b018e14cf4f9152fa793aac/modules/openapi/src/smithy4s/openapi/Smithy4sOpenApiProtocol.scala).

We should pick one (preferably DATE_TIME) and use it consistently within the protocol - I'm not sure if it can be configured in `smithy4s.smithy` or it just needs to be codified (the spec isn't [entirely clear](https://awslabs.github.io/smithy/1.0/spec/core/protocol-traits.html#timestampformat-trait) on what "implicit format" means).

yisraelU:
changed default to date-time for openapi protocol
modified header creator to set to http-date
resolves #43

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,0 +1,6 @@
+-J-XX:MaxMetaspaceSize=4G
+-J-XX:MaxInlineLevel=20
+-J-Xss2m
+-J-Xms512M
+-J-Xmx2G
+-J-XX:ReservedCodeCacheSize=256M
--- a/modules/example/resources/smithy4s.example.ObjectService.json
+++ b/modules/example/resources/smithy4s.example.ObjectService.json
@@ -98,7 +98,11 @@
                         "name": "X-Foo",
                         "in": "header",
                         "schema": {
-                            "$ref": "#/components/schemas/LowHigh"
+                            "type": "string",
+                            "enum": [
+                                "Low",
+                                "High"
+                            ]
                         }
                     }
                 ],
@@ -164,13 +168,6 @@
             },
             "GetObjectOutputPayload": {
                 "type": "string"
-            },
-            "LowHigh": {
-                "type": "string",
-                "enum": [
-                    "Low",
-                    "High"
-                ]
             },
             "NoMoreSpaceResponseContent": {
                 "type": "object",
--- a/modules/example/resources/smithy4s.example.ObjectService.json
+++ b/modules/example/resources/smithy4s.example.ObjectService.json
@@ -98,11 +98,7 @@
                         "name": "X-Foo",
                         "in": "header",
                         "schema": {
-                            "type": "string",
-                            "enum": [
-                                "Low",
-                                "High"
-                            ]
+                            "$ref": "#/components/schemas/LowHigh"
                         }
                     }
                 ],
@@ -168,6 +164,13 @@
             },
             "GetObjectOutputPayload": {
                 "type": "string"
+            },
+            "LowHigh": {
+                "type": "string",
+                "enum": [
+                    "Low",
+                    "High"
+                ]
             },
             "NoMoreSpaceResponseContent": {
                 "type": "object",
--- a/modules/openapi/src/software/amazon/smithy/openapi/fromsmithy/protocols/Smithy4sAbstractRestProtocol.scala
+++ b/modules/openapi/src/software/amazon/smithy/openapi/fromsmithy/protocols/Smithy4sAbstractRestProtocol.scala
@@ -18,7 +18,7 @@ package software.amazon.smithy.openapi.fromsmithy.protocols
 
 import cats.syntax.all._
 import smithy4s.api.UncheckedExamplesTrait
-import software.amazon.smithy.jsonschema.Schema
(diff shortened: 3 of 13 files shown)


modules/openapi/test/resources/foo.smithy
```diff
@@ -29,7 +29,10 @@ structure Person {
   name: String,
 
   @httpHeader("X-Bamtech-Partner")
-  partner: String
+  partner: String,
+
```
lewisjkl, line 33: It'd be good to add test cases for query params, path params, and body as well. Just to make sure that all the dates are ending up in the format they should

modules/openapi/src/software/amazon/smithy/openapi/fromsmithy/protocols/Smithy4sAbstractRestProtocol.scala
```diff
@@ -18,7 +18,7 @@ package software.amazon.smithy.openapi.fromsmithy.protocols
 
 import cats.syntax.all._
 import smithy4s.api.UncheckedExamplesTrait
-import software.amazon.smithy.jsonschema.Schema
+import _root_.software.amazon.smithy.jsonschema.Schema
```
kubukoz, line 21:
```suggestion
import software.amazon.smithy.jsonschema.Schema
```

modules/openapi/src/software/amazon/smithy/openapi/fromsmithy/protocols/Smithy4sAbstractRestProtocol.scala
```diff
@@ -196,11 +191,12 @@ abstract class Smithy4sAbstractRestProtocol[T <: Trait]
       context: Context[_ <: Trait],
       member: MemberShape
   ): Boolean = {
+
     if (
       member
         .getMemberTrait(context.getModel, classOf[TimestampFormatTrait])
         .isPresent
-    ) return false
+    ) return false;
```
kubukoz, line 199:
```suggestion
    ) return false
```

modules/openapi/test/resources/foo.smithy
yisraelU: @lewisjkl sure

Baccata:
You need to regenerate examples : `sbt examples/compile` will do. 

We're checking-in some generated code to make it easier to understand the impact of changes on the code generation pipeline during PR review

modules/example/resources/smithy4s.example.ObjectService.json
```diff
@@ -98,7 +98,11 @@
                         "name": "X-Foo",
                         "in": "header",
                         "schema": {
-                            "$ref": "#/components/schemas/LowHigh"
+                            "type": "string",
+                            "enum": [
+                                "Low",
+                                "High"
+                            ]
```
lewisjkl, line 105: Is this an intentional change to inline this enum? If so, why?

yisraelU: No
I compiled example and received that

yisraelU: or maybe my IDE did some wizardry, and i thought it was compiled output. Either  way should I remove?

Baccata: That's likely a result of some of your changes. Not sure which though, but I'd rather it wasn't inlined.

modules/openapi/src/software/amazon/smithy/openapi/fromsmithy/protocols/Smithy4sAbstractRestProtocol.scala
```diff
@@ -280,9 +277,22 @@ abstract class Smithy4sAbstractRestProtocol[T <: Trait]
           .foreach(param.examples)
       }
       val target = context.getModel.expectShape(member.getTarget)
-      val refSchema = context.inlineOrReferenceSchema(member)
```
Baccata, line 283: The removal of this is likely the reason why we've lost inlining. 

The change of behaviour related to timestamps should not impact the rest of the code. Moreover, I'm not quite sure why you moved the logic in here.

.sbtopts
```diff
@@ -0,0 +1,6 @@
+-J-XX:MaxMetaspaceSize=4G
```
Baccata, line 1: though I'm generally not a fan of .sbtopts because it impacts dev environments (I personally have `SBT_OPTS` set up in my bashrc) , I appreciate you taking the time to set it up considering the complexity of the build and the fact that these might help people who checkout the repo :+1:

Baccata (approved):

PR #140, opened 2022-03-15
Schema overhaul

Baccata:
This is an overhaul of the "schema/schematic" abstraction to make use of a GADT. We're trading off extensibility for ease of implementation for interpreters. In particular, having a GADT at hand allows us to peek into the structure of a schema (and the hints carried at each layer), several layers down, without a need for ad-hoc constructs such as `Hinted` and such. 

It also allows interpreters to delegate to other interpreters when convenient.


kubukoz:
random note, `withHints` is now probably redundant (?) since `addHints` is available directly on Schema

Baccata:
@kubukoz addressed your comments in fd29123

PR #153, opened 2022-03-29
Schema overhaul surjection

Baccata:
This adds the concept of `surjection` to the metamodel. A surjection is the association of partial function and a total function. Thus, it can capture concerns such as validation/refinement. 

The way surjections are modelled forces the association to a specific `ShapeTag` and associated instance. In other words, refinements **have to be associated with smithy traits**. This constraint is there to prevent capture of ad-hoc concerns in schemas that would be impossible to surface in smithy. 

Having surjections in the metamodel makes it easier for decoding schema compilers to be implemented, as the implementor just has to implement the `surjection` method. The trade-off is that it puts more pressure on the construction of the schema, as the logic that performs the validation (ie the partial function) has to be captured in the schema value. To that end, a number of constructs and method, such as `Validator` and `Schema#validated` are surfaced, allowing to pull some validation rules from the implicit scope. 

This is far from perfect, in the sense that the validation rules escape usecases such as the one expose by the dynamic module, because some static information has to be available to summon the validation rules. We will do our best to address this later on.


modules/core/src/smithy4s/internals/Hinted.scala
```diff
@@ -55,6 +55,20 @@ case class Hinted[F[_], A](hints: Hints, make: Hints => F[A]) {
...
 
+  def emap[B](
+      f: A => Either[schema.ConstraintError, B]
+  )(implicit C: Covariant[F]): Hinted[F, B] =
+    Hinted(
+      hints,
+      h =>
+        C.map(make(h))(a =>
+          f(a) match {
+            case Left(constraintError) => throw constraintError
```
kubukoz, line 66: is `throw` definitely safe here? If `F` doesn't suspend (e.g. Either) that might misbehave

Baccata: Dude, it's very much wip xD but yeah it's not the correct location to throw

modules/core/src/smithy4s/http/CodecAPI.scala
```diff
@@ -123,6 +123,15 @@ object CodecAPI {
 
       def writeToArray(value: B): Array[Byte] = self.writeToArray(from(value))
     }
+
+    def xmap[B](to: A => Either[ConstraintError, B], from: B => A): Codec[B] = {
+      // TODO, this is a hack that will be removed when we get around to adapting
```
kubukoz, line 128: what does adapting mean here?

kubukoz (approved):

modules/core/src/smithy4s/http/CodecAPI.scala
Baccata: Forgot to finish the sentence 🤦

Issue #167, opened 2022-04-05
AWS: requests with incorrect Content-Type header


andyjayne:
Related to #163 and #164.

Generated service for [Kinesis](https://github.com/aws/aws-sdk-js-v3/blob/v3.63.0/codegen/sdk-codegen/aws-models/kinesis.json) has requests rejected with the following error:

```
PayloadError(., expected = object, message=Expected JSON object, offset: 0x00000000, buf:
+----------+-------------------------------------------------+------------------+
|          |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f | 0123456789abcdef |
+----------+-------------------------------------------------+------------------+
| 00000000 | 3c 55 6e 6b 6e 6f 77 6e 4f 70 65 72 61 74 69 6f | <UnknownOperatio |
| 00000010 | 6e 45 78 63 65 70 74 69 6f 6e 2f 3e 0a          | nException/>.    |
+----------+-------------------------------------------------+------------------+)
```

Adding curl logging to the client middleware shows the requests are using an incorrect Content-Type header `application/octet-stream` instead of `application/x-amz-json-1.1`.

Here's an example for ListShards operation:

```sh
curl \
  --request POST \
  --url 'https://kinesis.us-east-1.amazonaws.com/' \
  --header 'Authorization: <REDACTED>' \
  --header 'host: kinesis.us-east-1.amazonaws.com' \
  --header 'X-Amz-Date: 20220405T171758Z' \
  --header 'X-Amz-Security-Token: <REDACTED>' \
  --header 'X-Amz-Target: Kinesis_20131202.ListShards' \
  --header 'Content-Length: 66' \
  --header 'Content-Type: application/octet-stream' \
  --data '{"StreamName":"<REDACTED>"}'
```

PR #168, opened 2022-04-05
Fix Content-Type for AWS service requests using http4s client

andyjayne:
Before this change requests were using an incorrect `Content-Type` header `application/octet-stream` rather than using the `Content-Type` header defined by the `AwsProtocol`. This was because `withEntity` on http4s client requests sets the headers according to the implicit `EntityEncoder` and the simple `HttpRequest` body is `Option[Array[Byte]]`.

This changes the request so that the headers from the simple `HttpRequest` take priority over those set by `withEntity`.

Fixes https://github.com/disneystreaming/smithy4s/issues/167

Here's an example for Kinesis ListShards operation:

<details>

<summary>Before</summary>

```sh
curl \
  --request POST \
  --url 'https://kinesis.us-east-1.amazonaws.com/' \
  --header 'Authorization: <REDACTED>' \
  --header 'host: kinesis.us-east-1.amazonaws.com' \
  --header 'X-Amz-Date: 20220405T171758Z' \
  --header 'X-Amz-Security-Token: <REDACTED>' \
  --header 'X-Amz-Target: Kinesis_20131202.ListShards' \
  --header 'Content-Length: 66' \
  --header 'Content-Type: application/octet-stream' \
  --data '{"StreamName":"<REDACTED>"}'
```

</details>

<details>

<summary>After</summary>

```sh
curl \
  --request POST \
  --url 'https://kinesis.us-east-1.amazonaws.com/' \
  --header 'Content-Length: 66' \
  --header 'Authorization: <REDACTED>' \
  --header 'Content-Type: application/x-amz-json-1.1' \
  --header 'host: kinesis.us-east-1.amazonaws.com' \
  --header 'X-Amz-Date: 20220405T174817Z' \
  --header 'X-Amz-Security-Token: <REDACTED>' \
  --header 'X-Amz-Target: Kinesis_20131202.ListShards' \
  --data '{"StreamName":"<REDACTED>"}'
```

</details>

### DIFF
--- a/modules/aws-http4s/src/smithy4s/aws/http4s/AwsHttp4sBackend.scala
+++ b/modules/aws-http4s/src/smithy4s/aws/http4s/AwsHttp4sBackend.scala
@@ -59,13 +59,14 @@ object AwsHttp4sBackend {
         case HttpMethod.PUT    => PUT
         case HttpMethod.DELETE => DELETE
       }
-      req = request.body.foldLeft(
-        Request[F](
-          method = method,
-          uri = endpoint,
-          headers = Headers(headers.map { case (k, v) => Header.Raw(k, v) })
-        )
-      )(_.withEntity(_))
+      req = request.body
+        .foldLeft(
+          Request[F](
+            method = method,
+            uri = endpoint
+          )
+        )(_.withEntity(_))
+        .putHeaders(Headers(headers.map { case (k, v) => Header.Raw(k, v) }))
     } yield req
   }
 


Baccata:
You're on fire ! 

Besides the curl call, does the call actually work ?

Baccata (approved):
Awesome. I'll merge and release. Anton had a pretty decent test suite on baguette that I'd like to port here eventually, which would probably have caught it, but something something priorities

kubukoz:
wow, thanks for this, is the tool you're working on public by chance? I was planning to build something like that.

PR #178, opened 2022-04-13
Support untagged union

daddykotex:
Adds support for untagged union as well as a bit of documentation regarding union encoding/decoding.

### DIFF
--- a/modules/core/src/smithy4s/internals/Hinted.scala
+++ b/modules/core/src/smithy4s/internals/Hinted.scala
@@ -138,21 +138,6 @@ object Hinted {
         f: (Option[H1], Option[H2]) => F[A]
     ): Hinted[F, A] =
       Hinted(Hints(), (h: Hints) => f(h.get[H1], h.get[H2]))
-    def onHintOneOf[H1: ShapeTag, H2: ShapeTag, A](
-        onH1: H1 => F[A],
-        onH2: H2 => F[A],
-        fallback: => F[A]
-    ): Hinted[F, A] =
-      Hinted(
-        Hints(),
-        (h: Hints) => {
-          h.get[H1]
-            .map(onH1)
-            .orElse(h.get[H2].map(onH2))
-            .getOrElse(fallback)
-        }
-      )
-
     def onHint[H: ShapeTag, A](default: H)(f: H => F[A]): Hinted[F, A] =
       Hinted(Hints(), (h: Hints) => f(h.get[H].getOrElse(default)))
     def static[A](fa: F[A]): Hinted[F, A] =
--- a/modules/core/src/smithy4s/internals/Hinted.scala
+++ b/modules/core/src/smithy4s/internals/Hinted.scala
@@ -138,6 +138,21 @@ object Hinted {
         f: (Option[H1], Option[H2]) => F[A]
     ): Hinted[F, A] =
       Hinted(Hints(), (h: Hints) => f(h.get[H1], h.get[H2]))
+    def onHintOneOf[H1: ShapeTag, H2: ShapeTag, A](
+        onH1: H1 => F[A],
+        onH2: H2 => F[A],
+        fallback: => F[A]
+    ): Hinted[F, A] =
+      Hinted(
+        Hints(),
+        (h: Hints) => {
+          h.get[H1]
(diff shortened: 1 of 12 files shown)


modules/protocol/resources/META-INF/smithy/smithy4s.smithy
```diff
@@ -56,5 +56,8 @@ structure UncheckedExample {
 structure uuidFormat {
 }
 
-@trait(selector: "union")
+@trait(selector: "union :not([trait|smithy4s.api#untagged])")
```
Baccata, line 59: 👍

lewisjkl (approved):

modules/core/src/smithy4s/internals/Hinted.scala
```diff
@@ -138,6 +138,21 @@ object Hinted {
         f: (Option[H1], Option[H2]) => F[A]
     ): Hinted[F, A] =
       Hinted(Hints(), (h: Hints) => f(h.get[H1], h.get[H2]))
+    def onHintOneOf[H1: ShapeTag, H2: ShapeTag, A](
```
Baccata, line 141: Interesting. I'll allow it 😄

modules/docs/src/03-protocols/02-simple-rest-json/01-overview.md
```diff
@@ -53,3 +67,83 @@ This protocol and its interpreters, are aware of the following traits provided o
 * [error trait](https://awslabs.github.io/smithy/1.0/spec/core/type-refinement-traits.html#error-trait)
 * [http traits](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html), including **http**, **httpError**, **httpLabel**, **httpHeader**, **httpPayload**, **httpQuery**, **httpPrefixHeaders**, **httpQueryParams**.
 * [timestampFormat trait](https://awslabs.github.io/smithy/1.0/spec/core/protocol-traits.html?highlight=timestampformat#timestampformat-trait)
+
+## Decoding and encoding unions
```
Baccata, line 71: 👍  awesome

Baccata (approved):
Some nitpick regarding name and placement of a method, consider it mergeable after addressed

modules/core/src/smithy4s/internals/Hinted.scala
kubukoz: personally I like to do sth like 

```scala

Hinted[F].from { hints =>
  H1(hints).map(onH1)
    ..orElse(H2(hints).map(onH2))
}
```

but there's still the `from` outside that may not be the most convenient :)

Baccata: Maybe we should generate custom extractors, and allow for : 

``` 
Hinted[F].from { 
    case H1(...) => onH1
    case H2(...) => onH2
} 
```

daddykotex: I tried a top level extractor in the companion object directly but it conflicts with case class default extractors. instead, I went for something that produce this:

```scala
package smithy4s.api

import smithy4s.syntax._

case class Untagged()
object Untagged extends smithy4s.ShapeTag.Companion[Untagged] {
  val id: smithy4s.ShapeId = smithy4s.ShapeId("smithy4s.api", "untagged")

    id,
    smithy.api.Trait(Some("union :not([trait|smithy4s.api#discriminated])"), None, None),
  )

  val schema: smithy4s.Schema[Untagged] = constant(Untagged()).withHints(hints)
  implicit val staticSchema : schematic.Static[smithy4s.Schema[Untagged]] = schematic.Static(schema)
  
  object extractHint { def unapply(h: smithy4s.Hints): Option[Untagged] = h.get[Untagged] }
}
```

wdyt?

daddykotex: see https://github.com/disneystreaming/smithy4s/pull/178/commits/778ac7296b55af18cb292688d73985cccf8f881f

daddykotex: got rid of it in favor of extractors for `ShapeTag.Companion` and `NewType`

Baccata (approved):
Great work

PR #179, opened 2022-04-13
Update/jsoniter scala core 2.13.13

daddykotex:


build.sbt
```diff
@@ -634,14 +639,22 @@ val isCE3 = settingKey[Boolean]("Is the current build using CE3?")
 
 lazy val Dependencies = new {
 
+  object ScalaJS {
+    val scalaJsJavaSecureRandom: Def.Initialize[ModuleID] = Def.setting(
+      ("org.scala-js" %%% "scalajs-java-securerandom" % "1.0.0").cross(
+        CrossVersion.for3Use2_13
```
daddykotex, line 645: I'm not sure if we should add this (I remember seeing that using `for3Use2_13` was not appropriate for libraries)

armanbilge: I recommend using `cats.effect.std.UUIDGen` and avoiding this dep altogether. See:
- https://github.com/typelevel/cats-effect/pull/2945

daddykotex: I will attempt to rewrite, but we depend on cats-effect transitively so I'll have to update the build a bit. In the mean time, FYI @Baccata :

Screen Shot 2022-04-13 at 11 17 15

daddykotex: my latest changes should do it!

modules/tests/src-ce2/UUIDGen.scala
```diff
@@ -0,0 +1,24 @@
...
+ *     https://disneystreaming.github.io/TOST-1.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package cats.effect.std
```
Baccata, line 17: hehehe nice 😄

Baccata (approved):
Approved, as long as the build is green. 

Good solution :)

modules/tests/src-ce2/UUIDGen.scala
daddykotex: fake it till you can make it

PR #185, opened 2022-04-14
Ensure hint are set and used for validation of union

daddykotex:


modules/json/test/src/smithy4s/http/json/SchematicJCodecTests.scala
```diff
@@ -162,6 +170,24 @@ object SchematicJCodecTests extends SimpleIOSuite {
     expect(str == jsonStr)
   }
 
+  pureTest("Valid union values are parsed successfuly") {
+    val jsonStr = """{"checked":"foo"}"""
+    val result = readFromString[CheckedOrUnchecked](jsonStr)
+    expect(result == CheckedOrUnchecked.CheckedCase(CheckedString("foo")))
+  }
+
+  pureTest("Invalid union values fails to parse") {
```
daddykotex, line 179: @Baccata 

I'm not 100% sure I understand the issue because, in bad TDD fashion, I wrote the test after, then checked if the test failed before my fix, but it does not.

to be clear, the test passes w/o the fix, so I'm not sure if it does anything

Baccata: Try moving the constraint from the string shape to one of the alternatives in the union definition in smithy (similarly to how you'd annotate fields in structure shapes)

daddykotex: got it, the issue happens if the annotation is placed _directly_ on a member of the union, the fixes works

Baccata:
Superseded by https://github.com/disneystreaming/smithy4s/pull/186

daddykotex:
Oh well... I forgot to push my changes after your advice. I'm sorry lollll

Baccata:
No worries

PR #189, opened 2022-04-16
update to smithy v2, update enum handling/codegen

lewisjkl:
Not exactly sure why the examples show some of the hints switching order, maybe we should add some sorting to those to make sure they are always the same (or obv we can figure out why they suddenly changed).

I decided to update the way enums were rendered to be based off of EnumShape rather than EnumTrait since EnumTrait is now deprecated.


modules/codegen/src/smithy4s/codegen/v1compat/V1EnumTransformation.scala
```diff
@@ -0,0 +1,75 @@
...
+  * field defined already OR the value matches a certain regex pattern. We want to convert all EnumTraits
+  * to EnumShapes, regardless of what the values are. We are able to do this since we already had logic
+  * that used values as names in the generated code using a toCamelCase function. This transformation
+  * preserves the same logic that existed before, thus causing no changes to the generated code.
+  */
+final class V1EnumTransformation extends ProjectionTransformer {
+  def getName(): String = "V1EnumTransformation"
+
+  @annotation.nowarn("msg=class EnumTrait in package traits is deprecated")
+  def transform(x: TransformContext): Model = {
```
lewisjkl, line 23: One potential con about this transformation is that it drops any traits that existed on the stringshape containing the enumtrait. This makes sense in some cases, such as with a trait that is only allowed on StringShapes and not EnumShapes.. but I imagine it could be problematic too. I'll have to look and see what the smithy core does to address this

modules/example/resources/smithy4s.example.ObjectService.json
```diff
@@ -140,8 +140,7 @@
                         "properties": {
                             "int": {
                                 "type": "number",
-                                "format": "int32",
-                                "nullable": true
+                                "format": "int32"
```
lewisjkl, line 143: Looks like EnumShapes are maybe rendered differently than the old EnumTraits were by our openapi conversion. Going to need to take a look.

lewisjkl: Does this actually have something to do with the changes to get rid of boxed types in smithy?

lewisjkl:
Whew got this rebased so it targets the `smithy2-main` branch (which was forked off of the surjection branch). Tomorrow I'll make some of the other updates we chatted about and get the CI happy again.

lewisjkl:
Added Hint to represent IntEnums, handled in document decoder/encoder. Need to add tests still

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
```diff
@@ -103,11 +138,43 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+          }
+          .toList
+        Enumeration(shape.name, shape.name, values).some
+      }
+
+      override def intEnumShape(shape: IntEnumShape): Option[Decl] = {
+        val values = shape
+          .getEnumValues()
+          .asScala
+          .zipWithIndex
```
Baccata, line 172: in the case of `IntEnum`, the value should become the ordinal, and I guess the `Name` should take the place of the string value.

modules/example/resources/smithy4s.example.ObjectService.json
Baccata: It might be the case indeed

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
```diff
@@ -103,11 +138,43 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+        case _ => getDefault(shape)
+      }
+
+      override def enumShape(shape: EnumShape): Option[Decl] = {
+        val values = shape
+          .getEnumValues()
+          .asScala
+          .zipWithIndex
+          .map { case ((name, value), index) =>
+            EnumValue(value, index, name)
```
Baccata, line 162: If smithy enum values contain traits (I'm not sure it's the case), the traits should be captured here, and rendered in the Scala code

modules/json/src/smithy4s/http/json/SchematicJCodec.scala
```diff
@@ -621,13 +621,52 @@ private[smithy4s] class SchematicJCodec(maxArity: Int)
...
+          case None        => in.enumValueError(str)
+        }
+      }
+
+      def encodeValue(x: A, out: JsonWriter): Unit =
+        out.writeVal(to(x)._1.toInt)
+
+      def decodeKey(in: JsonReader): A = {
+        val str = in.readKeyAsInt().toString
+        fromName.get(str) match {
```
Baccata, line 651: `fromOrdinal` should be used

modules/json/src/smithy4s/http/json/SchematicJCodec.scala
```diff
@@ -621,13 +621,52 @@ private[smithy4s] class SchematicJCodec(maxArity: Int)
...
+          case Some(value) => value
+          case None        => in.enumValueError(str)
+        }
+      }
 
+      def encodeKey(x: A, out: JsonWriter): Unit = out.writeKey(to(x)._1.toInt)
```
Baccata, line 657: `._2`

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
lewisjkl: Okay will make this change. Just to clarify, this will mean that the ordinal will be used to send over the wire whenever the IntEnum hint is present, correct?

lewisjkl: I guess the ordinal will no longer be guaranteed to be an actual ordinal, meaning that the ordinal values may not be sequential and may not start at 0.. seems like that breaks the principle of least surprise, wdyt?

lewisjkl:
Closing in favor of #352.

PR #193, opened 2022-04-19
proto command for testing purposes

yisraelU:
adds command to handle cli input as smithy string and attempts to generate scala code .
2 arguments , input and output directory(defaults to pwd)

p.s. 
i don't like the term `proto`,perhaps `test` would be better, or maybe you have a better term

### DIFF
--- a/modules/codegen-cli/src/smithy4s/codegen/cli/ProtoTypeCommand.scala
+++ b/modules/codegen-cli/src/smithy4s/codegen/cli/ProtoTypeCommand.scala
@@ -1,3 +1,19 @@
+/*
+ *  Copyright 2021 Disney Streaming
+ *
+ *  Licensed under the Tomorrow Open Source Technology License, Version 1.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     https://disneystreaming.github.io/TOST-1.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


Baccata:
Yeah `proto` is not a great name in that it would conflate with the proto IDL.

modules/codegen-cli/src/smithy4s/codegen/cli/ProtoTypeCommand.scala
```diff
@@ -0,0 +1,67 @@
...
+import os.Path
+import smithy4s.codegen.ProtoTypeArgs
+import smithy4s.codegen.cli.Smithy4sCommand.ProtoType
+
+object ProtoTypeCommand {
+
+  import Options._
+
+  val smithyString =
+    Opts.option[String](long = "input", help = "Smithy string ")
```
Baccata, line 32: Don't think this should be an `Opts.option` but rather an `Opts.argument`

Baccata:
I'm cool with the addition of a command, just needs a better name :)

Baccata:
Any thoughts of whether to update or close this ?

yisraelU:
I am going to close this for now, I think I might like something like a repl mode in the future

modules/codegen/src/smithy4s/codegen/CodegenArgs.scala
```diff
@@ -28,3 +28,8 @@ case class CodegenArgs(
     dependencies: List[String],
     transformers: List[String]
 )
+
+case class ProtoTypeArgs(
+                          smithyString:String,
```
kubukoz, line 33: Could we achieve the same outcome by using the `codegen` command with `-` as the input path, and piping the string through stdin?

PR #198, opened 2022-04-21
Better error handling on the client

daddykotex:
1. status matches only one error alternative, we pick that
2. discriminator header is found and match an alternative, we pick taht
3. best effort is applied (described below)

Best effort parsing happens if:

* no errorTypeHeader is found to discriminate
* status code matches multiple error type or no error type

In this case, the following logic happens:

* we try to decode for every union alternative
* first success wins
* order is "smart" (see below)

Ordering of the alternatives depends on the status code. If status
is in the 4xx, then generic client error are placed first, otherwise
it's server ++ client ++ the rest.

Note: if one of the alternative has only optional event, chances
are that parsing will be successful


modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
```diff
@@ -143,29 +191,80 @@ private[smithy4s] class SmithyHttp4sClientEndpointImpl[F[_], Op[_, _, _, _, _],
...
-          metadataPartial <- metadataDecoder.decode(metadata).liftTo[F]
+          metadataPartial <- metadataDecoder.decode(metadata).pure[F]
           bodyPartial <- response.as[BodyPartial[T]]
-        } yield metadataPartial.combine(bodyPartial)
+        } yield metadataPartial.map(_.combine(bodyPartial))
     }
   }
 }
+
+private final class ErrorAltPicker[E](alts: Vector[SchemaAlt[E, _]]) {
```
Baccata, line 231: This doesn't seem to be depending on any http4s-specific construct. Let's move it to core (in package `smithy4s.http` for instance)

Baccata:
Very good. Just need to move some non-http4s-aware code somewhere where it could be more easily re-used for other backends.

modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
daddykotex: I have a dependency on cats for `===`, should I just replace that to `==`?

daddykotex: used `==`

PR #201, opened 2022-04-22
Pull dependencies rather than scanning the build classpath

daddykotex:
TLDR:

- make `protocol` a java lib (move test in another module)
- depend on it and smithy in the model loader (avoid scanning the classpath)
- minor comment fixes, etc.

------

If you have jars in your sbt classpath that contains smithy files, the content is taken into account and codegen will produce code accordingly.

In some context, users can be completely unaware that they're pulling such dependencies and may not understand why codegen has produced such code. 

In other scenario, it straight up break the build because the generated code conflict with some stuff already on the classpath.

We avoid the issue here by using explicit dependency on : `protocol` (local project) and `smithy-model`.

We have to make sure that `protocol` is published locally for all jvm version before running the codegen scripted tests.


build.sbt
```diff
@@ -407,23 +416,27 @@ lazy val codegenPlugin = (projectMatrix in file("modules/codegen-plugin"))
...
   .in(file("modules/protocol"))
-  .jvmPlatform(buildtimejvmScala2Versions, jvmDimSettings)
+  .jvmPlatform(
+    autoScalaLibrary = false,
+    scalaVersions = Seq.empty,
+    settings = jvmDimSettings
+  )
+  .settings(
+    libraryDependencies += Dependencies.Smithy.model,
+    javacOptions ++= Seq("--release", "8")
```
Baccata, line 426: :+1:

Baccata (approved):
Great work 👍

build.sbt
daddykotex: at least, in the JVM 11, `javac -Xlint` did not seem to be a thing

PR #205, opened 2022-05-03
Interpolator

yisraelU:
I believe there are still more use cases to figure out


modules/codegen/src/smithy4s/codegen/Interpolator.scala
```diff
@@ -0,0 +1,33 @@
+package smithy4s.codegen
+
+import cats.implicits._
+object Interpolator {
+  class Interpolator(val sc: StringContext) extends AnyVal {
```
Baccata, line 5: make this an implicit class (and rename it to `LineInterpolator`)

modules/codegen/src/smithy4s/codegen/Interpolator.scala
```diff
@@ -0,0 +1,33 @@
+package smithy4s.codegen
+
+import cats.implicits._
+object Interpolator {
```
Baccata, line 4: rename this to `LineSyntax`

build.sbt
```diff
@@ -348,8 +348,9 @@ lazy val codegen = projectMatrix
       Dependencies.Weaver.cats.value % Test
     ),
     testFrameworks += new TestFramework("weaver.framework.CatsEffect"),
-    scalacOptions := scalacOptions.value
-      .filterNot(Seq("-Ywarn-value-discard", "-Wvalue-discard").contains)
+ scalacOptions := scalacOptions.value
+      .filterNot(Seq("-Ywarn-value-discard", "-Wvalue-discard","-Xfatal-warnings").contains)
```
Baccata, line 352: todo : revert before finalising the PR

modules/codegen/src/smithy4s/codegen/Interpolator.scala
```diff
@@ -0,0 +1,33 @@
...
+  class Interpolator(val sc: StringContext) extends AnyVal {
+    def render(renderables: Renderable.WithValue[_]*): RenderLine = {
```
Baccata, line 6: let's rename this method to `line`

modules/codegen/src/smithy4s/codegen/Renderable.scala
```diff
@@ -64,7 +107,26 @@ object Renderable {
   }
 }
 
+// Models
+case class RenderLine(imports: Set[String], line: String) {
```
Baccata, line 111: Let's rename this to `Line`

modules/codegen/src/smithy4s/codegen/Renderable.scala
```diff
@@ -64,7 +107,26 @@ object Renderable {
...
+  def tupled = (imports, line)
+  def modify(f: String => String) = RenderLine(imports, f(line))
+  def toRenderResult = RenderResult(imports, List(line))
+}
+object RenderLine {
```
Baccata, line 116: A `Line` should have a `Renderable` instance

modules/example/src/smithy4s/example/FooService.scala
```diff
@@ -5,7 +5,7 @@ import smithy4s.syntax._
 trait FooServiceGen[F[_, _, _, _, _]] {
   self =>
 
-  def getFoo() : F[Unit, Nothing, GetFooOutput, Nothing, Nothing]
+  def getFoo() : F[PrimitiveType(Unit),Ref(smithy4s.example,GetFooOutput),List()]
```
Baccata, line 8: Something is obviously off there, the `toString` is being called on `Type` somewhere it shouldn't be

modules/codegen/src/smithy4s/codegen/Renderable.scala
```diff
@@ -33,13 +33,56 @@ object Renderable {
 
   implicit val stringRenderable: Renderable[String] = (s: String) =>
     RenderResult(Set.empty, List(s))
-  implicit def listRenderable[A](implicit
-      A: Renderable[A]
-  ): Renderable[List[A]] = { (l: List[A]) =>
+  def typeRenderable(ns: String): Renderable[Type] = new Renderable[Type] {
```
Baccata, line 36: I think we could make this non-parameterised, and filter out the imports that could be removed "at the end of the world"

modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -338,11 +330,11 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
...
       hints: List[Hint]
   ): RenderResult = {
-    val decl = s"case class $name(${renderArgs(fields)})"
-    val imports = fields.foldMap(_.tpe.imports) ++ syntaxImport
+    val decl = render"case class $name(${renderArgs(fields)})"
+    val imports = decl.imports ++ syntaxImport
     lines(
       if (hints.contains(Hint.Error)) {
-        block(s"$decl extends Throwable") {
+        block(s"${decl.line} extends Throwable") {
```
Baccata, line 337: as you correctly inferred, let's make `block` take a `Line`, and have a method overload that'd take a `String`

modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -594,6 +592,27 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
 
     def renderAlgParams =
       s"$renderInput, $renderError, $renderOutput, $renderStreamedInput, $renderStreamedOutput"
+
+    def renderAlgParamsWithImports(genName: String) = {
```
Baccata, line 596: the implementation of this should look like this : 

``` 
import op._ 
line"$input, $errors, $output, $streamedInput, $streamedOutput"
```

Baccata:
After giving this some thoughts, we should clean the names/terminology. Here's a list of actions : 

- [ ] `RenderLine` => `Line`
- [ ]  `RenderResult` => `Lines` (plural) 
- [ ]  `def render` interpolator would become `def line`
- [ ] `Renderable` => `ToLines` 
- [ ]   We should create a `ToLine` expressing the same thing as `ToLines` but for a single line
- [ ]  We currently have `Renderable.WithValue`. Let's make it more generic, and go with : 

```scala 
case class WithValue[TC[_], A](typeclass: TC[A], value: A) 

object WithValue {
   implicit def to[TC[_], A](value: A)(implicit instance: TC[A]): WithValue[TC, A] =
      WithValue(value, instance)
}
```

This way, we should be able to use `WithValue` for both `ToLines` and `ToLine`

- [ ]  Let's have this non-orphan typeclass instance : `implicit def lineToLines[A : ToLine] : ToLines[A]`
- [ ]  Let's move all possible instances from `ToLines` to `ToLine` 
- [ ]  Let's make a `ToLine` instance for `Type` and `Primitive`
- [ ]  Let's remove `importsAndRender` plainly and simply

modules/example/src/smithy4s/example/FooService.scala
yisraelU: yeah forgot to checkin .

modules/codegen/src/smithy4s/codegen/Renderer.scala
yisraelU: problem I have is to return a Line , I must include the imports for Error , which require the Service Name . The problem is the Operation doesn't have access to the Service name, hence I was passing it in .
Should I leave out the Error imports and add them in via addImports function (which kind of returns us back to this ad hoc style..)

Baccata: The operations (and their error coproducts) are rendered within the service's companion object, so as long as we `import MyService._`  in the file of of the service, it should be okay really

Baccata:
Closing in favour of https://github.com/disneystreaming/smithy4s/pull/209

PR #211, opened 2022-05-12
Add options for rendering case class ADT members directly within union

lewisjkl:
Currently, smithy4s generates unions as sealed traits the members of which are all case classes of size 1, which is great when the smithy member targets a primitive, but sucks when it targets a structure.

The goal of this PR is to provide the users with a way to eliminate the superfluous layer.

This PR includes:

- The smithy definition for the adtMember trait
- A validator to make sure all usages of the adtMember trait are valid
- Updates the renderer to pay attention to and respect the adtMember trait


modules/codegen/src/smithy4s/codegen/IR.scala
```diff
@@ -114,13 +114,18 @@ object Field {
 
 }
 
-case class Alt(name: String, realName: String, tpe: Type, hints: List[Hint])
+case class Alt(
+    name: String,
+    realName: String,
+    tpe: Either[Product, Type],
```
lewisjkl, line 120: I'm definitely open to changing this to a custom ADT if people would prefer.

modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -331,13 +331,20 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
...
     val decl = line"case class $name(${renderArgs(fields)})"
     val imports = syntaxImport
+    val implct = if (makeSchemaImplicit) "implicit " else ""
+    val altLines = altString.map(lines(newline, _)).getOrElse(Lines.empty)
+
     lines(
       if (hints.contains(Hint.Error)) {
-        block(line"${decl} extends Throwable") {
+        val alsoExtend = extendsType.map(t => s" $t").getOrElse("")
+        block(line"${decl} extends Throwable$alsoExtend") {
```
lewisjkl, line 347: I'd like to add an example to make sure this works as expected

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
```diff
@@ -523,7 +535,7 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
         val ref = Type.Ref(shapeId.getNamespace(), shapeId.getName())
         val (name, node) = map.head // unions are encoded as objects
         val alt = union.alts.find(_.name == name).get
-        TypedNode.AltTN(ref, name, NodeAndType(node, alt.tpe))
+        TypedNode.AltTN(ref, name, NodeAndType(node, alt.tpe.toOption.get))
```
lewisjkl, line 538: I need to add some examples that hit this code path and sort out what to do here... Not exactly sure how to get around this other than changing the AltTN type a bit

modules/example/src/smithy4s/example/InStoreOrder.scala
```diff
@@ -0,0 +1,17 @@
+package smithy4s.example
```
lewisjkl, line 1: Need to update so this file is not rendered any more

modules/example/src/smithy4s/example/OrderType.scala
```diff
@@ -0,0 +1,41 @@
...
+
+    val schema: smithy4s.Schema[InStoreOrder] = struct(
+      OrderNumber.schema.required[InStoreOrder]("id", _.id).addHints(smithy.api.Required()),
+      string.optional[InStoreOrder]("locationId", _.locationId),
+    ){
+      InStoreOrder.apply
+    }.withId(id).addHints(hints)
+
+    val alt = schema.oneOf[OrderType]("inStore")
+  }
```
lewisjkl, line 26: Here's the new stuff

modules/protocol-tests/test/src/smithy4s/api/validation/AddMemberTraitValidatorSpec.scala
```diff
@@ -0,0 +1,224 @@
...
+        .severity(Severity.ERROR)
+        .message(
+          "test#struct is improperly referenced from [test#MyStruct2]"
+        )
+        .build()
+    )
+    expect(result == expected)
+  }
+
+  test("return error when structure has no members") {
```
Baccata, line 193: Can you elaborate on why this check is needed ?

modules/codegen/src/smithy4s/codegen/IR.scala
Baccata: Yeah, let's add a custom ADT. I think it should be a three-way thing : 

``` 
sealed trait UnionMember 
object UnionMember {
   case class EmptyProductCase(...)
   case class ProductCase(product: Product) 
   case class TypeCase(tpe: Type) 
}
```

modules/codegen/src/smithy4s/codegen/Renderer.scala
Baccata: Let's make the @adtMember trait conflict with the `@error` trait to avoid this case altogether.

modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -331,13 +331,20 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
       originalName: String,
       fields: List[Field],
       recursive: Boolean,
-      hints: List[Hint]
+      hints: List[Hint],
+      extendsType: Option[String] = None,
```
Baccata, line 335: rename field to `adtParent`, because we should distinguish between that and the future cases for mixins

modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -331,13 +331,20 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
...
+      makeSchemaImplicit: Boolean = true,
+      altString: Option[String] = None
```
Baccata, line 337: rename to `additionalLines` and take a `Lines` parameter

modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -331,13 +331,20 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
       originalName: String,
       fields: List[Field],
       recursive: Boolean,
-      hints: List[Hint]
+      hints: List[Hint],
+      extendsType: Option[String] = None,
+      makeSchemaImplicit: Boolean = true,
```
Baccata, line 336: I think this could be derived from the presence of `adtParent` parameter

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
Baccata: > other than changing the AltTN type 

Let's go ahead and do that

modules/protocol/resources/META-INF/smithy/smithy4s.meta.smithy
```diff
@@ -13,3 +13,7 @@ namespace smithy4s.meta
 
 @trait(selector: ":is(service, operation)")
 structure packedInputs {}
+
+@trait(selector: "structure")
+@idRef(failWhenMissing: true, selector: "union")
+string adtMember
```
Baccata, line 19: we should really add documentation directly inlined in the smithy spec as to what those do

modules/example/src/smithy4s/example/OrderType.scala
```diff
@@ -0,0 +1,41 @@
...
+
+import smithy4s.schema.Schema._
+
+sealed trait OrderType extends scala.Product with scala.Serializable
+object OrderType extends smithy4s.ShapeTag.Companion[OrderType] {
+  val id: smithy4s.ShapeId = smithy4s.ShapeId("smithy4s.example", "OrderType")
+
+  val hints : smithy4s.Hints = smithy4s.Hints.empty
+
+  case class OnlineCase(online: OrderNumber) extends OrderType
```
daddykotex, line 11: following stephen's question, could we get rid of the `Case` part?

modules/protocol-tests/test/src/smithy4s/api/validation/AddMemberTraitValidatorSpec.scala
daddykotex: At first I though that the rule was to ensure that the _union_ had atleast a member, but its not. I generated a sample and it looks like that:

```scala
package smithy4s.example

import smithy4s.schema.Schema._

sealed trait OrderType2 extends scala.Product with scala.Serializable
object OrderType2 extends smithy4s.ShapeTag.Companion[OrderType2] {
  val id: smithy4s.ShapeId = smithy4s.ShapeId("smithy4s.example", "OrderType2")

  val hints : smithy4s.Hints = smithy4s.Hints.empty

  case class NameCase(name: String) extends OrderType2
  case class OtherStoreOrder() extends OrderType2
  object OtherStoreOrder extends smithy4s.ShapeTag.Companion[OtherStoreOrder] {
    val id: smithy4s.ShapeId = smithy4s.ShapeId("smithy4s.example", "OtherStoreOrder")

    val hints : smithy4s.Hints = smithy4s.Hints.empty

    implicit val schema: smithy4s.Schema[OtherStoreOrder] = constant(OtherStoreOrder()).withId(id).addHints(hints)

    val alt = schema.oneOf[OrderType2]("inStore")
  }

  object NameCase {
    val hints : smithy4s.Hints = smithy4s.Hints.empty
    val schema: smithy4s.Schema[NameCase] = bijection(string.addHints(hints), NameCase(_), _.name)
    val alt = schema.oneOf[OrderType2]("name")
  }

  implicit val schema: smithy4s.Schema[OrderType2] = union(
    NameCase.alt,
    OtherStoreOrder.alt,
  ){
    case c : NameCase => NameCase.alt(c)
    case c : OtherStoreOrder => OtherStoreOrder.alt(c)
  }.withId(id).addHints(hints)
}
```

modules/protocol/src/smithy4s/meta/ShapeIdTrait.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.function.BiFunction;
+import software.amazon.smithy.model.FromSourceLocation;
+import software.amazon.smithy.model.SourceLocation;
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.node.StringNode;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.traits.AbstractTrait;
+
+/**
+ * Abstract trait class for traits that contain only a string value.
```
yisraelU, line 29: Do you mean proper shape Id value ?

modules/example/src/smithy4s/example/OrderType.scala
Baccata: I'm opposed to this, until we have a better mechanism for collision detection in place.

modules/protocol-tests/test/src/smithy4s/api/validation/AddMemberTraitValidatorSpec.scala
lewisjkl: Yeah, upon further inspection this isn't necessary. I'll remove it

modules/codegen/src/smithy4s/codegen/IR.scala
lewisjkl: I meant to ask earlier, what is empty product case for? I'm assuming it isn't a product type with zero fields, because that can just be modeled by Product with an empty list of fields, right?

Baccata: Not too sure : when Unit is targeted by a union member we should probably render a case object no matter what, whereas members targeting other empty structures might need a reference to the schema attached to those structures. 

We'll have to give it a think

modules/codegen/src/smithy4s/codegen/IR.scala
```diff
@@ -114,15 +118,31 @@ object Field {
 
 }
 
-case class Alt(name: String, realName: String, tpe: Type, hints: List[Hint])
+sealed trait UnionMember {
+  def update(f: Product => Product)(g: Type => Type): UnionMember = this match {
```
lewisjkl, line 122: What's the better name for this function? I can't think right now.. it is kind of a bimap, but not really. Maybe I could rewrite it as a bimap?

modules/codegen/src/smithy4s/codegen/IR.scala
```diff
@@ -217,6 +237,16 @@ object TypedNode {
...
   }
+  sealed trait AltValueTN[+A] {
+    def map[B](f: A => B): AltValueTN[B] = this match {
+      case ProductAltTN(value) => ProductAltTN(f(value))
+      case TypeAltTN(value)    => TypeAltTN(f(value))
+    }
+  }
+  object AltValueTN {
+    case class ProductAltTN[A](value: A) extends AltValueTN[A]
+    case class TypeAltTN[A](value: A) extends AltValueTN[A]
```
lewisjkl, line 248: Not sure if this is the right approach, but basically I was just giving myself a way to disambiguate between the two inside of the renderer.

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
```diff
@@ -59,13 +60,15 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
 
   def allDecls = allShapes
     .filter(_.getId().getNamespace() == namespace)
-    .map(_.accept(toIRVisitor))
+    .map(_.accept(toIRVisitor(renderAdtMemberStructures = false)))
```
lewisjkl, line 63: This is how I went about preventing the adtMember structures from being rendered at the top level (like they normally are)

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
```diff
@@ -480,9 +525,21 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+            case Some(adtMemberTrait) =>
+              val cId = shape.getId
+              val newNs =
+                cId.getNamespace + "." + adtMemberTrait.getValue.getName
+              shape.asStructureShape.get.toBuilder
+                .id(ShapeId.fromParts(newNs, cId.getName))
+                .build()
+            case _ => shape
+          }
+        }
```
lewisjkl, line 542: Alright.. so this is kind of hackish, so let me know if anyone can think of a better way. Basically, if the shape contains the adtmember trait, then I am rewriting its ShapeId so that it will be moved into the correct location (add the name of the union to the namespace). It works, but 😬

modules/example/src/smithy4s/example/TestString.scala
```diff
@@ -0,0 +1,13 @@
+package smithy4s.example
+
+import smithy4s.Newtype
+import smithy4s.schema.Schema._
+
+object TestString extends Newtype[String] {
+  val id: smithy4s.ShapeId = smithy4s.ShapeId("smithy4s.example", "TestString")
+  val hints : smithy4s.Hints = smithy4s.Hints(
+    smithy4s.example.TestTrait(Some(smithy4s.example.OrderType.InStoreOrder(smithy4s.example.OrderNumber(100), Some("someLocation")))),
```
lewisjkl, line 9: Example of a rendered trait

modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -331,10 +331,14 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
...
       fields: List[Field],
       recursive: Boolean,
-      hints: List[Hint]
+      hints: List[Hint],
+      adtParent: Option[String] = None,
+      additionalLines: Lines = Lines.empty
   ): Lines = {
     val decl = line"case class $name(${renderArgs(fields)})"
     val imports = syntaxImport
+    val implct = if (adtParent.isEmpty) "implicit " else ""
```
kubukoz, line 340: **issue:** the name could be more descriptive about where this implicit ends up, e.g. `schemaImplicit`

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
```diff
@@ -480,9 +525,21 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+        val maybeShape = model
           .getShape(ShapeId.fromParts(ns, name))
           .asScala
+        maybeShape.map { shape =>
+          shape.getTrait(classOf[AdtMemberTrait]).asScala match {
+            case Some(adtMemberTrait) =>
+              val cId = shape.getId
+              val newNs =
+                cId.getNamespace + "." + adtMemberTrait.getValue.getName
+              shape.asStructureShape.get.toBuilder
```
daddykotex, line 537: I see a `.get` I get goosebumps ;)

jokes aside, I think it would be an improvement to use `map / getOrElse { throw RelevantError }`

it would be equivalent (throwing on none), but at least the error message would make some sense

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
```diff
@@ -84,7 +87,10 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
         val rec = isRecursive(shape.getId(), Set.empty)
 
         val hints = traitsToHints(shape.getAllTraits().asScala.values.toList)
-        Product(shape.name, shape.name, shape.fields, rec, hints).some
+        val p = Product(shape.name, shape.name, shape.fields, rec, hints).some
+        if (shape.getTrait(classOf[AdtMemberTrait]).isPresent()) {
```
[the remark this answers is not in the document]
lewisjkl: What would that look like in this case? It seems to me it would be more complex that way, but perhaps I am missing what you mean

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
lewisjkl: If a shape has the AdtMemberTrait, it has to be a structure shape.. I could add logic to throw an error like you say, but it feels unnecessary when that error will never be thrown 🤷

modules/codegen/src/smithy4s/codegen/IR.scala
```diff
@@ -114,15 +118,31 @@ object Field {
...
+    case TypeCase(tpe)        => TypeCase(g(tpe))
+    case ProductCase(product) => ProductCase(f(product))
+  }
+}
+object UnionMember {
+  case class ProductCase(product: Product) extends UnionMember
+  case class TypeCase(tpe: Type) extends UnionMember
+}
+
+case class Alt(
```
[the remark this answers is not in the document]
lewisjkl: I agree, I didn't make it final since none of the other case classes in this file are final.. I felt awkward sticking out like that 😆  Happy to go either way on this

modules/codegen/src/smithy4s/codegen/IR.scala
lewisjkl: I thought that, but the fact that both functions are constrained to the same return type as the input type (e.g. Product => Product) made it feel like less of a map.. idk

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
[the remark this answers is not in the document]
daddykotex: Ahh it's my mistake, I don't know why but I thought the else was `None`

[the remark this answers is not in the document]
lewisjkl: No worries 🙂

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
daddykotex: we're saying that the error will never be thrown (because honestly it most likely never will in the context the *trait exists* to support structure special case), but there are always chances that this happen and an error like "`AdtMemberTrait` should be applied to structure only" is better than `NoSuchElementException`

it's very minor, and to be honest, I don't mind much so I'll trust your judgement, but I've search through a fair share of `no such element` exceptions that I never do that anymore

lewisjkl: Yeah, I mean doesn't hurt to add other than being a bit more verbose, but not enough for me to worry. I'll update so it has a more specific error message

modules/example/src/smithy4s/example/TestString.scala
Baccata: :+1:

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
```diff
@@ -435,10 +443,47 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
         .members()
         .asScala
         .map { member =>
-          (member.getMemberName(), member.tpe, hints(member))
+          val memberTarget =
+            model
+              .getShape(member.getTarget)
+              .get() // member target exists or model is invalid
```
Baccata, line 449: you can call `expectShape` instead

modules/codegen/src/smithy4s/codegen/IR.scala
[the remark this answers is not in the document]
Baccata: @lewisjkl feel free to address in a later PR and apply everywhere, it's not a piece of code that's meant to be used as library really.

Baccata (approved):

PR #213, opened 2022-05-13
Use a simpler mechanism to detect file changes

keynmol:
This greatly simplifies the caching logic and solves the problem I was having - if you run `~compile` on a module with smithy specs, they actually weren't re-generated and re-compiled if you changed a spec.

I've confirmed on my project that this change addresses the problem.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,11 @@ jobs:
               pushRemoteCache_$BUILD_KEY \
               -J-Xmx4G
 
+      - name: Run plugin tests
+        if: matrix.scalaVersion == '2_12' && matrix.scalaPlatform == 'jvm' && matrix.ceVersion == 'CE2'
+        run: |
+          sbt scripted
+
       - name: Run checks
         if: matrix.scalaVersion == '2_13' && matrix.scalaPlatform == 'jvm' && matrix.ceVersion == 'CE2'
         run: |
--- a/modules/codegen-plugin/src/sbt-test/codegen-plugin/custom-settings/project/build.properties
+++ b/modules/codegen-plugin/src/sbt-test/codegen-plugin/custom-settings/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.4.0
+sbt.version=1.5.5
--- a/modules/codegen-plugin/src/smithy4s/codegen/CodegenPlugin.scala
+++ b/modules/codegen-plugin/src/smithy4s/codegen/CodegenPlugin.scala
@@ -83,12 +83,28 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
 
   override def projectConfigurations: Seq[Configuration] = Seq(Smithy4s)
 
+  import sbt.nio.Keys.{fileInputs, fileOutputs}
+
   override lazy val projectSettings =
     Seq(
       Compile / smithy4sInputDir := (Compile / sourceDirectory).value / "smithy",
       Compile / smithy4sOutputDir := (Compile / sourceManaged).value,
       Compile / smithy4sOpenapiDir := (Compile / resourceManaged).value,
-      Compile / smithy4sCodegen := cachedSmithyCodegen(Compile).value,
+      Compile / smithy4sCodegen := cachedSmithyCodegen1(Compile).value,
+      Compile / smithy4sCodegen / fileInputs ++= {
+        Option((Compile / smithy4sInputDir).value.listFiles())
+          .getOrElse(Array.empty)
+          .toSeq
+          .map(_.toGlob)
+      },
+      Compile / smithy4sCodegen / fileOutputs ++=
+        Option(
+          (Compile / smithy4sOutputDir).value
+            .listFiles()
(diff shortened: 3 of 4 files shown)


modules/codegen-plugin/src/sbt-test/codegen-plugin/custom-settings/project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.4.0
+sbt.version=1.5.5
```
keynmol, line 1: I had to do this to run any scripted tests, otherwise it said this:

```
[error] [error] [launcher] error during sbt launcher: java.lang.UnsatisfiedLinkError: Can't load library: /Users/velvetbaldmime/Library/Caches/JNA/temp/jna4456582555411290288.tmp
[error] x codegen-plugin/defaults
sbt.internal.scripted.TestException: {line -1}  Reload for scripted batch execution failed.
  | => cat sbt.scriptedtest.ScriptedTests.$anonfun$runBatchedTests$5(ScriptedTests.scala:237)
	at sbt.scriptedtest.ScriptedTests.$anonfun$runBatchedTests$5$adapted(ScriptedTests.scala:231)
	at sbt.scriptedtest.ScriptedTests.$anonfun$commonRunTest$6(ScriptedTests.scala:325)
	at scala.util.control.Exception$Catch.apply(Exception.scala:228)
	at sbt.scriptedtest.ScriptedTests.commonRunTest(ScriptedTests.scala:324)
	at sbt.scriptedtest.ScriptedTests.$anonfun$runBatchedTests$3(ScriptedTests.scala:241)
	at sbt.scriptedtest.ScriptedTests.runOrHandleDisabled(ScriptedTests.scala:275)
	at sbt.scriptedtest.ScriptedTests.$anonfun$runBatchedTests$1(ScriptedTests.scala:245)
```

Which I think is a known JNA issue on M1 macs in older versions of SBT.

Which also makes me think none of y'all are running scripted tests locally 🤠

.github/workflows/ci.yml
```diff
@@ -65,6 +65,11 @@ jobs:
               pushRemoteCache_$BUILD_KEY \
               -J-Xmx4G
 
+      - name: Run plugin tests
```
keynmol, line 68: I don't think scripted tests were running at all - let me know if I missed it somewhere

modules/codegen-plugin/src/smithy4s/codegen/CodegenPlugin.scala
```diff
@@ -133,44 +139,33 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
...
-    cached(
-      (FilesInfo(inputFiles.map(FileInfo.hash(_)).toSet), dependencies)
-    )
+    val fileChangesReport = (conf / smithy4sCodegen).inputFileChanges
+
+    if (fileChangesReport.hasChanges) {
+
+      import fileChangesReport.*
+
+      val filePaths = modified ++ created ++ unmodified
```
keynmol, line 149: The general idea is simple - if there were _any_ changes, we just reconstruct the "live" set of smithy files (which means unmodified + modified - deleted)

If there were no changes, we need to return the list of output files still - otherwise SBT won't pick them up for future compilation.

modules/codegen-plugin/src/sbt-test/codegen-plugin/custom-settings/project/build.properties
Baccata: Considering the SBT plugin is just a small wrapper over the logic held by the `codegen` module, and that we test that via bootstrapping, I don't really care about the scripted tests that much 😄

Baccata: but thanks anyway

modules/codegen-plugin/src/smithy4s/codegen/CodegenPlugin.scala
Baccata: Would you mind inlining that as a comment in the code ? That's useful information to have when we read the files

Baccata (approved):
I think it all makes sense. Would like @daddykotex to give his review before we merge

modules/codegen-plugin/src/smithy4s/codegen/CodegenPlugin.scala
daddykotex: that's amazing I did not know sbt had `inputFileChanges` along with `fileInputs` and `fileOutputs`

daddykotex (approved):
Great PR

It looks like we're piggy backing on sbt's to handle that for us, that's great

also thanks for adding the scripted test to the CI

daddykotex:
@Baccata could I add the comment and push a commit on top of @keynmol 's PR?

is that something maintainers do?

Baccata:
> @Baccata could I add the comment and push a commit on top of @keynmol 's PR?
> is that something maintainers do? 

Sure, go ahead.

lewisjkl (approved):

PR #215, opened 2022-05-14
Recursively return all the generated files

keynmol:
Use `FileFunction` mechanism to more accurately cache the input and output files of the task, as well as react to file changes for more interactive experience.

This combination _seems_ to remove the issues I was seeing before with Zinc.

### DIFF
--- a/modules/codegen-plugin/src/smithy4s/codegen/CodegenPlugin.scala
+++ b/modules/codegen-plugin/src/smithy4s/codegen/CodegenPlugin.scala
@@ -106,12 +106,12 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
           .map(_.toGlob)
           .toSeq,
       Compile / smithy4sCodegenDependencies := List.empty: @annotation.nowarn,
-      Compile / sourceGenerators += (Compile / smithy4sCodegen).map(
-        _.filter(_.ext == "scala")
-      ),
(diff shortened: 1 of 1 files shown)


keynmol:
I think this needs a bit more work, as I'm getting 

```
[error] 6 |import smithy4s.syntax._
[error]   |       ^^^^^^^^^^^^^^^
[error]   |       value syntax is not a member of smithy4s
```

occasionally when modifying a broken compilation unit..

modules/codegen-plugin/src/smithy4s/codegen/CodegenPlugin.scala
```diff
@@ -172,6 +176,16 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
         .toList
 
       resPaths.map(path => new File(path.toString))
-    } else (conf / smithy4sOutputDir).value.listFiles().toSeq
+    } else {
+      val base = (conf / smithy4sOutputDir).value
+
+      val allOutputFiles = (base ** "*").get()
```
Baccata, line 182: I'm not certain that the logic is correct, considering smithy4s doesn't wipe the output folder between runs. Can't we use a similar mechanism to the one you used, but for outputs ?

Baccata:
@keynmol is this ready ?

keynmol:
I believe so, yes.
It's better than what's currently in main, so that's 1 reason to merge it :)

Baccata (approved):

PR #217, opened 2022-05-18
Rewrite the model loading logic to avoid having to depend on publishL…

Baccata:
The codegen module depends on the protocol module, so the smithy files are present on the classpath no matter what. 

Hardcoding their reference means that our build doesn't have to eagerly `publishLocal`, but also there's one less thing that coursier has to look for.


build.sbt
```diff
@@ -113,6 +113,8 @@ lazy val core = projectMatrix
       "smithy.waiters",
       "smithy4s.api"
     ),
+    genDiscoverModels := true,
+    Test / genDiscoverModels := true,
```
daddykotex, line 117: I had to set that to `true` to make the build pass, but I'm unsure because you had set it to false explictly. So keep an eye on it @Baccata

daddykotex (commented):
I added a few commits, but in general I think it's a no go. We have to set `discoverModels` to true in every instance where we load a smithy model that uses a non-prelude definitions:
* shapes from protocol
* shapes from aws.json

that happens in doc, the codegen plugin, benchmarking, tests, basically everywhere.

All that added complexity and I think the downstream user would have the original problem we solved with https://github.com/disneystreaming/smithy4s/pull/201

I'd be open to reverting #201 and then trying to find another solution (if the publishLocal before codegen is problem enough)

Baccata:
> We have to set discoverModels to true in every instance where we load a smithy model that uses a non-prelude definitions

Yeah, that wasn't acceptable. have a look at the latest state, it's a middle ground that should work okay by enforcing the presence of the smithy4s models, but still allows for tapping into `discoverModels`, which solves the aws models (json) for this project without having to hardcode them.

modules/codegen/src/smithy4s/codegen/ModelLoader.scala
```diff
@@ -138,12 +138,17 @@ object ModelLoader {
       assembler
     }
 
-    def maybeDiscoverModels(
+    def addClasspathModels(
         classLoader: ClassLoader,
         flag: Boolean
     ): ModelAssembler = {
+      val smithy4sResources = List(
```
daddykotex, line 145: Ahhh, I love it. it could even be a Buildinfo settings if we were maniac ;)

daddykotex (approved):
This is much better. I'm happy with what I see. I will have another PR to improve the doc

PR #219, opened 2022-05-19
Improve CLI docs

daddykotex:
We have just a few commands let's have them in the docs for a quick reference. Who knows who it might help.
localhost_4000_01-overview_04-cli md


Baccata (approved):
awesome

PR #225, opened 2022-05-20
adds SchemaVisitor implememtation of MetadataWriter

yisraelU:
WIP
I am requesting review as a check , to make sure that my understanding and implementation is correct before I proceed

### DIFF
--- a/modules/core/src/smithy4s/http/internals/SchemaVisitorMetadataWriter.scala
+++ b/modules/core/src/smithy4s/http/internals/SchemaVisitorMetadataWriter.scala
@@ -1,0 +1,172 @@
+package smithy4s
+package http
+package internals
+
+import smithy4s.http.HttpBinding
+import smithy4s._
+import smithy4s.http.internals.MetaEncode._
+import smithy4s.internals.InputOutput
+import smithy4s.schema.Alt.SchemaAndValue
+import smithy4s.schema._
+
+import java.util.Base64
+
+object SchemaVisitorMetadataWriter extends SchemaVisitor[MetaEncode] { self =>
+
+  override def primitive[P](
+      shapeId: ShapeId,
+      hints: Hints,
+      tag: Primitive[P]
+  ): MetaEncode[P] = {
+    tag match {
+      case Primitive.PBlob =>
+        StringValueMetaEncode[ByteArray](ba =>
+          Base64.getEncoder.encodeToString(ba.array)
+        )
+      case Primitive.PTimestamp =>
+        (
+          hints.get(HttpBinding).map(_.tpe),
+          hints.get(smithy.api.TimestampFormat)
+        ) match {
+          case (_, Some(timestampFormat)) =>
+            StringValueMetaEncode((timestamp: Timestamp) =>
+              timestamp.format(timestampFormat)
+            )
+          case (Some(HttpBinding.Type.QueryType), None) |
+              (Some(HttpBinding.Type.PathType), None) =>
+            // See https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html?highlight=httpquery#httpquery-trait
+            StringValueMetaEncode((timestamp: Timestamp) =>
+              timestamp.format(smithy.api.TimestampFormat.DATE_TIME)
+            )
+          case (Some(HttpBinding.Type.HeaderType), None) =>
+            // See https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html?highlight=httpquery#httpheader-trait
+            StringValueMetaEncode((timestamp: Timestamp) =>
+              timestamp.format(smithy.api.TimestampFormat.HTTP_DATE)
+            )
+          case _ => MetaEncode.empty
+        }
+      case Primitive.PDocument => MetaEncode.empty[P]
+      case Primitive.PUnit     => MetaEncode.empty[P]
+      case _                   => fromToString[P]
+    }
+  }
+
+  override def list[A](
+      shapeId: ShapeId,
+      hints: Hints,
+      member: Schema[A]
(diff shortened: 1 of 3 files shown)


lewisjkl (commented):
Looks like it is going in the right direction to me. Nice work!

yisraelU:
All green ! :)

modules/core/src/smithy4s/http/internals/SchemaVisitorMetadataReader.scala
```diff
@@ -0,0 +1,304 @@
...
+              Timestamp.showFormat(smithy.api.TimestampFormat.HTTP_DATE)
+            // See https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html?highlight=httpquery#httpheader-trait
+            MetaDecode.from(formatString)(str =>
+              Timestamp.parse(str, smithy.api.TimestampFormat.HTTP_DATE)
+            )
+          case (None, None) =>
+            MetaDecode.EmptyMetaDecode
+        }
+      case Primitive.PUnit =>
+        MetaDecode.StructureMetaDecode(
```
daddykotex, line 95: for my knowledge, is this different than `EmptyMetaDecode` ?

modules/core/src/smithy4s/http/internals/SchemaVisitorMetadataWriter.scala
```diff
@@ -0,0 +1,207 @@
...
+              (metadata, t: T) => updateFunction(metadata, t)
+            }
+          }
+          field.leftFolder(folderT)
+        }
+    }
+    val updateFunctions = fields.map(field => encodeField(field)).collect {
+      case Some(updateFunction) =>
+        updateFunction
+    }
```
lewisjkl, line 177:
```suggestion
    val updateFunctions = fields.flatMap(encodeField)
```
Would this be equivalent?

modules/core/src/smithy4s/http/internals/SchemaVisitorMetadataReader.scala
```diff
@@ -0,0 +1,304 @@
...
+          self(schema.addHints(Hints(binding) ++ hints))
+        val update = decoder
+          .updateMetadata(binding, label, field.isOptional, reservedQueries)
+        FieldDecode(label, binding, update)
+      }
+    }
+    val fieldUpdates: Vector[FieldDecode] =
+      fields.map(f => decodeField(f)).collect { case Some(fieldUpdate) =>
+        fieldUpdate
+      }
```
lewisjkl, line 203: Could also be replaced by a flatMap?

modules/core/src/smithy4s/http/internals/SchemaVisitorMetadataWriter.scala
```diff
@@ -0,0 +1,207 @@
...
+        .fromHints(field.label, hints, maybeInputOutput)
+        .map { binding =>
+          val folderT = new Field.LeftFolder[Schema, Metadata] {
+            override def compile[T](
+                label: String,
+                instance: Schema[T]
+            ): (Metadata, T) => Metadata = {
+              val encoder: MetaEncode[T] =
+                self(
+                  instance.addHints(Hints(binding) ++ hints).addHints(binding)
```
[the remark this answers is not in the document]
yisraelU: yes . thanks for catching that

modules/core/src/smithy4s/http/internals/SchemaVisitorMetadataReader.scala
yisraelU: yes, afaict, this is used for cases where the input is Unit

modules/core/src/smithy4s/http/internals/SchemaVisitorMetadataReader.scala
yisraelU: same as above,

daddykotex:
Feel free to mark as ready for review @yisraelU 
Unless we want to wait for @Baccata

modules/core/src/smithy4s/http/internals/SchemaVisitorMetadataReader.scala
```diff
@@ -0,0 +1,304 @@
...
+  }
+
+  override def set[A](
+      shapeId: ShapeId,
+      hints: Hints,
+      member: Schema[A]
+  ): MetaDecode[Set[A]] =
+    self(member) match {
+      case MetaDecode.StringValueMetaDecode(f) =>
+        MetaDecode.StringListMetaDecode[Set[A]] { iterator =>
```
daddykotex, line 126: the comment is related to the original code I guess (because I assume you ported that from the Schematic version), but do we really have to use a mutable `ListBuffer`?

just below, in the `map` implementation, we use `map` and `toMap`. Can't we just do `iterator.map(f).toList` and `iterator.map(f).toSet` in both case?

Baccata: fair point

modules/core/src/smithy4s/http/internals/SchemaVisitorMetadataWriter.scala
Baccata: I can't remember why I did it this way, might be for 2.12 compatibility but I'm not 100% sure. If Jeff's suggestion works on 2.12 feel free to apply it

Baccata:
might as well remove the previous implementations :)

Baccata (approved):

Baccata:
@daddykotex @lewisjkl please review carefully in case I missed anything

lewisjkl (approved):
Looks good!

PR #226, opened 2022-05-20
clarify @adtMember docs

lewisjkl:
The current docs are sort of confusing. Hopefully we can come up with some changes to clarify them. This is a start.


daddykotex (approved):
These changes make sense to me.

PR #234, opened 2022-05-25
Giter 8 template

daddykotex:
While working on porting untagged union support to Open API extension, I wanted to setup a quick project. Now, I've done it quite a few times and I figure a Giter8 template will make this fast thanks to sbt support.

`sbt new file://$PATH_TO_REPO/modules/smithy4s.g8` or even via the Github support and this branch: EDIT: this is not gonna work

The repository name has to end with `g8`: https://github.com/sbt/sbt-giter8-resolver/blob/master/src/main/scala/sbt/Giter8TemplateResolver.scala#L10

So let's not merge this but keep it around to put in it's own repo.


modules/smithy4s.g8/src/main/g8/src/main/scala/com/example/Main.scala
```diff
@@ -0,0 +1,44 @@
...
+  val run = Routes.all
+    .flatMap { routes =>
+      EmberServerBuilder
+        .default[IO]
+        .withPort(port"9000")
+        .withHost(host"localhost")
+        .withHttpApp(routes.orNotFound)
+        .build
+    }
+    .use(_ => IO.never)
```
kubukoz, line 42:
```suggestion
    .useForever
```

PR #236, opened 2022-05-29
Memorize document encoders

plokhotnyuk:
Currently smithy4s spends more than 1/3 of CPU cycles on creation of document encoders during serialization of data structures:
image

This small patch fixes that by storing document encoders locally in codecs and increases throughput of serialization of data structures on ~50%.

### DIFF
--- a/modules/json/src/smithy4s/http/json/SchematicJCodec.scala
+++ b/modules/json/src/smithy4s/http/json/SchematicJCodec.scala
@@ -784,16 +784,18 @@ private[smithy4s] class SchematicJCodec(maxArity: Int)
   ): JCodec[Z] =
     new JCodec[Z] {
 
-      val documentFields =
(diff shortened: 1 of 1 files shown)


Baccata:
Oh wow, this is a massive mistake on my end, thanks so much for the patch. 

Out of curiosity, does `private[this]` have an impact on performance ?  I'd assume it wouldn't be necessary in anonymous classes

Baccata (approved):

plokhotnyuk:
> Out of curiosity, does `private[this]` have an impact on performance ? I'd assume it wouldn't be necessary in anonymous classes

It shouldn't impact much on performance due to in-lining of wrapper methods by JIT compilers but it decreases the level of in-lining that could be measurable in some cases, especially on older JVMs with small `-XX:MaxInlineLevel` values.

PR #237, opened 2022-05-31
More efficient serialization of documents with ASCII keys

plokhotnyuk:
Before:
```
[info] Benchmark                      (size)   Mode  Cnt         Score        Error  Units
[info] GoogleMapsAPIWriting.smithy4s     N/A  thrpt    5     39816.307 ±    983.413  ops/s
[info] NestedStructsWriting.smithy4s     128  thrpt    5    403630.790 ±  10781.265  ops/s
[info] PrimitivesWriting.smithy4s        N/A  thrpt    5   5866595.809 ± 208961.868  ops/s
```
After:
```
[info] Benchmark                      (size)   Mode  Cnt         Score        Error  Units
[info] GoogleMapsAPIWriting.smithy4s     N/A  thrpt    5     46183.832 ±    877.871  ops/s
[info] NestedStructsWriting.smithy4s     128  thrpt    5    460022.818 ±  41012.061  ops/s
[info] PrimitivesWriting.smithy4s        N/A  thrpt    5   6737206.295 ± 516062.120  ops/s
```

### DIFF
--- a/modules/json/src/smithy4s/http/json/SchematicJCodec.scala
+++ b/modules/json/src/smithy4s/http/json/SchematicJCodec.scala
@@ -744,28 +744,50 @@ private[smithy4s] class SchematicJCodec(maxArity: Int)
       field: Field[JCodecMake, Z, A]
   ): (Z, JsonWriter) => Unit = {
     field.fold(new Field.Folder[JCodecMake, Z, (Z, JsonWriter) => Unit] {
-      val jLabel = jsonLabel(field)
       def onRequired[AA](
           label: String,
           instance: JCodecMake[AA],
           get: Z => AA
       ): (Z, JsonWriter) => Unit = {
         val codec = instance.get
-        (z: Z, out: JsonWriter) => {
-          out.writeKey(jLabel)
-          codec.encodeValue(get(z), out)
+        val jLabel = jsonLabel(field)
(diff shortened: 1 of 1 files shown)


Baccata (approved):

modules/json/src/smithy4s/http/json/SchematicJCodec.scala
```diff
@@ -744,28 +744,50 @@ private[smithy4s] class SchematicJCodec(maxArity: Int)
...
           label: String,
           instance: JCodecMake[AA],
           get: Z => AA
       ): (Z, JsonWriter) => Unit = {
         val codec = instance.get
-        (z: Z, out: JsonWriter) => {
-          out.writeKey(jLabel)
-          codec.encodeValue(get(z), out)
+        val jLabel = jsonLabel(field)
+        if (jLabel.forall(JsonWriter.isNonEscapedAscii)) {
```
daddykotex, line 754: interesting: performing the ascii check + ascii write is faster than just writekey

plokhotnyuk: `jLabel.forall(JsonWriter.isNonEscapedAscii)` is called only during creation-time of field handlers, which then are memorized

PR #241, opened 2022-05-31
Improve recursion handling in dynamic schemas

Baccata:
Size of the PR is explained by the inclusion of the dynamodb schema for testing purposes. I'll remove it before we merge.


kubukoz:
It works (although I think model dumping has changed somewhere between 0.12.16 and 0.13.0) :)

PR #247, opened 2022-06-02
Fix dynamic compiler applying field/alt hints at the wrong level

### LINKED ISSUE #244: Dynamically derived clients don't send headers in 0.13.4
I believe a regression was introduced in 0.13.4 (bisect narrows it down to 57b16c218129b6c1f8abe437eac69b7819dbc96f in https://github.com/disneystreaming/smithy4s/pull/225).

The problem appears to be that DynamicSchemaIndex's services no longer send headers. For example, for this service:

```
namespace example

@simpleRestJson
service ExampleService {
  operations: [ExampleOp]
}

@http(method:"GET", uri: "/greeting")
@readonly
operation ExampleOp {
  input: ExampleInput
}

structure ExampleInput {
  @required
  @httpHeader("X-GREETING")
  greeting: String
}
```

sending a request made from this document:

```scala
Document.obj(
  "greeting" -> Document.fromString("test")
)
```

results in the client sending a request that stringifies to:

```
Request(method=GET, uri=http://localhost:9090/greeting, httpVersion=HTTP/1.1, headers=Headers(Content-Length: 2, Content-Type: application/json, Host: localhost:9090))
```

Meanwhile, a static call to `.exampleOp("test")` sends:

```
Request(method=GET, uri=http://localhost:9090/greeting, httpVersion=HTTP/1.1, headers=Headers(X-GREETING: test, Host: localhost:9090))
```

Reproduction: `sbt run` in https://github.com/kubukoz/smithy4s-repro-dynamic-1 . Contains a static call as well (which works fine).

Baccata:
With the recent schema overhaul, fields and alts are supposed to carry hints directly on them. 

Fixes #244

@kubukoz, would you be so kind as to turning your repro repo (which I've tested against) into some unit tests in the dynamic module before we merge this ?

### DIFF
--- a/modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
+++ b/modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
@@ -396,20 +396,22 @@ private[dynamic] object Compiler {
           val lFields = {
             members.zipWithIndex
               .map { case ((label, mShape), index) =>
-                val memberHints = allHints(mShape.traits)
                 val lMemberSchema =
-                  schema(mShape.target).map(_.addHints(memberHints: _*))
-                if (
-                  mShape.traits
-                    .getOrElse(Map.empty)
-                    .contains(IdRef("smithy.api#required"))
-                ) {
-                  lMemberSchema.map(_.required[DynStruct](label, _(index)))
-                } else {
-                  lMemberSchema.map(
-                    _.optional[DynStruct](label, arr => Option(arr(index)))
-                  )
-                }
+                  schema(mShape.target)
+                val lField =
+                  if (
+                    mShape.traits
+                      .getOrElse(Map.empty)
+                      .contains(IdRef("smithy.api#required"))
+                  ) {
+                    lMemberSchema.map(_.required[DynStruct](label, _(index)))
+                  } else {
+                    lMemberSchema.map(
+                      _.optional[DynStruct](label, arr => Option(arr(index)))
+                    )
+                  }
+                val memberHints = allHints(mShape.traits)
+                lField.map(_.addHints(memberHints: _*))
               }
               .toVector
               .sequence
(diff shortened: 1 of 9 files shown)


modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
```diff
@@ -430,8 +432,8 @@ private[dynamic] object Compiler {
                 .map { case ((label, mShape), index) =>
                   val memberHints = allHints(mShape.traits)
                   schema(mShape.target)
-                    .map(_.addHints(memberHints: _*))
```
daddykotex, line 433: does the order make a difference here?

Baccata: It very much does

daddykotex: ahh yeah, in retrospec, my question did not make much sense

modules/dynamic/test/src-jvm/smithy4s/dynamic/PlatformUtils.scala
```diff
@@ -47,6 +47,10 @@ private[dynamic] trait PlatformUtils { self: Utils.type =>
       SModel
         .assembler()
         .addImport(s"./sampleSpecs/$fileName")
+        // this is probably not right
```
Baccata, line 50: I think it is

modules/tests/src/smithy4s/tests/PizzaClientSpec.scala
```diff
@@ -253,11 +258,15 @@ abstract class PizzaClientSpec extends IOSuite {
   }
 
   def router(ref: Compat.Ref[IO, State]) = {
-    def storeAndReturn(key: String, request: Request[IO]): IO[Response[IO]] = {
-      ref
-        .updateAndGet(_.saveRequest(key, request))
-        .flatMap(_.getResponse(key))
-    }
+    def storeAndReturn(key: String, request: Request[IO]): IO[Response[IO]] =
+      // Collecting the whole body eagerly to make sure we don't consume it after closing the connection
```
kubukoz, line 262: Surprised we didn't run into this earlier - I was getting `java.io.IOException: response was disposed` when this was just keeping requests as-is.

modules/dynamic/test/src-jvm/smithy4s/dynamic/PlatformUtils.scala
kubukoz: great, I'll drop the comment :)

kubukoz (approved):

PR #254, opened 2022-06-03
SchemaVisitor JCodec

daddykotex:
1. renamed SchematicJCodec => SchemaVisitorJCodec
2. most code is the same apart from a few changes (listed below)

noticeable change:
    1. untaggedUnion, discriminatedUnion and taggedUnion signature have changed and received all the alternatives rather than first and rest
    2. pre-computed document, various timestamp format codecs
    
I cherry picked @plokhotnyuk changes. But some of their changes are not formatted so you'll find some differences

### DIFF
--- a/modules/aws/src/smithy4s/json/AwsSchematicJCodec.scala
+++ b/modules/aws/src/smithy4s/json/AwsSchematicJCodec.scala
@@ -22,28 +22,37 @@ import com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter
 import smithy4s.Timestamp
 import smithy4s.http.json.Cursor
 import smithy4s.http.json.JCodec
-import smithy4s.http.json.SchematicJCodec
-import smithy4s.internals.Hinted
+import smithy4s.http.json.SchemaVisitorJCodec
+import smithy4s.schema.Primitive
 
 private[aws] object AwsSchematicJCodec
-    extends SchematicJCodec(maxArity = 1024) {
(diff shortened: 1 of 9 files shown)


daddykotex:
I think I got it this time. Two important changes:

* this commit where I replace some `instance.hints` to `.hints` on the field directly: https://github.com/disneystreaming/smithy4s/pull/254/commits/dea13fc230aadb2d34ed1134a7d446046b10bc03
* this commit where I move the hint mask logic at the call site, rather than at the codec definition https://github.com/disneystreaming/smithy4s/pull/254/commits/ec433b620b5b2d89813e2b853aadadf7d9453763

I'm not 100% sure I did the right think in both of these so please review carefully

Also, I accidentally pushed a [WIP: move hintmask to apply](https://github.com/disneystreaming/smithy4s/pull/254/commits/ec433b620b5b2d89813e2b853aadadf7d9453763), I can rebase if you want me too

daddykotex:
I took this as an opportunity to rebase and rearrange my commits. I removed the rename commit, it should be better now

daddykotex:
The diff is garbage still, if you want my opinion.

If you want to do a very careful review, feel free to close, and I will reopen (when I'm back) and use one commit per method of the visitor and I'll try to keep the initial ordering

Baccata:
@daddykotex yeah I think that's probably the best course of action. For many files I'm not fussed, but this one is particular in that a lot of care has been put into it, so keeping the ordering of the methods roughly the same to minimise the diff is probably a good idea. 

Thanks for being understanding :)

Issue #266, opened 2022-06-11
Codegen: redundant map newtype in generated hint


kubukoz:
For the following spec:

```
namespace example

use smithy.test#httpRequestTests
use smithy4s.api#simpleRestJson

service HelloService {
    operations: [SayHello]
}

@endpoint(hostPrefix: "{hostLabel}.prefix.")
@http(method: "POST", uri: "/")
@httpRequestTests([
    {
        id: "say_hello",
        protocol: simpleRestJson,
        params: {
            "hostLabel": "foo",
            "greeting": "Hi",
            "name": "Teddy",
            "query": "Hello there"
        },
        method: "POST",
        host: "example.com",
        resolvedHost: "foo.prefix.example.com",
        uri: "/",
        queryParams: [
            "Hi=Hello%20there"
        ],
        headers: {
            "X-Greeting": "Hi",
        },
        body: "{\"name\": \"Teddy\"}",
        bodyMediaType: "application/json"
    }
])
operation SayHello {
    input: SayHelloInput,
    output: Unit
}

@input
structure SayHelloInput {
    @required
    @hostLabel
    hostLabel: String,

    @httpHeader("X-Greeting")
    greeting: String,

    @httpQuery("Hi")
    query: String,

    name: String
}
```

I'm getting a compilation error:

```
[error] /Users/kubukoz/projects/smithy4s-demo/target/scala-2.13/src_managed/main/example/HelloService.scala:55:259: type mismatch;
[error]  found   : smithy.test.StringMap.Type
[error]  required: Map[String,String]
[error]       smithy.test.HttpRequestTests(List(smithy.test.HttpRequestTestCase("say_hello", "smithy4s.api#simpleRestJson", "POST", "/", Some("example.com"), Some("foo.prefix.example.com"), None, Some(List("Hi=Hello%20there")), None, None, Some(smithy.test.StringMap(Map("X-Greeting" -> "Hi"))), None, None, Some("{\"name\": \"Teddy\"}"), Some("application/json"), Some(smithy4s.Document.obj("hostLabel" -> smithy4s.Document.fromString("foo"),"greeting" -> smithy4s.Document.fromString("Hi"),"name" -> smithy4s.Document.fromString("Teddy"),"query" -> smithy4s.Document.fromString("Hello there"))), None, None, None, None, None))),
[error]  
```

Prettified generated code in question:

```scala
smithy.test.HttpRequestTestCase(
            "say_hello",
[...]

PR #268, opened 2022-06-12
Fix map rendering in traits (non-toplevel values)

Baccata:
Fixes https://github.com/disneystreaming/smithy4s/issues/266


kubukoz:
Mind adding a test?

kubukoz (approved):

PR #269, opened 2022-06-13
Unify ShapeTag across Scala versions

Baccata:
Making ShapeTag invariant is the right call anyway, because it unlocks
some things down the road. The cost of doing that is that coproduct
instances need widening before being passed to `Hints`.


daddykotex (approved):
Can we remove scala 3 implementation because the invariant approach is the one taken in scala 3 by default?

anyhow, I love this

Baccata:
There's a few things that allow for Scala 3 to not need contravariance, including the fact that the implicit resolver seems to be much better at doing the right thing, and kicking-in without redundant type ascriptions.

PR #275, opened 2022-06-15
Fix errors not respecting http bindings semantics

daddykotex:
I'm still working out a test, but if you're confident this fixes it and you want to go ahead and merge i would not mind

#### Oli's comment : 

Fix errors serialising what should be headers in the json body


modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
```diff
@@ -202,7 +203,10 @@ private[smithy4s] class SmithyHttp4sClientEndpointImpl[F[_], Op[_, _, _, _, _],
   ): F[Either[MetadataError, E]] = {
     val schema = oneOf.instance
     val errorMetadataDecoder = Metadata.PartialDecoder.fromSchema(schema)
-    implicit val errorCodec = entityCompiler.compilePartialEntityDecoder(schema)
+    implicit val errorCodec: EntityDecoder[F, BodyPartial[ErrorType]] =
```
daddykotex, line 206: I had to add the type ascription because of scala 3 (I think it's good practice for implicits anyway)

daddykotex:
I did a manual test and I can confirm that the simpler:
```
namespace smithy4s.hello

@simpleRestJson
service HelloWorldService {
  version: "1.0.0",
  operations: [Hello]
}

@http(method: "POST", uri: "/{name}", code: 200)
operation Hello {
  input: Person,
  output: Greeting
}

structure Person {
  @httpLabel
  @required
  name: String,

  @httpQuery("town")
  town: String
}

structure Greeting {
  @required
  message: String,

  @required
  @httpHeader("X-NAME")
  name: Name
}

string Name
```

render `name: Name` as a header correctly with this patch

modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
```diff
@@ -202,7 +203,10 @@ private[smithy4s] class SmithyHttp4sClientEndpointImpl[F[_], Op[_, _, _, _, _],
...
+      entityCompiler.compilePartialEntityDecoder(
+        schema.addHints(InputOutput.Output.widen)
```
kubukoz, line 208: Can we add this in the codegen instead? That's how we do it for input/output types of the operation. The upside is that it's a single place that reaches every interpreter.

(also wondering if `DumpModel` does this so that dynamic models see that trait)

Baccata:
@daddykotex, I've decided to merge your SchemaVisitor PR ahead of this fix, as it only changes package private things.

modules/core/src/smithy4s/http/HttpBinding.scala
```diff
@@ -106,10 +106,14 @@ object HttpBinding extends ShapeTag.Companion[HttpBinding] {
   private[smithy4s] def fromHints(
       field: String,
       fieldHints: Hints,
-      inputOutput: Option[InputOutput]
-  ): Option[HttpBinding] = inputOutput.flatMap {
-    case InputOutput.Input  => fromHintsInput(field, fieldHints)
-    case InputOutput.Output => fromHintsOutput(field, fieldHints)
+      shapeHints: Hints
```
Baccata, line 109: @kubukoz I've changed this. Rendering an `Output` on the error type wouldn't quite work, because what's being serialised currently in our http integrations, is not so much the coproduct, but the individual member. 

Instead, we're relying on the presence of the `Error` hint

modules/json/src/smithy4s/http/json/SchemaVisitorJCodec.scala
```diff
@@ -1122,58 +1121,49 @@ private[smithy4s] class SchemaVisitorJCodec(maxArity: Int)
...
-    nonPayloadStruct(fields, maybeInputOutput)(make, encode)
+    nonPayloadStruct(fields, structHints)(make, encode)
   }
   override def struct[S](
       shapeId: ShapeId,
       hints: Hints,
       fields: Vector[SchemaField[S, _]],
       make: IndexedSeq[Any] => S
   ): JCodec[S] = {
-    (
```
Baccata, line 1133: @daddykotex I've taken advantage of having to change this bit of code in order to reduce the nesting of the various pattern matches

Baccata:
Still need to add unit tests

modules/json/src/smithy4s/http/json/SchemaVisitorJCodec.scala
daddykotex: that's great

daddykotex: these are a few things I'll have to backport to my PR

Baccata: your PR has been merged

modules/tests/src/smithy4s/tests/PizzaSpec.scala
```diff
@@ -141,6 +141,7 @@ abstract class PizzaSpec
         )
       val discriminator = headers.get("X-Error-Type").flatMap(_.headOption)
 
+      expect(headers.get("X-CODE") == Some(List("1"))) &&
```
daddykotex, line 144: oh wow, that's why I was having trouble writing a test

I was like: damn I can't get the headers out of this client

but I was working in `PizzaClientSpec`, I had not even realize there was a similar one that worked with the http4s client directly, fuuuu

modules/json/src/smithy4s/http/json/SchemaVisitorJCodec.scala
daddykotex: lol just realized that as I was looking at the PR (did not do a round of notification lol)

Baccata:
Found why local tests were green : I was editing the copy of the smithy file in the target folder, as opposed to the source one.

daddykotex (commented):
I can't approve this as it's my own PR, but thanks for taking over and as I understand it, it will make things easier to maintain for different integration.

thanks

PR #276, opened 2022-06-16
Error headers fix

Baccata:
Literally same thing as this https://github.com/disneystreaming/smithy4s/pull/275, trying to see whether red tests are due to some caching on GHA ...


Baccata:
Closing, found the problem

PR #277, opened 2022-06-17
Switch weaver to munit in core Module.

Baccata:
This switches weaver to Munit. We attempt to reduce the diff via the
creation of an `expect` object that exposes signatures somewhat similar
to weaver's expectations.

The reason for doing this is the impending cross-building to scala-native

### DIFF
--- a/modules/core/test/src-js/smithy4s/Platform.scala
+++ b/modules/core/test/src-js/smithy4s/Platform.scala
@@ -1,0 +1,6 @@
+package smithy4s
+
+object Platform {
+  def isJS = true
+  def isJVM = false
+}
--- a/build.sbt
+++ b/build.sbt
@@ -114,6 +114,7 @@ lazy val docs =
 lazy val core = projectMatrix
   .in(file("modules/core"))
   .settings(
+    Test / fork := virtualAxes.value.contains(VirtualAxis.jvm),
     allowedNamespaces := Seq(
       "smithy.api",
       "smithy.waiters",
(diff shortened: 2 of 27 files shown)


keynmol (approved):

modules/core/test/src-js/smithy4s/Platform.scala
```diff
@@ -0,0 +1,6 @@
+package smithy4s
+
+object Platform {
```
daddykotex, line 3: minor in this instance: I like when platform or scala version specific implementation like that implement a common interface defined in the global source folder

that way when you work on the specific implementation, you know what you need to implement

Baccata: I tend to agree generally speaking, but, honestly, in this case I'm not bothered.

daddykotex (approved):
Looking good, thanks for the `expect object`, I just hope it's not going to be too hard to reuse in other modules (because it's in core). it's not if all modules already depend on core, but otherwise will add to add a `dependson(core -> "compile->compile;test->test")` or something like that

Baccata:
>  I just hope it's not going to be too hard to reuse in other modules (because it's in core)

Right now we're just gonna be aiming it at core and json (which is already set up exactly like how you mention). If it becomes a problem we'll extract it in its own module

Baccata:
The tests were hanging, so I've finished porting the json module and enabled `Test / fork := true` in the core module, which seems to fix it. It wouldn't surprise me if it was something to do with the hotloading of classes in SBT, but 🤷‍♂️ . If the problem occurs again in forked JVMs, we'll investigate deeper.

PR #279, opened 2022-06-23
Change the Field and Alt model to not hold schemas anymore

Baccata:
Description's in the title : this change is motivated by [these points](https://github.com/disneystreaming/smithy4s/pull/279#issuecomment-1164373975)


Baccata:
@benoitlouy, be advised : I'm making some more changes to our Schema model to revert back to "fields and alternatives do not hold the hints directly" : it's proven more error-prone than having the schema referenced by fields/alternatives carry this responsibility. 

I want to remove obvious footguns before enabling MiMa

kubukoz (approved):

PR #280, opened 2022-06-24
fix imports for newtypes from other packages

lewisjkl:
An issue was discovered where extra imports are added to services under the following conditions:

- Service has an operation which has an input parameter from another namespace
- Said input parameter is a newtype

The reason this happens is because the service unpacks the newtype to directly point at the inner type of the newtype. So once this unpacking happens, the import is no longer needed. This issue is masked in the rendering of structures, for example, because they use the newtype inside the construction of the schema.

This fix is kind of a patch, the real fix is to create the import resolution based on carrying fully qualified name information through the rendering process before cleaning up at the end (as we've discussed we want to do).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -567,7 +567,8 @@ lazy val example = projectMatrix
       "smithy4s.example",
       "smithy4s.example.import_test",
       "smithy4s.example.imp",
-      "smithy4s.example.error"
+      "smithy4s.example.error",
+      "smithy4s.example.common"
     ),
     smithySpecs := Seq(
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "example.smithy",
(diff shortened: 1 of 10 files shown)


kubukoz (approved):

Baccata:
@lewisjkl the bug was a symptom of a deeper problem, and recognising it led to a simpler fix : 

The problem was that the import that should have been computed from "what references do we need to compute a schema" were actually computed from "what references do we need to compute parameters", in a "happy accident" kind of way. We were just accidentally doing the right thing for many usecases. 

The correct fix is therefore to be more precise in the imports that we compute from parameters, by avoiding to pull the newtypes reference **at all**, since the parameters do not need it, but rather, ensure that the rendering of the schemas is accompanied with the correct imports. 

Thank you **very much** for adding an example schema replicating the problem.

PR #282, opened 2022-06-27
Document the design of the Service and Endpoint interfaces

Baccata:
Add a documentation page for the `smithy4s.Service` and the `smithy4s.Endpoint` interfaces.


modules/docs/src/05-design/03-services.md
```diff
@@ -0,0 +1,293 @@
...
+}
+
+def fromNaturalTransformation[Context[_]](run: KVStoreOp ~> Context) = new KVStore[Context]{
+  def put(key: String, value: String) = run(KVStoreOp.Put(key, value))
+  def get(key: String)                = run(KVStoreOp.Get(key))
+  def delete(key: String)             = run(KVStoreOp.Delete(key))
+}
+```
+
+This duality is heavily used by Smithy4s : finally-encoded interfaces are generally more natural to Scala developers, and are better supported in editors (autocompletion, etc). But from an implementation's perspective, the initial, data-base encoding is really interesting, because operations are reified as **datatypes** that can be associated with instances of generic typeclasses : it is possible to abstract over data, it is not possible to abstract over method calls.
```
lewisjkl, line 70:
```suggestion
This duality is heavily used by Smithy4s : finally-encoded interfaces are generally more natural to Scala developers, and are better supported in editors (autocompletion, etc). But from an implementation's perspective, the initial, data-based encoding is really interesting, because operations are reified as **datatypes** that can be associated with instances of generic typeclasses : it is possible to abstract over data, it is not possible to abstract over method calls.
```

modules/docs/src/05-design/03-services.md
```diff
@@ -0,0 +1,293 @@
...
+trait Interface[Context[_, _, _, _, _,]]{
+  def operation1(a: A, b: B) : Context[Input, Error, Output, StreamedInput, StreamedOutput]
+}
+```
+
+Let's address this awkwardness right away, by explaining the rationale behind this seemingly humongous signature :
+
+#### Input
+
+That's the input type of an operation. Typically, a case class that holds fields matching the method parameters. We keep track of it in the output type for several reasons:
```
lewisjkl, line 86:
```suggestion
That's the input type of an operation. Typically, a case class that holds fields matching the method parameters. We keep track of it in the return type for several reasons:
```

Wasn't sure if you meant output of the function or the `Output` parameter.. if that even makes sense 😆

modules/docs/src/05-design/03-services.md
```diff
@@ -0,0 +1,293 @@
...
+
+* In the internal logic of Smithy4s, It prevents having to prematurely shoe-horn kinds into other kinds by means of injection/projection, which helps both implementor and compiler alike
+* It will in handy for the implementation of some pagination-aware interpreters
```
lewisjkl, line 89: I don't have enough knowledge about the subject to understand how this plays in. Some more explanation here could be helpful 🤷

modules/docs/src/05-design/03-services.md
```diff
@@ -0,0 +1,293 @@
...
+
+Rather than polluting all layers of abstraction, we decided to just have the concept of operation be impacted and hold the streamed type in a separate type parameter. This allows for interpreters from various ecosystem to emerge. It also has the quality of allowing users to access the unary component of outputs (ie, data that is communicated in the headers of http responses) without necessarily allocating resources to consume the streamed component of outputs.
+
+NB: at the time of writing this, Smithy4s does not have any streaming-aware interpreter implemented. But streaming is such a fundamental notion in remote interactions, and we had to devise a plan to ensure that third party could decide to implement interpreters without waiting.
```
lewisjkl, line 107:
```suggestion
NB: at the time of writing this, Smithy4s does not have any streaming-aware interpreter implemented. But streaming is such a fundamental notion in remote interactions, and we had to devise a plan to ensure that third parties could decide to implement interpreters without waiting.
```

modules/docs/src/05-design/03-services.md
```diff
@@ -0,0 +1,293 @@
...
+
+  // ...
+}
+```
+
+Implementations of such interfaces are code-generated. This implies that any smithy `Service` shape gets translated as a finally-encoded interface, but also as an intially-encoded `GADT`
+
+## The high-level philosophy of Smithy4s
+
+The goal of Smithy4s is to allow users to derive client stubs and routers in various protocols, by running the generated code (or instances of generated interfaces) in some one-liner functions. To that end, Smithy4s surfaces a number of abstractions (such as `smithy4s.schema.Schema`) that allow for the implementation of (very) polymorphic interpreters. Thse interpreters operate on the generated code, which reflects what the user defines in their smithy Specs.
```
lewisjkl, line 140:
```suggestion
The goal of Smithy4s is to allow users to derive client stubs and routers in various protocols, by running the generated code (or instances of generated interfaces) in some one-liner functions. To that end, Smithy4s surfaces a number of abstractions (such as `smithy4s.schema.Schema`) that allow for the implementation of (very) polymorphic interpreters. These interpreters operate on the generated code, which reflects what the user defines in their smithy Specs.
```

modules/docs/src/05-design/03-services.md
```diff
@@ -0,0 +1,293 @@
...
+
+1. Assuming this method call : `kvstore.get("key")`
+2. turning the method call into a piece of data : `KVStoreOp.Get("key")` using the initially-encoded dual of the `KVStore` interface
+3. Retrieving the Smithy4s `Schemas` (input and output) associated to the `Get` operation
+4. Compiling the schema associated to the input of the `Get` operation into some encoding function : `GetInput => Request`
+5. Running the request through a low-level `Request => Response` function (like an http client)
+6. Running `Get` into some function that gives us its `GetInput` representation
+7. Compiling the schema associated to the output (`GetOutput` ~= `Option[String]`) of the `Get` operation into some decoding function `Response => Output`
+
+So we get `kvstore.get => KVStoreOp.Get => GetInput => Request => Response => GetOutput`, which gives us the full data flow, client side.
```
lewisjkl, line 156: These function signatures here and the server-side are fantastic 👏

modules/docs/src/05-design/03-services.md
```diff
@@ -0,0 +1,293 @@
...
+
+As a result, it is important for smithy4s to expose functions that generically enable the filtering of throwables against the `Error` type parameter of an operation, so that interpreters can intercept errors and apply the correct encoding (dictated via `Schema`) before communicating them back to the caller over the user. Conversely, it is important to expose a function that allows to go from the generic `Error` type parameter to `Throwable`, so that errors received via low-level communication channels can be turnd into `Throwable` at the client call site, in order to populate the relevant error channel when exposing monofunctor semantics.
```
lewisjkl, line 219:
```suggestion
As a result, it is important for smithy4s to expose functions that generically enable the filtering of throwables against the `Error` type parameter of an operation, so that interpreters can intercept errors and apply the correct encoding (dictated via `Schema`) before communicating them back to the caller over the user. Conversely, it is important to expose a function that allows to go from the generic `Error` type parameter to `Throwable`, so that errors received via low-level communication channels can be turned into `Throwable` at the client call site, in order to populate the relevant error channel when exposing monofunctor semantics.
```

lewisjkl (approved):
This is awesome! I mostly just commented with some super minor grammar things, take them or leave them 🙂

PR #285, opened 2022-06-28
Remove smithy files from package

yisraelU:
modified packageSrc mappingv to exclude all files that have a smithy directory in them


build.sbt
```diff
@@ -394,6 +398,7 @@ lazy val dynamic = projectMatrix
   .in(file("modules/dynamic"))
   .dependsOn(core, tests % "test->compile", http4s % "test->compile")
   .settings(
+    Compile / packageSrc / mappings := excludeSmithyFiles((Compile / packageSrc / mappings).value),
```
Baccata, line 401: This is not necessary, because there is no manifest file in this module that would make it possible for the smithy files it contains to be loaded via `discoverModels`

build.sbt
```diff
@@ -27,7 +27,10 @@ ThisBuild / version := {
   if (!sys.env.contains("CI")) "dev"
   else (ThisBuild / version).value
 }
-
+def excludeSmithyFiles(mappings:Seq[(File,String)]) = mappings.collect {
+  case (file, path) if !path.contains("smithy/") || file.ext.contains("smithy") =>
```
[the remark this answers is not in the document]
yisraelU: Interesting, thats good to know. Will do that .

build.sbt
```diff
@@ -367,6 +367,10 @@ lazy val protocol = projectMatrix
     settings = jvmDimSettings
   )
   .settings(
+    Compile / packageSrc / mappings := (Compile / packageSrc / mappings).value.collect {
```
Baccata, line 370: 2 things in a single comment : 

* can you run `scalafmtAll` (format checks are currently not set for *sbt files) 
* this collect could be written as a `filter` (or `filterNot`)

Baccata (approved):

Issue #150, opened 2022-03-26
Provide a way to render Swagger for multiple services


keynmol:
I.e.

```scala
smithy4s.http4s.swagger.docs[IO](HelloService, WorldService, ...)
```

Baccata:
Addressed in 0.14

PR #287, opened 2022-06-30
Improve swagger docs support

daddykotex:
Address: https://github.com/disneystreaming/smithy4s/issues/150

Also address a security issue with the `url` query param that was previously required.

The result looks like:
Screen Shot 2022-06-30 at 10 41 39
Screen Shot 2022-06-30 at 10 41 53
Screen Shot 2022-06-30 at 10 42 03

I'm still unsure about the user experience around `smithy4s.http4s.swagger.multipleDocs[IO](HelloWorldService, PizzaAdminService)(path = "docs")`

Let me know what you think


daddykotex:
I'll mark as draft, just because I would need to update doc but will only do so when we agree on the UI to build the swagger http4s route

Baccata:
This looks great ! To improve the UX, we can use the `partially-applied` pattern : 

```
def docs : PartiallyAppliedDocs = new PartiallyAppliedDocs("docs") 

class PartiallyAppliedDocs(path: Path) {
   def atPath(path: Path) = new PartiallyAppliedDocs(path)
   def apply(first: HasId, rest: HasId*)= ...
} 
```

That wouldn't be breaking the sources of most current usage I think.

modules/http4s-swagger/test/src-ce3/TestCompat.scala
```diff
@@ -28,5 +28,5 @@ trait TestCompat { self: BaseIOSuite =>
   }
 
   def docs(path: String) =
-    Docs[IO](service, path, swaggerUiPath = "swaggerui")
+    Docs.multiple[IO](path, swaggerUiPath = "swaggerui")(service)
```
Baccata, line 31: might as well remove the `multiple` method altogether and just document the new way of wiring things up 😄 



Also we might want to make the `swaggerUiPath` configurable in the PartiallyApplied construct

daddykotex: > Also we might want to make the swaggerUiPath configurable in the PartiallyApplied construct

I started doing it, but I found that cumbersome. I figured, if someone needs to go all the way down and customize the swagger ui, they can use `Docs.build[IO](path, swaggerUiPath = "swaggerui")(service)` like in this `TestCompat`

note `build` is the new name I used instead of multiples`

-----

will update the docs, but not much has changed from a user point of view

daddykotex: I can do that

daddykotex:
Updated docs:

localhost_3000_smithy4s_docs_protocols_simple-rest-json_docs

daddykotex:
wow, chrome screenshot does not like that

v2:
localhost_3000_smithy4s_docs_protocols_simple-rest-json_docs (1)

modules/http4s-swagger/src-jvm-ce2/Compat.scala
```diff
@@ -50,21 +55,19 @@ private[smithy4s] object Compat {
   }
 
   trait DocsCompanion extends SwaggerUiInit {
-    def apply[F[_]](
-        hasId: HasId,
+    def build[F[_]](
```
Baccata, line 58: This method should disappear plainly and simply in favour of "smart-setters" in the `PartiallyApplied` construct.

modules/http4s-swagger/src-jvm-ce3/Compat.scala
```diff
@@ -29,14 +30,20 @@ private[smithy4s] object Compat {
...
-    def docs[F[_]](
-        hasId: HasId,
-        path: String = "docs"
-    )(implicit
-        F: Sync[F]
-    ): HttpRoutes[F] = {
-      val docs = Docs[F](hasId, path)
-      docs.routes
+    def docs: PartiallyAppliedDocs = new PartiallyAppliedDocs("docs")
+    def atPath(path: String): PartiallyAppliedDocs =
```
Baccata, line 34: This method should disappear plainly and simply in favour of "smart-setters" in the `PartiallyApplied` construct.

modules/http4s-swagger/src-jvm-ce3/Compat.scala
```diff
@@ -29,14 +30,20 @@ private[smithy4s] object Compat {
...
+      new PartiallyAppliedDocs(path)
+
+    class PartiallyAppliedDocs(path: String) {
```
Baccata, line 37: This should receive smart setters to virtually "mutate" (read, immutable copy) of the `path` value and the `swaggerUiPath` value. The `apply` method should be the last thing the user call and should create the `HttpRoutes`

modules/http4s-swagger/src-jvm/smithy4s/http4s/swagger/Docs.scala
```diff
@@ -18,46 +18,52 @@ package smithy4s
...
+        case _                 => false
       }
     }
   }
   def routes: HttpRoutes[F] = HttpRoutes.of[F] {
     case r @ GET -> DocPath() if r.uri.query.isEmpty =>
-      PermanentRedirect(
-        Location(Uri.unsafeFromString(s"/$path/index.html?url=/$jsonSpec"))
-      )
+      Found(Location(Uri.unsafeFromString(s"/$path/index.html")))
```
Baccata, line 57: I'm curious (and also somewhat illiterate when it comes to 3xx statuses) : can you tl;dr the "why" of changing the status ?

daddykotex: I changed it because 308 is cached very aggressively (forever) and thus you can't change a redirect afterwards. But I can revert that, I don't have an opinion on the matter. On quick look at MDN suggest that it's mostly for SEO purposes.

In my case, it was annoying because I had run a smithy project locally before, accessed the docs (for say PizzaAdmin) and now my browser was always redirecting me to this old url with pizza spec in it

Baccata: Makes total sense, thanks !

Baccata:
@daddykotex I made some changes to improve the UX a little, but it's great work 👍

Baccata (approved):

build.sbt
```diff
@@ -520,32 +520,6 @@ lazy val `http4s-swagger` = projectMatrix
   )
   .http4sJvmPlatform(allJvmScalaVersions, jvmDimSettings)
 
-lazy val `doc-check` = projectMatrix
```
daddykotex: I crap, I forgot to remove my test module loll

Baccata: no worries :)

daddykotex:
Ahhh, two interesting things:

- it was my first time using the partially applied pattern, seen before, but never implemented it. I found it weird to have `F[_]` on `apply` rather that as a class type parameter. I see that you've replaced the former with the later, that's great
- did not think about moving the blocker as case class member instead of a parameter to the `apply` method. again, good to see you did that

thanks

PR #289, opened 2022-07-01
Scala native support

Baccata:
Adds Scala Native support (for Scala 3 only, to avoid the matrix increasing in size too much for something that is quite bleeding edge). 

The supported modules are : 

* smithy4s-core
* smithy4s-json 
* smithy4s-dynamic 

This PR also amends the dynamic module slightly to make it depend on the smithy-model library (on JVM only), facilitating the loading of dynamic services/schemas directly from a smithy Model. This makes tests a little bit easier to write and reduces the need for sharing some test-utilities across modules, which was leading to a cyclic module dependency that I was too lazy to solve via yet another indirection. 

Additionally, because smithy4s-dynamic is built on native now, I had to remove the dependency to CE/weaver in its test. I wanted to keep the diff smallish, so I ended up making a dummy IO construct (totally not pure) to reduce the noise a little. When CE builds on native, we'll switch back.


modules/dynamic/test/src-jvm/smithy4s/dynamic/Http4sDynamicPizzaClientSpec.scala
```diff
@@ -1,48 +0,0 @@
...
-            JsonIOProtocol
-              .fromJsonIO[PizzaAdminServiceGen, PizzaAdminServiceOperation](
-                JsonIOProtocol.toJsonIO(dynamicClient)(dsi.service)
-              )
-
-          }
-      }
-}
-
-object Http4sDynamicPizzaClientSpec extends smithy4s.tests.PizzaClientSpec {
```
Baccata, line 40: moved this one to the http4s module.

modules/dynamic/test/src/smithy4s/dynamic/DummyIO.scala
```diff
@@ -0,0 +1,39 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.dynamic
+import smithy4s.tests._
+
+object DummyIO {
```
Baccata, line 20: They call me Haxxor, destroyer of common-sense. 

Seriously though, it was easier to do that than turn everything into a `scala.util.Try`

modules/dynamic/test/src/smithy4s/dynamic/KVStoreImpl.scala
```diff
@@ -18,29 +18,25 @@ package smithy4s
...
+import DummyIO._
 
-class KVStoreImpl(ref: Ref[IO, Map[String, String]]) extends KVStore[IO] {
-  def delete(key: String): IO[Unit] = ref
-    .modify[Either[Throwable, Unit]] { map =>
-      map.get(key) match {
-        case None        => (map, Left(KeyNotFoundError(key)))
-        case Some(value) => (map - value, Right(()))
-      }
+class KVStoreImpl(map: scala.collection.mutable.Map[String, String])
```
Baccata, line 23: sue me.

build.sbt
```diff
@@ -423,10 +432,14 @@ lazy val dynamic = projectMatrix
   .jvmPlatform(
     allJvmScalaVersions,
     jvmDimSettings ++ Seq(
-      libraryDependencies += Dependencies.Smithy.model % Test
+      Test / fork := true,
+      // Forwarding cwd to tests
```
daddykotex, line 436: the comment is wrong in the sense that this is now the cwd

modules/dynamic/test/src/smithy4s/dynamic/DummyIO.scala
daddykotex: ahhaha that's clever

daddykotex:
Just asking, would you accept a PR where I update the copyright everywhere, that we would merge to main (and eventually here) so that we reduce the noise in this PR ?

Baccata:
I would certainly

build.sbt
Baccata: I'm not sure I follow. Care to elaborate ?

daddykotex:
> I would certainly

see https://github.com/disneystreaming/smithy4s/pull/292

build.sbt
```diff
@@ -96,6 +97,20 @@ lazy val docs =
     )
     .settings(Smithy4sPlugin.doNotPublishArtifact)
 
+val munitDeps = Def.setting {
+  if (virtualAxes.value.contains(VirtualAxis.native)) {
+    Seq(
+      Dependencies.MunitMilestone.core.value % Test,
```
daddykotex, line 103: what's the reason why we can't switch to the Milestone all together

modules/core/src-jvm-native/smithy4s/Timestamp.scala
```diff
@@ -17,11 +17,11 @@
 package smithy4s
 
 import smithy.api.TimestampFormat
-import smithy4s.Timestamp._
-import java.time._
+// import smithy4s.Timestamp._
```
daddykotex, line 20: can we get rid of the comment here?

project/Smithy4sPlugin.scala
```diff
@@ -445,15 +447,21 @@ object Smithy4sPlugin extends AutoPlugin {
           val platformAxis =
             if (projectId.endsWith(jsSuffix)) {
               projectId = projectId.dropRight(jsSuffix.length)
-
               "js"
-            } else "jvm"
+            } else if (projectId.endsWith(nativeSuffix)) {
+              projectId = projectId.dropRight(nativeSuffix.length)
```
daddykotex, line 452: gotta love the mutation :D

daddykotex (approved):
I'm ok with this PR.

I had one question regarding Munit but it's not a blocker and could be resolved w/ a comment I believe

PR #297, opened 2022-07-08
Optimize serialization for ArraySeq

plokhotnyuk:


daddykotex:
This seems so mechanical that I'm surprised it's not how foreach would be implemented. Is it because using `foreach` is just more expensive than a while loop because of some reason (boxing and what not?)?

daddykotex (approved):
thanks again

PR #302, opened 2022-07-11
Revises the "union" method in SchemaVisitor ...

Baccata:
... to make it easier to write memoised dispatching logic

The `union` method in SchemaVisitor is an obvious footgun, because it makes it a hard to memoise the compilation of schemas for each alternative. 

This PR revises the `union` method to expose a new construct `Alt.Dispatcher`, which does the heavylifting of memoizing the result of compiling schemas, and dispatching the union instance to the correct alternative. 

* Creates a `Alt.Dispatcher` construct that aims at streamlining the
implementation of the `union` methods in SchemaVisitors, which is the
hardest due to the fact that the implementor has to think about memoization.
* Creates an `EncoderK` typeclass (used by Alt.Dispatcher) to generalise
over what is an encoder, and streamline further the implementation of
union encoders.
* Extract a generalisation of SchemaVisitor, named SchemaCompiler, which could
be used to unify Schematic, which @kubukoz really really wants.
* Examplifies the use of `Alt.Dispatcher` in DocumentEncoderSchemaVisitor

This should remove the need for https://github.com/disneystreaming/smithy4s/pull/260


modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
```diff
@@ -45,33 +45,44 @@ import smithy4s.schema.Primitive.PUUID
...
-      def contramap[A, B](fa: DocumentEncoder[A])(
-          f: B => A
-      ): DocumentEncoder[B] = fa.contramap(f)
+  implicit val encoderKInstance: EncoderK[DocumentEncoder, Document] =
+    new EncoderK[DocumentEncoder, Document] {
+      def apply[A](fa: DocumentEncoder[A], a: A): Document = fa.apply(a)
+      def absorb[A](f: A => Document): DocumentEncoder[A] =
+        new DocumentEncoder[A] {
+          def apply: A => Document = f
+          override def canBeKey: Boolean = false
```
Baccata, line 78: this is cheating a little but we're planning on removing that `canBeKey` stuff in a subsequent PR

modules/core/src/smithy4s/http/internals/SchemaVisitorPathEncoder.scala
```diff
@@ -22,10 +22,10 @@ import smithy.api.TimestampFormat
 import smithy4s.http.PathSegment
 import smithy4s.http.PathSegment.{GreedySegment, LabelSegment, StaticSegment}
 import smithy4s.{Hints, Lazy, Refinement, ShapeId}
-import smithy4s.schema.Alt._
 import smithy.api.Http
 
-object SchemaVisitorPathEncoder extends SchemaVisitor[MaybePathEncode] { self =>
```
daddykotex, line 28: good one

Baccata:
> We can eventually rewrite most of our implementation and get a free improvement with it 

There's not that many that actually encode things in this repo, and it might be a little bit less efficient than the raw management that the jsoniter interpreter does.

modules/core/src/smithy4s/schema/Alt.scala
```diff
@@ -39,6 +41,59 @@ object Alt {
...
+  /**
+    * Precompiles an Alt to produce an instance of `G`
+    */
+  trait Precompiler[F[_], G[_]] { self =>
+    def apply[A](label: String, instance: F[A]): G[A]
+    def toPolyFunction[U]: PolyFunction[Alt[F, U, *], G] =
+      new PolyFunction[Alt[F, U, *], G] {
+        def apply[A](fa: Alt[F, U, A]): G[A] = self.apply(fa.label, fa.instance)
+      }
+  }
```
kubukoz, line 53: **suggestion:** some helper for converting from `F ~> G` (e.g. if you want to simply lift a visitor to Precompiler) would be great (it'd just ignore the `label`)

kubukoz: tbh I don't _actually_ have a need for this, so feel free to ignore

kubukoz (approved):

PR #296, opened 2022-07-07
Replace schematic implementation by schema visitor

daddykotex:

PR #311, opened 2022-07-13
DocumentKeyDecoder and DocumentKeyEncoder

daddykotex:
This is an attempt at https://github.com/disneystreaming/smithy4s/pull/296#discussion_r916620199 following @Baccata's initial advice.

I mark it as WIP because I'm not sure of the following:

- I had to implement the surject/biject/lazy for the decoder but not the encoder
- I did not integrate with the SchemaVisitorJCodec which I think I need to do (if I have the right approach)


modules/core/src/smithy4s/internals/KeyDecoder.scala
```diff
@@ -0,0 +1,166 @@
...
+import java.util.UUID
+
+import smithy4s._
+import smithy4s.Document._
+import smithy4s.schema.EnumValue
+import smithy4s.schema.Primitive
+import smithy4s.schema.Primitive._
+import smithy4s.schema.SchemaVisitor
+
+trait KeyDecoder[A] { self =>
```
kubukoz, line 29: **suggestion:** the name/package should probably mention that it's for document decoding

daddykotex: I was not sure because I would like to use it for the SchemaVisitorJCodec

kubukoz: hmm, it deals with the Document type directly though

daddykotex: you're absolutely right, I'll need to rethink that one, my bad

I guess I'll adjust if I need to expand to support the JCodec thingy, or rename if I don't

Baccata: > I was not sure because I would like to use it for the SchemaVisitorJCodec 

Let's rename. The JCodec thing doesn't need it as much. What this solves at the document level is that the logic was very hack-ish, encoding the value as a Document before ensuring it was a String. 

The JCodec construct does not suffer from the same ugliness. We could refactor into a `isSimpleShape` function that'd take a Schema and output whether it's a primitive, or a transformation targeting a primitive, which would allow us to remove the `canBeKey` method on JCodec, but it's much less important.

modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
```diff
@@ -136,28 +126,32 @@ object DocumentEncoderSchemaVisitor extends SchemaVisitor[DocumentEncoder] {
...
-        }
-        DObject(mapBuilder.result())
-      } else {
-        val arrayBuilder = IndexedSeq.newBuilder[Document]
-        keysAndValues.foreach { case (k, v) =>
-          arrayBuilder.+=(DObject(Map("key" -> k, "value" -> v)))
-        }
-        DArray(arrayBuilder.result())
+    from[Map[K, V]] { map =>
+      maybeKeyEncoder match {
```
Baccata, line 132: @daddykotex the pattern should happen before receiving the map : 

```
maybeKeyEncoder match {
   case Some(keyEncoder) => from[Map[K, V]] { ... } 
   case None => from[Map[K, V]] { ... } 
} 
```

Baccata (approved):

PR #330, opened 2022-07-21
Update/swagger UI dist 4.13.0

daddykotex:
Would be superseeding: https://github.com/disneystreaming/smithy4s/pull/329

### DIFF
--- a/modules/http4s-swagger/src-jvm/smithy4s/http4s/swagger/SwaggerInit.scala
+++ b/modules/http4s-swagger/src-jvm/smithy4s/http4s/swagger/SwaggerInit.scala
@@ -22,9 +22,14 @@ import cats.data.NonEmptyList
 import org.http4s._
 import org.http4s.dsl.Http4sDsl
 import org.http4s.headers.`Content-Type`
+import scala.io.Source
 
 object SwaggerInit {
   final case class SwaggerUrl(url: String, name: String)
+
+  final val swaggerInitializerJs = {
+    Source.fromResource("swagger-initializer.js").getLines().mkString("\n")
+  }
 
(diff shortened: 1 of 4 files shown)


Baccata:
I'm not sure I follow why that's needed. Can you elaborate ?

daddykotex:
Well, I tested the new release and it works out of the box. 

But I was afraid the content of `swagger-initializer.js` might have changed, so I head over to the swagger-ui repo and confirm it has not. But this was a manual process and I figured we would need to to it every time an update would be released.

This leads me to this custom task: check if our local copy is in sync with the file in the jars we pull.


-----

The TLDR is: that's not necessary but it makes sure updating the swagger-ui webjar does not break our integration (an integration that can only be tested in a web browser (with javascript & al))

modules/http4s-swagger/src-jvm/smithy4s/http4s/swagger/SwaggerInit.scala
```diff
@@ -22,10 +22,15 @@ import cats.data.NonEmptyList
 import org.http4s._
 import org.http4s.dsl.Http4sDsl
 import org.http4s.headers.`Content-Type`
+import scala.io.Source
 
 object SwaggerInit {
   final case class SwaggerUrl(url: String, name: String)
 
+  final val swaggerInitializerJs = {
```
daddykotex, line 30: this is the part I'm not sure about, there may be a better way to do this

daddykotex:
I think this could make sense as a unit test:

1. a test that pulls the content of the file in question from the jar
2. ensure that it matches the copy we have in our scala code

if not the test fail and we need to manually ensure the update is not breaking and update the test

I'll reopen on the latest PR for swagger-ui dist update

PR #333, opened 2022-07-22
Implement type-refinement 

lewisjkl:
- Add refined trait in smithy4s.meta
- Add example in sampleSpecs
- Add example type correlating to example in sampleSpecs


modules/example/src/smithy4s/example/Age.scala
```diff
@@ -0,0 +1,13 @@
+package smithy4s.example
```
lewisjkl, line 1: This obviously shouldn't be generated, but I didn't do any of the renderer work yet

modules/example/src/smithy4s/example/PersonAge.scala
```diff
@@ -0,0 +1,13 @@
+package smithy4s.example
```
lewisjkl, line 1: This will point at Age rather than Int once the renderer is updated

modules/example/src/smithy4s/example/TestItOut.scala
```diff
@@ -0,0 +1,17 @@
+package smithy4s.example
+
+import smithy4s.schema.Schema._
+
+case class TestItOut(age: Option[Age] = None, personAge: Option[PersonAge] = None)
```
lewisjkl, line 5: This will point to the user defined age type after the renderer is updated

modules/example/src/smithy4s/example/package.scala
```diff
@@ -30,8 +30,10 @@ package object example {
   type StreamedBlob = smithy4s.example.StreamedBlob.Type
   type SomeVector = smithy4s.example.SomeVector.Type
   type SomeValue = smithy4s.example.SomeValue.Type
+  type PersonAge = smithy4s.example.PersonAge.Type
   type TestString = smithy4s.example.TestString.Type
   type ObjectSize = smithy4s.example.ObjectSize.Type
+  type Age = smithy4s.example.Age.Type
```
lewisjkl, line 36: Renderer won't add this line anymore

lewisjkl:
Still have some TODOs that I need to resolve and generally do more testing. Also need to get it working on aggregate shapes like lists, unions, structs, etc. Currently it is just targeting primitives

modules/example/src/smithy4s/example/PersonAge.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import smithy4s.Newtype
+import smithy4s.schema.Schema._
+
+object PersonAge extends Newtype[Int] {
```
lewisjkl, line 6: Ah I did have this working where it was Newtype[Age] earlier, but I must have messed it up with another change I made. I'll resolve tomorrow.

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
```diff
@@ -253,23 +254,59 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+              .flatMap(_.getTrait(classOf[RefinedTrait]).asScala.flatMap {
+                refTrait =>
+                  if (
+                    refTrait.getCanonicalShapeId.asScala.contains(shape.getId)
+                  ) // Only return if is canonical shape
+                    Some(refTrait)
+                  else None
+              })
+          }
+          .headOption // TODO: Currently shapes could have multiple traits that have the refined trait. A validator should be added to prevent this
```
Baccata, line 276: good call :)

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
```diff
@@ -253,23 +254,59 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+            Type
+              .Alias(
+                shape.getId().getNamespace(),
+                shape.getId().getName(),
+                Type.PrimitiveType(primitive)
+              )
+              .some
+          } else Type.PrimitiveType(primitive).some
+
+        getExternalTypeInfo(shape) match {
```
Baccata, line 298: we may want to allow it on collection types too, to enable things like `NonEmptyList` and whatnot.

modules/example/src/smithy4s/example/TestItOut.scala
```diff
@@ -0,0 +1,19 @@
...
+object TestItOut extends smithy4s.ShapeTag.Companion[TestItOut] {
+  val id: smithy4s.ShapeId = smithy4s.ShapeId("smithy4s.example", "TestItOut")
+
+  val hints : smithy4s.Hints = smithy4s.Hints.empty
+
+  implicit val schema: smithy4s.Schema[TestItOut] = struct(
+    Age.underlyingSchema.optional[TestItOut]("age", _.age),
+    PersonAge.underlyingSchema.optional[TestItOut]("personAge", _.personAge),
+    FancyList.underlyingSchema.optional[TestItOut]("fancyList", _.fancyList),
+    Name.underlyingSchema.optional[TestItOut]("name", _.name),
```
lewisjkl, line 15: Wasn't sure whether these should be using the `underlyingSchema` or the `schema`. I couldn't quite tell when it is appropriate to use one over the other.

sampleSpecs/refined.smithy
```diff
@@ -5,18 +5,41 @@ use smithy4s.meta#refined
...
 @refined(
   targetClasspath: "smithy4s.example.refined.Age",
-  providerClasspath: "smithy4s.example.refined.Age.provider",
-  canonicalShape: Age
+  providerClasspath: "smithy4s.example.refined.Age.provider"
 )
 structure ageFormat {}
 
+@trait(selector: "list")
+@refined(
```
Baccata, line 13: let's call it "refinement", because it's not the trait that is refined, it's the type that the trait ends up being applied on.

modules/example/src/smithy4s/example/TestItOut.scala
Baccata: Right, so, I'd like for us to use newtypes by default, and therefore using `schema`. This implies that the fields in `TestItOut` would point to the newtypes instead of the wrapped types. 

In a subsequent PR, we'll add a `@unwrapped` trait in a subsequent PR, that will target members, so that the corresponding fields and referenced types in Scala will be the underlying type (as opposed to the newtype).

sampleSpecs/refined.smithy
Baccata: Addressed in [aad9b81](https://github.com/disneystreaming/smithy4s/pull/333/commits/aad9b813d4ab58e98e89aa63095574d1baa1e313)

modules/example/src/smithy4s/example/TestItOut.scala
Baccata: Addressed in [aad9b81](https://github.com/disneystreaming/smithy4s/pull/333/commits/aad9b813d4ab58e98e89aa63095574d1baa1e313)

PR #343, opened 2022-07-29
add middleware example docs

lewisjkl:
localhost_3000_smithy4s_docs_guides_extract-request-info


Baccata:
Good work !  Let's add a link to Davenport's [fiberlocal](https://github.com/davenverse/fiberlocal/) for people who want to work against abstract effect.

modules/docs/src/06-guides/extract-request-info.md
```diff
@@ -0,0 +1,205 @@
...
+```
+
+See our [getting started documentation](../01-overview/01-intro.md) for instructions on how to use this specification to generate scala code.
+
+### Service Implementation
+
+Let's start by creating a case class that we will use to hold the value of some headers from our request.
+
+```scala mdoc:silent
+case class RequestInfo(contentType: Option[String], userAgent: Option[String])
```
Baccata, line 76: Instead of having `RequestInfo` hold options (though, I'm not sure quite whether the Http4s model forces the presence of those two headers), let's have `IOLocal` hold an `Option[Request[Info]]`, and for the service to receive an `IO[RequestInfo]`. In other words, populated by : 

```scala 
ioLocal.get.flatMap {
   case Some(value) => IO.pure(value)
   case None => IO.raiseError(new IllegalAccessException("Tried to access the value outside of the lifecycle of an http request") 
} 
```

Baccata (approved):

PR #346, opened 2022-08-01
make refinement provider implicit

lewisjkl:
Updates refinement provider to be provided to the refined schema as an implicit.

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/Renderer.scala
@@ -678,8 +678,8 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
             underlyingTpe,
             hint
           ) =>
-        line"${underlyingTpe.schemaRef}.refined(${renderNativeHint(hint)})"
-          .addImport(maybeProviderImport.getOrElse(s"$fqn._"))
+        line"${underlyingTpe.schemaRef}.refined[$fqn](${renderNativeHint(hint)})"
+          .addImport(maybeProviderImport.getOrElse(""))
     }
 
     private def schemaRefP(primitive: Primitive): String = primitive match {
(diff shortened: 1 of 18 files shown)


modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -671,8 +671,15 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
...
-        line"${underlyingTpe.schemaRef}.refined(${renderNativeHint(hint)})($providerFqn)"
+      case Type.ExternalType(
+            _,
+            fqn,
+            maybeProviderImport,
+            underlyingTpe,
+            hint
+          ) =>
+        line"${underlyingTpe.schemaRef}.refined(${renderNativeHint(hint)})"
+          .addImport(maybeProviderImport.getOrElse(s"$fqn._"))
```
lewisjkl, line 682: If I don't add the import for the companion object of the `targetType` explicitly, I get a diverging implicit expansion. I think scala expects the implicit to be in the companion object of the constraint by default and doesn't check the target type companion object by default?

modules/example/src/smithy4s/example/FancyList.scala
```diff
@@ -1,11 +1,12 @@
 package smithy4s.example
 
 import smithy4s.Newtype
+import smithy4s.example.refined.FancyList._
```
lewisjkl, line 4: Example of the import I am referring to in my previous comment

Baccata: See [2ebcd04](https://github.com/disneystreaming/smithy4s/pull/346/commits/2ebcd046176466ced908069e4287221546f371a0)

modules/codegen/src/smithy4s/codegen/Renderer.scala
kubukoz: It might be a matter of annotating some types to get that type ineference to kick in

lewisjkl: Yeah, Oli fixed it actually. See: https://github.com/disneystreaming/smithy4s/pull/346#discussion_r934938553

Baccata (approved):

PR #352, opened 2022-08-02
Update to smithy 2.0; add int enum support

lewisjkl:
Ported over from #189.

*Note: This PR is created against the latest published snapshot of smithy 2.0 (version `1.21.0-rc1`). There have been a few improvements/changes since then that I will point out in the comments below.*


modules/core/src/smithy4s/http/internals/SchemaVisitorMetadataReader.scala
```diff
@@ -142,10 +142,19 @@ private[http] class SchemaVisitorMetadataReader()
...
       total: E => EnumValue[E]
   ): MetaDecode[E] = {
-    MetaDecode
-      .from(
-        s"Enum[${values.map(_.stringValue).mkString(",")}]"
-      )(string => values.find(_.stringValue == string).map(_.value))
+    if (hints.get[IntEnum].isDefined) {
+      MetaDecode
+        .from(
+          s"Enum[${values.map(_.stringValue).mkString(",")}]"
```
lewisjkl, line 148: Not sure if we want to be using the `stringValue` in cases like this.. probably not, but that's what I had in the other PR. The ordinal would look a bit more confusing, but would be more relevant

modules/example/src/smithy4s/example/AgeFormat.scala
```diff
@@ -7,7 +7,7 @@ object AgeFormat extends smithy4s.ShapeTag.Companion[AgeFormat] {
   val id: smithy4s.ShapeId = smithy4s.ShapeId("smithy4s.example", "ageFormat")
 
   val hints : smithy4s.Hints = smithy4s.Hints(
-    smithy.api.Trait(Some("integer"), None, None, None),
+    smithy.api.Trait(Some("integer"), None, None),
```
lewisjkl, line 10: This change is caused by the difference between `1.22.0` and `1.21.0-rc1`. Once we update to smithy 2 (the real one), this will change back.

modules/example/src/smithy4s/example/ArbitraryData.scala
```diff
@@ -7,7 +7,7 @@ import smithy4s.schema.Schema._
 object ArbitraryData extends Newtype[Document] {
   val id: smithy4s.ShapeId = smithy4s.ShapeId("smithy4s.example", "arbitraryData")
   val hints : smithy4s.Hints = smithy4s.Hints(
-    smithy.api.Trait(None, None, None, None),
+    smithy.api.Trait(None, None, None),
```
lewisjkl, line 10: See https://github.com/disneystreaming/smithy4s/pull/352/files#r935968506

modules/example/src/smithy4s/example/FancyListFormat.scala
```diff
@@ -7,7 +7,7 @@ object FancyListFormat extends smithy4s.ShapeTag.Companion[FancyListFormat] {
   val id: smithy4s.ShapeId = smithy4s.ShapeId("smithy4s.example", "fancyListFormat")
 
   val hints : smithy4s.Hints = smithy4s.Hints(
-    smithy.api.Trait(Some("list:test(> member > string)"), None, None, None),
+    smithy.api.Trait(Some("list:test(> member > string)"), None, None),
```
lewisjkl, line 10: See https://github.com/disneystreaming/smithy4s/pull/352/files#r935968506

modules/example/src/smithy4s/example/LowHigh.scala
```diff
@@ -12,7 +12,9 @@ sealed abstract class LowHigh(_value: String, _name: String, _ordinal: Int) exte
 object LowHigh extends smithy4s.Enumeration[LowHigh] with smithy4s.ShapeTag.Companion[LowHigh] {
   val id: smithy4s.ShapeId = smithy4s.ShapeId("smithy4s.example", "LowHigh")
 
-  val hints : smithy4s.Hints = smithy4s.Hints.empty
+  val hints : smithy4s.Hints = smithy4s.Hints(
+    smithy.api.Enum(List(smithy.api.EnumDefinition(smithy.api.NonEmptyString("Low"), Some(smithy.api.EnumConstantBodyName("LOW")), None, None, None), smithy.api.EnumDefinition(smithy.api.NonEmptyString("High"), Some(smithy.api.EnumConstantBodyName("HIGH")), None, None, None))),
```
lewisjkl, line 16: These are here since this enum is still using the v1 syntax. Not sure if we want to remove these or not.. probably doesn't impact usage, just readability.

modules/example/src/smithy4s/example/StructureWithRefinedTypes.scala
```diff
@@ -9,8 +9,8 @@ object StructureWithRefinedTypes extends smithy4s.ShapeTag.Companion[StructureWi
   val hints : smithy4s.Hints = smithy4s.Hints.empty
 
   implicit val schema: smithy4s.Schema[StructureWithRefinedTypes] = struct(
-    Age.schema.optional[StructureWithRefinedTypes]("age", _.age),
-    PersonAge.schema.optional[StructureWithRefinedTypes]("personAge", _.personAge),
+    Age.schema.optional[StructureWithRefinedTypes]("age", _.age).addHints(smithy.api.Default()),
+    PersonAge.schema.optional[StructureWithRefinedTypes]("personAge", _.personAge).addHints(smithy.api.Default()),
```
lewisjkl, line 13: Going to see why these showed up... they shouldn't have

sampleSpecs/enums.smithy
```diff
@@ -0,0 +1,31 @@
...
+    @enumValue(2)
+    QUEEN
+    @enumValue(3)
+    KING
+    @enumValue(4)
+    ACE
+    @enumValue(5)
+    JOKER
+}
+
```
lewisjkl, line 18: With the real smithy 2 release we can change these to:

```
intEnum {
JACK = 1
QUEEN = 2
... etc
```

modules/example/src/smithy4s/example/StructureWithRefinedTypes.scala
lewisjkl: The reason these showed up is that smithy 2.0 does the following when loading v1 models:

```
Inject an appropriate `@default` trait on structure members that targeted
      shapes with a default zero value and were not marked with the `@box`
      trait
```

Not sure how we want to proceed here.. especially since the default seems to be empty? I guess that indicates a null default? In that case, this could be problematic with required fields since they'll get the empty default trait as well.. which would indicate that a `null` default is valid, where it isn't.

modules/codegen/src/smithy4s/codegen/CollisionAvoidance.scala
```diff
@@ -244,7 +244,8 @@ object CollisionAvoidance {
     "BigDecimal",
     "Map",
     "List",
-    "Set"
+    "Set",
+    "None"
```
lewisjkl, line 248: Idk if this is the right solution here.. but it works. There was a collision between scala's Option#None and an enum member in the aws-http module

lewisjkl:
Alright so the plugin tests are failing with:

```
[info] [error] java.lang.UnsupportedOperationException: Member `aws.protocols#ChecksumAlgorithm$SHA256` cannot be added to software.amazon.smithy.model.shapes.StringShape$Builder
[info] [error] 	at software.amazon.smithy.model.shapes.AbstractShapeBuilder.addMember(AbstractShapeBuilder.java:209)
[info] [error] 	at software.amazon.smithy.model.loader.AbstractMutableModelFile.createShapes(AbstractMutableModelFile.java:204)
[info] [error] 	at software.amazon.smithy.model.loader.CompositeModelFile.createShapes(CompositeModelFile.java:133)
[info] [error] 	at software.amazon.smithy.model.loader.ModelAssembler.assemble(ModelAssembler.java:571)
[info] [error] 	at smithy4s.codegen.ModelLoader$.load(ModelLoader.scala:84)
[info] [error] 	at smithy4s.codegen.Codegen$.processSpecs(Codegen.scala:36)
[info] [error] 	at smithy4s.codegen.Smithy4sCodegenPlugin$.$anonfun$cachedSmithyCodegen$5(Smithy4sCodegenPlugin.scala:169)
[info] [error] 	at sbt.util.Tracked$.$anonfun$lastOutput$1(Tracked.scala:73)
[info] [error] 	at scala.Function$.$anonfun$untupled$1(Function.scala:110)
[info] [error] 	at sbt.util.Tracked$.$anonfun$inputChangedW$1(Tracked.scala:219)
[info] [error] 	at smithy4s.codegen.Smithy4sCodegenPlugin$.$anonfun$cachedSmithyCodegen$1(Smithy4sCodegenPlugin.scala:182)
[info] [error] 	at scala.Function1.$anonfun$compose$1(Function1.scala:49)
[info] [error] 	at sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:62)
[info] [error] 	at sbt.std.Transform$$anon$4.work(Transform.scala:68)
[info] [error] 	at sbt.Execute.$anonfun$submit$2(Execute.scala:282)
[info] [error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:23)
[info] [error] 	at sbt.Execute.work(Execute.scala:291)
[info] [error] 	at sbt.Execute.$anonfun$submit$1(Execute.scala:282)
[info] [error] 	at sbt.ConcurrentRestrictions$$anon$4.$anonfun$submitValid$1(ConcurrentRestrictions.scala:265)
[info] [error] 	at sbt.CompletionService$$anon$2.call(CompletionService.scala:64)
[info] [error] 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[info] [error] 	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
[info] [error] 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[info] [error] 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
[info] [error] 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
[info] [error] 	at java.base/java.lang.Thread.run(Thread.java:833)
```

I've been trying to figure out why this might be, so far no luck. Any ideas are appreciated!

modules/codegen/src/smithy4s/codegen/CollisionAvoidance.scala
Baccata: yeah that's fine

Baccata:
@lewisjkl, the scripted tests fail because some of them pull an incompatible version of aws-issued smithy libraries : 

https://github.com/disneystreaming/smithy4s/blob/main/modules/codegen-plugin/src/sbt-test/codegen-plugin/dependencies-only/build.sbt#L9

Baccata (approved):

PR #355, opened 2022-08-06
first pass at using mixins to create traits

lewisjkl:
The purpose of this PR is to use smithy 2.0 mixins to generate traits in the codegen.

Note this PR is based off of the `0.0.15-smithy2` branch.

See the examples in the PR for how traits look and are used.


modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -415,6 +433,66 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
...
+  ): Lines = {
+    block(line"trait $name$mixinExtensions") {
+      lines(
+        // TODO: noDefault is probably not the right move, just putting for quick sanity check
+        fields.map(f => line"def ${fieldToRenderLine(f, noDefault = true)}")
+      )
+    }
+  }
+
+  private def renderProduct(
```
Baccata, line 455: let's just pass the whole `Product` instance as a parameter to these methods, they're starting to be too wordy

Baccata (approved):

PR #357, opened 2022-08-09
Best-effort error-decoding fix

Baccata:
Remove the best-effort error-decoding logic
    
    Now, when a client gets an unsuccessful response from a server, it'll
    try to inspect the X-Error-Type header first, and then will try to match
    the received status code against the errors it knows about
    (provided the code in question is uniquely used)
    
    If the client cannot infer a unique alternative based on this
    information, it'll raise a generic "UnknownErrorResponse"

### DIFF
--- a/modules/core/src/smithy4s/ShapeId.scala
+++ b/modules/core/src/smithy4s/ShapeId.scala
@@ -24,6 +24,14 @@ final case class ShapeId(namespace: String, name: String) extends HasId {
 }
 
 object ShapeId extends ShapeTag.Has[ShapeId] {
+  def parse(string: String): Option[ShapeId] = {
+    if (!string.contains('#')) None
+    else {
+      val segments = string.split("#")
+      if (segments.length > 1) None
+      else Some(ShapeId(segments(0), segments(1)))
+    }
+  }
 
   final case class Member(shapeId: ShapeId, member: String)
 
--- a/build.sbt
+++ b/build.sbt
@@ -668,7 +668,7 @@ lazy val Dependencies = new {
 
   val Jsoniter =
     Def.setting(
-      "com.github.plokhotnyuk.jsoniter-scala" %%% "jsoniter-scala-core" % "2.13.38"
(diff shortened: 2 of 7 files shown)


modules/core/src/smithy4s/ShapeId.scala
```diff
@@ -24,6 +24,14 @@ final case class ShapeId(namespace: String, name: String) extends HasId {
 }
 
 object ShapeId extends ShapeTag.Has[ShapeId] {
+  def parse(string: String): Option[ShapeId] = {
```
daddykotex, line 27: if we really need this, I suggest we implemente it on top of `ShapeId.from(s: String)` from smithy-model itself. You may need to catch `ShapeIdSyntaxException`

daddykotex:
I'm having trouble figuring out what exactly changes from before just by looking at the code. can you help me out here?

daddykotex:
> I'm having trouble figuring out what exactly changes from before just by looking at the code. can you help me out here?

There is support for the `NameOnly` of the shape in the header which was not there before but I don't think that's it.

Baccata:
> I'm having trouble figuring out what exactly changes from before just by looking at the code. can you help me out here?

@daddykotex sure. So the behaviour change is that now, if we can't discriminate by neither `X-Error-Type` header value NOR status code (because several errors have the same status code in the smithy model), we raise an `UnkownError`. Previously, if we couldn't discriminate, we'd apply an untagged-union kind of decoding logic by doing a "is there a client-error that we can decode to ? if so, raise it", but it turned out to be a terrible idea, as teams are apparently using the same "structure" for a number of their errors, which led our logic to raise errors on the client side that were definitely incorrect. 

The previous behaviour was only matching the name of the error in `X-Error-Type`, I've added logic to future-proof it by accepting fully-qualified names (the case that references `ShapeId`), as using the full qualified name as a discriminator would be more precise.

daddykotex:
Ah ok, that was my high level understanding of what we had discussed. Makes sense to me.

daddykotex (approved):
LGTM,

thanks for the cleanup of my code :D

PR #358, opened 2022-08-09
add default support in several locations

lewisjkl:
Adding support for default values in the various codecs (document, metadata, json).

Metadata and json are still in progress as of the time of writing this description. Tests are in place, but not yet all passing.

Notice this PR is created against a snapshot version of smithy 2 that I published locally. This is because `1.21.0-rc1` doesn't handle default traits properly/completely.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -162,7 +162,8 @@ lazy val core = projectMatrix
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "errors.smithy",
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "example.smithy",
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "adtMember.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "enums.smithy"
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "enums.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "defaults.smithy"
     ),
     (Test / sourceGenerators) := Seq(genSmithyScala(Test).taskValue),
     testFrameworks += new TestFramework("weaver.framework.CatsEffect"),
@@ -300,6 +301,7 @@ lazy val codegen = projectMatrix
     buildInfoKeys := Seq[BuildInfoKey](version, scalaBinaryVersion),
     buildInfoPackage := "smithy4s.codegen",
     isCE3 := true,
+    resolvers += Resolver.mavenLocal,
     libraryDependencies ++= Seq(
       Dependencies.Cats.core.value,
       Dependencies.Smithy.model,
@@ -390,6 +392,7 @@ lazy val protocol = projectMatrix
       .filterNot { case (file, path) =>
         path.equalsIgnoreCase("META-INF/smithy/manifest")
       },
+    resolvers += Resolver.mavenLocal,
     libraryDependencies += Dependencies.Smithy.model,
     javacOptions ++= Seq("--release", "8")
   )
@@ -620,7 +623,8 @@ lazy val example = projectMatrix
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "brandscommon.smithy",
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "refined.smithy",
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "enums.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "mixins.smithy"
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "mixins.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "defaults.smithy"
     ),
     Compile / resourceDirectory := (ThisBuild / baseDirectory).value / "modules" / "example" / "resources",
     isCE3 := true,
@@ -673,7 +677,7 @@ lazy val Dependencies = new {
     )
 
   val Smithy = new {
-    val smithyVersion = "1.21.0-rc1"
+    val smithyVersion = "1.21.0-SNAPSHOT"
     val model = "software.amazon.smithy" % "smithy-model" % smithyVersion
     val build = "software.amazon.smithy" % "smithy-build" % smithyVersion
     val awsTraits =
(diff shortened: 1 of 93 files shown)


modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
```diff
@@ -190,13 +191,21 @@ object DocumentEncoderSchemaVisitor extends SchemaVisitor[DocumentEncoder] {
     ): (S, Builder[(String, Document), Map[String, Document]]) => Unit = {
       val encoder = apply(field.instance)
       (s, builder) =>
-        field.foreachT(s) { t =>
-          val jsonLabel = field.instance.hints
+        field.foreachT(s) { maybeT =>
+          val hints = field.instance.hints
```
Baccata, line 195: hotpath.

build.sbt
```diff
@@ -390,6 +392,7 @@ lazy val protocol = projectMatrix
       .filterNot { case (file, path) =>
         path.equalsIgnoreCase("META-INF/smithy/manifest")
       },
+    resolvers += Resolver.mavenLocal,
```
lewisjkl, line 395: Will remove when changing to use official smithy 2 release

modules/example/src/smithy4s/example/StreamedObjects.scala
```diff
@@ -66,7 +66,7 @@ object StreamedObjectsGen extends smithy4s.Service[StreamedObjectsGen, StreamedO
     val input: smithy4s.Schema[GetStreamedObjectInput] = GetStreamedObjectInput.schema.addHints(smithy4s.internals.InputOutput.Input.widen)
     val output: smithy4s.Schema[GetStreamedObjectOutput] = GetStreamedObjectOutput.schema.addHints(smithy4s.internals.InputOutput.Output.widen)
     val streamedInput : smithy4s.StreamingSchema[Nothing] = smithy4s.StreamingSchema.nothing
-    val streamedOutput : smithy4s.StreamingSchema[StreamedBlob] = smithy4s.StreamingSchema("GetStreamedObjectOutput", StreamedBlob.schema)
+    val streamedOutput : smithy4s.StreamingSchema[StreamedBlob] = smithy4s.StreamingSchema("GetStreamedObjectOutput", StreamedBlob.schema.addHints(smithy.api.Default(smithy4s.Document.fromString(""))))
```
lewisjkl, line 69: Afaik AWS automatically adds this default since the blob shape used to be boxed and now isn't, this would go away if we updated streaming.smithy to be $version 2.0

modules/json/src/smithy4s/http/json/SchemaVisitorJCodec.scala
```diff
@@ -1205,7 +1220,11 @@ private[smithy4s] class SchemaVisitorJCodec(maxArity: Int)
               if (f.isRequired) {
                 if (value == null) cursor.requiredFieldError(f.label, f.label)
                 value
-              } else Option(value)
+              } else {
+                Option(value).orElse {
+                  Option(defaults.get(f.label))
+                }
+              }
```
lewisjkl, line 1227: Not sure if orElse is as efficient here as other methods

modules/example/src/smithy4s/example/StreamedObjects.scala
daddykotex: interesting

modules/json/src/smithy4s/http/json/SchemaVisitorJCodec.scala
```diff
@@ -1155,6 +1155,21 @@ private[smithy4s] class SchemaVisitorJCodec(maxArity: Int)
...
+        new util.HashMap[String, Any](
+          documentFields.length << 1,
+          0.5f
+        ) {
+          documentFields.foreach { field =>
+            val decode: Document => Option[Any] =
+              Document.Decoder.fromSchema(field.instance).decode(_).toOption
+            field.getDefault.foreach { d =>
+              val result = decode(d)
+              result.foreach(put(jsonLabel(field), _))
```
Baccata, line 1168: you have a bug there, as you're using the `jsonLabel` and when looking up you're using the field's label. Moreover, safer to use the field object as a key in the map, as opposed to the label.

modules/json/src/smithy4s/http/json/SchemaVisitorJCodec.scala
Baccata: it's not, because of the cost of the Option allocation. Also you're only leveraging defaults values on optional fields there. 

It should be : 

``` 
fieldsWithDefaults.foreach { case (f, default) => 
 ....
  if (f.isRequired) {
      if(default == null) {
      ... 
      }  else {
        default 
      }
  } 
} 
```

lewisjkl: The reason I put it in the optional side is that smithy doesn't allow required fields to have default values (required and default traits are mutually exclusive). But I do get your broader point and will make the update

Baccata: @lewisjkl actually that's an interesting notion : in the smithy4s level, I think the presence of `@default` should translate to a `required` field, and in the case class, the field should not be wrapped in Option. Essentially `@default` means that an instance of the case class always has a value for the corresponding field, which means in the Schema, the field should be considered required.

lewisjkl: Ah, I see what you're saying. I will make that change 👍

modules/example/src/smithy4s/example/DefaultTest.scala
```diff
@@ -0,0 +1,18 @@
+package smithy4s.example
+
+import smithy4s.schema.Schema._
+
+case class DefaultTest(one: Int, two: String, three: List[String])
```
lewisjkl, line 5: Guessing we are going to want this to be:

```suggestion
case class DefaultTest(one: Int = 1, two: String = "test", three: List[String] = List.empty)
```

If so, I'll make the changes to facilitate that

Baccata: Yeah. But also we need to make sure that the ordering of the fields goes as such : 

1. fully required fields first 
2. required fields with default second 
3. optional fields last. 

The fields are already getting sorted to ensure that optional fields come AFTER required ones. Just change the ordering there .

Baccata (approved):

PR #364, opened 2022-08-11
Smithy 2.0 syntax is released in the 1.23.0 artifact

Baccata:


modules/example/src/smithy4s/example/AgeFormat.scala
```diff
@@ -7,7 +7,7 @@ object AgeFormat extends smithy4s.ShapeTag.Companion[AgeFormat] {
   val id: smithy4s.ShapeId = smithy4s.ShapeId("smithy4s.example", "ageFormat")
 
   val hints : smithy4s.Hints = smithy4s.Hints(
-    smithy.api.Trait(Some("integer"), None, None),
+    smithy.api.Trait(Some("integer"), None, None, None),
```
daddykotex, line 10: I think this is going to conflict with @lewisjkl PR but it should not be too hard to address

daddykotex: git should even be able to do it itself

PR #371, opened 2022-08-15
Use jsonName on Missing required field error

daddykotex:
On top of that, I include a change to improve the `message` of such error so that we know the path where a field was expected but not found.

### DIFF
--- a/modules/json/src/smithy4s/http/json/Cursor.scala
+++ b/modules/json/src/smithy4s/http/json/Cursor.scala
@@ -63,12 +63,7 @@ class Cursor private () {
       top -= 1
       list = stack(top) :: list
     }
-    val path = PayloadPath(list)
-    throw PayloadError(
-      path,
-      expecting,
-      s"Missing required field at $path"
-    )
+    throw PayloadError(PayloadPath(list), expecting, "Missing required field")
   }
 
   private def getPath(): PayloadPath = {
--- a/modules/json/src/smithy4s/http/json/Cursor.scala
+++ b/modules/json/src/smithy4s/http/json/Cursor.scala
@@ -63,7 +63,12 @@ class Cursor private () {
       top -= 1
       list = stack(top) :: list
     }
-    throw PayloadError(PayloadPath(list), expecting, "Missing required field")
+    val path = PayloadPath(list)
+    throw PayloadError(
+      path,
+      expecting,
+      s"Missing required field at $path"
+    )
   }
 
   private def getPath(): PayloadPath = {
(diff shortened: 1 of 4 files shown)


modules/json/src/smithy4s/http/json/Cursor.scala
```diff
@@ -63,7 +63,12 @@ class Cursor private () {
       top -= 1
       list = stack(top) :: list
     }
-    throw PayloadError(PayloadPath(list), expecting, "Missing required field")
+    val path = PayloadPath(list)
+    throw PayloadError(
```
Baccata, line 67: let's change the `getMessage` implementation of the `PayloadError` class instead.

Baccata: Concatenating the path to the message field of the case class means that the information gets duplicated in the json output when the PayloadError is serialised. 

Showing the path makes sense in all scenarios.

modules/json/src/smithy4s/http/json/SchemaVisitorJCodec.scala
```diff
@@ -1157,7 +1157,10 @@ private[smithy4s] class SchemaVisitorJCodec(maxArity: Int)
             stage2 += {
               val value = buffer.get(f.label)
               if (f.isRequired) {
-                if (value == null) cursor.requiredFieldError(f.label, f.label)
+                if (value == null) {
+                  val label = jsonLabel(f)
```
plokhotnyuk, line 1161: Can we avoid recalculation of JSON label in runtime?

daddykotex: I looked at it, and it felt super convoluted. I figured that if we end up in this branch, it's happing one and we throw an exception from here, so at this point I think optimizations are out the window

is my assumption correct?

Baccata: You just need to store the following bits of information in an array containing tuples, which you'd use in place of the `fields.foreach` call (line 1156) 

* jsonLabel 
* fieldLabel 
* required

modules/json/src/smithy4s/http/json/SchemaVisitorJCodec.scala
```diff
@@ -1215,7 +1218,10 @@ private[smithy4s] class SchemaVisitorJCodec(maxArity: Int)
...
+                  cursor.requiredFieldError(label, label)
```
Baccata, line 1223: whilst we're at it, the first string here should probably be something else

daddykotex: I noticed it but figured it was intended because it should be a `expected`, so I assume a type, but we were passing the label. My thinking was that a required field is not there so let's say we're `expected = fieldLabel` rather than `expected = int`, for example.

modules/json/src/smithy4s/http/json/SchemaVisitorJCodec.scala
daddykotex: made `fields: Fields[Z]` into a fields: LabelledFields[Z]` which is an alias for Vector[(Field[...], String)]` (the string being the json label)

we have label, and required within the field

Baccata (approved):

Issue #372, opened 2022-08-16
Mergify is out of sync


daddykotex:
I'm currently having a look at mergify and I noticed the content of https://github.com/disneystreaming/smithy4s/blob/c077c41fb312523d38930acb84f63a27d0a7eacd/.mergify.yml seems to be out of date after the addition of Scala Native.

After discussing with @Baccata , we decided adding a checkpoint task would be the best way to go. Suppose I add a job `matrix-success-checkpoint` and configure Mergify for the following conditions: `- status-success="matrix-success-checkpoint"` (see note below).

This looks like that in the UI for all successes: 

![Screen Shot 2022-08-16 at 11 11 42](https://user-images.githubusercontent.com/5230460/184915826-6debce92-8584-45f9-8429-64541c30f5a7.png)

And like that when a failure happens:
![Screen Shot 2022-08-16 at 11 11 28](https://user-images.githubusercontent.com/5230460/184915945-716bd022-8e96-4187-b632-5fd7286ba08e.png)

Note:

We use `status-success` in our configuration but the documentation says `check-success`: https://docs.mergify.com/conditions/?highlight=status+success#github-actions

Was it update or am I missing something?

Baccata:
I believe so

PR #373, opened 2022-08-16
Build checkpoint for mergify

daddykotex:
Add a job that acts as a checkpoint to signal to
mergify that all jobs of the matrix (in this case `build`) were
successful.

This would address https://github.com/disneystreaming/smithy4s/issues/372


Baccata (approved):
awesome

PR #376, opened 2022-08-17
Add support for decline interpreters 

yisraelU:
Adds a module to provide a CLI for all Smithy4s generated services


build.sbt
```diff
@@ -368,6 +370,20 @@ lazy val codegenPlugin = (projectMatrix in file("modules/codegen-plugin"))
     scriptedBufferLog := false
   )
 
+lazy val cli = (projectMatrix in file("modules/cli"))
+  .settings(
+    name := "smithy4s-cli",
+    testFrameworks += new TestFramework("weaver.framework.CatsEffect"),
```
kubukoz, line 376: I think this can be enabled in all modules by default. It doesn't actually add any dependencies.

yisraelU: I added it to the top of the build file

modules/cli/src/standalone/Smithy4sSimpleStandaloneCli.scala
```diff
@@ -0,0 +1,47 @@
...
+  * service, or if you just want to have full control of how you call the upstream.
+  *
+  * @param impl
+  *   Opts providing the implementation of the service. If you have that in a Resource, use use
+  *   things like [[smithy4s.cli.util.UnliftResource]] to convert. Only make the Opts require any
+  *   parameters if you need program-wide config that doesn't fit into a specific operation's input
+  *   (operation inputs are mapped to Opts automatically by all Smithy4sCli implementations).
+  */
+class Smithy4sSimpleStandaloneCli[Alg[_[_, _, _, _, _]], Op[_, _, _, _, _]](
+  impl: Opts[smithy4s.Monadic[Alg, IO]]
```
yisraelU, line 38: @kubukoz Is there a need to expose Opts here in this interface , it could just take in a Monadic  and get wrapped internally

kubukoz: It's here so that you could customize the interpreter with additional CLI options.

build.sbt
```diff
@@ -368,6 +366,19 @@ lazy val codegenPlugin = (projectMatrix in file("modules/codegen-plugin"))
     scriptedBufferLog := false
   )
 
+lazy val cli = (projectMatrix in file("modules/cli"))
```
Baccata, line 369: After consideration, I'd like this to be renamed to `decline` (as well as jvm package in which the code is located)

modules/docs/src/01-overview/06-advanced-usage.md
```diff
@@ -0,0 +1,67 @@
+---
+sidebar_label: Advanced Usage
+title: Advanced Usage
```
Baccata, line 3: Rename this to "Deriving CLIs"

modules/cli/src/core/OptsVisitor.scala
```diff
@@ -0,0 +1,341 @@
...
+import smithy4s.Hints
+import smithy4s.Lazy
+import smithy4s.Refinement
+import smithy4s.ShapeId
+import smithy4s.Timestamp
+import smithy4s.cli.core.CoreHints._
+import smithy4s.schema.Alt
+import smithy4s.schema.EnumValue
+import smithy4s.schema.Primitive
+import smithy4s.schema.Primitive.PBigDecimal
```
Baccata, line 35: you can use wild cards on `smithy4s.schema._` and `smithy4s.schema.Primitive._`

build.sbt
yisraelU: Should we rename the actual Scala code or stick with the Smithy4sCli convention

Baccata: that's what I mean by the "jvm package in which the code is located"

yisraelU:
Currently keeping Monovore Effect as we are providing an actual runnable cli via CommandIOApp .
I am not sure about @Baccata regarding removing the effect lib and just using core.
Perhaps it was to be removed and we would limit ourselves to providing a Command aka Argument parser leaving the user to implement the actual interface

modules/docs/src/03-protocols/04-deriving-cli.md
```diff
@@ -0,0 +1,87 @@
+---
```
daddykotex, line 1: I'd move this file under `modules/docs/src/03-protocols/simple-rest-json/04-deriving-cli.md `

so it is along the other simpleRestJson things

modules/docs/src/03-protocols/04-deriving-cli.md
```diff
@@ -0,0 +1,87 @@
...
+sidebar_label: Deriving CLIs
+title: Deriving CLIs
+---
+
+- The Smithy4s Decline module provides the capability to derive a [Decline](https://ben.kirw.in/decline/) Cli for your service.
+- Lets revisit our HelloWorld smithy definition from the [Quickstart](modules/docs/src/01-overview/02-quickstart.md)
+
+```kotlin
+namespace smithy4s.hello
```
daddykotex, line 10: unrelated: maybe we could leverage mdoc to have this example in one place and avoid having multiple copies of it

modules/docs/src/03-protocols/04-deriving-cli.md
daddykotex: Screen Shot 2022-08-19 at 09 15 11

modules/docs/src/03-protocols/04-deriving-cli.md
```diff
@@ -0,0 +1,87 @@
...
+  val simple = new HelloWorldService[IO]{
+    def hello(name: String, town: Option[String]):IO[Greeting] = {
+      IO{
+        Greeting(s"hello $name , who hails from $town")
+      }
+    }
+  }
+}
+```
+ - Now Using the ```decline``` module from smithy4s we can wrap the service instance in an instance of a Smithy4sCli.
```
daddykotex, line 53:
```suggestion
 - Now Using the ```decline``` module from smithy4s we can wrap the service instance in an instance of a ```Smithy4sCli```.
```

modules/docs/src/03-protocols/04-deriving-cli.md
```diff
@@ -0,0 +1,87 @@
...
+ - There is a convenient class ```Smithy4sSimpleStandaloneCli``` that you can extend and simply pass in the service wrapped in an Opts 
+   - ``` object Hello extends standalone.Smithy4sSimpleStandaloneCli(Opts(HelloWorldServiceInstance.simple)) ```
+ - ```Hello``` is now a runnable CommandIOApp and will provide the following interface
+
+```
+ Usage: hello-world-service hello [--output <output>] <name> [<town>]
```
daddykotex, line 59: this would be a great opportunity to leverage `mdoc` to generate this instead of hardcoding it

modules/docs/src/03-protocols/04-deriving-cli.md
```diff
@@ -0,0 +1,87 @@
...
+   - The Input structure is flattened to top level fields with no nesting
+     - All primitives are mapped to Positional `decline` Arguments , except for boolean fields
+     - Boolean fields are mapped to a `decline` Flag
+     - Blobs are mapped to take in either stdin(by passing in `-` ) or a full file path and will produce a Byte array 
+     - Lists and recursive types
+       - a top level list is converted to a repeated positional argument (or flag in case of Boolean)
+   - How is Nesting handled
+     - all nested fields are converted to Options 
+     - a nested list is converted to expect a json i.e. List[String] would expect ```'["foo","baz","qux"]'``` as a decline Option
+       - when the nested list contains blobs , it will expect a json of base64 encoded strings
```
daddykotex, line 81:
```suggestion
       - when the nested list contains blobs, it will expect a json list of base64 encoded strings 
```

modules/docs/src/03-protocols/04-deriving-cli.md
yisraelU: Why ?

modules/docs/src/03-protocols/04-deriving-cli.md
yisraelU: Will deal with all mdoc related stuff after this PR

modules/docs/src/03-protocols/04-deriving-cli.md
daddykotex: my understanding is that deriving clis works with this particular protocol, maybe I'm wrong though

yisraelU: no , deriving cli's is protocol agnostic

daddykotex: it may not be tied to simpleRestJson, but it sure seems to be built on http & json

yisraelU: Perhaps you can elaborate in standup. 
Currently IIUC , if there are http based traits that information is used for documentation . however its not a requirement.
json is a requirement iff your data is structured in a way that needs a better mechanism then positional or named arguments, this doesn't affect how or if the data is transmitted at all over the network or otherwise using any particular encoding

build.sbt
```diff
@@ -374,6 +372,19 @@ lazy val codegenPlugin = (projectMatrix in file("modules/codegen-plugin"))
     scriptedBufferLog := false
   )
 
+lazy val decline = (projectMatrix in file("modules/decline"))
+  .settings(
+    name := "decline",
+    isCE3 := true,
+    libraryDependencies ++= List(
+      "org.typelevel" %%% "cats-core" % "2.7.0",
```
daddykotex, line 380: One last thing: I see that we have things like `Dependencies.Weaver.cats.value` can you use them when possible?

I see that cats is there, and decline & decline effect could be moved in there (because decline is already in use by the cli)

modules/decline/src/core/OptsVisitor.scala
```diff
@@ -35,19 +42,23 @@ import java.util.UUID
...
   ): Opts[P] = {
     val fieldName = FieldName.require(hints)
     val isNested = IsNested.orFalse(hints)
-    val doc = hints.get(Documentation).fold("")(_.value)
+    val doc = hints.get(Documentation).map(_.value).toList ::: hints
+      .get(ExternalDocumentation)
+      .toList
+      .flatMap(_.value.map { case (description, link) =>
+        s"\n\n$description\n\n[See more]($link)"
+      }.toList)
```
kubukoz, line 57: **thought:** rendering markdown here sounds weird, given the output will end up in a terminal - I don't think many (if any) of these read markdown

modules/decline/src/core/OptsVisitor.scala
```diff
@@ -326,5 +340,9 @@ object OptsVisitor
     schema.compile[Opts](this).map(bijection.to)
 
   override def refine[A, B](schema: Schema[A], refinement: Refinement[A, B]) =
-    schema.compile[Opts](this).map(refinement(_).fold(message =>throw RefinementFailed(message),identity))
+    schema
+      .compile[Opts](this)
+      .map(
+        refinement(_).fold(message => throw RefinementFailed(message), identity)
+      )
 }
```
kubukoz, line 348: **thought:** can we use `mapValidated` or similar to, instead of throwing, fail the Opts value?

modules/decline/src/core/commons.scala
```diff
@@ -52,9 +52,8 @@ object commons {
 
     Argument.from("json")(parse(_).toValidatedNel)
   }
-  val byteArrayArgument: Argument[ByteArray] = Argument.from("base64")(
-    s =>
-     Valid(ByteArray(s.getBytes("UTF-8"))))
+  val byteArrayArgument: Argument[ByteArray] =
+    Argument.from("base64")(s => Valid(ByteArray(s.getBytes("UTF-8"))))
```
kubukoz, line 56: **thought:** should we decode the base64 to a plain array of bytes before passing it to the ByteArray?

yisraelU: yes it should

modules/decline/src/core/OptsVisitor.scala
yisraelU: lol copilot generated it .
will remove markdown syntax

build.sbt
yisraelU: done

kubukoz:
really happy to see this merged! :shipit:

yisraelU:
@kubukoz me too , really great feature

Issue #389, opened 2022-08-20
Using Set as operation name causes compilation failures


keynmol:
The quickest way to reproduce is with Scala CLI

**test.smithy**
```kotlin
$version: "2.0"

namespace hellosmithy4s.spec

use smithy4s.api#simpleRestJson

@simpleRestJson
service HelloService {
  version: "1.0.0",
  operations: [Set]
}

@http(method: "POST", uri: "/api/update/{key}", code: 204)
operation Set {
  input := {
    @httpLabel
    @required
    key: Key,

    value: Value
  }
}

string Key
integer Value
```

```
> cs launch --channel https://disneystreaming.github.io/coursier.json smithy4s -- generate test.smithy
Aug 20, 2022 5:02:18 PM software.amazon.smithy.openapi.fromsmithy.mappers.RemoveUnusedComponents removalRound
INFO: Removing unused OpenAPI components: [#/components/schemas/SetInput]
/private/var/folders/my/drt5584s2w59ncgxgxpq7v040000gn/T/tmp.07Z3NfFw/hellosmithy4s/spec/SetInput.scala
/private/var/folders/my/drt5584s2w59ncgxgxpq7v040000gn/T/tmp.07Z3NfFw/hellosmithy4s/spec/Value.scala
/private/var/folders/my/drt5584s2w59ncgxgxpq7v040000gn/T/tmp.07Z3NfFw/hellosmithy4s/spec/HelloService.scala
/private/var/folders/my/drt5584s2w59ncgxgxpq7v040000gn/T/tmp.07Z3NfFw/hellosmithy4s.spec.HelloService.json
/private/var/folders/my/drt5584s2w59ncgxgxpq7v040000gn/T/tmp.07Z3NfFw/hellosmithy4s/spec/package.scala
/private/var/folders/my/drt5584s2w59ncgxgxpq7v040000gn/T/tmp.07Z3NfFw/hellosmithy4s/spec/Key.scala
```

## Error

```
/v/f/m/d/T/tmp.07Z3NfFw > scala-cli compile . --dep "com.disneystreaming.smithy4s::smithy4s-http4s:0.15.0"
Compiling project (Scala 3.1.3, JVM)
[error] ./hellosmithy4s/spec/HelloService.scala:45:60: Found:    hellosmithy4s.spec.SetInput
[error] Required: hellosmithy4s.spec.Key
[error]     def _Set(key: Key, value: Option[Value] = None) = _Set(SetInput(key, value))
[...]

PR #393, opened 2022-08-22
Collision Avoidance additions

yisraelU:
add additional types to CollisionAvoidance list for reserved types
added additional Smithy spec to showcase handling of those cases
Fixed bug #389  when uncapitalizing name that starts with underscore


lewisjkl (approved):
Looks great, thanks for adding the example

daddykotex:
kudos @yisraelU

PR #394, opened 2022-08-22
Range collision avoidance

yisraelU:
since Range is automatically imported as part of the stdlib , it needs to be handled as part of Collision Avoidance, this resulted in a compilation error with smithy.api.Range that is referenced from RefinementProvider.
The reference was modified to refer to the protected name aka prefixed with underscore


kubukoz (approved):

lewisjkl (approved):

Baccata:
This change is source/binary breaking, as it changes the core library code, please revert and let's try to think a little harder about it. 

The underscore prefix was a poor man's solution. I think now that we have a better collision detection set up, we should strive to put a better solution in place.

PR #406, opened 2022-08-23
Dynamic smithy4s enums + allow IntEnum hint in JSON codecs

daddykotex:


modules/codegen-plugin/src/sbt-test/codegen-plugin/defaults/src/main/scala/Main.scala
```diff
@@ -24,4 +27,22 @@ object Main extends App {
...
   }
+
+  def buildSchemaIndex() =
+    ModelLoader
+      .load(
+        specs = Set.empty,
+        dependencies = List(
+          "com.disneystreaming.smithy4s::smithy4s-dynamic:0.15.1",
+          "com.disneystreaming.smithy4s::smithy4s-codegen:0.15.1"
+        ),
```
kubukoz, line 38: are these necessary? I don't think they bring in a lot of smithy

modules/codegen-plugin/src/sbt-test/codegen-plugin/defaults/src/main/scala/Main.scala
```diff
@@ -24,4 +27,22 @@ object Main extends App {
       println("failed")
       sys.exit(1)
   }
+
+  def buildSchemaIndex() =
+    ModelLoader
+      .load(
+        specs = Set.empty,
```
kubukoz, line 34: might wanna add some smithy file here too, to ensure its schemas are included in the loaded model

modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
```diff
@@ -210,6 +210,60 @@ private[dynamic] object Compiler {
...
+            .flatMap { case (k, m) =>
+              getTrait[smithy.api.EnumValue](m.traits).toList.map { v =>
+                (k, v)
+              }
+            }
+            .zipWithIndex
+            .map { case ((k, v), i) =>
+              v.value ->
+                EnumValue(
+                  k,
```
kubukoz, line 226: I believe this should be `v.value` downcasted into a string

modules/codegen-plugin/src/sbt-test/codegen-plugin/defaults/src/main/scala/Main.scala
daddykotex: ahh it's my bad, it's a forgotton attempt

modules/codegen-plugin/src/sbt-test/codegen-plugin/defaults/src/main/scala/Main.scala
daddykotex: with specs or not, this triggered the problem we observed initially

modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
```diff
@@ -210,6 +210,60 @@ private[dynamic] object Compiler {
...
+            getTrait[smithy.api.EnumValue](m.traits).toList.map { v =>
+              (k, v)
+            }
+          }
+          .collect { case (k, smithy.api.EnumValue(Document.DNumber(v))) =>
+            (k, v.toInt) // safe cast?
+          }
+          .map { case (k, v) =>
+            EnumValue(
+              k, // check
```
kubukoz, line 253: this sounds correct, I guess the `stringName` of int enums is the same as the enum entry's name. In codegen:

```scala
sealed abstract class FaceCard(_value: String, _name: String, _intValue: Int) extends Enumeration.Value {
  override val value: String = _value
  override val name: String = _name
  override val intValue: Int = _intValue
  override val hints: Hints = Hints.empty
  @inline final def widen: FaceCard = this
}
object FaceCard extends Enumeration[FaceCard] with ShapeTag.Companion[FaceCard] {
  val id: ShapeId = ShapeId("smithy4s.example", "FaceCard")

  val hints : Hints = Hints(
    smithy.api.Documentation("FaceCard types"),
    IntEnum(),
  )

  case object JACK extends FaceCard("JACK", "JACK", 1)
  case object QUEEN extends FaceCard("QUEEN", "QUEEN", 2)
  case object KING extends FaceCard("KING", "KING", 3)
  case object ACE extends FaceCard("ACE", "ACE", 4)
  case object JOKER extends FaceCard("JOKER", "JOKER", 5)

  val values: List[FaceCard] = List(
    JACK,
    QUEEN,
    KING,
    ACE,
    JOKER,
  )
  implicit val schema: Schema[FaceCard] = enumeration(values).withId(id).addHints(hints)
}
```

modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
```diff
@@ -210,6 +210,60 @@ private[dynamic] object Compiler {
...
+    override def intEnumShape(id: ShapeId, shape: IntEnumShape): Unit = {
+      val values: List[EnumValue[Int]] = shape.members.foldMap { m =>
+        m.toList
+          .flatMap { case (k, m) =>
+            getTrait[smithy.api.EnumValue](m.traits).toList.map { v =>
+              (k, v)
+            }
+          }
+          .collect { case (k, smithy.api.EnumValue(Document.DNumber(v))) =>
+            (k, v.toInt) // safe cast?
```
daddykotex, line 249: this is `BigDecimal` to `Int`

kubukoz, line 249: Smithy validates `intEnum` entry ranges so I think it's safe

modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
```diff
@@ -210,6 +210,60 @@ private[dynamic] object Compiler {
     override def blobShape(id: ShapeId, shape: BlobShape): Unit =
       update(id, shape.traits, bytes)
 
+    override def enumShape(id: ShapeId, shape: EnumShape): Unit = {
+      val values: List[(Document, EnumValue[Document])] =
```
kubukoz, line 214: Smithy 1.x-style enums have a runtime representation of Int (`intValue` of the entry). I think we can do the same in 2.x-style enums because the intValue is always unique within an enum.

modules/dynamic/test/src-jvm/smithy4s/dynamic/EnumSpec.scala
```diff
@@ -49,21 +66,65 @@ class EnumSpec extends FunSuite {
...
   test("dynamic enums have names if they're in the model") {
-    assertNames(ShapeId("example", "Element"), List("ICE", "FIRE"))
+    assertEnum(
+      ShapeId("example", "Element"),
+      names = List("ICE", "FIRE"),
+      values = List("Ice", "Fire")
+    )
   }
 
   test("dynamic enum names are derived if not present in the model") {
```
kubukoz, line 107: this should have some tests for the stringValue as well (same for intEnums)

modules/codegen-plugin/src/sbt-test/codegen-plugin/defaults/src/main/scala/Main.scala
kubukoz: yeah :)

modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
```diff
@@ -225,18 +279,23 @@ private[dynamic] object Compiler {
           // Using the intValue as a runtime value
```
kubukoz, line 279: note: this comment isn't accurate after the changes here

modules/json/src/smithy4s/http/json/codecs.scala
```diff
@@ -37,7 +37,9 @@ object codecs {
       Discriminated,
       Untagged,
       InputOutput,
-      DiscriminatedUnionMember
+      DiscriminatedUnionMember,
+      // TODO: add tests for `codecs` understanding int enums. Maybe pizza spec
+      IntEnum
```
kubukoz, line 42: **issue:** Must've slipped through the cracks, but I think because this was missing the HTTP codecs would never encode/decode an enum as an int. This will affect dynamic _and_ codegen.

@daddykotex would you be so kind and implement a test for this? Might be PizzaSpec+PizzaClientSpec worthy, but probably just checking it in a http4s spec would be good enough

modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
```diff
@@ -389,7 +484,7 @@ private[dynamic] object Compiler {
                     )
                   }
                 val memberHints = allHints(mShape.traits)
-                lField.map(_.addHints(memberHints: _*))
+                lField.map(_.addHints(memberHints.all.toSeq: _*))
```
kubukoz, line 487: this is one problematic case - I wanted to add an `addHints` overload to `SchemaFieldOps`, but that breaks binary compatibility on 2.12 (the existing extension is renamed due to [weird 2.12 compiler behavior](https://github.com/scala/scala/pull/7896)). We can consider it for 0.16.0, in the meantime here's the workaround.

modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
```diff
@@ -210,6 +224,82 @@ private[dynamic] object Compiler {
     override def blobShape(id: ShapeId, shape: BlobShape): Unit =
       update(id, shape.traits, bytes)
 
+    override def enumShape(id: ShapeId, shape: EnumShape): Unit = {
+      val values: List[EnumValue[Int]] =
```
daddykotex, line 228: 👍

modules/json/src/smithy4s/http/json/codecs.scala
daddykotex: will do

kubukoz (approved):
Let's gooooo

PR #410, opened 2022-08-24
Update 04-deriving-cli.md

yisraelU:
fixed relative link to overview

### DIFF
--- a/modules/docs/src/03-protocols/04-deriving-cli.md
+++ b/modules/docs/src/03-protocols/04-deriving-cli.md
@@ -4,7 +4,7 @@ title: Deriving CLIs
 ---
 
 - The Smithy4s Decline module provides the capability to derive a [Decline](https://ben.kirw.in/decline/) Cli for your service.
-- Lets revisit our HelloWorld smithy definition from the [Quickstart](modules/docs/src/01-overview/02-quickstart.md)
+- Lets revisit our HelloWorld smithy definition from the [Quickstart](../01-overview/02-quickstart.md)
 
 ```kotlin
 namespace smithy4s.hello


modules/docs/src/03-protocols/04-deriving-cli.md
```diff
@@ -4,7 +4,7 @@ title: Deriving CLIs
 ---
 
 - The Smithy4s Decline module provides the capability to derive a [Decline](https://ben.kirw.in/decline/) Cli for your service.
-- Lets revisit our HelloWorld smithy definition from the [Quickstart](modules/docs/src/01-overview/02-quickstart.md)
+- Lets revisit our HelloWorld smithy definition from the [Quickstart](../01-overview/02-quickstart.md)
```
kubukoz, line 7:
```suggestion
- Let's revisit our HelloWorld smithy definition from the [Quickstart](../01-overview/02-quickstart.md)
```

daddykotex:
oh yeah, I had just caught that
I found it via another [PR](https://github.com/disneystreaming/smithy4s/pull/404)

I tweaked the checks so they don't go under the radar. What's weird for me is the `mdoc` correctly told us that the links were dead but as a warning and does not fail the task, so it went unnoticed.

There is an option to disable link-hygiene, but not make it fail

daddykotex:
> That seems like oversight. Did you check for existing/closed issues in mdoc?

I had a quick look but could not find anything related. I will see if I can open an issue and a subsequent PR
Screen Shot 2022-08-24 at 11 49 50

daddykotex:
ref: https://github.com/scalameta/mdoc/issues/679

Issue #255, opened 2022-06-03
Operation error name conflict


kubukoz:
This Smithy spec generates code that doesn't compile:

```
namespace example

service A {
  operations: [MyOp]
}

operation MyOp {
  errors: [MyOpError]
}

@error("client")
structure MyOpError {}
```

Relevant code:

```scala
def unliftError(e: MyOpError) : Throwable = e match {
  case MyOpError.MyOpErrorCase(e) => e //here
}
```

Error:

```
type mismatch;
 found   : example.AGen.MyOpError
 required: Throwable
```

As I understand this is caused by a name conflict, in which `MyOpError` is both the name of the error trait *and* the name of the subtype wrapped by `MyOpErrorCase`.

Baccata:
Maybe we could emit a `Danger` event in the smithy validator ?

daddykotex:
In this case, isn't something that we can fix w/o requiring use input ?

the code:

Baccata:
The problem is not the `e`, it's the fact that `MyOpError` (synthetised by smithy4s) is colliding with a user type and the compiler cannot distinguish them in this particular scope.

daddykotex:
ahh I see

kubukoz:
A fully qualified name would do it, but that's probably not what we want to do (ugly codegen)

yisraelU:
#412 fixes this

PR #412, opened 2022-08-24
Collision detection

yisraelU:
Draft to address collisions with stdlib types and keywords
keyword protection is done with underscore as using backticks (which would be idiomatic scala) creates a new problem where we append to names "Gen" or "Op" etc.. a backtick must wrap the entire identifier
I found this to be sufficient reason to avoid switching to backticks as it would involve significant code shuffling .
This draft is to be reviewed for correctness . I will be making further commits to address code conciseness and quality


Baccata:
Ain't got time to review proper before next week, but, I think the most desirable solution would be to remove the underscore-logic altogether in favour of a local rename. 

Say you have the following in a smithy model 

```
namespace foo 

@uniqueItems 
list Set {
    member: String
} 
```

then you'll have `foo.Set` colliding with `scala.Set`. When handling such things manually, what we'd do is : 

``` 
import scala.{Set => SSet} 
```

Doing this would be facilitated by the fact that we're retaining structural information for longer, meaning we can transform the `NameRefs` everywhere they need to be before collapsing the structure into text.

yisraelU:
@Baccata I'm not using underscore for collision renaming.
It's only being used for keyword renaming i.e. a val name type

yisraelU:
this will fix #255 too

modules/core/test/src/smithy4s/ReservedTypeSmokeSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package smithy4s
+
+import munit._
+import smithy4s.example.collision.ReservedNameService
+
+class ReservedTypeSmokeSpec() extends FunSuite {
+
+  test("Idiomatic ADTs can be generated from unions") {
```
daddykotex, line 24: what exactly is being tested here?

daddykotex: ahh I guess the test name is garbage, but we're just ensuring that `ReservedNameService` is available (because codegen & compilation was successful (so was collision handling))

Baccata: Yeah I copied another spec and forgot to rename the test. 

We're testing that the generated code actually uses types from the Scala stdlib, and that the methods are not renamed with underscores or whatnot.

lewisjkl (approved):

PR #414, opened 2022-08-25
customization of default rendering

lewisjkl:
Allows rendering defaults in three different modes: FULL, OPTION_ONLY, NONE.

See docs in PR for full explanation.


kubukoz (approved):

kubukoz (approved):

modules/docs/src/04-codegen/01-customisation.md
```diff
@@ -325,3 +325,92 @@ string Email
...
+The default is `FULL`.
+
+#### FULL
+
+`FULL` means that default values are rendered for all field types. For example:
+
+```kotlin
+use smithy4s.meta#defaultRender
+
+@defaultRender(mode: "FULL")
```
Baccata, line 346: not quite sure it should be a trait. The preference probably applies to all case classes that users generate, so I'm thinking it should be a piece of metadata instead.

daddykotex: two things:
1.  >preference probably applies to all case classes

The opposite could be just as true I guess, some people may want to change the behaviour for one specific case class.

2. >  the rendering of trait values relies on the ordering of parameters

I don't understand what you mean by the `rendering of trait values` in this context, do you have an example?

Baccata: 1) we have optimise user-facing features in terms of what most users will benefit from. Though you are not technically incorrect, I'm pretty sure that if someone dislikes default values in one case, they'll dislike default values in all cases, because default values have very tangible downsides to them. Therefore, it's likely that a piece of UX that would allow to disable defaults altogether across the whole code-gen pipeline, people who dislike defaults will be happy about the low-boilerplate solution. 
2) I meant rendering of hints.

lewisjkl: Good points, I forgot that we had talked about using the metadata for this. I'll update the PR with that and using named parameters in hints 👍

lewisjkl: Done in: https://github.com/disneystreaming/smithy4s/pull/414/commits/0b9d512413465421fc8d1c32403adf1adf85a1e0

lmk if this doesn't match what you are talking about wrt the named parameters in hints

modules/docs/src/04-codegen/01-customisation.md
```diff
@@ -325,3 +325,91 @@ string Email
...
+The default is `FULL`.
+
+This value is set using metadata which means that the setting will be applied to all the rendering done by smithy4s.
+
+#### FULL
+
+`FULL` means that default values are rendered for all field types. For example:
+
+```kotlin
+metadata defaultRenderMode = "FULL"
```
Baccata, line 346: All the metadata bits that smithy4s surfaces/uses should be prefixed with `smithy4s` one way or another. If `.` are allowed in metadata keys, we could make it `smithy4s.defaultRenderMode`

modules/docs/src/04-codegen/01-customisation.md
Baccata: > lmk if this doesn't match what you are talking about wrt the named parameters in hints 

It is indeed

modules/docs/src/04-codegen/01-customisation.md
lewisjkl: Unfortunately, dots aren't allowed. But I changed it to `smithy4sDefaultRenderMode`

sampleSpecs/defaults.smithy
```diff
@@ -11,3 +13,10 @@ structure DefaultTest {
   two: String = "test"
   three: StringList = []
 }
+
+structure DefaultOrderingTest {
```
Baccata, line 17: Mmm, I think it might be a good opportunity to add unit tests in the `codegen` module, since relying on our usual smoke-test way of doing things ain't gonna work (as metadata is meant to be unique across a smithy model). 

Would you mind adding such a unit test, a little bit like the smithy-translate ones, where we dynamically parse a model and run it through the codegen pipeline, and assert that the output code is what we think it should be ?

lewisjkl: Yeah that's a good idea, will add 👍

Baccata (approved):

PR #415, opened 2022-08-26
Fix mima command to run more quickly

lewisjkl:
Adds a matrix-specific command to allow mima to run more specifically

### DIFF
--- a/project/MimaVersionPlugin.scala
+++ b/project/MimaVersionPlugin.scala
@@ -17,7 +17,7 @@ object MimaVersionPlugin extends AutoPlugin {
       ProjectMatrixPlugin &&
       plugins.JvmPlugin
 
-  override def trigger = noTrigger
+  override def trigger = allRequirements
 
   object autoImport {
     val ReleaseTag = """^v((?:\d+\.){2}\d+(?:-.*)?)$""".r
@@ -25,6 +25,8 @@ object MimaVersionPlugin extends AutoPlugin {
     lazy val mimaReportBinaryIssuesIfRelevant = taskKey[Unit](
       "A wrapper around the mima task which ensures publishArtifact is set to true"
     )
+    lazy val isMimaEnabled =
+      settingKey[Boolean]("setting to enable mima checks")
   }
   import autoImport._
 
@@ -74,6 +76,7 @@ object MimaVersionPlugin extends AutoPlugin {
     )
 
   override def projectSettings: Seq[Setting[_]] = Seq(
+    isMimaEnabled := false,
     mimaReportBinaryIssuesIfRelevant := filterTaskWhereRelevant(
       mimaReportBinaryIssues
     ).value,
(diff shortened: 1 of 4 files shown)


project/MimaVersionPlugin.scala
```diff
@@ -87,13 +87,24 @@ object MimaVersionPlugin extends AutoPlugin {
...
-      if (sbtPlugin.value || !isJvm) {
+      val isScalaNative = virtualAxes.?.value
+        .getOrElse(List.empty)
+        .contains(
+          VirtualAxis.native
+        )
+
+      val suffix =
+        if (isScalaJs) "_sjs1"
+        else if (isScalaNative) "_native0.4"
```
lewisjkl, line 104: Anyone know of a way to get these values from the build directly? CrossVersion something?

armanbilge: Here's what sbt-typelevel does so it doesn't have to play games with suffixes and stuff :)

https://github.com/typelevel/sbt-typelevel/blob/cedbed34d087ac2a0f24112eaa5f128c906deae0/mima/src/main/scala/org/typelevel/sbt/TypelevelMimaPlugin.scala#L88

lewisjkl: Awesome, thanks! I updated to use that same mechanism.

daddykotex (approved):

Baccata:
I'm actually not sure I want this enabled on JS and Native. MiMa is adding 4 minutes of computation on each matrix cell, and although I think it's important to have it enabled to help people who do not instinctively spot bincompat problems, I'm okay with the idea that running it on JVM only is probably good enough at this point in the life of the library, in particular when some contributors run a LOT of commits on each PR ... 

I know Scala isn't the most "green" language, far from it, but I'd rather we tried to keep computation usage low-ish.

armanbilge:
4 minutes is kind of outrageous. From a brief glance at your logs I wonder if it's a misconfiguration: for example the MiMa check for 2.13+js seems to be compiling your entire project.

Baccata:
@armanbilge, you're right. 

@lewisjkl, in case you're not aware, we create cell-specific command [here](https://github.com/disneystreaming/smithy4s/blob/ed15fce6769be1aeaa4d4b70ddad0a3d13b4439d/project/Smithy4sPlugin.scala#L477-L485) which let us invoke specific things for the particular combo run by CI. I'd expect a MiMa command to appear in this list, and for GHA to have `run: sbt mimaReportBinaryIssuesIfRelevant_$BUILD_KEY` in it.

lewisjkl:
Looks like I need to tweak it a bit still, but I've got to head out for the day. I'll wrap it up on Monday.

lewisjkl:
I disabled mima checks as per the convo above, happy to change that back if we want. I also updated so it uses a command from the desiredCommands list.

Baccata (approved):

PR #419, opened 2022-08-30
httpResponseCode improvements & test

daddykotex:


daddykotex:
note: it was already implemented, but for some reason I did not commit my test, so here it is

Baccata (approved):

Issue #422, opened 2022-08-31
Codegen: mixin with default field causes downstream compilation errors


kubukoz:
```
$version: "2"

namespace hello

@mixin
structure HasContext {
  context: String = "default"
}

structure Input with [HasContext] {}
```

The generated code:

```scala
package hello


trait HasContext {
  def context: String = "default"
}
```

```scala
//pretty-printed


```

lewisjkl:
Ah yeah, the default value shouldn't be there on the mixin-related trait. We already handled this for optional types, but need to do it for defaulted members as well

kubukoz:
Related: if the mixin does _not_ have a default but the implementation does, there's another issue:

```
$version: "2"

namespace hello

@mixin
structure HasContext {
  context: String
}

structure Input with [HasContext] {}

apply Input$context @default("test")

```

```scala
// Input.scala
package hello

import smithy4s.Schema
import smithy4s.Hints
import smithy4s.schema.Schema.string
import smithy4s.ShapeId
import smithy4s.schema.Schema.struct
import smithy4s.ShapeTag

// doesn't compile, HasContext's `context` is an Option
case class Input(context: String = "test") extends HasContext
object Input extends ShapeTag.Companion[Input] {
  val id: ShapeId = ShapeId("hello", "Input")

  val hints : Hints = Hints.empty

  implicit val schema: Schema[Input] = struct(
    string.required[Input]("context", _.context).addHints(smithy.api.Default(smithy4s.Document.fromString("test"))),
  ){
    Input.apply
  }.withId(id).addHints(hints)
}
```

(in 0.15.2)

kubukoz:
would this be an appropriate solution to that last one? `context: Option[String] = Some("test")`

lewisjkl:
So initially, I was going to solve the latter half of this issue by making it so:

```kotlin
@mixin()
structure A {
 a: String
}

structure B with [A] {
 a: String = "test"
}
```

would become:

```scala
trait A {
 def a: Option[String]
}

case class B(a: Option[String] = Some("test")) extends A
```

However, then I realized the following is also possible:

```kotlin
@mixin()
structure A {
  a: String
}

structure B with [A] {
  @required
  a: String
}
```

At this point, the only thing I can think of doing here is to render this as:

```scala
trait A {
 def a: Option[String]
}

// does not extend A
case class B(a: String)
```
What do you think?

lewisjkl:
[...]

PR #423, opened 2022-08-31
fix issue with default in mixin

lewisjkl:
Handles part of #422. Fixes the issue where mixins were rendering defaults when they shouldn't

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/Renderer.scala
@@ -642,10 +642,11 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
       case Field(name, _, tpe, required, hints) =>
         val line = line"$tpe"
         val tpeAndDefault = if (required) {
-          val maybeRenderedDefault = hints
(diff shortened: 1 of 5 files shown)


modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -642,9 +642,10 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
       case Field(name, _, tpe, required, hints) =>
         val line = line"$tpe"
         val tpeAndDefault = if (required) {
-          val maybeDefault = hints
+          val maybeRenderedDefault = hints
             .collectFirst { case d @ Hint.Default(_) => d }
             .map(renderDefault)
```
daddykotex, line 647: not sure if it would have been better, but an alternative would have been to add `.filterNot(_ => noDefault)`

daddykotex (approved):

modules/codegen/src/smithy4s/codegen/Renderer.scala
lewisjkl: I like it, updated 👍

PR #424, opened 2022-08-31
Generate tests from test traits

daddykotex:


daddykotex:
I'll mark it as draft, because it seems like it never ends on CE2.
Also, I'd like to add more test and possibly refactor a little bit, but I need to take a step back a little bit, I was too close for too long.

Feel free to add test cases and refactor as you want in the meantime

daddykotex:
Oh well, it's because of CE3 and CE2 differences with Deferred, do we want to support both for `weaverTests` module?

I would need to add a `Compat` object

Baccata:
> Oh well, it's because of CE3 and CE2 differences with Deferred, do we want to support both for weaverTests module?

Yeah, for the time being, until we stop supporting CE2

daddykotex:
> > Oh well, it's because of CE3 and CE2 differences with Deferred, do we want to support both for weaverTests module?
> 
> Yeah, for the time being, until we stop supporting CE2

will do that right away

daddykotex:
This is atleast ready for review, I'm pretty sure it can be improved but it's a good foundation

modules/weaver-tests/src/smithy4s/weavertests/WeaverHttpRequestTestCase.scala
```diff
@@ -0,0 +1,249 @@
...
+      testCase.method.toLowerCase(),
+      request.method.name.toLowerCase()
+    )
+
+    List(
+      bodyAssert,
+      headerAssert.pure[IO],
+      uriAssert.pure[IO],
+      methodAssert.pure[IO]
+    ).sequence.map(_.reduce(_ && _))
```
kubukoz, line 138:
```suggestion
    ).combineAll
```

modules/weaver-tests/src/smithy4s/weavertests/WeaverHttpRequestTestCase.scala
```diff
@@ -0,0 +1,249 @@
...
+                new smithy4s.Interpreter[Op, IO] {
+                  def apply[I_, E_, O_, SE_, SO_](
+                      op: Op[I_, E_, O_, SE_, SO_]
+                  ): IO[O_] = {
+                    val (in, endpointInternal) = service.endpoint(op)
+
+                    if (endpointInternal.id == endpoint.id)
+                      requestDeferred.complete(in.asInstanceOf[I]) *>
+                        IO.pure(dummyOutput.asInstanceOf[O_])
+                    else IO.raiseError(new Throwable("Wrong endpoint called"))
```
kubukoz, line 168: I think we might want to make this go through Document in the future instead of casting, so that it works in Dynamic. It's fine right now

modules/weaver-tests/src/smithy4s/weavertests/WeaverHttpRequestTestCase.scala
```diff
@@ -0,0 +1,249 @@
...
+      testCase: HttpRequestTestCase,
+      inputFromDocument: Document.Decoder[I]
+  ): GeneratedTest = {
+    GeneratedTest(
+      name = endpoint.id.toString + "(server|request): " + testCase.id,
+      assertions = {
+        val dummyOutput = DefaultSchemaVisitor(endpoint.output)
+        val input = testCase.params
+          .map { inputFromDocument.decode(_).liftTo[IO] }
+          .getOrElse(IO.unit.asInstanceOf[IO[I]])
```
kubukoz, line 152: Can you do the cast on `Unit` level instead of `IO`?

kubukoz: alternatively we could decode a `Document.obj()` with `inputFromDocument` normally.

modules/weaver-tests/src/smithy4s/weavertests/WeaverHttpResponseTestCase.scala
```diff
@@ -0,0 +1,218 @@
...
+              actual
+                .map { values =>
+                  val actualValue = values.map(_.value).fold
+                  assert.eql(value, actualValue)
+                }
+                .getOrElse(
+                  failure(s"Header $name was not found in the response.")
+                )
+            }
+            .reduce(_ && _)
```
kubukoz, line 102:
```suggestion
            .toList.combineAll
```

modules/weaver-tests/src/smithy4s/weavertests/WeaverHttpResponseTestCase.scala
```diff
@@ -0,0 +1,218 @@
...
+        }
+        .getOrElse(success)
+
+    val statusAssert = assert.eql(testCase.code, response.status.code)
+
+    List(
+      bodyAssert,
+      headerAssert.pure[IO],
+      statusAssert.pure[IO]
+    ).sequence.map(_.reduce(_ && _))
```
kubukoz, line 112:
```suggestion
    ).combineAll
```

modules/weaver-tests/src/smithy4s/weavertests/WeaverHttpResponseTestCase.scala
```diff
@@ -0,0 +1,218 @@
...
+
+      val dummyInput = DefaultSchemaVisitor(endpoint.input)
+      val output =
+        testCase.params
+          .map {
+            outputFromDocument
+              .decode(_)
+              .liftTo[IO]
+          }
+          .getOrElse(IO.unit.asInstanceOf[IO[O]])
```
kubukoz, line 132: same idea: I'd decode an empty object `Document`

daddykotex: empty object is the same a `Unit`? I would see object  being rendered as `{}` where unit is ``

Baccata: > empty object is the same a Unit? 

It is indeed.

modules/weaver-tests/src/smithy4s/weavertests/WeaverHttpRequestTestCase.scala
```diff
@@ -0,0 +1,250 @@
...
+          .flatMap { case (requestDeferred, input) =>
+            type R[I_, E_, O_, SE_, SO_] = IO[O_]
+
+            val fakeImpl: smithy4s.Monadic[Alg, IO] =
+              service.transform[R](
+                new smithy4s.Interpreter[Op, IO] {
+                  def apply[I_, E_, O_, SE_, SO_](
+                      op: Op[I_, E_, O_, SE_, SO_]
+                  ): IO[O_] = {
+                    val (in, endpointInternal) = service.endpoint(op)
```
Baccata, line 164: we could make this easier by adding an unwrap method on the `Endpoint` interface, that would go like : 

``` 
def unwrap(op: Op[_,_,_,_,_]) : Option[I]
```

As for the output, I think this is one of the rare occurrences where it's fine to run the visitor dynamically (to avoid the cast)

modules/weaver-tests/src/smithy4s/weavertests/WeaverTests.scala
```diff
@@ -0,0 +1,108 @@
...
+    private def appliesTo[TC](testCases: List[TC])(
+        value: AppliesTo,
+        appliesTo: TC => Option[AppliesTo]
+    ): List[TC] =
+      testCases.collect {
+        case tc if appliesTo(tc).forall(_ == value) => tc
+      }
+  }
+
+  def recordTests(implicit ce: CompatEffect) = {
```
Baccata, line 103: I think this should be applied at construction without the user having to call it.

modules/weaver-tests/src/smithy4s/weavertests/WeaverTests.scala
```diff
@@ -0,0 +1,108 @@
...
+import smithy4s.Endpoint
+import smithy4s.Service
+import smithy4s.ShapeTag
+import smithy4s.UnsupportedProtocolError
+import weaver._
+
+case class GeneratedTest(name: String, assertions: IO[Expectations])
+
+abstract class WeaverTests[P, Alg[_[_, _, _, _, _]], Op[_, _, _, _, _]](
+    protocolTag: ShapeTag[_],
```
Baccata, line 34: I assume the `P` type parameter is meant to capture the protocol, so you can do 

``` 
abstract class WeaverTests[P: ShapeTag, Alg[_[_, _, _, _, _]], Op[_, _, _, _, _]](
  protocol: P, 
  ... 
```

modules/weaver-tests/src/smithy4s/weavertests/WeaverTests.scala
```diff
@@ -0,0 +1,108 @@
...
+    client: (
+        HttpApp[IO],
+        Uri
+    ) => Either[UnsupportedProtocolError, smithy4s.Monadic[Alg, IO]],
+    server: smithy4s.Monadic[Alg, IO] => Either[
```
Baccata, line 39: the client and server classes should be separate classes : protocol implementors might not be able to implement both sides, and they still need to be able to benefit from it

build.sbt
```diff
@@ -622,6 +625,30 @@ lazy val tests = projectMatrix
   )
   .http4sPlatform(allJvmScalaVersions, jvmDimSettings)
 
+lazy val weaverTests = projectMatrix
+  .in(file("modules/weaver-tests"))
```
Baccata, line 629: let's rename it to something else. The fact that we're using weaver is a detail, and the module doesn't aim at providing facilities for users of smithy4s to test their code in the context of the weaver framework.   

It could be `compliance` or something of the sort

kubukoz: At some point I think it'd be pretty great to provide test trait support that'd be test-framework-agnostic, too. Doesn't have to be in the initial appearance though

Baccata: The tedious thing is that there aren't any standard assertions. It means you have to come up with some data structure that does roughly the same thing as test-framework assertions. It's going be more boilerplaty (you have to create custom failure messages).

@daddykotex, if you're up for a little challenge, I think we should actually try and pursue it in this PR. It'd mean that you'd be producing something like a `List[ComplianceTest[F]]` that the framework of choice could run however it pleases . 

```
case class ComplianceTest[F[_]](name: String, run: F[ComplianceResult]) 
```

daddykotex: Ahh, ok, I did not think we were looking for something like that

Basically, a user would:

1. add dependency on this module: `smithy4s-compliance-tests`
2. add a depeendency on their model (smithy stuff)
3. the user can then generate a `List[ComplianceTest]` that they can register with the testing framework they want: (munit, weaver, w/e)

Right now though, we're only testing smithy4s stuff. What I mean is that we use a smithy4s generated http4s client along with a smithy4s generated http4s route and we _wire_ together to run assertion. That's great, but I'd argue it's not that useful to a user: we test the generated code.

I think that a more useful compliance test would allow the user to wire up an external server (maybe something not implemented with smithy4s) and/or an external client.

Does that make any sense?

Baccata: > I think that a more useful compliance test would allow the user to wire up an external server (maybe something not implemented with smithy4s) and/or an external client. 

> What I mean is that we use a smithy4s generated http4s client along with a smithy4s generated http4s route and we wire together to run assertion.

Well even if you're technically correct in that we're only testing smithy4s-derived (not-generated) constructs, the same tests can be used to test against several http protocols and backends, which might be implemented by third parties. It doesn't cost us much to lose the weaver dependency on a published artifact. As a matter of fact, despite being the maintainer of weaver, I feel more comfortable not publishing an artifact that has a dependency on weaver. 

So even if we are likely the only ones using this artifact, I'd much rather treat it like any other published artifact we are issuing in this library and prevent tight-coupling when not necessary. This philosophy has paid off handsomely so far and has avoided a large number of headaches.

daddykotex: my comment did not mean to imply that dropping weaver was not a good idea. I'll do it, this whole suggestion make sense, I meant something else:

currently, we test http4s client with http4s router, no true server  to client communication. should we expose a way for user to spawn the clients test but to target their own server located at for example http://localhost:8080

(basically, a way for the user to target an existing server (or the opposite, to spawn a server that they can hit with their client))

Baccata: Oh right, sorry, I see what you mean. 

We should copy what is done in PizzaSpec/PizzaClientSpec. Namely, [this entrypoint](https://github.com/disneystreaming/smithy4s/pull/424/files#diff-09b109d2d2c3403953cd05ef5d2c8966037aa476f4eb05d8caaea7c6b00e9577R62-R69) should be replace by something that takes 

``` 
client: Either[HttpApp[IO] => Resource[IO, smithy4s.Monadic[Alg, F]], Int => Resource[IO,smithy4s.Monadic[Alg, IO]]], 
server: smithy4s.Monadic[Alg, F] => Resource[F, (Client[F], Uri)] 
```

In other words, on the client side, it is : 

"give me either a function that takes HttpApp or a port and returns a Algebra, or a function that takes a port and returns an algebra, and I'll be able to spin a server to test your client"

On the server side, it is : "give me a function that, provided an algebra, spins a server and returns me with a client and the URI at which the server is running, and I'll use that client to test your server". 

This would be sufficient for client tests to target existing servers. Regarding server tests, we can obviously not test the implementation of the server-side of the protocol without controlling the business logic of server in question.

daddykotex: I think this is what is sort of blocking me. Let's focus on the **client request tests**.

Looking at the current implementation of the PizzaClientSpec :

```scala
def makeClient: Either[
    HttpApp[IO] => Resource[IO, PizzaAdminService[IO]],
    Int => Resource[IO, PizzaAdminService[IO]]
  ]
```
I went ahead and did most of the work using that, but then, when it came to it, I tried to wire it all together: but I cant implement the right side of this function. Whats the body of the following:

```scala
object Http4sPizzaClientSpec2 extends smithy4s.tests.PizzaClientSpec {
  def makeClient: Either[
    HttpApp[IO] => Resource[IO, PizzaAdminService[IO]],
    Int => Resource[IO, PizzaAdminService[IO]]
  ] = Right { port =>
    ???
  }
}
```

I have a feeling I'll feel very dumb in about just a minute, but that's okay :D

Baccata: @daddykotex the Right-hand-side is there for testing clients that are not backed by http4s. The port lets the implementor formulate a `localhost:$port` URI and that to some function, alongside an http-client, that will create a smithy4s instance. 

When the implementor provides a `Right`, the spec should instantiate an http-server at a random port, then use the lambda contained by the `Right` to get a client that will exercise that http-server. 

Hope that makes sense.

daddykotex: oh wow, ok, that's not how I was understanding it
I'll make more progress on that tomorrow morning

that makes a lot of sense

------

edit: wrote that last night, forgot to send :D

daddykotex:
To recap what I did lately:
* dropped weaver for assertion (implemented a poor man `eql` and use `Either[String, Unit]` as my compliance result
* dropped server test for the time being (I will include them in a separate PR once we land on a design for the client tests)

daddykotex:
hopefully we squash this when we merge :P

build.sbt
```diff
@@ -137,6 +138,7 @@ lazy val core = projectMatrix
     isMimaEnabled := true,
     allowedNamespaces := Seq(
       "smithy.api",
+      "smithy.test",
```
kubukoz, line 141: possible question: should `smithy.test` traits be included in smithy4s-core? Looks like they will.

kubukoz (approved):

daddykotex:
> @daddykotex I'm tagging this ready for review. I presume that server-tests would come in another PR ?

absolutely

build.sbt
daddykotex: oh, I did not think of that, that's a good point. I'll try and move the codegen for this namespace in the complaince-test module instead

daddykotex: did that! good catch

kubukoz:
I'll merge main into this to make sure it runs on Native all right.

daddykotex:
Good catch :P

kubukoz:
We could consider enabling "require branch to be up to date with main before merging", that'd avoid this kind of accident reaching main

kubukoz:
I'll take care of this - @Baccata I think we can apply the same idea as we did with the other tests, right? Get rid of concrete ember and implement it as `IO.fail` in our own tests?

modules/compliance-tests/test/src-ce2/smithy4s/compliancetests/WeaverComplianceTest.scala
```diff
@@ -0,0 +1,52 @@
...
+ *     https://disneystreaming.github.io/TOST-1.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.compliancetests
```
kubukoz, line 17: Why do we have two of the same file? Given we have Compat outsourced, could we just have one `src/` implementation of `WeaverComplianceTest`?

daddykotex: I've changed a lot of things over the lifetime of this PR, at some point I was extending `SomeClassThatNeededCompat` and I was not able to implement it w/o having essentially the same file in the two source trees. But now that's gone, I'm pretty sure you're right and this can be moved to the main source tree

modules/compliance-tests/src/smithy4s/compliancetests/ClientHttpComplianceTestCase.scala
```diff
@@ -0,0 +1,215 @@
...
+        .map(tc => clientRequestTest(endpoint, tc))
+    }
+  }
+
+  private val randomInt =
+    Resource.eval(IO(scala.util.Random.nextInt(9999)))
+
+  private val randomPort = randomInt.map(_ + 50000)
+
+  private def emberServer(app: HttpApp[IO]): Resource[IO, Int] =
```
kubukoz, line 199: I wanted to make this abstract lke we did for `PizzaClientSpec`, but this class is a concrete one.

It feels odd to have this kind of interface:

```scala
class ClientHttpComplianceTestCase[...](
...,
    getClient: Either[
      HttpApp[IO] => Resource[IO, smithy4s.Monadic[Alg, IO]],
      Int => Resource[IO, smithy4s.Monadic[Alg, IO]]
    ],
    mkServer: HttpApp[IO] => Resource[IO, Int]
)
```

because we might as well merge these two parameters into 

```scala
    getClient: Either[
      HttpApp[IO] => Resource[IO, smithy4s.Monadic[Alg, IO]],
      (HttpApp[IO] => Resource[IO, Int], Int => Resource[IO, smithy4s.Monadic[Alg, IO]])
    ],
```

which is almost exactly like

```scala
    getClient: Either[
      HttpApp[IO] => Resource[IO, smithy4s.Monadic[Alg, IO]],
      HttpApp[IO] => Resource[IO, Resource[IO, smithy4s.Monadic[Alg, IO]]
    ],
```

which could very well be 

```scala
    getClient: 
      HttpApp[IO] => Resource[IO, smithy4s.Monadic[Alg, IO]],
```

and that doesn't really help us.

How about this: we make this class abstract, make the server method abstract, then provide a concrete implementation that does http4s+ember but we don't use that implementation in our own tests?

In that case I'd actually have a single parameter here:

```scala
    getClient: 
      HttpApp[IO] => Resource[IO, smithy4s.Monadic[Alg, IO]],
```

and make it an `Either` in the http4s-specific implementation.

@Baccata @daddykotex

kubukoz: dunno how to make a PR to a fork from a fork, guess I'll just leave it here: https://github.com/kubukoz/smithy4s/tree/compliance-native-fix - this does what I described above.

daddykotex: thanks for the branch that helped to understand what you were saying above. I think you should push that to my branch, no worries.

that makes sense to me, and if you can get rid of the duplicated test classes (for compat) go ahead

daddykotex: I had to add a `locally` to avoid an unused import turned into a warning

I tried to use `@nowarn` but could not get it to work

Baccata: I'm confused as to the addition of an hard-coupling to Ember, tbh. In particular, doesn't that pose a problem in Native ?

kubukoz: It does not (for us) because the ember specific class isn't accessible from the test suite. For others it does (eg for an Akka http client implementation)

Baccata: I think we may want for it to remain abstract. It adds a little bit more burden downstream (like, 10 more lines), but it removes one dependency from the artifact. I do like the 

``` 
getClient: HttpApp[IO] => Resource[IO, smithy4s.Monadic[Alg, IO]]
```

As long as we document it proper, I'm happy to get rid of the Blaze-specific implementation. WDYT ?

kubukoz: (jfc why is the resolve button so accessible)

yeah, let's. We can do the http4s thingy in our own tests.

kubukoz: > As long as we document it proper

@daddykotex wanna take this one? 😅

daddykotex: yeah sure,
let me pick up every detail from this conversation :{

daddykotex: as I am writing the docs, I realize: would it make sense for the module to be called `compliance-testket`? maybe even `http-compliance-testkit` or something like that?

daddykotex: localhost_3000_smithy4s_docs_protocols_compliance-tests

In the screenshot, the side menu is misaligned, please disregard that, it's the screenshot engine that fails when you trigger the process and you're half way down the page

kubukoz: image

something sus going on here

daddykotex: ahh it's because I was on a watch loop, copied the file (so mdoc produced a file) then I renamed it (so mdoc produced a second file, leaving the copy behind in the output dir)

daddykotex: fixed 
Screen Shot 2022-09-29 at 09 49 18

kubukoz (approved):

kubukoz:
@Baccata you cool with this?

Baccata:
will review this week when I'm back to work

modules/compliance-tests/src/smithy4s/compliancetests/ClientHttpComplianceTestCase.scala
```diff
@@ -0,0 +1,181 @@
...
+                .map(fs2.Stream.emits(_))
+                .map(req.withBodyStream(_))
+                .flatMap(requestDeferred.complete(_))
+                .as(Response[IO]())
+            }
+            .orNotFound
+
+          getClient(app).use { client =>
+            // avoid blocking the test forever...
+            val recordedRequest = requestDeferred.get.timeout(1.second)
```
Baccata, line 151: you may want to add the timeout to to the whole computation, instead of the request recording, because if the `endpoint.wrap(in)` was to never return, you'd be deadlocked. 

Unless you do `parProductR` (assuming it exists)

Baccata, line 151: you may want to add the timeout to to the whole computation, instead of the request recording, because if the `endpoint.wrap(in)` was to never return, you'd be deadlocked. 

Unless you do `parProductR` (assuming it exists)

modules/compliance-tests/src/smithy4s/compliancetests/ClientHttpComplianceTestCase.scala
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  def allClientTests(
+  ): List[ComplianceTest[IO]] = {
+    service.endpoints.flatMap { case endpoint =>
+      endpoint.hints
+        .get(HttpRequestTests)
+        .map(_.value)
+        .getOrElse(Nil)
+        .filter(_.protocol == protocolTag.id.toString())
```
Baccata, line 176: Let's call `.show` instead of `.toString()`, we should never trust `toString`

PR #429, opened 2022-09-01
removed decline-effect lib , relying only on decline-core

yisraelU:
CommandIOApp is removed from Smithy4sCli . a method command was added to return a Command[IO[ExitCode]]
The alternative from what i could see is to return a List[Opts[_]] , I think this makes more sense , TY @lewisjkl


daddykotex:
I thought you had decided not to do it, glad you could pull it off in the end

yisraelU:
@daddykotex yeah thanks to @lewisjkl it ended up being a very minor change , still including the suspended effects inside the Opts.

daddykotex (approved):

modules/dynamic/src/smithy4s/dynamic/internals/DynamicEndpoint.scala
```diff
@@ -22,7 +22,7 @@ private[internals] case class DynamicEndpoint(
     id: ShapeId,
     input: Schema[DynData],
     output: Schema[DynData],
-    override val errorable: Option[Errorable[DynData]],
+    override val errorable: Option[Errorable[DynData]] = None,
```
Baccata, line 25: Revert this change.

modules/decline/src/Smithy4sCli.scala
```diff
@@ -128,7 +120,17 @@ class Smithy4sCli[Alg[_[_, _, _, _, _]], Op[_, _, _, _, _]](
       }
   }
 
-  def main: Opts[IO[ExitCode]] = service.endpoints
+  private def opts: Opts[IO[ExitCode]] = service.endpoints
     .foldMapK(endpointSubcommand(_))
 
+  def command: Command[IO[ExitCode]] = {
```
Baccata, line 126: if we're getting rid of the `CommandIOApp` thing, couldn't this be polymorphic ?

Baccata:
I presume this change breaks what's documented in https://disneystreaming.github.io/smithy4s/docs/protocols/deriving-cli, it'd be nice if have the documentation of the "end of the world" snippet be compiled, at the very least.

modules/dynamic/src/smithy4s/dynamic/internals/DynamicEndpoint.scala
yisraelU: sorry my bad

modules/decline/src/Smithy4sCli.scala
yisraelU: yes I believe so

yisraelU:
I think Smithy4s Cli should be renamed to Command as its more appropriate now

build.sbt
```diff
@@ -724,7 +724,6 @@ lazy val Dependencies = new {
 
   object Decline {
     val core = Def.setting("com.monovore" %%% "decline-effect" % "2.3.0")
```
Baccata, line 726: this is still pointing to decline-effect

modules/decline/src/standalone/Smithy4sSimpleStandaloneCli.scala
```diff
@@ -34,14 +35,16 @@ import smithy4s.decline.util.PrinterApi
   *   parameters if you need program-wide config that doesn't fit into a specific operation's input
   *   (operation inputs are mapped to Opts automatically by all Smithy4sCli implementations).
   */
-class Smithy4sSimpleStandaloneCli[Alg[_[_, _, _, _, _]], Op[_, _, _, _, _]](
-    impl: Opts[smithy4s.Monadic[Alg, IO]]
+class Smithy4sSimpleStandaloneCli[Alg[_[_, _, _, _, _]], Op[_, _, _, _, _], F[
```
Baccata, line 38: So, this is kind of rendered useless by not extending `IOApp` anymore : it doesn't have a main method ... therefore I'm not too sure about this PR. We could have a `decline-effect` module that would re-create the CommandIOApp UX, otherwise this class ought to go and the documentation should showcase an example of using CommandIOApp in combination with the polymorphic construct.

modules/docs/src/03-protocols/04-deriving-cli.md
```diff
@@ -39,7 +40,8 @@ structure Greeting {
 }
 ```
 
-Lets implement the HelloWorld service
+Lets implement the HelloWorld service , we will use cats.effect.IO for our effect type.
+
 ```scala
```
Baccata, line 45: I'd like for the examples in this documentation to be using `mdoc` to validate that the code is indeed valid.

build.sbt
yisraelU: that is probably a copy/paste error. ill fix

modules/decline/src/standalone/Smithy4sSimpleStandaloneCli.scala
yisraelU: Why is it useless, its a pure convenience that handles Entrypoint creation and printer default.
I can switch to a factory method in companion object, but I still see value.
As far as docs are concerned , we currently have IOApp documented . I think CommandIOApp is not necessary as we are providing a command , one can call CommandIOApp.run and pass in the Command , but I think thats a separate concern

modules/docs/src/03-protocols/04-deriving-cli.md
yisraelU: I will look into how to do that

modules/decline/src/Smithy4sCli.scala
```diff
@@ -16,11 +16,12 @@
 
 package smithy4s.decline
 
-import cats.effect.ExitCode
-import cats.effect.IO
 import cats.implicits._
+import cats.instances.all._
```
kubukoz, line 20: This import is redundant with `cats.implicits._`

modules/decline/src/Smithy4sCli.scala
```diff
@@ -111,24 +103,49 @@ class Smithy4sCli[Alg[_[_, _, _, _, _]], Op[_, _, _, _, _]](
...
-              .flatMap(printer.printOutput)
-              .as(ExitCode.Success)
-              .recoverWith {
-                case e if endpoint.errorable.flatMap(_.liftError(e)).nonEmpty =>
-                  printer
-                    .printError(e)
-                    .as(ExitCode.Error)
-              }
+
+          FO.flatMap(printer.printOutput)
```
kubukoz, line 112: Doesn't `printer.printOutput.flatMap` work? It should.

modules/decline/src/Smithy4sCli.scala
```diff
@@ -111,24 +103,49 @@ class Smithy4sCli[Alg[_[_, _, _, _, _]], Op[_, _, _, _, _]](
...
+        .get[Documentation]
+        .map(_.value)
+        .getOrElse(s"Command line interface for ${service.id.show}"),
+      helpFlag = true
+    )(opts)
+  }
+}
+
+object Smithy4sCli {
+  def Smithy4sSimpleStandaloneCli[Alg[_[_, _, _, _, _]], Op[_, _, _, _, _], F[
```
kubukoz, line 138: Maybe let's rename this to `def standalone` - all the other words (except simple) appear in the context

kubukoz (changes requested):

modules/decline/src/Smithy4sCli.scala
yisraelU: It should but it wasnt compiling , Ill try to figure that out

modules/docs/src/03-protocols/04-deriving-cli.md
```diff
@@ -62,7 +62,7 @@ object HelloWorldServiceInstance{
...
 object app extends IOApp.Simple{
   override def run(args: List[String]): IO[Unit] = {
-    val helloCommand:Command[IO[Unit]] =Smithy4sSimpleStandaloneCli(Opts(HelloWorldServiceInstance.simple)).command
+    val helloCommand:Command[IO[Unit]] = Smithy4sCli.standalone(Opts(HelloWorldServiceInstance.simple)).command
```
kubukoz, line 65: Needs formatting

modules/docs/src/03-protocols/04-deriving-cli.md
```diff
@@ -57,8 +57,17 @@ object HelloWorldServiceInstance{
...
+ - To utilize the helper method , wrap the service in an instance of Opts and pass it into ```Smithy4sCli.standalone```
+ - there are 2 methods available on the ```Smithy4sCli``` instance
+    - ```opts``` which will  provide an Opts[F[Unit]] for the service
+    - ```command``` which will provide a Command[F[Unit]] for the service . This uses defaults from the Smithy spec
+      - command name will use be the service name 
+      - if documentation comments are available on the service , they will be used as the command help text
+ ```scala mdoc:silent
+   val serviceWrappedInOpts = Opts(HelloWorldServiceInstance.simple)
+   val helloCommand :Command[IO[Unit]] =Smithy4sCli.standalone(serviceWrappedInOpts).command
+```
```
daddykotex, line 70:
```suggestion
 - There is a convenient helper method ```Smithy4sCli.standalone``` to make it easier to construct Cli's using  defaults.
 - To utilize the helper method, wrap the service in an instance of `Opts` and pass it into ```Smithy4sCli.standalone```
 - there are 2 methods available on the ```Smithy4sCli``` instance
    - ```opts``` which will provide an `Opts[F[Unit]]` for the service
    - ```command``` which will provide a `Command[F[Unit]]` for the service. This uses defaults from the Smithy spec
      - command name will use the service name 
      - if documentation comments are available on the service , they will be used as the command help text
```scala mdoc:silent
   val serviceWrappedInOpts = Opts(HelloWorldServiceInstance.simple)
   val helloCommand: Command[IO[Unit]] = Smithy4sCli.standalone(serviceWrappedInOpts).command
```
```

**Review by daddykotex — APPROVED** [review]
ready to go in for me. kudos

**Review by lewisjkl — APPROVED** [review]

File: modules/decline/src/Smithy4sCli.scala (hunk 7 above)
Comment (daddykotex) [reply modules/decline/src/Smithy4sCli.scala:110]: basically @kubukoz :

```
[info] [decline3] compiling 6 Scala sources to /Users/David.Francoeur/workspace/dev/smithy4s/modules/decline/target/jvm-3/classes ...
[error] [decline3] -- [E008] Not Found Error: /Users/David.Francoeur/workspace/dev/smithy4s/modules/decline/src/Smithy4sCli.scala:110:15 
[error] [decline3] 106 |            service
[error] [decline3] 107 |              .asTransformation[GenLift[F]#λ](entrypoint.interpreter)(
[error] [decline3] 108 |                endpoint.wrap(input)
[error] [decline3] 109 |              )
[error] [decline3] 110 |              .flatMap(printer.printOutput)
[error] [decline3]     |            ^
[error] [decline3]     |value flatMap is not a member of smithy4s.GenLift[F]#λ[I, E, O, endpoint.SI, endpoint.SO].
[error] [decline3]     |An extension method was tried, but could not be fully constructed:
[error] [decline3]     |
[error] [decline3]     |    cats.implicits.toFlatMapOps[
[error] [decline3]     |      ([SO] =>> smithy4s.GenLift[F]#λ[I, E, O, endpoint.SI, SO])
[error] [decline3]     |    , A](
[error] [decline3]     |      Smithy4sCli.this.service.asTransformation[
[error] [decline3]     |        smithy4s.GenLift[Smithy4sCli.this.F]#λ
[error] [decline3]     |      ](entrypoint.interpreter).apply[I, E, O, endpoint.SI, endpoint.SO](
[error] [decline3]     |        endpoint.wrap(input)
[error] [decline3]     |      )
[error] [decline3]     |    )(evidence$1)
[error] [decline3]     |
[error] [decline3]     |where:    F is a type in class Smithy4sCli with bounds <: [_] =>> Any
[error] [decline3] one error found
[error] [decline3] one error found
[error] [decline3] (decline3 / Compile / compileIncremental) Compilation failed
```

Comment (yisraelU) [reply modules/decline/src/Smithy4sCli.scala:110]: exactly , i wasnt either . both ugly

File: modules/decline/src/core/OptsVisitor.scala [hunk 11]
```diff
@@ -229,10 +229,9 @@ object OptsVisitor extends SchemaVisitor[Opts] { self =>
       case b: BijectionSchema[a, b] =>
         list(shapeId, hints, b.underlying).map(_.map(b.bijection.to))
       case s: Schema.RefinementSchema[a, b] =>
-        list(shapeId, hints, s.underlying).map(
-          _.map(value =>
-            s.refinement(value)
-              .fold(message => throw RefinementFailed(message), identity)
+        list(shapeId, hints, s.underlying).mapValidated(
```
daddykotex, line 232: yessir

modules/decline/test/src/smithy4s/decline/OptsSchematicSpec.scala
```diff
@@ -34,41 +32,35 @@ import smithy4s.schema.Schema._
...
-  def parseOpts[A](
-                    schema: Schema[A]
-                  )(
-                    input: String*
-                  ): Either[Help, EitherThrow[A]] = Command("test-command", "test-header")(
-    schema.compile[OptsVisitor.OptsF[EitherThrow, *]](compiler)
-  )
-    .parse(input)
+  def parseOpts[A](schema: Schema[A])(input: String*): Either[Help, A] =
+    Command("test-command", "test-header")(schema.compile[Opts](OptsVisitor)).parse(input) match {
```
daddykotex, line 41: you probably want to drop the following

also, I learnt about that not too long ago: https://www.scala-lang.org/api/2.13.x/scala/util/ChainingOps.html

modules/decline/test/src/smithy4s/decline/OptsSchematicSpec.scala
```diff
@@ -16,14 +16,12 @@
 
 package smithy4s.decline
```
daddykotex, line 17: maybe move to `smithy4s.decline.core` and rename to `OptsVisitor`

PR #430, opened 2022-09-02
Hosttrait

yisraelU:
Going to add methods to create smithy4s client from host trait , allowing user to define scheme(http/s) in next commit


Baccata:
Does not fit the scope of the open-source project.

Issue #420, opened 2022-08-31
Sharing specifications: Unable to resolve smithy traits


anderha:
Hi,

I am currently facing a problem sharing smithy files in sbt projects. 

I define the api of project A in smithy files and would like to reuse them in another project B to generate client code. 

To publish the smithy files I followed the docs on [sharing specifications](https://disneystreaming.github.io/smithy4s/docs/overview/sharing-specs/). Unfortunately the compile step in project B fails with an error: `Unable to resolve trait 'smithy4s.api#simpleRestJson'. If this is a custom trait, then it must be defined before it can be used in a model.`

The smithy files are defined as follows:
```
namespace com.example.api.definition
use smithy4s.api#simpleRestJson

@simpleRestJson
service PingAPIController {
    version: "1.0.0",
    operations: [ping]
}

@http(method: "GET", uri: "/", code: 200)
@readonly
operation ping {}
```

Compiling the smithy files in project A works like a charm. In project B I have other local smithy files for which the codegen does also work.

Technical details
- Smithy4sCodegenPlugin version: 0.14.2
- Scala version: 2.13.8

Am I missing something?

Baccata:
@anderha, closing because this is a duplicate of https://github.com/disneystreaming/smithy4s/issues/391

Baccata:
@anderha actually I'm not 100% sure it's a duplicate. Would you mind providing a minimal reproduction in the form of a github repository we could have a look at ?

anderha:
Hey @Baccata I have created a sample project: https://github.com/anderha/smithy-package-sample

I have placed a manual into the README.md to replicate the error

Steps to make it work:
- remove the `@simpleRestJson` in https://github.com/anderha/smithy-package-sample/blob/main/modules/api-definition/src/main/resources/META-INF/smithy/healthAPIController.smithy#L5 
- publish another version with `sbt publishLocal` in `modules/api-definition`
  - increase version number in https://github.com/anderha/smithy-package-sample/blob/main/modules/api-definition/build.sbt#L1
  - reference increased version number in https://github.com/anderha/smithy-package-sample/blob/main/build.sbt#L12
 - `sbt compile` in root
 
It might be a duplicate, but I am not sure also

Baccata:
Right yeah it's a partial overlap with the other issue. Head over [there](https://disneystreaming.github.io/smithy4s/docs/overview/sharing-specs#packaging-specifications-in-order-to-share-them) to see understand what's missing. 

[...]

Issue #391, opened 2022-08-22
Sharing smithy files in a multi-project build


danielkarch:
Hi, 

I have a project with two submodules, `moduleA` and `moduleB`. Both use smithy4s and have their smithy files in their respective `src/main/smithy` directory. Now with the introduction of refinements, I would like to create a new module, `refinements`, such that both `moduleA` and `moduleB` depend on `refinements`. 
Unfortunately I cannot get this to work. I defined the `smithy4sInputDir` in the `smithyRefinements` module as
```
Compile / smithy4sInputDir := baseDirectory.value / "src/main/resources/META-INF/smithy"
```
and defined a file `refinements.smithy`
```
namespace refinements

structure Foo {}
```
as well as a `manifest` referring to that file. The `smithyRefinements / smithy4sCodegen` task runs fine and generates `Foo.scala`, i.e., smithy4s finds the `refinements.smithy` file.

In the `build.sbt` I let `moduleA` depend on `smithyRefinements`:
```scala
lazy val smithyRefinements = ...
lazy val moduleA = (project in file("modules/moduleA"))
  .dependsOn(smithyRefinements)
  ...
  .enablePlugins(Smithy4sCodegenPlugin)
```
In `modules/moduleA/src/main/smithy/ModuleA.smithy` I depend on `refinements#Foo`:
```
namespace moduleA

use refinements#Foo

structure Blah {
    foo: Foo
}
```
but when I run `moduleA / smithy4sCodegen`, I get the error
```
[error] (moduleA / Compile / smithy4sCodegen) software.amazon.smithy.model.validation.ValidatedResultException: Result contained ERROR severity validation events: 
[error] [ERROR] moduleA#Blah$foo: member shape targets an unresolved shape `refinements#Foo` | Target /.../modules/moduleA/src/main/smithy/ModuleA.smithy:8:5
```

What am I missing? I was able to depend on smithy files defined in another library, as described [here](https://github.com/disneystreaming/smithy4s/blob/main/modules/docs/src/01-overview/05-sharing-specs.md), but not when they are defined in another module in the same project.

daddykotex:
This is tricky. What you're essentially trying to is to share the code generation output from one project (refinements) into another project ( A and/or B). This works, but you get blocked before because the code generation process from the module A & B can't import the smithy model. When you define a dependency on the refinement model, it does not tell smithy4s that there are `smithy` file in there what would need to be added to the model to resolve it.

[...]

PR #432, opened 2022-09-02
Facilitate sharing specs across multi-module builds.

Baccata:
This adds the necessary mechanisms and logic to facilitate sharing smithy specs in multi-module settings.

The first important part is the addition of a resource generation mechanism that adds the local smithy files to the output resources, and synthesise smithy files containing only metadata, keeping track of the namespaces that have been generated by smithy4s. Because this extends the openapi generation to be able to generate more resources, I've renamed/revamped the arguments in `CodegenArgs` and `CodegenCommand` to be more fitting. 

Because the metadata in question is an array node, it is monoidal in nature and the values are aggregated when loading models. These values are used to skip code generation in downstream Smithy4s calls, as the downstream modules can trust that the upstream one will have generated the relevant package.

Then, the SBT plugin changes to load internal dependency outputs as local jars. This has the effect of pulling the models from upstream modules in the current model assembler, allowing users to depend on the shapes it contains.

resolves https://github.com/disneystreaming/smithy4s/issues/420
resolves https://github.com/disneystreaming/smithy4s/issues/391 

TODOS : 

- [x] Documentation 
- [x] Opt-out SBT setting 
- ~[ ] Think about whether some other metadata should be added to track the version of Smithy4s that generated the sources, to try and protect against incompatibilities, as this mechanism could be extended to use Smithy4s in the same style as protobuf, where specs are packaged alongside the generated code.~ in a subsequent PR
- [x] Think really hard about the implications of all of this before committing : let's get it out there and collect feedback


modules/codegen/src/smithy4s/codegen/Codegen.scala
```diff
@@ -36,34 +37,42 @@ object Codegen { self =>
...
     )
 
-    val scalaFiles = if (!args.skipScala) {
-      Codegen.generate(model, args.allowedNS, args.excludedNS).map {
-        case (relPath, name, outputString) =>
-          val fileName = name + ".scala"
-          val scalaFile = (args.output / relPath / fileName)
-          os.write.over(scalaFile, outputString, createFolders = true)
-          scalaFile
+    val (scalaFiles, smithyResources) = if (!args.skipScala) {
```
daddykotex, line 40: should we read:

```
val scalaFiles = if (args.skipScala) { Nil } else { .. }
val openApiFiles = if (args.skipOpenapi) { Nil } else { .. }
val resources = if (args.skipResources) { Nil } else { .. }
```

What I mean, is there, it the current setup `--skip scala` will not produce resources files, which seems counter-intuitive. Maybe there should not be a way to skip `resources. They would be skipped if you `--skip scala`

modules/codegen/src/smithy4s/codegen/SmithyResources.scala
```diff
@@ -0,0 +1,61 @@
...
+private[smithy4s] object SmithyResources {
+
+  def produce(
+      resourceOutputFolder: os.Path,
+      localSmithyFiles: List[os.Path],
+      namespaces: List[String]
+  ): List[os.Path] = {
+
+    val smithyFolder = resourceOutputFolder / "META-INF" / "smithy"
+    val namespaceTrackingSmithyFiles = namespaces.map { ns =>
```
daddykotex, line 19: would it be problematic to write on smithy file per module instead? If we have 2 modules, and each modules has N namespaces, we'll write 2 * N smithy file with a metadata entry with an array containing one element (namespace). Why not writing one file per module, with an array containing N elements (namespace)

modules/codegen/src/smithy4s/codegen/SmithyResources.scala
```diff
@@ -0,0 +1,61 @@
...
+/**
+  * This construct aims at adding metadata information to the classpath to let Smithy4s
+  * know about code that may have been generated in upstream modules.
+  *
+  * A "smithy4sGenerated" metadata holds a list of all namespaces that will have already
+  * been generated by smithy4s.
+  */
+private[smithy4s] object SmithyResources {
+
+  def produce(
```
daddykotex, line 12: I like when this kind of syncing logic, wipe the target folder and start from scratch. This version will leave leftover files behind if some namespace disappear from a module

modules/codegen-cli/src/smithy4s/codegen/cli/CodegenCommand.scala
```diff
@@ -53,21 +54,17 @@ object CodegenCommand {
...
-
-  val skipScalaOpt =
-    Opts
-      .flag(
-        long = "skip-scala",
-        help = "Indicates that scala code generation should be skipped"
+      .options[String](
+        long = "skip",
+        help =
+          "Indicates that some files typesshould be skipped during generation"
```
kubukoz, line 62:
```suggestion
          "Indicates that some files types should be skipped during generation"
```

modules/codegen/src/smithy4s/codegen/SmithyResources.scala
kubukoz: that's the status quo really, if you regenerate you may end up with leftover files that should no longer be generated

kubukoz (commented):
I'm too braindead to read through this today, but might take a peek later

modules/codegen-plugin/src/sbt-test/codegen-plugin/multimodule/project/plugins.sbt
```diff
@@ -0,0 +1,11 @@
+sys.props.get("plugin.version") match {
+  case Some(x) =>
+    addSbtPlugin("com.disneystreaming.smithy4s" % "smithy4s-sbt-codegen" % x)
+  case _ =>
```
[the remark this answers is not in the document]
Baccata: PR is obviously still in draft state and hasn't gone through clean-up yet

modules/codegen/src/smithy4s/codegen/Codegen.scala
Baccata: Nah, the resources in question are MEANT to be tied to the presence of Scala code, in particular because one of them literally exists to say "the Scala code for these particular namespaces was already generated, you can skip it downstream". People who want to package resources without Scala code should simply disable the Smithy4s plugin and manually create the resources, like we do in the protocol module. 

These options will be documented anyway, and I don't expect users to tweak them often.

modules/codegen/src/smithy4s/codegen/SmithyResources.scala
Baccata: It's problematic to wipe because other source/resource generators may exist and tap into the same directories (which some editors, like Intellij, happen to hardcode in their Scala support). That's why you have to return the list of files you've generated in source/resource generators, so that the build tool can ignore potential leftovers.

modules/codegen/src/smithy4s/codegen/SmithyResources.scala
Baccata: I chose to write one file per namespace because otherwise what would you name the aggregating file ? If you choose a fixed name, thinking "it's okay because there will only be one per module", then you put your users who build fat jars through the pain of having to set assembly strategies for that particular file, and it's not trivial to aggregate several smithy files into one. 

Then you could ask "why not write a manifest file then, that can be simply concatenated during assembly ?". Well for starters, it still puts the users through the hassle, and secondly, it means that we need to write bespoke code to load the manifest file in question. Even if it's not that hard to write, it's still annoying to have to do it when smithy is perfectly able to aggregate metadata in from several files without us having to do anything. 

Moreover, these files are not really meant to be seen by anyone, so it really doesn't matter if there are N of them. The prepending of "smithy4s" reduces the risk of collision with actual files from userland.

modules/codegen-cli/src/smithy4s/codegen/cli/CodegenCommand.scala
```diff
@@ -53,21 +54,17 @@ object CodegenCommand {
       )
       .orNone
 
-  val skipOpenapiOpt =
+  val skipOpts =
```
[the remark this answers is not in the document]
Baccata: That suggestion doesn't offer an objective argument as to why it is better than what is currently in place, which is arguably good enough. 

> Likewise, we could have a default SkipOptions.default which would be much more explicit than a Set.empty

I'm not sure I see the value of making it more explicit, CodegenArgs are not something that is manipulated much by users, we instantiate it manually for the SBT plugin, and for the Mill plugin (that we'll release at some point), so trying to bring more clarity to it (when named arguments make it perfectly clear) is something I consider rather superficial.

modules/codegen/src/smithy4s/codegen/SmithyResources.scala
Baccata: After a thought, rendering the smithy manifest creates this problem as well. I'm reconsidering your ask, not for the reason you're putting forward (because that particular reason does not present a particularly good rationale from a technical standpoint)

daddykotex: I did not think about the name nor the assembly issues, that's for sure!

daddykotex (approved):
I pushed a commit to address a doc issue, but otherwise it LGTM

build.sbt
```diff
@@ -823,7 +825,7 @@ def genSmithyImpl(config: Configuration) = Def.task {
   val outputDir = (config / genSmithyOutput).?.value
     .getOrElse((config / sourceManaged).value)
     .getAbsolutePath()
-  val openapiOutputDir =
+  val resourceOutputDir =
     (config / genSmithyOpenapiOutput).?.value
```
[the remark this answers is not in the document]
daddykotex: I can always address this on my mill PR

did my comment make sense to you?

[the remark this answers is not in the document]
Baccata: Yeah sure you can go ahead and rename in your PR :)

PR #438, opened 2022-09-05
Exhaustively capture the codegen input for cache invalidation

Baccata:
This exhaustively captures the codegenArgs as a way to invalidate SBT's cache to re-trigger codegen every time something changes.

This uses SBT's JsonFormat, converting os-lib's paths into file hashes. In theory, when a local file changes (say, a dependency snapshot), the cache should now be invalidated and trigger a regeneration.

Additionally, this disable jar caching in the Smithy Model assembler that loads local jars. This prevents users who experiment with snapshot artifacts from having to exit/restart SBT when issuing a new local snapshot . 

Unfortunately, this is pretty hard to test in an automated manner.

### DIFF
--- a/modules/codegen-plugin/src/smithy4s/codegen/JsonConverters.scala
+++ b/modules/codegen-plugin/src/smithy4s/codegen/JsonConverters.scala
@@ -1,0 +1,69 @@
+package smithy4s.codegen
+
+import sjsonnew._
+import BasicJsonProtocol._
+import sbt.FileInfo
+import sbt.HashFileInfo
+import sjsonnew._, LList.:*:
+
+private[smithy4s] object JsonConverters {
+
+  implicit val pathFormat: JsonFormat[os.Path] = {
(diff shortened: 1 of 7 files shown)


modules/codegen-plugin/src/smithy4s/codegen/JsonConverters.scala
```diff
@@ -0,0 +1,86 @@
...
+ *  limitations under the License.
+ */
+
+package smithy4s.codegen
+
+import sjsonnew._
+import BasicJsonProtocol._
+import sbt.FileInfo
+import sbt.HashFileInfo
+import sjsonnew._
```
lewisjkl, line 23:
```suggestion
import sjsonnew._
import BasicJsonProtocol._
import sbt.FileInfo
import sbt.HashFileInfo
```
nitpick, redundant import

lewisjkl (approved):

PR #441, opened 2022-09-08
Remove private modifier

yisraelU:
Includes merge from main
add val for service inside ServiceBuilder 
https://github.com/disneystreaming/smithy4s/blob/edcadcd5f24da800372cf647f34cb42b775bf780/modules/http4s/src/smithy4s/http4s/SimpleProtocolBuilder.scala#L58


kubukoz:
I'd say let's do the merge from main separately

PR #442, opened 2022-09-08
support root imports in refinements

lewisjkl:
Allow imports and classpaths in refinements to contain `_root_` at the beginning of them.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -79,13 +79,13 @@ lazy val docs =
     .settings(
       mdocIn := (ThisBuild / baseDirectory).value / "modules" / "docs" / "src",
       mdocVariables := Map(
-        "VERSION" -> (if (isSnapshot.value)
-                        previousStableVersion.value.getOrElse(
-                          throw new Exception(
-                            "No previous version found from dynver"
-                          )
-                        )
-                      else version.value),
+        // "VERSION" -> (if (isSnapshot.value)
+        //                 previousStableVersion.value.getOrElse(
+        //                   throw new Exception(
+        //                     "No previous version found from dynver"
+        //                   )
+        //                 )
+        //               else version.value),
         "SCALA_VERSION" -> scalaVersion.value,
         "HTTP4S_VERSION" -> Dependencies.Http4s.http4sVersion.value
       ),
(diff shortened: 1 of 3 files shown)


daddykotex:
can you add a test case here: https://github.com/disneystreaming/smithy4s/blob/main/modules/protocol-tests/test/src/smithy4s/api/validation/RefinementTraitValidatorSpec.scala

lewisjkl:
> can you add a test case here: https://github.com/disneystreaming/smithy4s/blob/main/modules/protocol-tests/test/src/smithy4s/api/validation/RefinementTraitValidatorSpec.scala

I am going to need to give this more thought and will come up with something tomorrow. Using the model assembler out of the box doesn't check things like that pattern trait when creating the model. I think I can either use the validator for that directly or think of another way though

Baccata:
>  Using the model assembler out of the box doesn't check things like that pattern trait when creating the model.  

Weird. It should though.

Baccata (approved):

lewisjkl:
CleanShot 2022-09-09 at 11 33 52
Yeah it is definitely possible that I am missing something, but that's the behavior I've seen

lewisjkl:
I realized that my issue was that I was calling it `providerInstance` when it was actually `providerImport` and the field was optional. That's why it wasn't validating it (as it shouldn't have)

lewisjkl:
One more thing, I originally added these regexes because then users will be able to see in their editors if they have the wrong syntax or not. But at the same time, the error is not super helpful since it just spits out a regex at the user.. might be worth switching this to a validator OR removing the check altogether since the compiler will check it any way.. idk thoughts?

daddykotex:
Woaw, kind of a bummer nobody noticed that! I looked at the doc, it's correct, it uses: `providerImport`

lewisjkl:
> Woaw, kind of a bummer nobody noticed that! I looked at the doc, it's correct, it uses: `providerImport`

Yeah for sure, one of those things that is hard to catch. I wonder/wish if there is a way in smithy to error if you have fields in a trait that aren't mapped to anything in the trait definition...

daddykotex:
> Yeah for sure, one of those things that is hard to catch. I wonder/wish if there is a way in smithy to error if you have fields in a trait that aren't mapped to anything in the trait definition...

maybe that's implementable as a validator 🗡️

PR #447, opened 2022-09-09
Servicebuilder revised Feedback Requested

yisraelU:
Ok so this is obviously not backwards compatible at all as it changes the interface that is exposed to users
I'm marking it draft to signal that this is more about a direction then a production ready PR
If we go ahead with this PR, we can keep the existing methods and mark deprecated , those methods will use the builder internally.
the benefits of this interface is that its fluent and achieves some code reuse.
Changes made

- I removed HttpApp from method args since its trivial for a user to wrap their HttpApp themselves and it really makes the underlying code more convoluted. We cannot run the conversion for them as we currently have a EffectCompat constraint on F while Client.fromHttpApp needs F[_]:Async. (@kubukoz says it doesnt really need it 🤷 ) .That said this does come with some implications to consider as this now blocks a user from utilizing an HttpApp where their F is doesnt have an instance of Async. I leave that for discussion about whether this is a serious problem
- I defaulted the URI to localhost:8080(though I can understand why this might be unacceptable)) to provide an initial valid state and not have to deal with mutation/effects as one might have to with standard builders. As a side effect this actually makes it nicer when testing as you get a standard localhost:8080 ..
-

### DIFF
--- a/modules/docs/src/01-overview/02-quickstart.md
+++ b/modules/docs/src/01-overview/02-quickstart.md
@@ -146,10 +146,10 @@ object ClientImpl extends IOApp.Simple {
 
   val helloWorldClient: Resource[IO, HelloWorldService[IO]] = for {
     client <- EmberClientBuilder.default[IO].build
-    helloClient <- SimpleRestJsonBuilder(HelloWorldService).clientResource(
-      client,
-      Uri.unsafeFromString("http://localhost:9000")
-    )
+    helloClient <- SimpleRestJsonBuilder(HelloWorldService)
+    .client(client)
+    .uri(Uri.unsafeFromString("http://localhost:9000"))
+    .resource
   } yield helloClient
 
   val run = helloWorldClient.use(c =>
--- a/modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
+++ b/modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
@@ -21,7 +21,6 @@ package internals
 import cats.syntax.all._
 import org.http4s.EntityDecoder
 import org.http4s.EntityEncoder
-import org.http4s.HttpApp
 import org.http4s.Request
 import org.http4s.Response
 import org.http4s.Uri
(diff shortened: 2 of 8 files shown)


modules/docs/src/01-overview/02-quickstart.md
```diff
@@ -146,10 +146,10 @@ object ClientImpl extends IOApp.Simple {
...
   val helloWorldClient: Resource[IO, HelloWorldService[IO]] = for {
     client <- EmberClientBuilder.default[IO].build
-    helloClient <- SimpleRestJsonBuilder(HelloWorldService).clientResource(
-      client,
-      Uri.unsafeFromString("http://localhost:9000")
-    )
+    helloClient <- SimpleRestJsonBuilder(HelloWorldService)
+    .client(client)
+    .uri(Uri.unsafeFromString("http://localhost:9000"))
+    .resource
```
lewisjkl, line 152: This is really nice, I like this api better than the prior one

modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
```diff
@@ -61,25 +60,20 @@ private[smithy4s] object SmithyHttp4sClientEndpoint {
...
-    clientOrApp match {
-      case Left(client) =>
-        client
-          .run(inputToRequest(input))
-          .use { response =>
-            outputFromResponse(response)
-          }
-      case Right(httpApp) =>
-        httpApp.run(inputToRequest(input)).flatMap(outputFromResponse)
-    }
```
lewisjkl, line 82: Love getting rid of clientOrApp

lewisjkl (commented):
I really like the direction 👍

Baccata:
Not bad at all, @yisraelU. Will give it a proper review when I come back from covid

Baccata:
@yisraelU, I'm happy for us to roll with it, provided you put in place what you suggested : 

> If we go ahead with this PR, we can keep the existing methods and mark deprecated , those methods will use the builder internally.

Baccata:
> If we go ahead with this PR, we can keep the existing methods and mark deprecated , those methods will use the builder internally.
 
 Yeah, happy to do that actually 👍 

> I defaulted the URI to localhost:8080(though I can understand why this might be unacceptable)) to provide an initial valid state and not have to deal with mutation/effects as one might have to with standard builders. 

Yeah that one is a bit trickier but I think acceptable tradeoff considering the larger benefits.

Baccata:
(sorry, misclick)

yisraelU:
@Baccata 
I updated accordingly
I will add more docs/clarity tomorrow .
The only caveat is that even with the deprecated methods we will still not be able to include the HttpApp based methods

Baccata:
> The only caveat is that even with the deprecated methods we will still not be able to include the HttpApp based methods 

That's fine I think, the method was mostly exposed to facilitate testing, at a point when using `Client.fromHttpApp` wasn't working (probably because of something we were doing wrong, but I'm failing to remember what). 

I think it's good to go, you just need to run `scalafmt` again. I know it's annoying, we might look into setting up a [pre-commit hook](https://medium.com/zyseme-technology/code-formatting-scalafmt-and-the-git-pre-commit-hook-3de71d099514) in the future, to even prevent local commits if the code is ill formatted.

Baccata (approved):

PR #451, opened 2022-09-13
Fix collision

yisraelU:
This PR resolves a bug that causes collisions to go undetected when the identifier contains a Path
This affects all schema ref's as the name is A.schema or A.underlyingSchema


modules/http4s/src/smithy4s/http4s/SimpleProtocolBuilder.scala
```diff
@@ -55,7 +55,7 @@ abstract class SimpleProtocolBuilder[P](val codecs: CodecAPI)(implicit
   class ServiceBuilder[
       Alg[_[_, _, _, _, _]],
       Op[_, _, _, _, _]
-  ] private[http4s] (service: smithy4s.Service[Alg, Op]) {
```
Baccata, line 58: probable mis-hap ?

yisraelU: OMG , definitely yes. will fix

Baccata:
@yisraelU, you can't use `distinctBy` as it's not part of the Scala 2.12 stdlib

Baccata (approved):

PR #455, opened 2022-09-15
Cached Schema Visitor

yisraelU:
will work on tests next
 Used type Any for the value part of Map instead of Existential type for compatibility with Scala 3


modules/core/src/smithy4s/schema/SchemaVisitor.scala
```diff
@@ -18,6 +18,7 @@ package smithy4s
 package schema
 
 import Schema._
+import scala.collection.mutable.{Map => MMap}
```
Baccata, line 21: use java HashMap instead, it's more performant.

modules/core/src/smithy4s/schema/SchemaVisitor.scala
```diff
@@ -45,6 +46,13 @@ trait SchemaVisitor[F[_]] extends (Schema ~> F) { self =>
 
 }
 
+abstract class CachedSchemaVisitor[F[_]] extends SchemaVisitor[F] {
+  self =>
+  private val cache: MMap[ShapeId, Any] = MMap.empty
+
+  override def apply[A](schema: Schema[A]): F[A] = cache.getOrElseUpdate(schema.shapeId, super.apply(schema)).asInstanceOf[F[A]]
```
Baccata, line 53: 👍

modules/core/src/smithy4s/schema/SchemaVisitor.scala
```diff
@@ -45,6 +46,13 @@ trait SchemaVisitor[F[_]] extends (Schema ~> F) { self =>
 
 }
 
+abstract class CachedSchemaVisitor[F[_]] extends SchemaVisitor[F] {
+  self =>
+  private val cache: MMap[ShapeId, Any] = MMap.empty
```
Baccata, line 51: don't cache by ShapeId, the key should be the schema itself. That is because several schemas can have the same ShapeId, such as Lazy/Bijection/Refinements

yisraelU:
@Baccata should we use scala.collection.concurrent.TrieMap
I noticed thats whats used in the memoize function on PolyFunction

daddykotex:
10 000 feets measurements, this pr:

CI / Test CE2 2_12 (jvm) - in 6m
CI / Test default 2_12 (jvm) - in 4m
CI / Test CE2 2_13 (jvm) - in 3m
CI / Test default 2_13 (jvm) - in 8m 
CI / Test default 2_13 (js) - in 5m
CI / Test default 3_0 (jvm) - in 8m [biggest diff here]
CI / Test default 3_0 (js) - in 7m
CI / Test default 3_0 (native) - in 5m

PR on main:
CI / Test CE2 2_12 (jvm) - in 6m
CI / Test default 2_12 (jvm) - in 5m
CI / Test CE2 2_13 (jvm) - in 3m
CI / Test default 2_13 (jvm) - in 8m
CI / Test default 2_13 (js) - in 6m
CI / Test default 3_0 (jvm) - in 11m 
CI / Test default 3_0 (js) - in 8m
CI / Test default 3_0 (native) - in 6m

it can be **** though, take that with a grain of salt

lewisjkl:
> or maybe a java HashMap would be better actually

Then we'd need to use something different for native/JS, right?

Baccata:
> 10 000 feets measurements, this pr:

Once again : there is hardly anything in our tests that exercises big schemas, which is the usercase that benefits the most from this. 

This is an issue that is experienced by our users who have big schemas (talking tens of operations involving hundreds of datatypes). 

The biggest schemas in our codebase are the AWS ones, and we don't exercise them in any capacity at runtime

yisraelU:
> @yisraelU in this case, a mutable map is sufficient (or maybe a java HashMap would be better actually).
> 
> The reason `TrieMap` is used in the PolyFunction is that the access might be happening concurrently, because we don't control the access patterns. In this case, the schemas are very much visited in a single synchronous thread. That's what should happen even if we don't control all the implementations of SchemaVisitor.

I think we'd get best perf from AnyRefMap , however this would require a type constraint on F

daddykotex:
> 

They're never materialized in any tests?

Baccata:
TBH, thinking twice about it we don't need much perf out of it as all the access doesn't happen on the hot-path. So whatever uses the least memory would be a good call.

Baccata:
>  They're never materialized in any tests?

Not that I know of

kubukoz:
> This is an issue that is experienced by our users who have big schemas (talking tens of operations involving hundreds of datatypes).

You can have even one operation and one deeply nested schema. You'd still get a huge improvement out of this

modules/core/test/src/smithy4s/CachedSchemaVisitorSpec.scala
```diff
@@ -25,6 +27,7 @@ class CachedSchemaVisitorSpec() extends FunSuite {
         tag: CollectionTag[C],
         member: Schema[A]
     ): Option[C[A]] = {
+      self(member)
```
kubukoz, line 30: why not `this(member)`?

modules/core/test/src/smithy4s/CachedSchemaVisitorSpec.scala
```diff
@@ -0,0 +1,160 @@
...
+    val counter = new AtomicInteger(0)
+    val visitor = new TestSchemaVisitor(counter)
+    val schema = Schema.int
+    val result: Option[Int] = visitor(schema)
+    assertEquals(result, None)
+    assertEquals(counter.get(), 1)
+  }
+
+  test(
+    "should hit the cache on second visit with the same schema resulting in counter = 1"
```
kubukoz, line 117: Another idea - checking instance equality of the result of two visits?

modules/core/test/src/smithy4s/CachedSchemaVisitorSpec.scala
yisraelU: i just use self instead of this by habit , there is no difference in this case

modules/core/test/src/smithy4s/CachedSchemaVisitorSpec.scala
yisraelU: I guess I would need to change my visitor for that , it currently returns None

Baccata (approved):

PR #457, opened 2022-09-16
0.16.0 feature branch

Baccata:


daddykotex (approved):
Sorry again for the botched initial release

I still consider myself lucky that I noticed it lol, only because I was trying to write pretty release notes with link to the site.

PR #462, opened 2022-09-17
Add smithy4s models to upstream model validation

Baccata:
If users publish smithy4s generated code, we package the models in the artifacts for sharing with downstream projects. The problem has to do with Smithy4s not including its own models when loading upstream models, which do not necessarily depend on smithy4s-protocol.

This change adds the smithy4s-protocol model when loading upstream models, just in case.

Fixes https://github.com/disneystreaming/smithy4s/issues/420

### DIFF
--- a/modules/codegen-plugin/src/sbt-test/codegen-plugin/multimodule-staged/foo/src/main/smithy/foo.smithy
+++ b/modules/codegen-plugin/src/sbt-test/codegen-plugin/multimodule-staged/foo/src/main/smithy/foo.smithy
@@ -1,0 +1,12 @@
+$version: "2.0"
+
+namespace foo
+
+use smithy4s.api#uuidFormat
+
+structure Foo {
+  a: Integer
+}
+
+@uuidFormat
+string MyUUID
--- a/modules/codegen/src/smithy4s/codegen/ModelLoader.scala
+++ b/modules/codegen/src/smithy4s/codegen/ModelLoader.scala
@@ -59,6 +59,7 @@ object ModelLoader {
       val upstreamModel = Model
         .assembler()
         .discoverModels(upstreamClassLoader)
+        .addClasspathModels(currentClassLoader, false)
         // disabling cache to support snapshot-driven experimentation
         .putProperty(ModelAssembler.DISABLE_JAR_CACHE, true)
         .assemble()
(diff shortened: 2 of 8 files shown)


modules/codegen-plugin/src/sbt-test/codegen-plugin/multimodule-staged/foo/src/main/smithy/foo.smithy
```diff
@@ -0,0 +1,12 @@
+$version: "2.0"
+
+namespace foo
+
+use smithy4s.api#uuidFormat
```
daddykotex, line 5: instead of a fully fledged test, can we reuse the ones in `modules/codegen-plugin/src/sbt-test/codegen-plugin/multimodule/` ? the change is minimal and that would avoid one extra sbt-test

modules/codegen/src/smithy4s/codegen/ModelLoader.scala
```diff
@@ -59,6 +59,7 @@ object ModelLoader {
       val upstreamModel = Model
         .assembler()
         .discoverModels(upstreamClassLoader)
+        .addClasspathModels(currentClassLoader, false)
```
daddykotex, line 62: is the comment on line 57 still relevant?

basically, we load to models: 
- one with upstream deps + smithy4s (via current class loader)
- one with everything on the classpath (with discovery enabled)

kubukoz (approved):

kubukoz:
What do we need to merge this?

modules/codegen/src/smithy4s/codegen/ModelLoader.scala
Baccata: you could be cleaned up

Baccata:
@daddykotex we'll remove the comment in a subsequent PR, no need to trigger lots of compilation for it.

modules/codegen/src/smithy4s/codegen/ModelLoader.scala
kubukoz: Sounds like a threat lmao

Baccata: 😆 I meant 'yeah, could be cleaned up'

Issue #421, opened 2022-08-31
Dynamic: unknown discriminator apply


kubukoz:
```
$version: "2"

namespace demo

@mixin
structure B {
  id: String
}

structure A with [B] {}

apply A$id @documentation("foo")
```

Running `dump-model` on this results in the following shape being generated (among others):

```json
{
  "type": "apply",
  "traits": {
    "smithy.api#documentation": "foo"
  }
}
```

Decoding it fails with:

```
Unknown discriminator: apply (path: .shapes.demo#A$id)
```

I _think_ that shape should not be generated, but instead the target field should have a trait added to it. Currently it's impossible to load a dynamic model that was dumped from anything with an `apply` in it.

Baccata:
wut, I wasn't aware `apply` was a possibility in the json syntax. You are correct in what needs to happen though. Mind PR-ing it in ?

kubukoz:
If nobody beats me to it :) I'll assign myself here if I pick it up

kubukoz:
Looks like #463 resolves this for my purpose (`apply` simply doesn't get generated, it seems), but since it's an actual decoding failure (and not simply an ignored field) I'd say we should still do something about this

Baccata:
We need a pre-processing step, that would inline the apply. We also need a pre-processing step that would inline the mixins. 
It's really doable, just a bit tedious.

kubukoz:
Just to be clear, this would be part of the dynamic model compiler? Essentially replicating what we're now doing in the part that serializes the jvm smithy model?

Baccata:
Yes.

PR #463, opened 2022-09-18
added ModelTransformer to  flatten and merge all mixins in the model

### LINKED ISSUE #453: Mixins are not inherited in Dynamic Model
Issue : https://github.com/disneystreaming/smithy4s/blob/b257e4cbfab5d46b9484fa8682a6619776e1eb93/modules/dynamic/src-jvm/DynamicSchemaIndexPlatform.scala#L31

Summary: 
When using the software.amazon.smithy.model.shapes.ModelSerializer from AWS , any members added in ( aka inherited) via a Mixin will be removed from members object and stored in a separate json object for Mixins. This makes the ModelSerializer a bad candidate for piping information to the Dynamic model, as it currently stands Mixins will not be included in the Dynamic model
See the Serialized Model below ,  MyStruct does not contain a member MyInt

```

    "smithy": "2.0",
    "shapes": {
        "test#MyInt": {
            "type": "structure",
            "members": {
                "theInt": {
                    "target": "smithy.api#Integer"
                }
            },
            "traits": {
                "smithy.api#mixin": {}
            }
        },
        "test#MyStruct": {
            "type": "structure",
            "mixins": [
                {
                    "target": "test#MyInt"
                }
            ],
            "members": {
                "str": {
                    "target": "smithy.api#String"
                }
            }
        }
    }
}
```

yisraelU:
per https://github.com/awslabs/smithy/issues/1408 adding a Model Transformer sends the mixins to where they belong 
closes #453

### DIFF
--- a/modules/dynamic/src-jvm/DynamicSchemaIndexPlatform.scala
+++ b/modules/dynamic/src-jvm/DynamicSchemaIndexPlatform.scala
@@ -18,6 +18,7 @@ package smithy4s.dynamic
 
 import smithy4s.http.PayloadError
 import software.amazon.smithy.model.shapes.ModelSerializer
+import software.amazon.smithy.model.transform.ModelTransformer
 
 private[dynamic] trait DynamicSchemaIndexPlatform {
   self: DynamicSchemaIndex.type =>
@@ -28,7 +29,9 @@ private[dynamic] trait DynamicSchemaIndexPlatform {
   def loadModel(
       model: software.amazon.smithy.model.Model
   ): Either[PayloadError, DynamicSchemaIndex] = {
-    val node = ModelSerializer.builder().build.serialize(model)
+    val flattenedModel =
+      ModelTransformer.create().flattenAndRemoveMixins(model);
+    val node = ModelSerializer.builder().build.serialize(flattenedModel)
     val document = NodeToDocument(node)
     smithy4s.Document
       .decode[smithy4s.dynamic.model.Model](document)


kubukoz:
After merging https://github.com/disneystreaming/smithy4s/pull/462 into this, I can confirm it makes mixins work in Smithy Playground: https://github.com/kubukoz/smithy-playground/pull/95

kubukoz:
I guess this will have to be consistent between mixins and [apply](https://github.com/disneystreaming/smithy4s/issues/421).

> per use case basis

I assume it would still have to be done in DynamicCompiler. But in general I think you're onto something - if we were to, for example, support generating code from a dynamic model, we'd need mixins intact/

kubukoz:
> 5.2. Mixins are an implementation detail of the model
> Mixins are an implementation detail of models and are only intended to reduce duplication in Smithy shape definitions. Mixins do not provide any kind of runtime polymorphism for types generated from Smithy models. Smithy model transformations like code generation or converting to other model formats like OpenAPI SHOULD completely elide mixins by flattening the model.

From [this page](https://awslabs.github.io/smithy/2.0/spec/mixins.html#mixins-are-an-implementation-detail-of-the-model).

I'd say if we were following these rules, we could flatten the mixins right here... but we're not following that in the codegen so I'm not sure.

Baccata:
>  we're not following that in the codegen so I'm not sure. 

We're not following this rule in codegen to be able to generate inheritance, which users have been asking for. But that aspect is utterly irrelevant in a dynamic context, as this information is not used at a protocol level.

kubukoz (approved):

PR #470, opened 2022-09-25
Native Http4s

kubukoz:


modules/aws-kernel/src-jvm/smithy4s/aws/SysEnvPlatformCompat.scala
```diff
@@ -17,5 +17,5 @@
 package smithy4s.aws.kernel
 
 trait SysEnvPlatformCompat {
-  def readEnvVar(envVar: String): Option[String] = sys.env.get(envVar)
+  def readEnvVar(envVar: String): Option[String] = Option(System.getenv(envVar))
```
kubukoz, line 20: This isn't strictly necessary, but I think when SecurityManager is enabled, reading an individual variable instead of the whole environment might be better for permission scope.

modules/aws-kernel/src-native/smithy4s/aws/CryptoPlatformCompat.scala
```diff
@@ -0,0 +1,95 @@
...
+  def toHexString(binary: Binary): String =
+    binary.map("%02x".format(_)).mkString
+
+  def sha256HexDigest(message: Array[Byte]): String =
+    unsafe.Zone { implicit z =>
+      val outLength: Ptr[CUnsignedInt] = unsafe.alloc[CUnsignedInt]()
+      val digest: Ptr[CUnsignedChar] =
+        unsafe.alloc[CUnsignedChar](CryptoPlatformCompat.EVP_MAX_MD_SIZE)
+      val dataLen = message.length.toLong
+      val cInput: CString = arrayToCString(message)
```
kubukoz, line 38: Notable change from the original source of this file: the parameter was a String and this was calling `toCString`.

modules/aws-kernel/src-native/smithy4s/aws/CryptoPlatformCompat.scala
```diff
@@ -0,0 +1,95 @@
...
+    unsafe.Zone { implicit z =>
+      val outLength: Ptr[CUnsignedInt] = unsafe.alloc[CUnsignedInt]()
+      val digest: Ptr[CUnsignedChar] =
+        unsafe.alloc[CUnsignedChar](CryptoPlatformCompat.EVP_MAX_MD_SIZE)
+
+      val cKey = arrayToCString(key)
+      val keylen = key.length.toLong
+
+      val cData: CString = toCString(data)
+      val datalen = data.length.toUInt
```
kubukoz, line 64: this is also different - here and in other places I had to do `.toUInt` to convert to `CSize` - apparently it's a change from SN 0.4 that made the type unsigned: https://github.com/scala-native/scala-native/pull/1949/

modules/aws-kernel/src-native/smithy4s/aws/CryptoPlatformCompat.scala
```diff
@@ -29,21 +29,67 @@ trait CryptoPlatformCompat {
...
+        cKey,
+        keylen,
+        cData,
+        datalen,
+        digest,
+        outLength
+      )
+
+      val out = Array.fill[Byte]((!outLength).toInt)(0)
+      for (i <- 0 until (!outLength).toInt) out(i) = digest(i.toLong).toByte
```
Baccata, line 77: we should make this a while loop

modules/aws-kernel/src-jvm/smithy4s/aws/SysEnvPlatformCompat.scala
Baccata, line 20: 👍

modules/aws-kernel/src-native/smithy4s/aws/CryptoPlatformCompat.scala
```diff
@@ -0,0 +1,95 @@
...
+      val dataLen = message.length.toLong
+      val cInput: CString = arrayToCString(message)
+
+      val ctx = crypto.EVP_MD_CTX_new()
+      crypto.EVP_DigestInit(ctx, crypto.EVP_sha256())
+      crypto.EVP_DigestUpdate(ctx, cInput, dataLen.toUInt)
+      crypto.EVP_DigestFinal(ctx, digest, outLength)
+
+      val out = Array.fill[Byte]((!outLength).toInt)(0)
+      for (i <- 0 until (!outLength).toInt) out(i) = digest(i.toLong).toByte
```
Baccata, line 46: ditto (while loop to avoid allocating a range)

Baccata (approved):

kubukoz:
this breaks BC for the tests module, does that warrant 0.17 or is the tests module exempt from BC guarantees?

Baccata:
> this breaks BC for the tests module, does that warrant 0.17 or is the tests module exempt from BC guarantees?

Nah, we're not guaranteeing bincompat on the test module as this time, we're only guaranteeing bincompat all the way to the http4s module.

daddykotex:
interesting PR, good work @kubukoz

PR #472, opened 2022-09-26
Bump mdoc and use check-link-hygiene

daddykotex:


kubukoz:
I've noticed that `--check` kills the entire sbt when it fails, so this probably does as well...

daddykotex:
yeah, I've heard that in some places, although the logs show that mdoc did not produce any errors

Screen Shot 2022-09-26 at 12 10 56

daddykotex:
well that's it, and it sucks, will fix upstream, what the hell

kubukoz:
please fix `--check` too and I'll give you... a dollar, maybe two! :D

daddykotex:
loll, look at that: https://github.com/scalameta/mdoc/pull/704/files#diff-2cc284217cfe2824b47da67d6861a834438a39647641d6c7493e1604c464dfb6L73

PR #475, opened 2022-09-27
Decline: add native support

kubukoz:
I think this is the last module we can actually add Native support to - all the others have some JVM-only dependencies we won't be able to get rid of.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -782,8 +782,11 @@ lazy val Dependencies = new {
   }
 
   object Decline {
-    val core = Def.setting("com.monovore" %%% "decline" % "2.3.1")
-    val effect = Def.setting("com.monovore" %%% "decline-effect" % "2.3.0")
+    val declineVersion = "2.3.1"
+
+    val core = Def.setting("com.monovore" %%% "decline" % declineVersion)
+    val effect =
+      Def.setting("com.monovore" %%% "decline-effect" % declineVersion)
   }
   object Fs2 {
     val core: Def.Initialize[ModuleID] =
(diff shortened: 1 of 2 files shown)


kubukoz:
can't reproduce the failure locally :(

build.sbt
```diff
@@ -781,7 +782,7 @@ lazy val Dependencies = new {
   }
 
   object Decline {
-    val core = Def.setting("com.monovore" %%% "decline" % "2.3.0")
+    val core = Def.setting("com.monovore" %%% "decline" % "2.3.1")
     val effect = Def.setting("com.monovore" %%% "decline-effect" % "2.3.0")
```
daddykotex, line 786: should we bump the one below also? or introduce a val to share

plokhotnyuk:
> can't reproduce the failure locally :(

Please consider replacing the following regex by some simple validation function:
https://github.com/disneystreaming/smithy4s/blob/main/modules/core/test/src-jvm/smithy4s/TimestampSpec.scala#L124

build.sbt
kubukoz: good catch, it is indeed used in docs. I'll bring in a val

build.sbt
```diff
@@ -47,27 +47,27 @@ lazy val root = project
...
+  protocol,
+  protocolTests,
+  openapi,
+  `aws-kernel`,
+  aws,
+  `aws-http4s`,
+  `codegen-cli`,
+  dynamic,
+  testUtils
+).flatMap(_.projectRefs)
```
lewisjkl, line 70: This is some 3022 thinking right here 🎉

lewisjkl (approved):

PR #476, opened 2022-09-28
Bump mdoc and use check-link-hygiene

daddykotex:
superseeds https://github.com/disneystreaming/smithy4s/pull/472


daddykotex:
that was a quick turn around

lewisjkl (approved):

PR #480, opened 2022-09-30
Pin docusaurus-lunr-search

### LINKED ISSUE #479: Search bar is broken
The search bar doesn't go off the "loading state". Tried several browsers. Asked somebody else too, same result. 

Probably a library update broke it for us. 

Screenshot 2022-09-30 at 07 12 28

daddykotex:
I can't explain why it does not work anymore. There are a bunch of open issues but I could not make sense of them.

I proceeded as follow:

* revert one dependabot commit
* `rm -rf .docusaurus node_modules/`
* run `yarn && yarn swizzle docusaurus-lunr-search SearchBar --danger --wrap && yarn build` (--wrap appears at some point)
* `yarn serve` (yarn start does not seem to work)

Closes #479

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,7 +144,7 @@ jobs:
         run: sbt "docs/mdoc"
 
       - name: Build Microsite
-        run: cd ./modules/website && yarn && yarn swizzle docusaurus-lunr-search SearchBar --danger && yarn build
+        run: cd ./modules/website && yarn && yarn swizzle docusaurus-lunr-search SearchBar --danger --wrap && yarn build
 
       - name: Publish Microsite
         uses: peaceiris/actions-gh-pages@v3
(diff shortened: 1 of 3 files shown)


daddykotex:
Well to be fair, this is the exact problem: https://github.com/praveenn77/docusaurus-lunr-search/issues/85

Changes in some utils.js file generate invalid file names

yisraelU (approved):

daddykotex:
Would it make sense to do microsite publish outside the release workflow (because it only happens when we create a new tag)

lewisjkl:
> Would it make sense to do microsite publish outside the release workflow (because it only happens when we create a new tag)

I'm fine with doing that. Do we have an easy way to trigger that?

Baccata:
I'll release from local when I'm home (still travelling back home, been awake for 25 hours and am not home yet). We should set up a dedicate job for it.

Baccata:
This PR doesn't conclude the issue though, the lunr dependency needs to be ignored for dependabot updates. Alternatively, we could replace it with  : https://github.com/easyops-cn/docusaurus-search-local

PR #485, opened 2022-10-03
Fix codegen resource outputs for Smithy files in subdirectories

kubukoz:
In 0.16.2, it seems like putting a Smithy file in a module breaks `smithy4sCodegen` of downstream modules. This will (hopefully) demonstrate that downstream modules fail to run codegen when such a situation arises.

The error is:

```
software.amazon.smithy.model.loader.ModelImportException: Unable to open Smithy model import URL: jar:file:/tmp/sbt_b32e5aac/foo/target/scala-2.13/foo_2.13-0.1.0-SNAPSHOT.jar!/META-INF/smithy/foodir

caused by

java.io.FileNotFoundException: JAR entry META-INF/smithy/foodir not found in /tmp/sbt_b32e5aac/foo/target/scala-2.13/foo_2.13-0.1.0-SNAPSHOT.jar
```

You can also reproduce this issue by trying to compile [this](https://github.com/kubukoz/demos/tree/smithy4s-subdir).

The generated manifest for this file structure:

```
smithy
- dir
| - coredir.smithy
```

is:

```
smithy4s.tracking.smithy
dir
```

### DIFF
--- a/modules/mill-codegen-plugin/src/smithy4s/codegen/mill/Smithy4sModule.scala
+++ b/modules/mill-codegen-plugin/src/smithy4s/codegen/mill/Smithy4sModule.scala
@@ -74,7 +74,7 @@ trait Smithy4sModule extends ScalaModule {
   def smithy4sCodegen: T[(PathRef, PathRef)] = T {
 
     val specFiles = if (os.exists(smithy4sInputDir().path)) {
-      os.walk(smithy4sInputDir().path, skip = _.ext != "smithy")
+      os.list(smithy4sInputDir().path)
     } else Seq.empty
 
     val scalaOutput = smithy4sOutputDir().path
--- a/modules/mill-codegen-plugin/src/smithy4s/codegen/mill/Smithy4sModule.scala
+++ b/modules/mill-codegen-plugin/src/smithy4s/codegen/mill/Smithy4sModule.scala
@@ -73,9 +73,7 @@ trait Smithy4sModule extends ScalaModule {
 
   def smithy4sCodegen: T[(PathRef, PathRef)] = T {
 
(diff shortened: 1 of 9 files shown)


Baccata:
Now that you've exhibited the problem, care to offer a fix ? The code that needs amending is probably [there](https://github.com/disneystreaming/smithy4s/blob/main/modules/codegen/src/smithy4s/codegen/SmithyResources.scala)

kubukoz:
yeah will take a glance later on

modules/mill-codegen-plugin/src/smithy4s/codegen/mill/Smithy4sModule.scala
```diff
@@ -74,7 +74,7 @@ trait Smithy4sModule extends ScalaModule {
   def smithy4sCodegen: T[(PathRef, PathRef)] = T {
 
     val specFiles = if (os.exists(smithy4sInputDir().path)) {
-      os.walk(smithy4sInputDir().path, skip = _.ext != "smithy")
+      os.list(smithy4sInputDir().path)
```
Baccata, line 77: why list at all ? What happens if we just pass the folder ?

kubukoz: if we just pass the folder, we need extra work to strip `smithy` from the directory structure.

You could argue it's better because we move more stuff out of the plugins into the shared code (`SmithyResources`)...

pushed a commit that does that, lemme know if it looks better. I also added a `.filter(os.isFile(_))` in the `os.walk` call so that we never list directories in `smithy/manifest`.

Baccata (approved):

PR #486, opened 2022-10-04
Extract primitive string encoding/decoding logic. 

Baccata:
We're gonna need this exact same logic for XML parsing/serialisation.


kubukoz (approved):

lewisjkl (approved):

daddykotex:
that's a great cleanup

PR #491, opened 2022-10-05
Opt out of local dependencies

### LINKED ISSUE #484: Codegen forces upstream module compilation
In a project like this:

```scala
val core = project

val root = project
  .in(file("."))
  .dependsOn(core)
```

Running `smithy4sCodegen` on root level triggers compilation of `core`. You can prove this by putting non-compiling Scala code in `core`, in which case `smithy4sCodegen` will fail. No smithy files necessary. `.dependsOn(core % "compile->compile")` doesn't change this.

Can't reproduce this in 0.15.3. My guess is that this is caused by https://github.com/disneystreaming/smithy4s/pull/432/ and might be a deliberate change in behavior (but it's not documented as such).

[Reproduction repo](https://github.com/kubukoz/demos/tree/smithy4s-compilation-trigger)

kubukoz:
- Allow opting out of dependency aggregation
- Add temporary implementation for Mill

Closes #484

### DIFF
--- a/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
+++ b/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
@@ -144,11 +144,10 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
 
     val localDependencyJars = Def.taskIf {
       if ((conf / smithy4sAggregateLocalDependencies).value) {
-        (conf / internalDependencyAsJars)
-          .map(
-            _.map(_.data).map(os.Path(_)).toList
-          )
-          .value
+        (conf / internalDependencyAsJars).value
+          .map(_.data)
+          .map(os.Path(_))
+          .toList
       } else List.empty[os.Path]
     }.value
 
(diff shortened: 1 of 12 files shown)


modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -131,13 +142,18 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
...
       (conf / smithy4sExcludedNamespaces).?.value.map(_.toSet)
     val updateReport = (conf / update).value
-    val internalDependencyJars =
-      (conf / internalDependencyAsJars).value.seq
-        .map(_.data)
-        .map(os.Path(_))
-        .toList
+
+    val localDependencyJars = Def.taskIf {
+      if ((conf / smithy4sAggregateLocalDependencies).value) {
```
Baccata, line 147: I wonder if rather than having a flag setting, we should make a TaskKey for this and give users the control over the task itself. This would increase consistency between SBT and Mill and would give more control to the user, if they wanted to add some unrelated jars, or have a more fine-grain filtering than "all or nothing"

kubukoz: yeah, that makes is easier on our side too. I suppose it's better this way because the non-default option is just `Nil`, not another task or something.

Baccata (approved):

PR #493, opened 2022-10-06
Support more HTTP methods, keep more details about HTTP matching failures

### LINKED ISSUE #492: Not all HTTP methods are supported
e.g. `@http(method: "HEAD")` results in `HttpMethod.fromString` failing, thus causing e.g. the creation of a http4s client to throw with `s"Operation ${endpoint.name} is not bound to http semantics"`.

kubukoz:
Closes #492.


kubukoz:
http4s supports many more methods, maybe we should list them all? Or even support arbitrary strings for method names and try to parse them into http4s at derivation time?

Baccata:
@kubukoz yeah we could make it an open-enumeration maybe ? Have a catch-all `Other` case ?

kubukoz:
sounds alright - maybe I'll remove the ones that we don't explicitly use and move them into the open.

kubukoz:
> sounds alright - maybe I'll remove the ones that we don't explicitly use and move them into the open.

Did it for the new ones only - didn't want to break bincompat by dropping existing ones.

kubukoz:
Not sure why removing the methods from `smithy4s.internal` (latest commit, f543008) didn't require an issue filter in MiMa...

kubukoz:
yo github wtf, I just pushed a merge from series/0.17 into this 🤔

kubukoz:
Looks like that did it...

build.sbt
```diff
@@ -231,8 +231,7 @@ lazy val scalacheck = projectMatrix
       Dependencies.collectionsCompat.value,
       Dependencies.Scalacheck.scalacheck.value
     ),
-    libraryDependencies ++= munitDeps.value,
-    testFrameworks += new TestFramework("weaver.framework.CatsEffect")
```
kubukoz, line 235: This is set by the project plugin

modules/core/src/smithy4s/http/internals/ErrorCodeSchemaVisitor.scala
```diff
@@ -20,7 +20,7 @@ package internals
 
 import smithy4s.schema._
 
-private[smithy4s] class ErrorCodeSchemaVisitor(
+private[http] class ErrorCodeSchemaVisitor(
```
kubukoz, line 23: More narrow visibility scope here - let me know if that makes sense

modules/http4s/src/smithy4s/http4s/SimpleProtocolBuilder.scala
```diff
@@ -60,23 +60,6 @@ abstract class SimpleProtocolBuilder[P](val codecs: CodecAPI)(implicit
 
     def client[F[_]: EffectCompat](client: Client[F]) =
       new ClientBuilder[Alg, Op, F](client, service)
-    @deprecated(
```
kubukoz, line 63: These were deprecated in 0.16.x, we should drop them now

kubukoz: although I'll roll this back, we can deal with deprecations in another PR

project/MimaVersionPlugin.scala
```diff
@@ -86,13 +85,7 @@ object MimaVersionPlugin extends AutoPlugin {
       // Focusing on backward compat as opposed to forward, for now.
       ProblemFilters.exclude[ReversedMissingMethodProblem]("smithy4s.*"),
       ProblemFilters.exclude[Problem]("*.internals*"),
-      ProblemFilters.exclude[IncompatibleResultTypeProblem]("*fromSchema"),
-      ProblemFilters.exclude[DirectMissingMethodProblem](
```
kubukoz, line 90: since we're able to break BC in 0.17 anyways. @Baccata can we drop the `*fromSchema` exclusion too?

Baccata: Sure

kubukoz: done in 06bedfbb8092ebf0ce856e44f38a6d1f939e3271

kubukoz:
alright pls review @baccata <3

modules/decline/src/Smithy4sCli.scala
```diff
@@ -53,6 +53,7 @@ class Smithy4sCli[Alg[_[_, _, _, _, _]], F[_]: MonadThrow](
   ): List[String] =
     HttpEndpoint
       .cast(endpoint)
+      .toOption
```
daddykotex, line 56: minor: since we're throwing a few lines below, would it be worth it to throw here also rather than just swallow the error?

daddykotex (approved):

modules/decline/src/Smithy4sCli.scala
kubukoz: yeah lemme do that

kubukoz: actually, I don't think so...

If `HttpEndpoint.cast` fails, that means the endpoint is _not_ a HTTP one and this method should just return `Nil`. We're throwing iff `HttpEndpoint.cast` succeeds and the `Http` hint is missing, which would be a completely invalid state.

The purpose of `protocolSpecificHelp` is to generate a list of additional help strings based on... protocol-specific info 😅 
so if you're implementing a CLI for a service that _only_ exposes a CLI (and no HTTP interface), you want `Nil` here.

kubukoz: I do the same in Playground and I've been thinking about it recently: https://github.com/kubukoz/smithy-playground/issues/139 - basically, if protocol-specific interpreters are allowed to prefix URLs, maybe we shouldn't trust the `Http` hint. Or `HttpEndpoint`, for that matter. Not sure what to think about this, but it's a separate problem from this PR I think.

PR #494, opened 2022-10-06
Change formatter to align margins in multiline strings

Baccata:


daddykotex:
minor and mostly informative in case you don't know. you can write commit  sha to a `.git-blame-ignore-revs` with content like:

```
# Update scalafmt config
825ba00bb95c0a9bb801ccb4c6dd74ec7b59b59e
```

And this will help looking at the history

PR #502, opened 2022-10-08
Add missing config scopes in codegen

### LINKED ISSUE #500: Local jar dependency between Compile and Test causes redundant codegen
```scala
val root = project
  .in(file("."))
  .settings(
    scalaVersion := "2.13.8",
    libraryDependencies ++= Seq(
      "com.disneystreaming.smithy4s" %% "smithy4s-core" % smithy4sVersion.value
    ),
    Smithy4sCodegenPlugin.defaultSettings(Test)
  )
```

In a setup like this, running `Test/smithy4sCodegen` seems to have zero caching - it'll re-run the actual action every time you call it. It's easier to showcase with a large model, such as an AWS spec: [here's a repro with AWS Lambda](https://github.com/kubukoz/demos/tree/smithy4s-local-config-dependency).

I checked and the last version that didn't have this issue was 0.15.3. In my experience, to surface this issue you need to have both Smithy sources in `main/smithy`, but you don't need Scala sources.

`Test / smithy4sLocalJars := Nil` is a workaround, but it's not going to be a viable one if you _do_ want the `Compile` sources to be used.

I checked and this doesn't seem to happen with multiple modules (I don't think we have scripted tests for it, but it'd be nice to - not sure how we could check that a task used the cache, though) - just between configurations.

kubukoz:
Closes #500.

### DIFF
--- a/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
+++ b/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
@@ -111,7 +111,8 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
     config / resourceGenerators += (config / smithy4sCodegen).map(
       _.filter(_.ext != "scala")
     ),
-    cleanFiles += (config / smithy4sOutputDir).value,
+    config / cleanFiles += (config / smithy4sOutputDir).value,
+    config / cleanFiles += (config / smithy4sResourceDir).value,
     config / smithy4sModelTransformers := List.empty
   )
 
@@ -154,7 +155,7 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
         m.root
       }
     val transforms = (conf / smithy4sModelTransformers).value
-    val s = streams.value
+    val s = (conf / streams).value
     val skipResources: Set[FileType] =
       if ((conf / smithy4sSmithyLibrary).value) Set.empty
       else Set(FileType.Resource)


modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -111,7 +111,8 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
     config / resourceGenerators += (config / smithy4sCodegen).map(
       _.filter(_.ext != "scala")
     ),
-    cleanFiles += (config / smithy4sOutputDir).value,
+    config / cleanFiles += (config / smithy4sOutputDir).value,
+    config / cleanFiles += (config / smithy4sResourceDir).value,
```
kubukoz, line 115: These aren't required to solve the issue, but thought it'd be a good idea.

Added the resource dir and scoped these to be consistent with the rest (everything is scoped by config).

`clean` without qualifiers seems to still work.

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -154,7 +155,7 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
         m.root
       }
     val transforms = (conf / smithy4sModelTransformers).value
-    val s = streams.value
+    val s = (conf / streams).value
```
kubukoz, line 158: The culprit - seems like the configs shared the cache (`_global` streams directory), which made Test see Compile's input hash and thus consider it different from its own. Not sure why that doesn't in turn break `Compile`'s caching (in this case it should be seeing what `Test` wrote last time it ran), but the scoping seems to resolve the issue anyway.

Baccata: Awesome find, and thank you so much for fixing !

Baccata (approved):

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
kubukoz: > Not sure why that doesn't in turn break Compile's caching (in this case it should be seeing what Test wrote last time it ran), but the scoping seems to resolve the issue anyway.

Update, it _did_ break Compile's caching apparently. So we should be all good now

PR #57, opened 2022-01-19
What's Play?

kubukoz:

PR #505, opened 2022-10-09
Local jars using AWS

kubukoz:


modules/codegen/src/smithy4s/codegen/ModelLoader.scala
```diff
@@ -170,7 +170,11 @@ object ModelLoader {
     ): ModelAssembler = {
       val smithy4sResources = List(
         "META-INF/smithy/smithy4s.smithy",
-        "META-INF/smithy/smithy4s.meta.smithy"
+        "META-INF/smithy/smithy4s.meta.smithy",
+        "META-INF/smithy/aws.protocols.smithy",
```
Baccata, line 174: I don't think this is gonna be the right solution. If you think about it, AWS is just one symptom of a deeper problem, and we need to solve the deeper problem really.

kubukoz: yeah I was about to write a comment about how I don't like this much

kubukoz: although I do think this is specific to AWS, because AWS traits are explicitly pulled in by the codegen plugin. For user-defined specs from fetched from deps this might not be an issue?

Baccata:
Addressed in https://github.com/disneystreaming/smithy4s/pull/57

PR #506, opened 2022-10-11
Amends cache key in Cached schema visitor

Baccata:
In order to reduce risk of memory leaks when people do not have stable references for the routers (ie `def` instead of `val`), this PR changes the cache key for the Cached schema visitor to avoid it being impacted by bits that are different when performing equality checks, in particular `Lazy`. 

This isn't a panacea, Schemas cannot be accurately compared for equality, for the sheer fact that two schemas might be semantically the same without being equal, due to the language limitation of not being able to compare thunks in Lazy.

Therefore, we mitigate the problem by providing a custom hashing function for schemas that applies a reasoned behaviour when encountering `Lazy`, and use that (conjointly with the ShapeId) as a key during memoization of schema compilation.

Additionally, this PR provides : 

* an implementation for the equals/hashcode methods of the Hints map
* a bunch of hash-code stable function constructs for the usecase of Dynamic. 

- [ ] Add a bunch of tests, in particular in the Dynamic usecase, to make sure that two dynamic schemas coming from the same model are not re-creating entries in the memoization map.

### DIFF
--- a/modules/core/src/smithy4s/schema/HashVisitor.scala
+++ b/modules/core/src/smithy4s/schema/HashVisitor.scala
@@ -1,0 +1,115 @@
+package smithy4s.schema
+
+import smithy4s.Lazy
+
+import smithy4s.Bijection
+import smithy4s.Refinement
+
+import smithy4s.{Hints, ShapeId}
+
+private[schema] class HashVisitor() extends SchemaVisitor[Lambda[A => Unit]] {
+  private var result: Int = 1
+  private val prime: Int = 31
+  private val visitedLazy: java.util.HashSet[ShapeId] =
+    new java.util.HashSet[ShapeId]()
+
+  def getResult: Int = result
+
+  def primitive[P](shapeId: ShapeId, hints: Hints, tag: Primitive[P]): Unit = {
+    result = result * prime + 1
+    result = result * prime + shapeId.hashCode()
+    result = result * prime + hints.hashCode()
+    result = result * prime + tag.hashCode()
+  }
+
+  def collection[C[_], A](
+      shapeId: ShapeId,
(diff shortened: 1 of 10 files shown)


modules/core/src/smithy4s/schema/HashVisitor.scala
```diff
@@ -0,0 +1,142 @@
...
+    this(schema)
+  }
+
+  def refine[A, B](schema: Schema[A], refinement: Refinement[A, B]): Unit = {
+    result = result * prime + 8
+    result = result * prime + refinement.hashCode()
+    this(schema)
+  }
+
+  def lazily[A](suspend: Lazy[Schema[A]]): Unit = {
```
Baccata, line 133: This is the whole reason for doing this. There is not way of implementing a reasoned `equals` method on the `Lazy` construct. Therefore the only way we can get something that will resist cache-miss is by running the Lazy in an educated fashion to produce some sort of hash value.

modules/dynamic/src/smithy4s/dynamic/internals/DynamicLambdas.scala
```diff
@@ -0,0 +1,62 @@
...
+package smithy4s.dynamic.internals
+
+import smithy4s.schema.Alt
+import smithy4s.schema.Schema
+
+/**
+  * A bunch of hash-code stable lambdas that are less likely to break memoization
+  * that may be able to SchemaVisitors.
+  */
+private[internals] object DynamicLambdas {
```
Baccata, line 26: The idea here is that we can provide functions with stable equals/hashcode values, which should avoid a lot of cache-miss when compiling dynamic schemas.

PR #508, opened 2022-10-12
protocol: support given import in refinements

ahjohannessen:
- change pattern for `Import` to account for Scala 3 given imports.


Baccata:
Thank you. We'll perform a release containing this patch later this week.

daddykotex:
great work @ahjohannessen

ahjohannessen:
Thank you :) Great 👍

PR #509, opened 2022-10-12
Reduce the risk of mem leaks

Baccata:
Okay, so this is an interesting one : the Cached schema visitor uses a HashMap, which only works if we have proper hashCode/equality methods. 

Now the thing is, because we use `transformHintsTransitively` during the course of some interpreters, it puts the user at risk if they're not properly memoizing the interpreter, as re-instantiation of interpreters would lead to re-transformation of hints, which would lead into new Schemas that wouldn't be referentially-equal to the previous iteration. 

To be clearer, any lambda present in a data structure makes it inherently hard to compare / check for equality)

This PR attempts at mitigating the problem by reducing the risk of cache-miss when transforming schemas using the same transformations. This is performed by wrapping the schema in a Wrapper construct that has a reasonably stable hash-code and equal method with regards with the constraints at end. It is not devoid of risk, if users started creating schemas in weird fashions, but it greatly mitigates the problem in "standard Smithy4s usage", and protects users would may use a `def` accidentally to store the smithy4s routing.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -943,7 +943,7 @@ def genSmithyImpl(config: Configuration) = Def.task {
       .map(_.data)
 
   val mc = "smithy4s.codegen.cli.Main"
-  val s = streams.value
+  val s = (config / streams).value
 
   def untupled[A, B, C](f: ((A, B)) => C): (A, B) => C = (a, b) => f((a, b))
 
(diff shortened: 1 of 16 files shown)


Baccata:
Still unsound ... I think the only way is to pass a cache contextually

PR #514, opened 2022-10-14
Amends the caching mechanism to avoid relying on global variables

Baccata:
This is changing the SchemaVisitor caching mechanism to allow to inject a mutable cache rather than relying on global ones, which can lead to memory when users apply unpredicted usage patterns.

Injecting a mutable cache means we can scope the caching to the lifecycle of interpreters, which is a lot safer.

The problem really stems from the fact that Schemas are not stable (with regards to hashCode/equality) methods when re-instantiated or when undergoing hint transformations, therefore cache miss are likely to occur even when semantically a cache hit should happen.

Additionally, this changes the Dynamic module to try and yield as-stable-as-possible schemas to further reduce the risk of memory leaks whilst benefiting from the caching mechanism.

I add to add a few MiMa filters for things that MiMa seems to be confused with, and things that were unfixable due to default parameters (note to self : remove default parameters when relevant in favour of method overloads)


modules/core/src/smithy4s/Hints.scala
```diff
@@ -67,6 +67,11 @@ object Hints {
     }
     override def toString(): String =
       s"Hints(${all.mkString(", ")})"
+
+    override def hashCode(): Int = toMap.hashCode()
+    override def equals(obj: Any): Boolean = {
```
Baccata, line 72: created proper hashChode and equality methods which were desperately missing on the Hints construct.

modules/core/src/smithy4s/http/CodecAPI.scala
```diff
@@ -31,6 +31,9 @@ import internals.StringAndBlobCodecSchemaVisitor
 trait CodecAPI {
 
   type Codec[A]
+  type Cache
+
+  def createCache(): Cache
```
Baccata, line 36: Right, so this "hides" the nature of the cache to CodecAPI callers. This is because the Cache might actually be caching constructs that are not for public consumption, so we're relying on a type-member to hide the detail, and passing it around. 

The reason we're not just making a `CachedCodecAPI` implementation is that doing so would be harder to respect binary compatibility, as the "end of the world" currently relies on a `val CodecAPI` being present on the service builders. Because the ServiceBuilder are Global, this wouldn't play too well without breaking bincompat

modules/core/src/smithy4s/internals/SchemaDescriptionDetailed.scala
```diff
@@ -43,7 +43,7 @@ private[internals] trait SchemaDescriptionDetailedImpl[A]
 }
 
 private[internals] object SchemaDescriptionDetailedImpl
-    extends SchemaVisitor.Cached[SchemaDescriptionDetailedImpl] { self =>
+    extends SchemaVisitor[SchemaDescriptionDetailedImpl] { self =>
```
Baccata, line 46: this is not critical enough that we'd have to cache it, I think.

modules/core/src/smithy4s/schema/CachedSchemaCompiler.scala
```diff
@@ -0,0 +1,47 @@
...
+  type Cache
+  def createCache(): Cache
+  def fromSchema[A](schema: Schema[A]): F[A]
+  def fromSchema[A](schema: Schema[A], cache: Cache): F[A]
+
+}
+
+object CachedSchemaCompiler {
+
+  private[smithy4s] abstract class Impl[F[_]] extends CachedSchemaCompiler[F] {
```
Baccata, line 30: This is basically a reusable helper that saves a bunch of boilerplate

modules/core/src/smithy4s/schema/CachedSchemaCompiler.scala
```diff
@@ -0,0 +1,47 @@
...
+    protected type Aux[_]
+    type Cache = CompilationCache[Aux]
+
+    override final def fromSchema[A](schema: Schema[A]): F[A] =
+      fromSchema(schema, CompilationCache.nop[Aux])
+
+    def createCache(): Cache = CompilationCache.make[Aux]
+
+    private val globalCache: Cache = createCache()
+    implicit def derivedImplicitInstance[A](implicit
```
Baccata, line 40: Here we're trusting that it's okay enough to use a global cache, which is consistent in the previous usage. This is because when using implicits, the schemas are somewhat guaranteed to be tied to types, therefore they are "uniquely" instantiated for that usecase, whereas in our dynamic interpreters, we often apply a hint-transformation pre-processing step, which implies non-unique instances of schemas, which is the reason for the memory leak.

modules/core/src/smithy4s/schema/CompilationCache.scala
```diff
@@ -0,0 +1,62 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s
+package schema
+import scala.collection.mutable.{Map => MMap}
+
+trait CompilationCache[F[_]] {
```
Baccata, line 21: The construct that actually abstract over the mutable variable we use for caching. Can be implemented as a no-op, which is useful to retain the API sane.

modules/core/test/src/smithy4s/CachedSchemaVisitorSpec.scala
```diff
@@ -21,156 +21,187 @@ import smithy4s.schema._
...
+  }
+
+  test(header + "enum") {
+    sealed abstract class FooBar(val stringValue: String, val intValue: Int)
+        extends smithy4s.Enumeration.Value {
+      val name = stringValue
+      val value = stringValue
+      val hints = Hints.empty
     }
-    val lazySchema = Schema.recursive(schema)
```
Baccata, line 170: Using `Schema.recursive` will not showing any problem if the schema is actually not recursive.

modules/dynamic/src/smithy4s/dynamic/internals/DynamicLambdas.scala
```diff
@@ -0,0 +1,62 @@
...
+package smithy4s.dynamic.internals
+
+import smithy4s.schema.Alt
+import smithy4s.schema.Schema
+
+/**
+  * A bunch of hash-code stable lambdas that are less likely to break memoization
+  * that may be able to SchemaVisitors.
+  */
+private[internals] object DynamicLambdas {
```
Baccata, line 26: Right, so this is a bit of a side-thing, but I wanted to ensure that several instances of the "same" dynamic schemas were "equal" when facing caching. To that end, we make all accessors serialisable, to make the natural `.hashCode` and `.equals` on `Schema` behave like we intuitively think they should.

modules/http4s/src/smithy4s/http4s/internals/CompilerContext.scala
```diff
@@ -0,0 +1,42 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.http4s.internals
+
+import smithy4s.http4s.EntityCompiler
+import smithy4s.http.Metadata
+
+private[http4s] trait CompilerContext[F[_]] {
```
Baccata, line 22: This is a helper trait binding the caches together, and also ensuring that the entity cache is consistent (at the type level) with the entity compiler, because we only want to instantiate a single entity cache per router/reverse-router.

project/MimaVersionPlugin.scala
```diff
@@ -77,6 +82,18 @@ object MimaVersionPlugin extends AutoPlugin {
 
   override def projectSettings: Seq[Setting[_]] = Seq(
     isMimaEnabled := false,
+    mimaBinaryIssueFilters ++= Seq(
```
Baccata, line 85: A few minimal MiMa filters. Some are there because MiMa seemed to be confused by some of the code, the `default` one is there because I made the mistake of using default parameters somewhere I shouldn't have. This does imply that there's one method (in a non internals package) that's actually not binary compatible in there, but really, it should not be considered API method by anyone.

lewisjkl (approved):
All looks good to me, great work 🚀

build.sbt
```diff
@@ -943,7 +943,7 @@ def genSmithyImpl(config: Configuration) = Def.task {
       .map(_.data)
 
   val mc = "smithy4s.codegen.cli.Main"
-  val s = streams.value
+  val s = (config / streams).value
```
kubukoz, line 946: for future generations: this is our local version of the fix in #502 that I forgot to port here.

modules/core/src/smithy4s/internals/SchemaDescription.scala
```diff
@@ -28,7 +28,7 @@ import smithy4s.schema.{
 }
 import smithy4s.schema.Primitive.PTimestamp
 
-object SchemaDescription extends SchemaVisitor.Cached[SchemaDescription] {
+object SchemaDescription extends SchemaVisitor[SchemaDescription] {
```
kubukoz, line 31: I forgot this existed, good to know

PR #521, opened 2022-10-17
Apply hint mask transitively in json codecs

kubukoz:
Previously, only top-level hints would be modified with the hint mask, so even if the mask was empty you could affect serialization with something like `@jsonName` or `@discriminated`.

This PR enables hint masks to work on all levels. This may possibly break existing behavior in users' code if they rely on the mask being applied only on the surface.

For such users, a workaround would be to use the `HintMask.allAllowed` mask in the codecs, but apply `.transformHintsLocally` with the old hint mask on the Schema directly.


kubukoz:
glad we have these IntEnum tests in pizza

Baccata (approved):

Issue #515, opened 2022-10-14
`java.io.IOException: Broken pipe` occurring on client calls with empty bodies 


ahjohannessen:
Seems like there is a bug in how empty body is handled. I have a repro [here](https://github.com/daddykotex/smithy4s-client-middleware/compare/main...ahjohannessen:smithy4s-client-middleware:wip-repo-empty-body-json). The smithy spec show cases differences in payload. Details are found [here](https://github.com/disneystreaming/smithy4s/discussions/513#discussioncomment-3878241).

Baccata:
The fact that the body is empty is normal on requests/responses that transmit all their data through metadata. The fact that the client raises an exception is not.

daddykotex:
This is interesting to look at. I guess other issues are coming into play here. If I keep the service modification that you've made @ahjohannessen in [your repro]() but adjust the code to this:

```scala
package com.example

import smithy4s.hello._
import cats.effect._
import cats.effect.syntax.resource._
import cats.implicits._
import org.http4s.implicits._
import org.http4s.ember.server._
import org.http4s._
import com.comcast.ip4s._
import smithy4s.http4s.SimpleRestJsonBuilder
import org.http4s.ember.client.EmberClientBuilder
import org.http4s.client.Client

object HelloWorldImpl extends HelloWorldService[IO] {

  def hello(
      name: String,
      town: Option[String]
  ): IO[HelloOutput] =
    IO.pure { HelloOutput("hello") }

  def hello2(
      name: String,
      age: Option[Int],
      town: Option[String]
  ): IO[Hello2Output] =
    IO.pure { Hello2Output("hello2") }
}

object Routes {
  private val example: Resource[IO, HttpRoutes[IO]] =
    SimpleRestJsonBuilder.routes(HelloWorldImpl).resource

  private val docs: HttpRoutes[IO] =
    smithy4s.http4s.swagger.docs[IO](HelloWorldService)

  val all: Resource[IO, HttpRoutes[IO]] = example.map(_ <+> docs)
}

abstract class Setup extends IOApp.Simple {
  // client middleware
  val authMiddleware: org.http4s.client.Middleware[IO] = { client =>
    Client { req =>
      client.run(
        req.withHeaders(
          headers.Authorization(Credentials.Token(AuthScheme.Bearer, "TOKEN"))
        )
      )
    }
  }

  val client = EmberClientBuilder.default[IO].build.map { client =>
[...]

PR #522, opened 2022-10-17
Fix to send empty body, and consume it

daddykotex:
Closes https://github.com/disneystreaming/smithy4s/issues/515

I've decided to include a e2e test for ember because when fixed on both, client and server, the test is not flaky and I think its a long term win.

Equipped with https://github.com/http4s/http4s/issues/4935#issuecomment-1279647970, I implemented both the server fix and the client fix


modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
```diff
@@ -102,7 +103,7 @@ private[smithy4s] class SmithyHttp4sClientEndpointImpl[F[_], Op[_, _, _, _, _],
     val baseRequest = Request[F](method, uri, headers = headers)
     if (inputHasBody) {
       baseRequest.withEntity(input)
-    } else baseRequest
+    } else baseRequest.withEntity(ByteVector.empty)
```
ahjohannessen, line 106: Should this not only apply to PUT and POST methods?

daddykotex: that's a good question, let me see what facilities exists here to distinguish what method allow or not bodies

daddykotex: it annoys me that I can't find a public equivalent for this and avoid copy/pasting that method Set out of ember :/

modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sServerEndpoint.scala
```diff
@@ -125,7 +125,10 @@ private[smithy4s] class SmithyHttp4sServerEndpointImpl[F[_], Op[_, _, _, _, _],
   private val extractInput: (Metadata, Request[F]) ==> I = {
```
kubukoz, line 125: Are we getting anything out of that kleisli? feels like it could be just a function.

modules/http4s/test/src-jvm-ce3/smithy4s/http4s/Http4sEmberPizzaClientSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+  def server(app: HttpApp[IO]): Resource[IO, Int] =
+    cats.effect.std.Random
+      .scalaUtilRandom[IO]
+      .flatMap(_.betweenInt(50000, 60000))
+      .toResource
+      .flatMap(port =>
+        Port
+          .fromInt(port)
```
kubukoz, line 59: haven't checked but wouldn't you be able to bind to `0` and get a random port? except in this case you'd have a guarantee that it's free (again, not sure if ember behaves like this)

modules/http4s/test/src-jvm-ce3/smithy4s/http4s/Http4sEmberPizzaClientSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+          .map(_ => port.value)
+      }
+
+  test("empty body") { client =>
+    (client.book("name") *> client.book("name2")).as(success)
+  }
+
+  private val dummyImpl = new PizzaAdminService[IO]() {
+    // format: off
+    override def addMenuItem(restaurant: String, menuItem: MenuItem): IO[AddMenuItemResult] = ???
```
kubukoz, line 80: `IO.stub` if you want

modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sServerEndpoint.scala
daddykotex: just removed it, good catch

modules/http4s/test/src-jvm-ce3/smithy4s/http4s/Http4sEmberPizzaClientSpec.scala
daddykotex: TIL 👍

modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sServerEndpoint.scala
```diff
@@ -119,26 +119,22 @@ private[smithy4s] class SmithyHttp4sServerEndpointImpl[F[_], Op[_, _, _, _, _],
       errorTransformation(other).flatMap(F.raiseError)
   }
 
-
-
-  // format: off
-  private val extractInput: (Metadata, Request[F]) ==> I = {
+  private def extractInput(metadata: Metadata, request: Request[F]): F[I] = {
```
kubukoz, line 122: This isn't exactly what I meant 😅 

This should still be a `val`. The problem with the current shape is that you compile the input decoder on every request, whereas previously it would've happened only once when the `val` was initiated.

I'm thinking of the shape

`private val extractInput: (Metadata, Request[F]) => F[I]`

where the function would be returned as the result of the pattern match.

daddykotex: got it. honestly when I rewrote I was like: they're the same let's use a def

but definitely here, a val allows us to do the pattern match ahead of time

modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
armanbilge: > it annoys me that I can't find a public equivalent for this and avoid copy/pasting that method Set out of ember :/

Ah yeah, an issue or PR to fix this would be great. Seems like we just need to add a boolean to `Method`.

daddykotex: happy to open it

modules/tests/src/smithy4s/tests/PizzaClientSpec.scala
```diff
@@ -174,8 +173,7 @@ abstract class PizzaClientSpec extends IOSuite {
       expected: E
   )(implicit
       loc: SourceLocation,
-      ct: scala.reflect.ClassTag[E],
-      show: Show[E] = Show.fromToString[E]
```
Baccata, line 178: let's keep this.

modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
daddykotex: Following the discussions in the [PR](https://github.com/disneystreaming/smithy4s/pull/522), I think this should be merged as in. I'm not sure there can be an always accurate way of telling whether a particular method should have a body or not.

Considering it works for our current use cases, we should move forward with it. If we get and issue later on that sending and empty body is problematic under certain situation, we should revisit.

daddykotex:
I've rebased on top of series/0.17 rather than trying to reconcile the two branches

But I think the code in series/0.17 is somewhat problematic, in the sense, that my local githooks were failing for copyright notices missing on some generated files

we should update the ci so that PR to 0.17 can trigger a CI run

Baccata:
@daddykotex I've fixed it in my PR, you can apply the same change to `ci.yml` (glob pattern problem) to get it running.

daddykotex:
I think I'll mark as draft and marge `series/0.17` in this branch when #567 is merged

Baccata:
@daddykotex I think it's time to sync this and tag it ready

modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sServerEndpoint.scala
```diff
@@ -121,26 +121,24 @@ private[smithy4s] class SmithyHttp4sServerEndpointImpl[F[_], Op[_, _, _, _, _],
...
+          request.body.compile.drain *>
+            totalDecoder.decode(metadata).liftTo[F]
       case None =>
         // NB : only compiling the input codec if the data cannot be
         // totally extracted from the metadata.
-        implicit val inputCodec = entityCompiler.compilePartialEntityDecoder(inputSchema, entityCache)
-        Kleisli { case (metadata, request) =>
+        (metadata, request) =>
+          implicit val inputCodec =
+            entityCompiler.compilePartialEntityDecoder(inputSchema, entityCache)
```
kubukoz, line 135: randomly scrolling, I think this still has the laziness issue (you only compile after the request). It's probably fine for perf now that we cache the compilers, but doing the compilation ahead of time seems more consistent with how we do it in other places

daddykotex: resolved

Baccata (approved):

PR #524, opened 2022-10-18
Fix hint transformations and adds a test suite

Baccata:
The transitive hints transformation was not accurately propagating hints wherever it needed. 

Additionally, this fixes the union Dispatcher construct, by caching on the labels as opposed to the instance, which aren't stable wrt hint transformation. 

This fixes the bug that makes https://github.com/disneystreaming/smithy4s/pull/521 red

### DIFF
--- a/modules/core/src/smithy4s/schema/Alt.scala
+++ b/modules/core/src/smithy4s/schema/Alt.scala
@@ -91,9 +91,10 @@ object Alt {
       ): G[U] = {
         val precompiledAlts =
           precompile.toPolyFunction
-            .unsafeCacheBy(
+            .unsafeCacheBy[String](
               alts.map(smithy4s.Existential.wrap(_)),
-              (alt: Alt[F, U, _]) => alt.label
+              (alt: Existential[Alt[F, U, *]]) =>
+                alt.asInstanceOf[Alt[F, U, _]].label
             )
 
         encoderK.absorb[U] { u =>
--- a/modules/core/src/smithy4s/schema/Alt.scala
+++ b/modules/core/src/smithy4s/schema/Alt.scala
@@ -18,6 +18,7 @@ package smithy4s
 package schema
 
 import smithy4s.capability.EncoderK
+import scala.annotation.nowarn
 
 /**
   * Represents a member of coproduct type (sealed trait)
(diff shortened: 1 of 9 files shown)


modules/core/src/smithy4s/schema/Alt.scala
```diff
@@ -36,7 +37,14 @@ final case class Alt[F[_], U, A](
 }
 object Alt {
 
-  final case class WithValue[F[_], U, A](alt: Alt[F, U, A], value: A) {
+  @nowarn
+  final case class WithValue[F[_], U, A](
+      @deprecated(
```
Baccata, line 42: In this PR, I'm deprecating access to the `alt` val, as it requires implementors to create their own caching mechanism, which is inherently risky due to how schemas are not stable (you need to cache using labels as keys). 

A fair chunk of this PR is a consequence of this deprecation, as interpreters/compilers are amended to make use of `Alt.Dispatcher` instead, which takes care of memoization for the implementor in the correct way, and offers a more elegant way to implement encoders of unions, by hiding the ugliness.

modules/json/src/smithy4s/http/json/SchemaVisitorJCodec.scala
```diff
@@ -802,41 +810,44 @@ private[smithy4s] class SchemaVisitorJCodec(maxArity: Int)
           }
         } else in.decodeError("Expected JSON object")
 
-      private[this] val altCache =
```
daddykotex, line 805: that's nice we can get rid of all the custom caches.

awesome

modules/core/src/smithy4s/schema/Alt.scala
```diff
@@ -82,7 +91,11 @@ object Alt {
       ): G[U] = {
         val precompiledAlts =
           precompile.toPolyFunction
-            .unsafeCache(alts.map(smithy4s.Existential.wrap(_)))
+            .unsafeCacheBy[String](
+              alts.map(smithy4s.Existential.wrap(_)),
+              (alt: Existential[Alt[F, U, *]]) =>
```
kubukoz, line 96: I've been away from polymorphic code so long that I almost forgot what `*` means here...

kubukoz (approved):

lewisjkl (approved):

daddykotex (approved):

PR #526, opened 2022-10-18
Removing hint transformation in union's "total" function ...

Baccata:
... for performance reasons, as this happens on the "hot-path". 

It's fine to remove because, due to the `Dispatcher` mechanism, the alt values should not be queried, which means the hints that may be transformed should not be queried either (at that level, anyway). It's even discouraged to query it because the alt field is now deprecated.


daddykotex:
This is really hard to spot. How did you find it?

I was on the look out for these kinds of things when I reviewed the PR and it completely flew over my head.

daddykotex (approved):

Baccata:
> This is really hard to spot. How did you find it?

Just popped in my head as I was preparing to go to bed.

PR #527, opened 2022-10-19
Client httpReponseTest for output

daddykotex:
I'm uncertain about the approach I've used here. Implementing test response case for client without the server interpretor (and vice- versa for the client) is tricky.

For example in this PR, where only client tests are available. I have to grab the codecAPI to get the mediatype and set it on the response (if not provided in the test case).

On the server side it's worse because I have to craft a request out of thin air. Previsouly I had managed to do it because to have compliance tests we required both the client and server side of the protocol to be implemented. But because, we split them, it's much harder.

I still have trouble to put in writing how I struggle with this feature lol.

Anyway I'll drop what I wrote in markdown in a comment on this PR, let me know if it makes sense


daddykotex:
For a client request test, we build an `I` instance with the `params` field of the `HttpRequestTestCase`. We can call the `Alg` with this input and that'll turn it into a `Request`. The test validates properties of the `Request` via the `HttpRequestTestCase` fields:

* method
* uri
* queryParams
* headers
* body
* etc.

For a server test, we use the `HttpRequestTestCase` fields listed above to construct a `Request` that we run through the server. The server route/decode that `Request` in to a `I`. The test ensures that the produced `I` instance matches an `I` instance constructed with the `params` field of the `HttpRequestTestCase`.

Hopefully that makes some sense. In summary:

* for client => we validate that a function from `I` to `Request` behaves correctly
* for server => we validate that a function from `Request` to `I` behaves correctly

On the client, this works because we're given a way to hook the client into a fake server where we capture the incoming request, and short-circuit the rest of the process with an exception.

In pseudocode:

`Input => Request => [Server] =/> Response => Output`

`=/>` is where we short-circuit.

On the server, this works because we're given a function to turn a `Alg` (the implementation of your operations) into a server. We supply a fake `Alg` that just records the received `input: I` and short-circuit the rest of the execution with an exception.

`Request => [Server] => Input => [Alg] =/> Output => Response`

Again `=/>` is where we short-circuit.

In both case, it works because we can generate `Input` (on client) and `Request` (on server) from the HttpResquestTestCase definition.


---------------

On the response test case side, things are more complex. The HttpResponseTestCase is esssentially a way to test a function that turns a `O` instance or an `E` (error) into a protocol response. 

Unfortunately, what we have is the full stack (request => implementation => response) which makes it hard to test just the tail (response consumption (client) or response generation (server)) because we have to provide the input (to be generated from thin air).

--------

It's more of a brain dump than anything. But whenever I think about this feature I'm like oh yeah, that'll be easy, it makes sense. But when I come to implement it. I hit all sorts of road blocks (real or not, I'm not sure :P ).

Baccata:
The approach you describe (generating input out of thin-air) is exactly the one I'd have taken.

daddykotex:
In that case, this implementation for client is all right.

But for the server implementation, I need to reimplement some of the logic that already exists in the client code. Turning an input an the metadata (path, query, etc.) into a request that I can run my routes

Baccata:
@daddykotex you may be able to cheat and use IOLocal to short-circuit any incoming requests and inject dummy inputs in place of what would normally be passed to the algebra by the Smithy4s router

daddykotex:
When I reach the algebra, I ultimately don't mind about the input in that case. My problem is routing through the http layer _to get to_ the algebra. This is the code I have right now:

```
  private[compliancetests] def serverResponseTest[I, E, O, SE, SO](
      endpoint: Endpoint[Op, I, E, O, SE, SO],
      testCase: HttpResponseTestCase
  ): ComplianceTest[IO] = {
    def makeRequest(): Request[IO] = {
      ???
    }
    type R[I_, E_, O_, SE_, SO_] = IO[O_]

    ComplianceTest[IO](
      name = endpoint.id.toString + "(server|response): " + testCase.id,
      run = {
        val outputDecoder = Document.Decoder.fromSchema(endpoint.output)
        val output = outputDecoder
          .decode(testCase.params.getOrElse(Document.obj()))
          .liftTo[IO]

        output.flatMap { output =>
          val fakeImpl: smithy4s.Monadic[Alg, IO] =
            service.transform[R](
              new smithy4s.Interpreter[Op, IO] {
                def apply[I_, E_, O_, SE_, SO_](
                    op: Op[I_, E_, O_, SE_, SO_]
                ): IO[O_] = {
                  // todo error structures
                  IO.pure(output.asInstanceOf[O_])
                }
              }
            )

          getServer(fakeImpl)
            .use { server =>
              server.orNotFound
                .run(makeRequest())// my problem is here, I get a 404 if I don't supply a valid request
                .flatMap { resp =>
                  resp.body
                    .through(utf8Decode)
                    .compile
                    .foldMonoid
                    .tupleRight(resp.status)
                }
                .map { case (actualBody, status) =>
                  val bodyAssert = testCase.body
                    .map(body => assert.eql(body, actualBody))
                  val assertions =
                    bodyAssert.toList :+
                      assert.eql(status.code, testCase.code)
                  assertions.combineAll
                }
            }
        }
      }
    )
  }
```

daddykotex:
I've tried your suggestion, and after a bit of time (some of it lost to a tricky stack overflow lol), I think it's not working: https://github.com/disneystreaming/smithy4s/commit/9568f09bf3d0ceee11540774f32acdad3b855604

Because yes, I can redefine the service and its endpoint, but ultimately the service that is used to build the http4s routes is selected in user code (when implementing `getServer`) :/

Baccata:
What if you change the `getServer` signature to a more generic : 

```scala 
def getServer[SomeAlg[_[_,_,_,_,_]]](impl: smithy4s.Monadic[SomeAlg, IO])(implicit service: Service[SomeAlg, _]): Resource[IO, HttpRoutes[IO]]
```

Baccata:
@daddykotex yup, feel free to open that in a separate PR :)

Baccata (approved):

daddykotex:
I added a few test, but also support for client tests on error structure annotated with `@httpResponseTests`

daddykotex:
ahh rip: https://github.com/disneystreaming/smithy4s/actions/runs/3330248676/jobs/5508490525#step:5:73

daddykotex:
I've added the following, as a workaround: https://github.com/disneystreaming/smithy4s/pull/527/commits/79717dd7589500450025d56fa0556d48ca60b402

wdyt?

cc @plokhotnyuk @Baccata 


basically, the JCodec turns a `Timestamp(epochSecond = n, 0)` into a `timestamp: n.0` on the JVM but into a `timestamp: n` on ScalaJS

Baccata:
I'm cool with the amendment

Issue #523, opened 2022-10-17
Upgrading GH actions


daddykotex:
When looking at the workflows, we see warnings about deprecated usage of node12 in some of our GitHub actions.

![Screen Shot 2022-10-17 at 15 08 20](https://user-images.githubusercontent.com/5230460/196262132-fe1f153e-fc57-4e6e-a562-a4dfbea3379e.png)

Baccata:
Yeah happy to upgrade.

PR #528, opened 2022-10-19
Move to coursier/setup-action

daddykotex:
There is also coursier/cache-action that needs upgrading but we'll open a PR when https://github.com/coursier/cache-action/pull/418 is merged

References https://github.com/disneystreaming/smithy4s/issues/523


Baccata (approved):

daddykotex:
v6 was released, build on node 16 so future actions workflow should be fine
https://github.com/coursier/cache-action/releases/tag/v6

PR #529, opened 2022-10-19
Render "Default" (aka Stub) classes

Baccata:
Adds logic and the necessary typelevel black-magic to facilitate the creation of implementations where methods receive a default/stub implementations (which can be overridden)

- [x] documentation

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/Renderer.scala
@@ -170,7 +170,7 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
         block(
           line"object ${NameRef(name)} extends $Service_.Provider[$nameGen, ${name}Operation]"
         )(
-          line"type Default[F[+_]] = $nameGen.Default[smithy4s.StubLift[F]#Stub]",
+          line"type $Default_[F[+_]] = $nameGen.Default[smithy4s.StubLift[F]#Stub]",
           line"def apply[F[_]](implicit F: ${NameRef(name)}[F]): F.type = F",
           line"def service: $Service_[$nameGen, ${name}Operation] = $nameGen",
           line"val id: $ShapeId_ = service.id"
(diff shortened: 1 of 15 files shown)


modules/core/test/src/smithy4s/DefaultServiceSmokeSpec.scala
```diff
@@ -0,0 +1,30 @@
...
+ *  limitations under the License.
+ */
+
+package smithy4s
+
+import munit._
+class DefaultServiceSmokeSpec() extends FunSuite {
+
+  test("Default stubs do compile") {
+    object stub extends smithy4s.example.Weather.Default[Option](None)
```
daddykotex, line 23: so if `F` were `IO` here we could do something like: `smithy4s.example.Weather.Default[IO](IO.stub)`

am I right?

modules/docs/src/01-overview/06-stubs.md
```diff
@@ -0,0 +1,28 @@
...
+object StubbedHelloWorld extends HelloWorldService.Default[IO](IO.stub)
+```
+
+Obviously, the generated methods can be overridden.
+
+```scala mdoc:silent
+import smithy4s.hello._
+import cats.effect._
+
+object OverriddenHelloWorld extends HelloWorldService.Default[IO](IO.stub){
```
daddykotex, line 23: got my answer here

daddykotex (approved):
If this gets merge before my #522 , I'll update it in consequence

modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -257,6 +258,14 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
           }
         },
         newline,
+        block(
+          line"class Default[P[-_, +_, +_, -_, +_]](default: => P[Any, Nothing, Nothing, Any, Nothing])"
```
yisraelU, line 262: same as above

modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -170,6 +170,7 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
         block(
           line"object ${NameRef(name)} extends $Service_.Provider[$nameGen, ${name}Operation]"
         )(
+          line"type Default[F[+_]] = $nameGen.Default[smithy4s.StubLift[F]#Stub]",
```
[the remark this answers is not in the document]
Baccata: `NameDef` but you're right ! Good catch

Baccata:
Will recreate against the 0.17 feature branch

PR #539, opened 2022-10-21
Use a concurrent map on JVM for the compilation cache

Baccata:
When schema visitors are used as means to formulate typeclass instances, a global cache is used. It is risky to use non-thread-safe maps when that's the case. 

NB : I've not managed to trigger a bug, doesn't mean it wouldn't happen.

### DIFF
--- a/modules/core/src/smithy4s/schema/CompilationCache.scala
+++ b/modules/core/src/smithy4s/schema/CompilationCache.scala
@@ -16,7 +16,7 @@
 
 package smithy4s
 package schema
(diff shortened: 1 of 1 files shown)


daddykotex:
Note: This is an alias for a `TrieMap`

daddykotex (approved):

Baccata:
> Note: This is an alias for a TrieMap

On JVM it is, on JS it's a simple mutable map (cause multithreading ain't a thing there)

PR #540, opened 2022-10-21
Export a SMITHY_VERSION so that publish can use it

daddykotex:


.github/workflows/ci.yml
```diff
@@ -159,6 +159,11 @@ jobs:
       - name: Cache
         uses: coursier/cache-action@v6
 
+      - name: Resolve latest smithy4s-core version
+        # export the version only when we trigger publish manually
+        if: ${{ github.event_name == 'workflow_dispatch' }}
+        run: echo "SMITHY_VERSION=$(cs resolve com.disneystreaming.smithy4s:smithy4s-core_2.13:latest.stable | head -1 | awk -F':' '{print $3}')"  >> $GITHUB_ENV
```
kubukoz, line 165: This should be SMITHY4S_VERSION so that it's not confused with the AWS smithy library's version

kubukoz (changes requested):

.github/workflows/ci.yml
daddykotex: absolutely

lewisjkl (commented):
Awesome, thanks for adding!

kubukoz (approved):

Baccata (approved):

PR #555, opened 2022-10-26
Polyfunction revamp

Baccata:
This is a revamp of a few things, in an effort to increase consistency in the various kinds of polymorphic functions that Smithy4s (and its users) deal with. In particular, it unifies the patterns between `PolyFunction` and what used to be `Transformation` (now called PolyFunction5) by means of build-time generation. 

It also recycles the term `Transformation`, making it a function-like capability that allows to provide a unified user-experience across polymorphic things without requiring the user to memorise ungodly names like "mapK, mapK5", etc ... : generated Services all receive a `transform` method that can take anything that has an associated `Transformation`, be that Polyfunction5, PolyFunction, and in the future cat's `FunctionK`, or other things such as error amendments, or transformations allowing to go from monofunctors to bifunctors (or vice-versa). 

* Things in the `kinds` package are not meant to be used/viewed by most users. It's mostly for us, implementors of interpreter layers. **bikeshedding around names is welcome**, naming these things is hard, so I want to make sure our small group of maintainers is roughly on the same page.  

* Try to make working with various kinds more consistent
* Use code generation to generate consistent versions of PolyFunction for various kinds
* Add FunctorK typeclasses for all kinds we're interested to deal with.
* Change the concept of Transformation altogether to be able to offer a unified developer experience where algebra could be transformed using various functions and natural transformations, without the end user having to remember complicated names.
 * Change Service interface to expose `toPolyFunction` and `fromPolyFunction` functions, allowing to turn an Algebra into a PolyFunction5 and vice-versa.
 * Paves the road for an easier BiFunctor integration (EitherT, ZIO, MonixBIO, etc) by adding a few constructs and aliases.
 
 #### TODO 

- [ ] some scaladoc around the various constructs and types 
- [ ] going through the docs and amend stale/obsolete reference to names 
- [ ] more transformation tests


modules/core/test/src/smithy4s/TransformationSmokeSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  test("transform method can be called with poly functions") {
+    object stub extends Weather[Option] {
+      def getCurrentTime(): Option[GetCurrentTimeOutput] = None
+    }
+
+    case object Empty extends Throwable
+
+    // Not ascribing the type to get a
+    val transformed = stub.transform(new PolyFunction[Option, Try] {
```
Baccata, line 33: This is one of the motivations for this PR : we want for users to be able to apply polymorphic functions of various shapes with a single `transform` method, and for it to work reasonably well wrt type inference. 

- [ ] TODO : need to complete the comment above this line

Baccata:
@jona7o, I'd like to extend you the courtesy of asking for your opinion on naming and stuff, as this is both source and binary breaking and will induce some changes to your libraries (nothing fundamental, mostly mechanical things).

modules/core/src/smithy4s/capability/Transformation.scala
```diff
@@ -0,0 +1,37 @@
...
+import kinds._
+
+/**
+  * Heterogenous function construct, allows to abstract over various kinds of functions
+  * whilst providing an homogenous user experience without the user having to
+  * manually lift functions from one kind to the other.
+  *
+  * Used to reduce the noise of transformations
+  */
+trait Transformation[Func, Input, Output] {
```
Baccata, line 13: This is probably the most complicated thing here : essentially this is a construct that says "I, Func, am a polymorphic function of some kind,  and when given an input, I can produce an output". For example, we use it to express the fact that we know how to apply a good old `F ~> G` (with `F[_] ` and `G[_]`) on something that looks like `Foo[F[_, _, _, _, _]]`. 

This allows to make it **simpler** for the user to apply generic transformations on the instances of the generated interfaces, without having to worry about the kinds at hand. 

TBH : I'm not fully certain it's a good idea. I just want to avoid having the users having to worry about the humongous kinds we deal with in this library, and for cat's FunctionK to just "work" with it modulo a simple import.

modules/http4s/src/smithy4s/http4s/SimpleProtocolBuilder.scala
```diff
@@ -39,15 +40,15 @@ abstract class SimpleProtocolBuilder[P](val codecs: CodecAPI)(implicit
...
+      impl: FunctorAlgebra[Alg, F]
   )(implicit
       serviceProvider: smithy4s.Service.Provider[Alg, Op],
       F: EffectCompat[F]
   ): RouterBuilder[Alg, Op, F] = {
     val service = serviceProvider.service
     new RouterBuilder[Alg, Op, F](
       service,
-      service.asTransformation[GenLift[F]#λ](impl),
+      service.toPolyFunction[Kind1[F]#toKind5](impl),
```
daddykotex, line 51: I love this

project/Boilerplate.scala
```diff
@@ -0,0 +1,341 @@
...
+/**
+ * Copied, with some modifications, from https://github.com/milessabin/shapeless/blob/master/project/Boilerplate.scala
+ *
+ * Generate a range of boilerplate classes, those offering alternatives with 0-22 params
+ * and would be tedious to craft by hand
+ *
+ * @author Miles Sabin
+ * @author Kevin Wright
+ */
+object Boilerplate {
```
lewisjkl, line 13: 🙌 nice

project/Smithy4sPlugin.scala
```diff
@@ -160,7 +160,12 @@ object Smithy4sPlugin extends AutoPlugin {
     val base =
       if (scalaVersion.startsWith("3."))
         filterScala3Options(commonCompilerOptions)
-      else commonCompilerOptions
+      else
+        commonCompilerOptions
+    // ++ Seq(
+    //   "-Xsource:3",
+    //   "-P:kind-projector:underscore-placeholders"
+    // )
```
lewisjkl, line 168: Extra comment?

lewisjkl (approved):
I'm sure you're already thinking along these lines, but since one motivation for the PR is to make transformations (er polyfunctions) easier to implement in user-land, we should make sure we document what that looks like. Happy for that to be in a different PR though.

yisraelU (approved):

modules/core/src-2/kinds/package.scala
```diff
@@ -0,0 +1,21 @@
+package smithy4s
+
+package object kinds {
+
+  // format: off
+  type FunctorAlgebra[Alg[_[_, _, _, _, _]], F[_]] = Alg[Kind1[F]#toKind5]
+  type FunctorInterpreter[Op[_, _, _, _, _], F[_]] = PolyFunction5[Op, Kind1[F]#toKind5]
+  type BiFunctorAlgebra[Alg[_[_, _, _, _, _]], F[_, _]] = Alg[Kind2[F]#toKind5]
+  type BiFunctorInterpreter[Op[_, _, _, _, _], F[_,_]] = PolyFunction5[Op, Kind2[F]#toKind5]
```
MoeQuadrat, line 9: Whats your take on that FunctorTransformer/FunctorTransformation would be a more fitting type alias for FunctorInterpreter 
@Baccata

MoeQuadrat:
> @jona7o, I'd like to extend you the courtesy of asking for your opinion on naming and stuff, as this is both source and binary breaking and will induce some changes to your libraries (nothing fundamental, mostly mechanical things).

Was no hassle to built in the current polyfunction revamp in our library.

modules/core/src-2/kinds/package.scala
MoeQuadrat: I like the idea of `FunctorProjection5` and `BiFunctorProjection5`.  Seems more fitting than Interpreter in this use case. Taking the free-monad in consideration the current naming also makes sense.

Baccata:
> but since one motivation for the PR is to make transformations (er polyfunctions) easier to implement in user-land, we should make sure we document what that looks like

Yup, will defo add some docs in a subsequent PR. As a matter of fact I'm thinking about having a `smithy4s-cats` module and offer a few OOTB transformations there. The documentation should tie into that

modules/core/src-2/kinds/package.scala
lewisjkl: I like the idea of FunctorProjection as well. I am usually fine with long names though if they are more accurate or descriptive.

kubukoz:
LGTM (just make sure to update with main before merging). I'd just suggest that we have a simple lookup map for old->new names in the release notes.

sth like:


| old                  | new                               |
| -------------------- | --------------------------------- |
| smithy4s.Monadic     | smithy4s.kinds.FunctorAlgebra     |
| smithy4s.Interpreter | smithy4s.kinds.FunctorInterpreter |
| asTransformation     | toPolyFunction                    |

Playground seems happy with just the name swap: https://github.com/kubukoz/smithy-playground/pull/129

> As a matter of fact I'm thinking about having a `smithy4s-cats` module and offer a few OOTB transformations there.

I'm even thinking about an `-effect` module so that we could include stuff like this (requires CE-kernel):

```scala
def liftFunctorInterpreterResource[Op[_, _, _, _, _], F[_]: MonadCancelThrow](
  fir: Resource[F, FunctorInterpreter[Op, F]]
): FunctorInterpreter[Op, F] =
  new FunctorInterpreter[Op, F] {

    def apply[I, E, O, SI, SO](
      fa: Op[I, E, O, SI, SO]
    ): F[O] = fir.use(_.apply(fa))

  }
```

modules/core/src-2/kinds/package.scala
kubukoz: I like `FunctorInterpreter` because I feel like the "interpreter" part sort of explains what it _does_, not just what it _is_ technically - a more semantic name than "projection". I'm fine with either though

Baccata:
Alright, after deliberation, I think I'm gonna stay with `FunctorInterpreter`, for this PR anyway. I'm still open to changing, but considering it's just an alias, it's not the end of the world, and the intended usage is very much to be able to `extend FunctorInterpreter`, which communicates the intent better than `FunctorProjection`. 

I might make it package private before the release, to encourage third parties to apply whatever aliases float their boats. Not quite sure.

Baccata:
Merging this to create a basis for 0.17, will address the todos in later PRs

PR #561, opened 2022-10-28
Migrate smithy4s.api -> alloy

lewisjkl:
Updates usages of smithy4s.api, including simpleRestJson, to be pulled from [alloy](https://github.com/disneystreaming/alloy). Mainly is just replacing imports and namespaces. There are some differences since alloy is a dependency being pulled rather than coming from the colocated protocol module. These differences are slight.

Alloy also contains an openapi module that translates simpleRestJson services into openapi. This is now used for the http4s-swagger module and the colocated openapi module in smithy4s is removed.

In addition to usual testing, I've tried these changes out with https://github.com/lewisjkl/smithy4s-minimal-example/pull/1 and made sure it works fine there (including with different union types)


lewisjkl:
Looks like CI doesn't run if branches aren't targeting main?

modules/aws-http4s/test/src-jvm/smithy4s/aws/http4s/Main.scala
```diff
@@ -29,7 +29,7 @@ object Main extends IOApp.Simple {
       .listTables(limit = Some(ListTablesInputLimit(10)))
       .run
       .flatMap(IO.println(_))
-  // FIXME: Lambda uses @restJson1 which is currently unsupported: https://github.com/disneystreaming/smithy4s/issues/53
+  // FIXME: Lambda uses @simpleRestJson1 which is currently unsupported: https://github.com/disneystreaming/smithy4s/issues/53
```
lewisjkl, line 32: Todo: fix

modules/codegen/src/smithy4s/codegen/Codegen.scala
```diff
@@ -116,7 +117,7 @@ object Codegen { self =>
         namespaces
           .filterNot(_.startsWith("aws."))
           .filterNot(_.startsWith("smithy."))
-          .filterNot(reserved)
+          .filterNot(ns => reserved.exists(ns.startsWith))
```
lewisjkl, line 120: Changed to this so that if reserved contains `alloy` it will also filter out `alloy.proto` and future additions we may make.

modules/aws-http4s/test/src-jvm/smithy4s/aws/http4s/Main.scala
lewisjkl, line 32:
```suggestion
  // FIXME: Lambda uses @restJson1 which is currently unsupported: https://github.com/disneystreaming/smithy4s/issues/53
```

daddykotex (approved):
That's awesome, excellent work

daddykotex:
Checking out to test

PR #567, opened 2022-11-02
Move Op from parameter to member in smithy4s.Service

Baccata:
Similarly to what happened with `cats.Parallel`, the Operation type of `smithy4s.Service` can be moved from type-parameter to type-member. This simplifies type signatures for a number of public methods, and helps implicit search in some cases.


modules/core/src/smithy4s/Service.scala
```diff
@@ -34,33 +34,44 @@ import kinds._
   *   around makes it drastically easier to implement logic generically, without involving
   *   metaprogramming.
   */
-trait Service[Alg[_[_, _, _, _, _]], Op[_, _, _, _, _]] extends FunctorK5[Alg] with Service.Provider[Alg, Op] {
-  implicit val serviceInstance: Service[Alg, Op] = this
-  val service = this
+trait Service[Alg[_[_, _, _, _, _]]] extends FunctorK5[Alg] with Service.Provider[Alg] {
+  type Operation[E, I, O, SI, SO]
```
daddykotex, line 38: allows for a long name :D (not that it was not allowed of course, just that the long version in the type parameter as just so annoying

kubukoz (approved):

kubukoz:
sooo why didn't CI fail on any of the last commits?

lewisjkl:
> sooo why didn't CI fail on any of the last commits?

The CI doesn't run unless the PR is targeting main (afaik)

kubukoz:
we should probably change that, then :)

Baccata:
Problem was that PR trigger was using the `*` glob pattern when it should be using the `**` (there might be slashes in the branch name)

modules/core/src/generated/schema/PartiallyAppliedStruct.scala
```diff
@@ -19,1174 +19,39 @@
 package smithy4s
 package schema
 
-class PartiallyAppliedStruct[S] protected[schema] (placeholder: ShapeId) {
+class PartiallyAppliedStruct[S] protected[schema](placeholder: ShapeId) {
```
Baccata, line 22: This was previously being formatted by scalafmt when it shouldn't have been.

modules/core/src/smithy4s/Service.scala
```diff
@@ -34,33 +34,48 @@ import kinds._
   *   around makes it drastically easier to implement logic generically, without involving
   *   metaprogramming.
   */
-trait Service[Alg[_[_, _, _, _, _]], Op[_, _, _, _, _]] extends FunctorK5[Alg] with Service.Provider[Alg, Op] {
-  implicit val serviceInstance: Service[Alg, Op] = this
-  val service = this
+trait Service[Alg[_[_, _, _, _, _]]] extends FunctorK5[Alg] with Service.Provider[Alg] {
```
Baccata, line 37: Most of the changes derive from this line : the `Op` is moved from type-parameter to type-member.

modules/core/src/smithy4s/Service.scala
```diff
@@ -34,33 +34,48 @@ import kinds._
...
+  type Operation[I, E, O, SI, SO]
+  type Endpoint[I, E, O, SI, SO] = smithy4s.Endpoint[Operation, I, E, O, SI, SO]
+  type Interpreter[F[_, _, _, _, _]] = PolyFunction5[Operation, F]
+  type FunctorInterpreter[F[_]] = PolyFunction5[Operation, kinds.Kind1[F]#toKind5]
+  type BiFunctorInterpreter[F[_, _]] = PolyFunction5[Operation, kinds.Kind2[F]#toKind5]
```
Baccata, line 42: the type aliases above are path-dependant (by virtue of depending on the Operation abstract type member) , it means we can ascribe `service.Endpoint`, `service.Interpreter`, etc etc. This allows to clean up a number of verbose ascriptions here and there.

modules/http4s/src/smithy4s/http4s/SimpleProtocolBuilder.scala
```diff
@@ -80,33 +78,34 @@ abstract class SimpleProtocolBuilder[P](val codecs: CodecAPI)(implicit
 
     def routes[F[_]: EffectCompat](
         impl: FunctorAlgebra[Alg, F]
-    ): RouterBuilder[Alg, Op, F] =
-      new RouterBuilder[Alg, Op, F](
+    ): RouterBuilder[Alg, F] =
+      new RouterBuilder[Alg, F](
         service,
-        service.toPolyFunction[Kind1[F]#toKind5](impl),
```
Baccata, line 86: the implementation was prematurely turned into a polymorphic function

modules/core/src/smithy4s/Service.scala
```diff
@@ -34,33 +34,48 @@ import kinds._
...
 
 }
 
 object Service {
-  trait Provider[Alg[_[_, _, _, _, _]], Op[_, _, _, _, _]] extends HasId {
-    def service: Service[Alg, Op]
+
+  def apply[Alg[_[_, _, _, _, _]]](implicit ev: Service[Alg]): ev.type = ev
+
+  type Aux[Alg[_[_, _, _, _, _]], Op[_, _, _, _, _]] = Service[Alg]{ type Operation[I, E, O, SI, SO] = Op[I, E, O, SI, SO] }
```
Baccata, line 63: the `Aux` pattern allows to "recover" the type parameter that was moved, in locations where it matters

PR #568, opened 2022-11-03
Build http4s routers lazily

### LINKED ISSUE #501: Building a client or server throws before checking the protocol
We have code like this:

```scala
checkProtocol(service, protocolTag)
	.as(
      new SmithyHttp4sReverseRouter[Alg, Op, F](
        uri,
        service,
        client,
        EntityCompiler
          .fromCodecAPI[F](codecs)
      )
    )
```

`as` is eager, so if the argument throws (like it currently does if any of the operations in the service doesn't have `@http` or a proper method), the whole thing throws instead of being a `Left`.

This should be fixed, presumably with something like `.map(_ =>` instead of `as`.

Relevant code: `ClientBuilder`, `RouterBuilder` in `smithy4s-http4s`.

To reproduce the issue outside of the repository, write a service like this:

```
service MyService {
  operations: [MyOperation]
}
```

and try to generate a client or server for it, using `SimpleRestJsonBuilder` - here are [some docs that showcase the process](https://disneystreaming.github.io/smithy4s/docs/overview/quickstart/).

Ideally, this would be tested with a unit test in the http4s module, using one of the protocol-less example services that we currently generate, like [`KVStore`](https://github.com/disneystreaming/smithy4s/blob/71fbf89363a2cb73ef05599d09c8fb050dd0328d/sampleSpecs/kvstore.smithy) (might not be available in the http4s tests right now, so maybe not the best choice) or [`Weather`](https://github.com/disneystreaming/smithy4s/blob/71fbf89363a2cb73ef05599d09c8fb050dd0328d/sampleSpecs/weather.smithy#L6) (it's available in the http4s tests).

Also, we don't have any tests for returning `UnsupportedProtocolError` (the case where `checkProtocol` returns a Left), so these could be added as well. One service that should fail it is [`DummyService`](https://github.com/disneystreaming/smithy4s/blob/71fbf89363a2cb73ef05599d09c8fb050dd0328d/sampleSpecs/metadata.smithy#L7-L10) (it has operations that are `@http` but it doesn't have a protocol) or [`EmptyService`](https://github.com/disneystreaming/smithy4s/blob/71fbf89363a2cb73ef05599d09c8fb050dd0328d/sampleSpecs/misc.smithy#L5) (it has no operations and no protocol).

kubukoz:
Closes #501. The main issue would've been solved by #493, but even that PR doesn't address the eagerness problem of `as` - we shouldn't be building the whole router if it's not even going to be returned.

### DIFF
--- a/modules/http4s/src/smithy4s/http4s/SimpleProtocolBuilder.scala
+++ b/modules/http4s/src/smithy4s/http4s/SimpleProtocolBuilder.scala
@@ -103,7 +103,9 @@ abstract class SimpleProtocolBuilder[P](val codecs: CodecAPI)(implicit
 
     def use: Either[UnsupportedProtocolError, Monadic[Alg, F]] = {
       checkProtocol(service, protocolTag)
-        .as(
+        // Making sure the router is evaluated lazily, so that all the compilation inside it
+        // doesn't happen in case of a missing protocol
+        .map { _ =>
           new SmithyHttp4sReverseRouter[Alg, Op, F](
             uri,
             service,
@@ -111,7 +113,7 @@ abstract class SimpleProtocolBuilder[P](val codecs: CodecAPI)(implicit
             EntityCompiler
               .fromCodecAPI[F](codecs)
           )
-        )
+        }
         .map(service.transform[GenLift[F]#λ](_))
     }
   }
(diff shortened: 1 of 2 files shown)


kubukoz:
yo @daddykotex I see you're active, can you please take a look? :)

daddykotex (approved):
Looks good
That's the hacktober issue right

kubukoz:
thanks!

Issue #531, opened 2022-10-19
Big json arrays?


Quafadas:
It would appear that a smithy client which requests a large JSON array, is presented with this response. 

```
PayloadError(.data, expected = list, message=Input list exceeded max arity of 1024)
```
According to a DM with @Baccata , the most likely reason, is that the parser is intentionally set, to limit the parsing of large arrays for security reasons. 

Proposed Outcome:  Offer a mechansim via which the client would attempt to parse this unreasonably large array. Perhaps by offering a "turn off the security" flag?

yisraelU:
@Quafadas can you pull latest version and verify that the new feature allowing passing in max arity solves your problem

PR #569, opened 2022-11-03
allow passing in the MaxArity for the JsonSchemaVisitor

yisraelU:
I added a method build to restJsonBuilder companion to disambiguate between the apply method which has a default value. 
Im trying to minimize impact on UX.
RE: #531


modules/example/src/smithy4s/example/Launcher.scala
```diff
@@ -31,7 +31,7 @@ object Launcher extends IOApp {
     val serverRes = for {
       impl <- Resource.eval(ObjectServiceImpl.makeIO)
       docs = smithy4s.http4s.swagger.docs[IO](ObjectService)
-      service <- SimpleRestJsonBuilder.routes(impl).resource
+      service <- SimpleRestJsonBuilder.build.routes(impl).resource
```
lewisjkl, line 34: Hmm not sure I like the idea of building the builder like this. Seems like the ideal UX would be something like:

```
SimpleRestJsonBuilder.routes(impl).withMaxJsonArity(2048).resource
```
where the call to `withMaxJsonArity` is not necessary if you want to use the default.

We could also take the more flexible approach of:

```
SimpleRestJsonBuilder.routes(impl).withJsonCodec(myJsonCodecApi).resource
```

What do you think?

yisraelU: I definitely agree,
The problem is that SimpleProtocolBuilder doesnt have access to the JsonCodecApi as its typed to CodecApi, so we cant access the existing codec to make a copy .
We can create a new instance of JsonCodecApi inside the builder

kubukoz: does it even need to be typed to `JsonCodecApi`? Couldn't it be just `CodecApi`?

yisraelU: @kubukoz does it make sense to allow any codec on a RestJsonBuilder ?

kubukoz: hmm now that you said it out loud...

modules/http4s/src/smithy4s/http4s/SimpleRestJsonBuilder.scala
```diff
@@ -19,12 +19,17 @@ package http4s
 
 import smithy4s.internals.InputOutput
 
-object SimpleRestJsonBuilder
+object SimpleRestJsonBuilder extends SimpleRestJsonBuilder(1024) {
+  def withMaxArity(maxArity: Int): SimpleRestJsonBuilder =
+    new SimpleRestJsonBuilder(maxArity) {}
+}
+abstract class SimpleRestJsonBuilder(maxArity: Int)
```
daddykotex, line 26: minor: does it need to be abstract or what's the gain in the case?

yisraelU: It doesnt .
I left it abstract for now to have a discussion, simply because SimpleProtocolBuilder is abstract and it doesn't need to be either.

modules/http4s/src/smithy4s/http4s/SimpleRestJsonBuilder.scala
```diff
@@ -19,11 +19,8 @@ package http4s
 
 import smithy4s.internals.InputOutput
 
-object SimpleRestJsonBuilder extends SimpleRestJsonBuilder(1024) {
-  def withMaxArity(maxArity: Int): SimpleRestJsonBuilder =
-    new SimpleRestJsonBuilder(maxArity) {}
-}
-abstract class SimpleRestJsonBuilder(maxArity: Int)
+object SimpleRestJsonBuilder extends SimpleRestJsonBuilder(1024)
+case class SimpleRestJsonBuilder(maxArity: Int)
```
Baccata, line 23: let's go back to a normal class + the `withMaxArity` method. Case classes are notoriously bad for binary compatibility (see [this](https://github.com/scala/improvement-proposals/pull/50) for instance) and this builder is something very public-facing.

kubukoz: thanks for the link, I forgot about that SIP!

yisraelU: Thanks, will take a look at that

modules/http4s/src/smithy4s/http4s/SimpleRestJsonBuilder.scala
```diff
@@ -19,12 +19,17 @@ package http4s
 
 import smithy4s.internals.InputOutput
 
-object SimpleRestJsonBuilder
+object SimpleRestJsonBuilder extends SimpleRestJsonBuilder(1024) {
+  def withMaxArity(maxArity: Int): SimpleRestJsonBuilder =
```
Baccata, line 23: the method ought to go on the class, not the object.

yisraelU: I thought it a bit odd to be on the class ,being that the class has a maxArity parameter in the constructor.

Baccata: Yeah, that's the "fluent builder" pattern. Each `withMethod` is basically equivalent to a `copy` modifying just one field, but has the quality of being much more nicer wrt bincompat.

The point of having the method on the class is that we can add more methods later on if we need to amend the builder to carry more customisation.

Baccata:
Also this PR needs to be sync'd with `series/0.17` to trigger CI

modules/json/src/smithy4s/http/json/codecs.scala
```diff
@@ -43,13 +44,18 @@ object codecs {
...
   ): SchemaVisitor[JCodec] =
-    new SchemaVisitorJCodec(maxArity = 1024, cache)
+    new SchemaVisitorJCodec(maxArity, cache)
 
   private[smithy4s] val schemaVisitorJCodec: SchemaVisitor[JCodec] =
-    new SchemaVisitorJCodec(maxArity = 1024, CompilationCache.nop[JCodec])
+    new SchemaVisitorJCodec(
+      maxArity = defaultMaxArity,
+      CompilationCache.nop[JCodec]
+    )
```
lewisjkl, line 59: What is this version currently used for? From a quick look it seems like only in tests. Maybe we can remove.

lewisjkl (commented):
Let's add some documentation showing how to increase the maxArity and what that means

modules/json/src/smithy4s/http/json/codecs.scala
yisraelU: Im not sure what you mean by version?

yisraelU: > Let's add some documentation showing how to increase the maxArity and what that means

good point

lewisjkl: > Im not sure what you mean by version?

Yeah sorry that is super confusing! I meant the "Val" vs the "def" above. The val I'm not sure is used anywhere. Thought we could remove it if that is in fact the case.

PR #570, opened 2022-11-03
Error handling improvements

### LINKED ISSUE #413: Generated client does not map HTTP error to custom type
Hi, 

I have defined a custom error type and an operation that returns that error
```
namespace acme.api

@simpleRestJson
service HelloWorldService {
  version: "1.0.0",
  operations: [Boom]  
}

@http(method: "POST", uri: "/boom", code: 200)
operation Boom {
  input: Unit,
  output: Unit,
  errors: [MyError]
}

@error("client")
structure MyError {
}
```
Note that I did not specify _which_ client error should be mapped here (i.e., there is no `@httpError` annotation). Further I have generated a client for the `HelloWorldService`: 
```scala
HelloWorldService.simpleRestJson.clientResource(
    httpClient,
    Uri.unsafeFromString("http://google.com/giveme404")
)
```
As you can see, the client will call a URL that returns a 404. According to [the smithy documentation](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html)

> The httpError trait is used to set a custom HTTP response status code. By default, error structures with no httpError trait use the default HTTP status code of the [error trait](https://awslabs.github.io/smithy/1.0/spec/core/type-refinement-traits.html#error-trait).
>
> -  400 is used for "client" errors
> -  500 is used for "server" errors

Hence I would expect that smithy4s would map all client errors to `MyError`. This is the case with `v0.14.2`, but not with `v0.15.1`. Now it will instead raise an `UnknownErrorResponse` error. I can provide a repository with a small example if necessary.

I believe I found the [piece of code](https://github.com/disneystreaming/smithy4s/blob/c9839dbdf2b7f81478e233beae0853ad499a0c2a/modules/core/src/smithy4s/http/ErrorAltPicker.scala#L44) that is responsible for this behaviour, but before I try to "fix" it I would like to know whether this is indeed a bug. I found the [commit](https://github.com/disneystreaming/smithy4s/commit/092b4beadd1141cc1a0db89088ebd7187dc78adc) where this was changed and am not sure if maybe it is now behaving as intended.

lewisjkl:
Updates error handling to close #413.

See attached docs in markdown file in PR for more information.


daddykotex:
note: I explicitly removed all that codegen noise from my other PR because I know @Baccata already has those changes in #567

lewisjkl:
> note: I explicitly removed all that codegen noise from my other PR because I know @Baccata already has those changes in #567

Which codegen noise are you referring to?

daddykotex:
Screen Shot 2022-11-03 at 16 20 44

You touched like 4 files, and like 100 lines or so but your PR is almost 2k lines

lewisjkl:
Ahh I see, I didn't even look at the github diff (my bad) but yeah I will address that. Thanks!

lewisjkl:
Marking as draft until we get 0.17 CI passing in general, then will update this branch and re-mark as ready for review

modules/docs/src/03-protocols/02-simple-rest-json/03-client.md
```diff
@@ -42,3 +42,109 @@ object Clients {
...
+
+```kotlin
+@httpError(404)
+@error("client")
+structure AnotherNotFoundError {
+  message: String
+}
+```
+
+Then I will get the following:
```
daddykotex, line 140: minor: most of the documentation is impersonal

by that I mean that we don't have many instances of `I do this` or `I get this`. I think we should keep the same style.

Otherwise good job on the explanations, it's well presented

lewisjkl: Good catch, I slipped into first person there. I will fix!

sampleSpecs/pizza.smithy
```diff
@@ -116,6 +116,12 @@ structure FallbackError {
   error: String
 }
 
+@error("client")
+structure FallbackError2 {
+  @required
+  error: String
+}
+
```
lewisjkl, line 124: Adding this was the only way to keep a current PizzaClientSpec test passing. We could also remove the test if we prefer. But in order to get an `UnknownErrorResponse` now, there must be two errors on an operation that have no status code with only a `@error("client")` annotation. Before this PR, it would never fallback to errors with no status code which is why the test case used to pass.

daddykotex: This is perfect for me

maybe a comment on top of the shape that explains it serves as a way to add ambiguity to the error picking algorithm so that it falls back to `UnknownError` ?

lewisjkl: Yeah good idea, will add!

Baccata (approved):
Great work, loving the docs and the tests 👍

modules/core/src/smithy4s/http/ErrorAltPicker.scala
```diff
@@ -41,29 +41,38 @@ final class ErrorAltPicker[E](alts: Vector[SchemaAlt[E, _]]) {
...
+      if (results.length == 1) results.headOption else None
+    }
+    lazy val fallbackError: Option[SchemaAlt[E, _]] = {
+      val matchingAlts = alts.filter { alt =>
+        alt.hints.get(Error) match {
+          case Some(e) if alt.hints.get(HttpError).isEmpty =>
+            // If Error trait is present (must be 'client' or 'server') and
+            // the status code is in the client/server error range, then use that alt
+            val isClientError =
+              e.value.toLowerCase == "client" && inputStatus >= 400 && inputStatus < 500
```
Baccata, line 57: nitpick : if `e == Error.Client`

modules/core/src/smithy4s/http/ErrorAltPicker.scala
```diff
@@ -41,29 +41,38 @@ final class ErrorAltPicker[E](alts: Vector[SchemaAlt[E, _]]) {
...
-          .get(HttpError)
-          .map(_.value)
-          .orElse(alt.hints.get(Error).map {
-            case Error.CLIENT => 400
-            case Error.SERVER => 500
-          })
-      }
-      .collect {
-        case (Some(key), values) if values.size == 1 => key -> values.head
+  private val byStatusCode: Int => Option[SchemaAlt[E, _]] = { inputStatus =>
```
Baccata, line 44: nitpick : hotpath

daddykotex: second attempt:

```diff
diff --git a/modules/core/src/smithy4s/http/ErrorAltPicker.scala b/modules/core/src/smithy4s/http/ErrorAltPicker.scala
index 91c297df..cb57cdad 100644
--- a/modules/core/src/smithy4s/http/ErrorAltPicker.scala
+++ b/modules/core/src/smithy4s/http/ErrorAltPicker.scala
@@ -41,29 +41,51 @@ final class ErrorAltPicker[E](alts: Vector[SchemaAlt[E, _]]) {
     .map(alt => alt.instance.shapeId.name -> alt)
     .toMap[String, SchemaAlt[E, _]]
 
-  private val byStatusCode: Int => Option[SchemaAlt[E, _]] = { inputStatus =>
-    val errorForStatus: Option[SchemaAlt[E, _]] = {
-      val results =
-        alts.filter(_.hints.get(HttpError).exists(_.value == inputStatus))
-      if (results.length == 1) results.headOption else None
-    }
-    lazy val fallbackError: Option[SchemaAlt[E, _]] = {
-      val matchingAlts = alts.filter { alt =>
-        alt.hints.get(Error) match {
-          case Some(e) if alt.hints.get(HttpError).isEmpty =>
-            // If Error trait is present (must be 'client' or 'server') and
-            // the status code is in the client/server error range, then use that alt
-            val isClientError =
-              e.value.toLowerCase == "client" && inputStatus >= 400 && inputStatus < 500
-            val isServerError =
-              e.value.toLowerCase == "server" && inputStatus >= 500 && inputStatus < 600
-            isClientError || isServerError
-          case _ => false
-        }
+  // build a map: status code to alternative
+  // exclude all status code that are used on multiple alternative
+  // in essence, it gives a `Map[Int, SchemaAlt[E, _]]` that's used
+  // for the lookup
+  private val byStatusCode: Int => Option[SchemaAlt[E, _]] = {
+    val perStatusCode: Map[Int, SchemaAlt[E, _]] = alts
+      .flatMap { alt =>
+        alt.hints.get(HttpError).map { he => he.value -> alt }
+      }
+      .groupMap(_._1)(_._2)
+      .collect {
+        // Discard alternative where another alternative has the same http status code
+        case (status, allAlts) if allAlts.size == 1 => status -> allAlts.head
+      }
+    val errorForStatus: Int => Option[SchemaAlt[E, _]] = perStatusCode.get
+
+    lazy val fallbackError: Int => Option[SchemaAlt[E, _]] = {
+      // grab the alt that's annotated with the expected `Error` hint
+      // only if there is only one
+      def forErrorType(expected: Error): Option[SchemaAlt[E, _]] = {
+        val matchingAlts = alts
+          .flatMap { alt =>
+            alt.hints
+              .get(HttpError)
+              .fold(
+                alt.hints.get(Error).collect {
+                  case e if e == expected => alt
+                }
+              )(_ => None)
+
+          }
+        if (matchingAlts.size == 1) matchingAlts.headOption else None
+      }
+      val clientAlt: Option[SchemaAlt[E, _]] = forErrorType(Error.CLIENT)
+      val serverAlt: Option[SchemaAlt[E, _]] = forErrorType(Error.SERVER)
+
+      { intStatus =>
+        if (intStatus >= 400 && intStatus < 500) clientAlt
+        else if (intStatus >= 500 && intStatus < 600) serverAlt
+        else None
       }
-      if (matchingAlts.size == 1) matchingAlts.headOption else None
     }
-    errorForStatus.orElse(fallbackError)
+
+    inputStatus =>
+      errorForStatus(inputStatus).orElse(fallbackError(inputStatus))
   }
 
   def getPreciseAlternative(
```

side note: I'm not sure we should consider error paths as hot path

daddykotex: I can push it if you like it @lewisjkl

Issue #504, opened 2022-10-09
AWS: putting specs in a local jar dependency makes validation fail downstream


kubukoz:
Consider this build:

```scala
Global / onChangedBuildSource := ReloadOnSourceChanges

val core = project
  .enablePlugins(Smithy4sCodegenPlugin)
  .settings(
    scalaVersion := "2.13.8",
    libraryDependencies ++= Seq(
      "com.disneystreaming.smithy4s" %% "smithy4s-aws" % smithy4sVersion.value
    )
  )

val root = project
  .in(file("."))
  .enablePlugins(Smithy4sCodegenPlugin)
  .settings(
    scalaVersion := "2.13.8"
  )
  .dependsOn(core)
```

If you put one AWS spec, say `kinesis.json` into `core`, `core/compile` will succeed but `root/compile` will fail at the `smithy4sCodegen` step:

```
sbt:root> compile
[error] stack trace is suppressed; run last Compile / smithy4sCodegen for the full output
[error] (Compile / smithy4sCodegen) software.amazon.smithy.model.validation.ValidatedResultException: Result contained ERROR severity validation events: 
[error] [ERROR] com.amazonaws.lambda#Lambda1: Unable to resolve trait `aws.protocols#restJson1`. If this is a custom trait, then it must be defined before it can be used in a model. | Model jar:file:/Users/kubukoz/projects/demos/core/target/scala-2.13/core_2.13-0.1.0-SNAPSHOT.jar!/META-INF/smithy/kinesis.json:7:36
[error] Total time: 0 s, completed Oct 9, 2022, 4:44:10 PM
```

Extra info:

- This also happens if you split the dependencies between `Compile` and `Test` of a single module.
- It doesn't happen if `core` is a remote dependency (tested with `core/publishLocal` and replacing `dependsOn` with a `libraryDependencies` entry)
- As I understand, the definitions that aren't being found are runtime dependencies of the sbt plugin, and should be seen by the validators.
- It works in 0.15.3, doesn't work in either 0.16.1, 0.16.2, 0.16.3 or a snapshot from 1fdb46b105cddb8f48609a4f695e505bddf5e3b9

~~I'll open a PR with a scripted test for this soon.~~ #505

Workaround: add to `root`:

```scala
libraryDependencies ++= Seq(
  "software.amazon.smithy" % "smithy-aws-traits" % smithy4s.codegen.BuildInfo.smithyVersion % Smithy4s
)
```

[...]

PR #573, opened 2022-11-04
[series/0.17] fix CI trigger (and build)

Baccata:
Fixes the blob pattern used by the CI trigger to fix workflows not running on PRs targeting the series/0.17 branch

closes https://github.com/disneystreaming/smithy4s/issues/504

#### Note about the large diff 

This is only due to the fact that the scalafmt.conf was off and some generated sources should not have been formatted, ever. 

#### Regarding change of assumed dependencies 

Avoid assuming dependencies during codegen (except for alloy)

* Only dependency assumed is alloy-core
* Assumption happens via build-tools default settings being set by
plugins/modules
* If users want to generate code from AWS, they need to manually
ascribe the dependencies containing AWS traits (updated the documentation in consequence)

Alloy can evolve independently from Smithy4s, and we cannot assume the files it contains will remain at stable locations across versions, so hardcoding a reference to the smithy files contained by alloy artifacts would be ill advised. Likewise, we cannot force the use "discoverModels" because it can backfire easily. The only solution is to have the plugins force the fetching of alloy (by default) at the build-tool level.

#### Mitigating the problem 

The user-declared dependencies are now propagating transitively. This means that, internally to a project, users can declare a smithy dependency once, and downstream projects will see it too. It's not perfect but alleviate the pain a little. 

#### Alternative considered 

Over the past few days I've done a fair bit of research to see whether we could bake something that would involve modules tracking the dependencies their used during their code generation, so that downstream could use that information to save the users having to worry about where some traits came from. My conclusion is that, though this would be ideal and provide the best user experience, it would require a lot of fiddling in non-trivial build-tool-specific layers (implying duplication of efforts between mill and SBT). Essentially, we'd need to bake some extra information in poms issued by projects that have smithy4s plugins enabled, and handcraft some resolution logic (respecting each build tool semantics) to do yet another pass of resolution after resolving what the user stated. If that would be reasonably easy on mill, it's much harder on SBT that effectively hides coursier behind layers and layers of indirections and doesn't give us a handle to the information we need in the APIs it exposes (like, we can't get a reference to the raw pom.xml when using SBT's resolution mechanism)". 

So, though this problem is very much solvable, the benefit/cost ratio is too low to pull the trigger on this at this time. Requiring more information from the users who want to generate AWS specs, or compliance-tests, is more acceptable (at this point in time, anyway). 

#### Edit

Some folks gave me some hints that may make the alternative described above possible. If that's the case, it'll come in a later PR


build.sbt
```diff
@@ -995,15 +990,27 @@ def genSmithyImpl(config: Configuration) = Def.task {
...
+              val discoverModelsOpt =
+                (if (discoverModels) List("--discover-models") else Nil)
+              val allowedNsOpt =
+                if (allowedNS.isDefined)
+                  List("--allowed-ns", allowedNS.get.mkString(","))
+                else Nil
+              val skipOpt = skip.flatMap(s => List("--skip", s))
+              val dependenciesOpt =
+                if (smithy4sDependencies.nonEmpty)
+                  List("--dependencies", smithy4sDependencies.mkString(","))
```
Baccata, line 1005: added this option

modules/core/src-2/generated/kinds/functorK.scala
```diff
@@ -19,58 +19,35 @@
 package smithy4s
 package kinds
 
-trait FunctorK[Alg[_[_]]] {
+trait FunctorK[Alg[_[_]]]{
```
Baccata, line 22: just formatting changes

modules/core/src-2/generated/kinds/kinds.scala
```diff
@@ -19,20 +19,17 @@
 package smithy4s
 package kinds
 
-object Kind1 {
+object Kind1{
```
Baccata, line 22: just formatting changes

modules/core/src/generated/schema/PartiallyAppliedStruct.scala
```diff
@@ -19,1174 +19,39 @@
 package smithy4s
 package schema
 
-class PartiallyAppliedStruct[S] protected[schema] (placeholder: ShapeId) {
+class PartiallyAppliedStruct[S] protected[schema](placeholder: ShapeId) {
```
Baccata, line 22: just formatting changes in this file

build.sbt
daddykotex: I love that you got rid of most genDiscoverModels

Baccata:
@kubukoz , I'm tagging this PR as closing this : https://github.com/disneystreaming/smithy4s/issues/504 : the `smithy-aws-traits` now has to provided explicitly by the users, but now propagates correctly to downstream project. 

I'm aware that the explicit providing is not ideal, but trying to automate the propagation through "maven boundaries" is way too involved for what I want to go in the next release.

modules/codegen/src/smithy4s/codegen/ModelLoader.scala
```diff
@@ -25,82 +25,67 @@ import software.amazon.smithy.model.Model
...
+    // Loading the upstream model
+    val upstreamModel = Model
       .assembler()
+      // disabling cache to support snapshot-driven experimentation
+      .putProperty(ModelAssembler.DISABLE_JAR_CACHE, true)
       .addClasspathModels(currentClassLoader, discoverModels)
+      .addImports(modelsInJars)
       .assemble()
       .unwrap()
-      .toBuilder()
```
Baccata, line 54: this logic is made simpler by the fact that we're removing the assumption that (alloy) models are communicated through the classpath. The plugins (SBT, mill, CLI) are now responsible for setting the dependency to alloy as default.

modules/core/src-3/generated/kinds/kinds.scala
```diff
@@ -19,21 +19,17 @@
 package smithy4s
 package kinds
 
-object Kind1 {
+object Kind1{
```
Baccata, line 22: just formatting changes

modules/mill-codegen-plugin/src/smithy4s/codegen/mill/Smithy4sModule.scala
```diff
@@ -62,8 +70,12 @@ trait Smithy4sModule extends ScalaModule {
   def smithy4sSmithyLibrary: T[Boolean] = true
 
   def smithy4sTransitiveIvyDeps: T[Agg[Dep]] = T {
-    smithy4sIvyDeps() ++ T
-      .traverse(moduleDeps)(_.transitiveIvyDeps)()
```
Baccata, line 66: was not recursively calling the correct task there

modules/mill-codegen-plugin/src/smithy4s/codegen/mill/Smithy4sModule.scala
```diff
@@ -62,8 +70,12 @@ trait Smithy4sModule extends ScalaModule {
...
+    smithy4sAllDeps() ++ T
+      .traverse(moduleDeps) {
+        case m if m.isInstanceOf[Smithy4sModule] =>
```
daddykotex, line 75: interesting

daddykotex:
> I'm satisfied with the content of the PR, but I'll build it locally and update some example projects that I have locally and see if anything breaks in unexpected ways

Works fine for a dummy project, made me realize that we have to update: https://github.com/disneystreaming/smithy4s.g8/pull/14

lewisjkl (approved):

PR #583, opened 2022-11-09
Merge main into 0.17

kubukoz:


modules/http4s/src/smithy4s/http4s/SimpleProtocolBuilder.scala
```diff
@@ -105,15 +105,17 @@ abstract class SimpleProtocolBuilder[P](val codecs: CodecAPI)(implicit
...
             uri,
             service,
             client,
             EntityCompiler
               .fromCodecAPI[F](codecs)
           )
-        )
+        }
         .map(service.fromPolyFunction[Kind1[F]#toKind5](_))
     }
```
kubukoz, line 120: There was a conflict here that I had to resolve locally

Baccata (approved):

PR #587, opened 2022-11-09
Support multiple input dirs (sbt, mill)

### LINKED ISSUE #464: sbt/codegen: multiple input dirs? (`sourceDirectory` vs `sourceDirectories`)
An sbt project doesn't necessarily have exactly one source directory. For example, in an sbt-typelevel project with `CrossType.Full`, it makes a huge difference:

image

Apparently, in this setup, if I have both JS and JVM enabled, I need to copy-paste (or symlink, at least) the Smithy files so that they are seen in each variant's main `sourceDirectory`.

A workaround is setting `smithy4sInputDir` explicitly, but I feel like I'm missing something - or that it would be a good idea to support multiple `sourceDirectories` entries in the plugin itself.

lewisjkl:
This change will allow builds that use sbt-crossproject to have smithy specs located in a shared folder and reused across different variants (e.g. JVM, JS, Native).

Deprecates the `smithy4sInputDir` setting in favor of the `smithy4sInputDirs` setting that takes a list. This new setting's default uses the unmanagedSources settingkey to determine where to look for smithy specs. All smithy specs found will be combined into a single model at the time of codegen.

I need to think about if we will need something similar for mill or not. I kinda think it wouldn't be necessary though.

Closes #464.


lewisjkl:
Also, I tested this locally with a project I created using the `sbt-typelevel` g8 template. Can push that project if anyone is interested.

kubukoz:
FWIW it's not just sbt-crossproject, people might just have multiple source unmanaged source directories for other (maybe crossbuilding, maybe not) reasons :)

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -97,8 +102,17 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
   override def projectConfigurations: Seq[Configuration] = Seq(Smithy4s)
 
   // Use this with any configuration to enable the codegen in it.
+  @annotation.nowarn(
```
kubukoz, line 105: wait, this works in 2.12???

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -33,11 +33,16 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
     val smithy4sVersion =
       settingKey[String]("Smithy4sVersion")
 
+    @deprecated("Deprecated in 0.17, use `smithy4sInputDirs` instead.")
```
kubukoz, line 36: I'm not sure if we need to deprecate it given we're about to break BC. Might as well just remove it, breaking plugin (only for the users who have a custom value here). Those who have an explicit smithy4sInputDir can migrate very easily.

kubukoz:
> I need to think about if we will need something similar for mill or not. I kinda think it wouldn't be necessary though.

imo it's the same situation, there might be reasons to have more than 1 local source of Smithy specs

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
lewisjkl: Apparently lol 🤷

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
lewisjkl: Yeah I'm cool with that personally. What say you @Baccata?

lewisjkl:
> imo it's the same situation, there might be reasons to have more than 1 local source of Smithy specs

Yeah you may be right, I'll give the mill plug-in a look tomorrow and see what I can do

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
Baccata: Yeah, happy to remove

PR #588, opened 2022-11-09
Use get default func

### LINKED ISSUE #382: SchemaField#default: provide variant that decodes the document internally
Something like 

```
//in implicit class for SchemaField[S, A]
def getDefaultValue: Option[A] = getDefault.map(getCachedDocumentDecoder[A].apply(_))
```

lewisjkl:
Use the functions introduced in #580.

Once this is merged, we can close #382.

### DIFF
--- a/modules/core/src/smithy4s/internals/DocumentDecoderSchemaVisitor.scala
+++ b/modules/core/src/smithy4s/internals/DocumentDecoderSchemaVisitor.scala
@@ -21,7 +21,6 @@ import java.util.Base64
 import java.util.UUID
 
 import smithy.api.JsonName
-import smithy.api.Default
 import smithy.api.TimestampFormat
 import smithy.api.TimestampFormat.DATE_TIME
 import smithy.api.TimestampFormat.EPOCH_SECONDS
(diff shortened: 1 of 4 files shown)


modules/core/src/smithy4s/internals/DocumentDecoderSchemaVisitor.scala
```diff
@@ -315,7 +312,7 @@ class DocumentDecoderSchemaVisitor(
     ) => Unit = {
       val jLabel = jsonLabel(field)
 
-      val maybeDefault = getDefault(field)
+      val maybeDefault = field.instance.getDefault
```
kubukoz, line 315: I _think_ this might not be covered by tests.

Previously this would be `Option[Document]`, now it's `Option[A]`... if that doesn't cause an error I'm not sure it's even being exercised.

kubukoz (changes requested):

modules/core/src/smithy4s/internals/DocumentDecoderSchemaVisitor.scala
lewisjkl: I can double check either way, but `getDefault` returns an `Option[Document]` while `getDefaultValue` returns an `Option[A]`

kubukoz (approved):

PR #595, opened 2022-11-10
Render "Default" classes that can be used to stub an implementation

Baccata:


modules/docs/markdown/01-overview/06-stubs.md
```diff
@@ -0,0 +1,26 @@
+---
+sidebar_label: Stubbed implementations
+title: Stubbed implementations
+---
+
+For various reasons, such as testing/mocking, you may want to instantiate a stub implementation of generated service interfaces. Smithy4s makes it easy, by generating a `Default` class in the companion object of each service.
+
+This class has a constructor parameter that require a value. This value is what is returned when invoking any of the methods
```
lewisjkl, line 8:
```suggestion
This class has a constructor parameter that requires a value. This value is what is returned when invoking any of the methods
```

lewisjkl (approved):

modules/core/test/src/smithy4s/StubSmokeSpec.scala
```diff
@@ -0,0 +1,30 @@
...
+ *  limitations under the License.
+ */
+
+package smithy4s
+
+import munit._
+class DefaultServiceSmokeSpec() extends FunSuite {
+
+  test("Default stubs do compile") {
+    object stub extends smithy4s.example.Weather.Default[Option](None)
```
kubukoz, line 23: Now that I think about it... why not make Default a trait with an abstract member?

daddykotex:
Can I add a commit and replace some places I know could benefit from a stub? (compliance-tests work that I have done)

modules/core/test/src/smithy4s/StubSmokeSpec.scala
kubukoz: That's what I was thinking :)

daddykotex: I naively tried to convert `Constant` to a `trait` but because it takes a parameter and that parameter is forwarded to the extends `Transformed` class, I could not pull it off

daddykotex (approved):

modules/core/test/src/smithy4s/StubSmokeSpec.scala
Baccata: @daddykotex is correct.

kubukoz: ah, I see the issue now. had to try it by myself to understand :)

```scala
line"class $Constant_[P[-_, +_, +_, +_, +_]](value: P[Any, Nothing, Nothing, Nothing, Nothing]) extends $Transformed_[$opTraitNameRef, P](reified, $const5_(value))",
```

PR #596, opened 2022-11-10
Catch PayloadErrors in complete decoding in CodecAPI

### LINKED ISSUE #594: JSON decoding throws on top-level struct fields missing
For the following model:

```smithy
namespace hello

structure Event {
    @required
    payload: Payload
}

structure Payload {
    @required
    id: String
}
```

This code throws:

```scala
import hello._

object Main extends App {
  val capi =
    smithy4s.http.json.codecs.apply(smithy4s.http.json.codecs.defaultHintMask)

  val codec =
    capi.compileCodec(
      Event.schema
    )

  println {
    // this is an Either[PayloadError, Event]
    capi.decodeFromByteArray(
      codec,
      """{}""".getBytes
    )
  }
}
```

Exception:

```
PayloadError(.payload, expected = payload, message=Missing required field)
```

Notes:

- This happens in 0.16.7, 0.15.2 and even 0.12.0, so I think it's always been like that.
- if you try to decode `{"payload": {}}`, you get a `Left` as expected
- Unions don't seem to be affected

kubukoz:
Closes #594. Targetting main because this can be released in a patch release.

### DIFF
--- a/modules/json/test/src/smithy4s/http/json/JsonCodecApiTests.scala
+++ b/modules/json/test/src/smithy4s/http/json/JsonCodecApiTests.scala
@@ -25,4 +25,24 @@ class JsonCodecApiTests extends FunSuite {
 
     assertEquals(encodedString, """{"a":"test"}""")
   }
+
+  test(
+    "struct codec with a required field should return a Left when the field is missing"
+  ) {
+    val schemaWithRequiredField =
+      Schema
+        .struct[String]
+        .apply(
+          Schema.string
+            .required[String]("a", identity)
+        )(identity)
+
+    val capi = codecs(HintMask.empty)
+
+    val codec = capi.compileCodec(schemaWithRequiredField)
+
+    val decoded = capi.decodeFromByteArray(codec, """{}""".getBytes())
+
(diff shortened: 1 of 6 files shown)


modules/json/test/src/smithy4s/http/json/JsonCodecApiTests.scala
```diff
@@ -25,4 +25,24 @@ class JsonCodecApiTests extends FunSuite {
...
+        .apply(
+          Schema.string
+            .required[String]("a", identity)
+        )(identity)
+
+    val capi = codecs(HintMask.empty)
+
+    val codec = capi.compileCodec(schemaWithRequiredField)
+
+    val decoded = capi.decodeFromByteArray(codec, """{}""".getBytes())
```
plokhotnyuk, line 44: Will it return `Left` for empty or invalid JSON input?

kubukoz: Yep. That might be covered by other tests... but we certainly didn't have a test for this :)

daddykotex (approved):
Good findings

Issue #256, opened 2022-06-03
Generate Scaladoc 


kubukoz:
Originally mentioned in https://github.com/disneystreaming/smithy4s/issues/245 : the `@deprecated` trait should generate a `@deprecated` annotation in the Scala code, and the `@documentation` / `@externalDocumentation` traits should result in some Scaladoc being generated.

Baccata:
I wholefully agree with this

kubukoz:
Started some work on the deprecations part of this.

zetashift:
This looks like "fun" good-first-issue. If my noobie questions aren't too much hassle, I'll gladly pick this up.
https://github.com/disneystreaming/smithy4s/pull/599 should provide some tips on how to implement scaladoc in the codegen no?

I believe something like this:
- Add a case class`Documentation` which extends `Hint` to the IR
-  Then add a way to render scaladocs in `Renderer`

daddykotex:
> This looks like "fun" good-first-issue. If my noobie questions aren't too much hassle, I'll gladly pick this up. #599 should provide some tips on how to implement scaladoc in the codegen no?
> 
> I believe something like this:
> 
> * Add a case class`Documentation` which extends `Hint` to the IR
> * Then add a way to render scaladocs in `Renderer`

I think you got that right! Good luck and don't hesitate to reach out here or on Discord if you need assistance

And thanks, really

kubukoz:
yeah, I think you can draw some inspiration from #599. I left some hints for working on codegen code here, might be useful for a good feedback loop: https://github.com/disneystreaming/smithy4s/issues/496#issuecomment-1312563839

> I believe something like this:

that sounds about right

kubukoz:
Mostly closed in #731, with some new tickets spawned to further improve this. Thanks @zetashift!

PR #599, opened 2022-11-12
Render deprecation hints

kubukoz:
Made a separate commit for the actual implementation so that you can see how it affects the deprecated shapes:

[files in `modules/example` of this commit](https://github.com/disneystreaming/smithy4s/pull/599/commits/d5b5a818c078bfd29da959cc0772c130ccd55b84#diff-f6f31a8e21b138548e86cdaeadb003dd61d64575ca94b9a8ad5baa265f248556)

I think this doesn't have to go into 0.17 as it wouldn't make breaking changes.

Edit: about the above, I think this might be source breaking in a way (people might have fatal warnings on). Let's stick it into 0.17.

Solves the deprecation part of #256

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -269,7 +269,13 @@ lazy val `aws-kernel` = projectMatrix
       "aws.protocols"
     ),
     Compile / sourceGenerators := Seq(genSmithyScala(Compile).taskValue),
-    Test / envVars ++= Map("TEST_VAR" -> "hello")
+    Test / envVars ++= Map("TEST_VAR" -> "hello"),
+    scalacOptions ++= Seq(
+      "-Wconf:msg=class AwsQuery in package aws.protocols is deprecated:silent",
+      "-Wconf:msg=class RestXml in package aws.protocols is deprecated:silent",
+      "-Wconf:msg=value noErrorWrapping in class RestXml is deprecated:silent",
+      "-Wconf:msg=class Ec2Query in package aws.protocols is deprecated:silent"
+    )
   )
   .jvmPlatform(latest2ScalaVersions, jvmDimSettings)
   .jsPlatform(
--- a/build.sbt
+++ b/build.sbt
@@ -725,6 +725,8 @@ lazy val complianceTests = projectMatrix
 /**
  * Example application using the custom REST-JSON protocol provided by
  * smithy4s.
+ *
+ * All Scala code in this module is generated!
  */
 lazy val example = projectMatrix
   .in(file("modules/example"))
@@ -742,6 +744,7 @@ lazy val example = projectMatrix
     ),
     smithySpecs := Seq(
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "example.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "deprecations.smithy",
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "errors.smithy",
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "streaming.smithy",
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "operation.smithy",
--- a/build.sbt
+++ b/build.sbt
@@ -726,7 +726,7 @@ lazy val complianceTests = projectMatrix
  * Example application using the custom REST-JSON protocol provided by
  * smithy4s.
  *
- * All Scala code in this module is generated!
+ * (almost) all Scala code in this module is generated! The ones that aren't should have a note stating so.
  */
 lazy val example = projectMatrix
   .in(file("modules/example"))
@@ -769,7 +769,9 @@ lazy val example = projectMatrix
     ),
     genSmithyOutput := ((ThisBuild / baseDirectory).value / "modules" / "example" / "src"),
     genSmithyResourcesOutput := (Compile / resourceDirectory).value,
-    smithy4sSkip := List("resource")
+    smithy4sSkip := List("resource"),
+    // Ignore deprecation warnings here - it's all generated code, anyway.
+    scalacOptions += "-Wconf:cat=deprecation:silent"
   )
   .jvmPlatform(List(Scala213), jvmDimSettings)
   .settings(Smithy4sBuildPlugin.doNotPublishArtifact)
--- a/modules/codegen/src/smithy4s/codegen/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/Renderer.scala
@@ -745,13 +745,12 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
       renderHintsVal(hints),
       newline,
       values.map { case e @ EnumValue(value, intValue, _, hints) =>
+        val valueName = NameRef(e.name)
+        val valueHints = line"$Hints_(${memberHints(e.hints)})"
+
         lines(
           deprecationAnnotation(hints),
-          line"""case object ${NameRef(
-            e.name
-          )} extends $name("$value", "${e.name}", $intValue, $Hints_(${memberHints(
-            e.hints
-          )}))"""
+          line"""case object $valueName extends $name("$value", "${e.name}", $intValue, $valueHints)"""
         )
       },
       newline,
--- a/modules/codegen/src/smithy4s/codegen/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/Renderer.scala
@@ -715,7 +715,7 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
           )
         }
 
-        deprecationAnnotation(hints).appendUnlessEmpty(Line.space) +
+        deprecationAnnotation(hints).appendIf(_.nonEmpty)(Line.space) +
           line"$name: " + tpeAndDefault
     }
   }
@@ -731,12 +731,12 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
   ): Lines = lines(
     deprecationAnnotation(hints),
     block(
-      line"sealed abstract class ${name.name}(_value: String, _name: String, _intValue: Int) extends $Enumeration_.Value"
+      line"sealed abstract class ${name.name}(_value: String, _name: String, _intValue: Int, _hints: $Hints_) extends $Enumeration_.Value"
     )(
       line"override val value: String = _value",
       line"override val name: String = _name",
       line"override val intValue: Int = _intValue",
-      line"override val hints: $Hints_ = $Hints_.empty",
+      line"override val hints: $Hints_ = _hints",
       line"@inline final def widen: $name = this"
     ),
     obj(name, ext = line"$Enumeration_[$name]", w = line"${shapeTag(name)}")(
@@ -749,7 +749,9 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
           deprecationAnnotation(hints),
           line"""case object ${NameRef(
             e.name
-          )} extends $name("$value", "${e.name}", $intValue)"""
+          )} extends $name("$value", "${e.name}", $intValue, $Hints_(${memberHints(
+            e.hints
+          )}))"""
         )
       },
       newline,
--- a/modules/codegen/src/smithy4s/codegen/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/Renderer.scala
@@ -132,13 +132,32 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
       renderTypeAlias(shapeId, ta.nameRef, tpe, recursive, hints)
     case enumeration @ Enumeration(shapeId, _, values, hints) =>
       renderEnum(shapeId, enumeration.nameRef, values, hints)
-    case _ => Lines.empty
+  }
+
+  private def deprecationAnnotation(hints: List[Hint]): Line = {
+    hints
+      .collectFirst { case h: Hint.Deprecated => h }
+      .foldMap { dep =>
+        val messagePart = dep.message
+          .map(msg => line"message = ${renderStringLiteral(msg)}")
+        val versionPart =
+          dep.since.map(v => line"since = ${renderStringLiteral(v)}")
+
+        val args = List(messagePart, versionPart).flatten.intercalate(comma)
+
+        val argListOrEmpty = if (args.nonEmpty) line"($args)" else line""
+
+        line"@deprecated$argListOrEmpty"
+      }
   }
 
   def renderPackageContents: Lines = {
     val typeAliases = compilationUnit.declarations.collect {
-      case TypeAlias(_, name, _, _, _, _) =>
-        line"type $name = ${compilationUnit.namespace}.${name}.Type"
+      case TypeAlias(_, name, _, _, _, hints) =>
+        lines(
+          deprecationAnnotation(hints),
+          line"type $name = ${compilationUnit.namespace}.${name}.Type"
+        )
     }
 
     val blk =
@@ -166,6 +185,7 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
       val name = s.name
       val nameGen = NameRef(s"${name}Gen")
       lines(
+        deprecationAnnotation(s.hints),
         line"type ${NameDef(name)}[F[_]] = $FunctorAlgebra_[$nameGen, F]",
         line"val ${NameRef(name)} = $nameGen"
       )
@@ -186,12 +206,16 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
     val opTraitNameRef = opTraitName.toNameRef
 
     lines(
+      deprecationAnnotation(hints),
       block(line"trait $genName[F[_, _, _, _, _]]")(
         line"self =>",
         newline,
         ops.map { op =>
-          line"def ${op.methodName}(${op.renderArgs}) : F[${op.renderAlgParams(genNameRef.name)}]"
-
+          lines(
+            deprecationAnnotation(op.hints),
+            line"def ${op.methodName}(${op.renderArgs}) : F[${op
+              .renderAlgParams(genNameRef.name)}]"
+          )
         },
         newline,
         line"def transform : $Transformation.PartiallyApplied[$genName[F]] = new $Transformation.PartiallyApplied[$genName[F]](this)"
@@ -498,18 +522,24 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
       additionalLines: Lines = Lines.empty
   ): Lines = {
     import product._
-    if (isMixin)
-      renderProductMixin(
-        product,
-        adtParent,
-        additionalLines
-      )
-    else
-      renderProductNonMixin(
-        product,
-        adtParent,
-        additionalLines
-      )
+    val base =
+      if (isMixin)
+        renderProductMixin(
+          product,
+          adtParent,
+          additionalLines
+        )
+      else
+        renderProductNonMixin(
+          product,
+          adtParent,
+          additionalLines
+        )
+
+    lines(
+      deprecationAnnotation(product.hints),
+      base
+    )
   }
 
   private def renderGetMessage(field: Field) = field match {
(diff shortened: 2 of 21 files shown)


modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -581,23 +612,27 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
...
+          case Alt(_, realName, UnionMember.ProductCase(struct), altHints) =>
             val additionalLines = lines(
               newline,
               line"""val alt = schema.oneOf[$name]("$realName")"""
             )
             renderProduct(
-              struct,
+              // putting alt hints first should result in higher priority of these.
+              // might need deduplication (although the Hints type will take care of it, just in case)
+              struct.copy(hints = altHints ++ struct.hints),
```
kubukoz, line 635: (I'll add this as a comment later)

I'm putting the deprecation note on the struct because there's nothing else to put it on.

modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -672,7 +707,8 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
           )
         }
 
-        line"$name: " + tpeAndDefault
+        deprecationAnnotation(hints).appendUnlessEmpty(Line.space) +
```
kubukoz, line 710: class fields are fun because they're single line (so I didn't do a line break). For this reason, if the deprecation note is present we need to append a space.

kubukoz: If someone knows a better combination of words than "unless empty", I'll take it. `appendIfPresent` didn't have a nice ring to it.

kubukoz: Maybe this would be good enough? `.appendIf(_.nonEmpty)(Line.space)`

modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -699,10 +736,13 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
       newline,
       renderHintsVal(hints),
       newline,
-      values.map { case e @ EnumValue(value, intValue, _, _) =>
-        line"""case object ${NameRef(
-          e.name
-        )} extends $name("$value", "${e.name}", $intValue)"""
+      values.map { case e @ EnumValue(value, intValue, _, hints) =>
+        lines(
+          deprecationAnnotation(hints),
```
kubukoz, line 741: It was at this moment that I realized we ignore all hints on enum values. Somehow it doesn't feel right 😬 

related... you can actually deprecate a list member. Feels useless, but the model accepts it. There are also (more) valid collection member hints on (e.g. documentation), maybe we should be generating these...

modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -712,6 +752,7 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
     )
   )
 
+  // also known as newtypes
```
kubukoz, line 755: personally I found it a bit confusing that it said "type alias". I could rename the method if it feels like a better name: `renderNewtype`

modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -943,18 +985,8 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
...
-        val str = Literal(Constant(raw))
-          .toString()
-          // Replace sequences like "\\uD83D" (how Smithy specs refer to unicode characters)
-          // with unicode character escapes like "\uD83D" that can be parsed in the regex implementations on all platforms.
-          // See https://github.com/disneystreaming/smithy4s/pull/499
-          .replace("\\\\u", "\\u")
-
-        line"$str"
-      }
+      case Primitive.Uuid   => uuid => line"java.util.UUID.fromString($uuid)"
```
kubukoz, line 988: random Q: shouldn't uuids be refinements now?

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
```diff
@@ -273,7 +273,8 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
                     value.getName().asScala,
                     value.getValue,
                     index
-                  )
+                  ),
+                  hints = Nil
```
kubukoz, line 277: AFAIK 1.x enums don't allow hints on values.

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
```diff
@@ -287,18 +288,30 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
           .asScala
           .zipWithIndex
           .map { case ((name, value), index) =>
-            EnumValue(value, index, name)
+            val member = shape.getMember(name).get()
+
+            // random note: any other traits that are allowed on enum values? cuz they don't generate hints.
+            // known: @deprecated, @documentation
```
kubukoz, line 294: IMO @documentation is reason enough to support this. We'd have to change the runtime enum value model, I assume.

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
```diff
@@ -287,18 +288,30 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+        ).some
       }
 
       override def intEnumShape(shape: IntEnumShape): Option[Decl] = {
         val values = shape
           .getEnumValues()
           .asScala
           .map { case (name, value) =>
-            EnumValue(name, value, name)
+            val member = shape.getMember(name).get()
```
kubukoz, line 312: feels a bit odd to do the lookup like this, but it's not _that_ much repetition and it _should_ be safe at all times.

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
```diff
@@ -722,7 +737,10 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
 
   private def traitsToHints(traits: List[Trait]): List[Hint] = {
     val nonMetaTraits =
-      traits.filterNot(_.toShapeId().getNamespace() == "smithy4s.meta")
+      traits
+        .filterNot(_.toShapeId().getNamespace() == "smithy4s.meta")
+        // not sure about this
+        .filterNot(_.toShapeId().getNamespace() == "smithy.synthetic")
```
kubukoz, line 743: I had to do it because `unfoldTrait` was throwing on a `None.get` when dealing with `smithy.synthetic.enum` (when I started calling `hints(shape)` with a 2.x enum shape). It might be possible to do it better.

project/Smithy4sBuildPlugin.scala
```diff
@@ -107,7 +107,9 @@ object Smithy4sBuildPlugin extends AutoPlugin {
     semanticdbVersion := scalafixSemanticdb.revision,
     testFrameworks += new TestFramework("weaver.framework.CatsEffect"),
     Test / fork := virtualAxes.?.value.forall(_.contains(VirtualAxis.jvm)),
-    Test / javaOptions += s"-Duser.dir=${sys.props("user.dir")}"
+    Test / javaOptions += s"-Duser.dir=${sys.props("user.dir")}",
+    // Ignores warnings in code using the deprecated Enum trait.
+    scalacOptions += "-Wconf:msg=object Enum in package api is deprecated:silent"
```
kubukoz, line 112: FYI here's what Enum now generates as:

```scala
@deprecated(message = "The enum trait is replaced by the enum shape in Smithy 2.0", since = "2.0")
object Enum extends Newtype[List[EnumDefinition]] {
  val id: ShapeId = ShapeId("smithy.api", "enum")
  val hints : Hints = Hints(
    smithy.api.Documentation("Constrains the acceptable values of a string to a fixed set\nof constant values."),
    smithy.api.Trait(selector = Some("string :not(enum)"), structurallyExclusive = None, conflicts = None, breakingChanges = Some(List(smithy.api.TraitDiffRule(change = smithy.api.TraitChangeType.PRESENCE.widen, severity = smithy.api.TraitChangeSeverity.ERROR.widen, path = None, message = None)))),
    smithy.api.Deprecated(message = Some("The enum trait is replaced by the enum shape in Smithy 2.0"), since = Some("2.0")),
  )
  val underlyingSchema : Schema[List[EnumDefinition]] = list(EnumDefinition.schema).withId(id).addHints(hints).validated(smithy.api.Length(min = Some(1), max = None))
  implicit val schema : Schema[Enum] = recursive(bijection(underlyingSchema, asBijection))
}
```

modules/codegen/src/smithy4s/codegen/Renderer.scala
Baccata: It's such a commonly used type that it warrants being first class, in particular because of optimisation at the json parsing level

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
Baccata: Good catch. The runtime enum value model [does support hints](https://github.com/disneystreaming/smithy4s/blob/main/modules/core/src/smithy4s/schema/EnumValue.scala#L25) 😉

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
Baccata: Mmm I'm not sure I understand. Can you share an example model that was leading it to throw ?

modules/codegen/src/smithy4s/codegen/Renderer.scala
Baccata: `appendIf` looks neat

modules/codegen/src/smithy4s/codegen/Renderer.scala
Baccata: > maybe we should be generating these

yeah. Let's do that in a later PR

modules/codegen/src/smithy4s/codegen/Renderer.scala
Baccata: Yeah happy to change the name

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
Baccata: That is correct

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
kubukoz: Any 2.x enum with any hints. I can drop something later

kubukoz: @Baccata question mark for when you have a sec

kubukoz: Even this (already on the 0.17 branch) causes it:

```
enum SwitchState {
  ON
  OFF
}
```

Try [this commit](https://github.com/disneystreaming/smithy4s/commit/e2c090e741aa854e9381c3c7ad86ac61d4fd6445) and `example/smithy4sCodegen`.

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
kubukoz: `EnumValue` does, true. I thought it didn't because the generated enum classes don't:

```scala
sealed abstract class FaceCard(_value: String, _name: String, _intValue: Int)

case object JACK extends FaceCard("JACK", "JACK", 1)
```

so we'll just have to change the codegen of these. If it doesn't turn out to be a can of worms I'll do that here

kubukoz:
I'll get back to this after the weekend, most likely

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
Baccata: Okay, so turns out traits from the `smithy.synthetic` package do not have associated shapes in the model. I think the filter is okay, just change the comment to explain it.

modules/codegen/src/smithy4s/codegen/Renderer.scala
kubukoz: #600

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
kubukoz: f274458cb44de0b126e9deeec60e5aba8cbff1b9

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
kubukoz: f8331897f3495a92440f6d7edd3d53cea4483788 + adding it to Dynamic e18ba3a8f944a70d4d9ecbe1030b26cfe8134066

modules/codegen/src/smithy4s/codegen/Renderer.scala
kubukoz: 0bc490d7248790ba975c11dc4b3ea4700bf807f9

modules/codegen/src/smithy4s/codegen/Renderer.scala
kubukoz: f8331897f3495a92440f6d7edd3d53cea4483788

build.sbt
```diff
@@ -725,6 +725,8 @@ lazy val complianceTests = projectMatrix
 /**
  * Example application using the custom REST-JSON protocol provided by
  * smithy4s.
+ *
+ * (almost) all Scala code in this module is generated! The ones that aren't should have a note stating so.
```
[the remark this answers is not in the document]
kubukoz: "nice to have" + "good enough" == "i'm not gonna"

kubukoz:
@baccata I think this is ready to go from my side

modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -132,13 +132,32 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
       renderTypeAlias(shapeId, ta.nameRef, tpe, recursive, hints)
     case enumeration @ Enumeration(shapeId, _, values, hints) =>
       renderEnum(shapeId, enumeration.nameRef, values, hints)
-    case _ => Lines.empty
+  }
+
+  private def deprecationAnnotation(hints: List[Hint]): Line = {
```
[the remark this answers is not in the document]
daddykotex: the only thing I'd change here is have an extra map, before the that's already there where I call `renderStringLiteral` before building the `line` to avoid `${renderStringLiteral(msg)}`

modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -691,24 +729,30 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
...
       renderHintsVal(hints),
       newline,
-      values.map { case e @ EnumValue(value, intValue, _, _) =>
-        line"""case object ${NameRef(
-          e.name
-        )} extends $name("$value", "${e.name}", $intValue)"""
+      values.map { case e @ EnumValue(value, intValue, _, hints) =>
+        lines(
+          deprecationAnnotation(hints),
+          line"""case object ${NameRef(
```
daddykotex, line 750: this comment is similar to the one above: in string interpolation, I try to avoid function (especially when they involve arguments). they usually get a newline by the formatter and it makes it hard to read

modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -988,4 +1023,15 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
       case _ => _ => line"null"
     }
 
+  private def renderStringLiteral(raw: String): Line = {
+    import scala.reflect.runtime.universe._
```
daddykotex, line 1027: what is the import used for?

modules/codegen/src/smithy4s/codegen/Renderer.scala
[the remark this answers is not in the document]
kubukoz: tbh I actually like that `renderStringLiteral` is being interpolated (as long as it doesn't break the line)

modules/example/src/smithy4s/example/DeprecatedStructure.scala
```diff
@@ -0,0 +1,26 @@
+package smithy4s.example
+
+import smithy4s.Schema
+import smithy4s.Hints
+import smithy4s.schema.Schema.string
+import smithy4s.ShapeId
+import smithy4s.schema.Schema.struct
+import smithy4s.ShapeTag
+
+@deprecated(message = "A compelling reason", since = "0.0.1")
```
daddykotex, line 10: hahaha

daddykotex (approved):
great job, it's more work that I had anticipated
kudos on the work on dynamic, this PR serves as a great reference if for changes that would need to apply there as well

modules/codegen/src/smithy4s/codegen/Renderer.scala
kubukoz: `Literal` and `Constant`

modules/codegen/src/smithy4s/codegen/Renderer.scala
kubukoz: I can get behind that

modules/example/src/smithy4s/example/Letters.scala
```diff
@@ -7,21 +7,21 @@ import smithy4s.ShapeId
...
 }
 object Letters extends Enumeration[Letters] with ShapeTag.Companion[Letters] {
   val id: ShapeId = ShapeId("smithy4s.example", "Letters")
 
   val hints : Hints = Hints.empty
 
-  case object A extends Letters("a", "A", 0)
-  case object B extends Letters("b", "B", 1)
-  case object C extends Letters("c", "C", 2)
+  case object A extends Letters("a", "A", 0, Hints(smithy.api.EnumValue(smithy4s.Document.fromString("a"))))
```
Baccata, line 22: I think we could filter out the `smithy.api.EnumValue` trait from hints rendering considering it's captured as static information in the enumeration.

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
```diff
@@ -287,18 +288,30 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
           .asScala
           .zipWithIndex
           .map { case ((name, value), index) =>
-            EnumValue(value, index, name)
+            val member = shape.getMember(name).get()
+
+            // random note: any other traits that are allowed on enum values? cuz they don't generate hints.
```
Baccata, line 293: is this comment still up to date ?

kubukoz: no I don't think it is

modules/example/src/smithy4s/example/Letters.scala
kubukoz: yeah, I was wondering if you'd mention that

kubukoz: that'd also have to be done for dynamic. I'll get to it later

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
```diff
@@ -742,6 +740,8 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
         // traits from the synthetic namespace, e.g. smithy.synthetic.enum
         // don't have shapes in the model - so we can't generate hints for them.
         .filterNot(_.toShapeId().getNamespace() == "smithy.synthetic")
+        // enumValue can be derived from enum schemas anyway, so we're removing it from hints
+        .filterNot(_.toShapeId().toString() == "smithy.api#enumValue")
```
Baccata, line 744: thy shall not use `.toString` when there is an alternative (in this case, `ShapeId.fromParts("smithy.api", "enumValue")` or probably even `EnumValueTrait.ID` or something

kubukoz: now you're just nitpicking :P

Baccata (approved):

modules/codegen/src/smithy4s/codegen/SmithyToIR.scala
Baccata: Nah, just .toString trauma 😄

kubukoz: :heavy-nodding:

PR #607, opened 2022-11-16
Add a dependency tracking mechanism

Baccata:
This facilitates the tracking of dependencies without imposing additional burden on the user to know what dependencies were used to generate code in upstream libraries/projects. 

Essentially, what it does is add a line in the manifest of project/modules that have the Smithy4s plugin enabled, tracking dependencies locally declared with `% Smithys`. A bit of logic is added so that downstream projects that also have `Smithy4s` enabled automatically resolve these dependencies before running code-generation. 

Additionally, this PR changes the build plugins so that libraryDependencies are taken into account for codegen :
    
Now, by default, any library declared as a "normal" dependency, by means of `ivyDeps` in mill or `libraryDependencies` in sbt are used during the code-generation process. This makes Smithy4s more consistent with itself, as the same behaviour is applied to internal dependencies. It is also argueably more intuitive to do it this way.

#### Todo 

- [x] Add the same mechanism in the Mill integration 

- [x] Dogfooding : make sure that the artifacts published by this build (which do not use the Smithy4sCodegenPlugin, cause it's also published by this build) contain the correct line in the manifest, so that users don't have to add explicitly to their builds dependencies to protocol-test-traits, aws-traits, etc ...


modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -134,9 +199,78 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
...
+      }
+      getJars(smithy4sDependencies)
     }
 
+  private lazy val simple = raw"([^:]*):([^:]*):([^:]*)".r
+  private lazy val cross = raw"([^:]*)::([^:]*):([^:]*)".r
+  private def extract(jarFile: java.io.File): Seq[ModuleID] = {
+    val jar = new JarFile(jarFile)
+    Option(
+      jar.getManifest().getMainAttributes().getValue("smithy4sDependencies")
```
kubukoz, line 262: let's extract `smithy4sDependencies` to a constant

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -134,9 +199,78 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
...
+            streams.value.log
+          )
+          .map(_.allFiles)
+          .fold(uw => throw uw.resolveException, identity)
+      }
+      getJars(smithy4sDependencies)
     }
 
+  private lazy val simple = raw"([^:]*):([^:]*):([^:]*)".r
+  private lazy val cross = raw"([^:]*)::([^:]*):([^:]*)".r
```
kubukoz, line 258: just spitballing, could we use coursier's api to parse the dependency strings?

kubukoz:
good stuff

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
Baccata: something something draft

modules/docs/markdown/01-overview/05-sharing-specs.md
```diff
@@ -151,17 +151,61 @@ libraryDependencies += "organisation" % "artifact" % "version" % "smithy4s,compi
...
+    libraryDependencies ++= Seq(
+      "software.amazon.smithy" % "smithy-aws-iam-traits" % "1.14.1" % Smithy4s
+    )
+  )
+```
+
+and publish this project to an artifact repository, the Jar manifest will contain the following line :
+
+```
+smithy4sDependencies: software.amazon.smithy:smithy-aws-iam-traits:1.14.1
```
daddykotex, line 187: may be worth noting that multiple dependencies are comma separated

modules/mill-codegen-plugin/test/src/smithy4s/codegen/mill/Smithy4sModuleSpec.scala
```diff
@@ -207,14 +209,101 @@ class Smithy4sModuleSpec extends munit.FunSuite {
     taskWorks(bar.smithy4sCodegen, barEv)
   }
 
+  test("multi-module staged codegen works".only) {
```
daddykotex, line 212: just adding a note to make we don't forget it when we remove the draft status

project/Dependencies.scala
```diff
@@ -0,0 +1,128 @@
+import sbt._
+import org.portablescala.sbtplatformdeps.PlatformDepsPlugin.autoImport._
+import Smithy4sBuildPlugin.autoImport.isCE3
+
+object Dependencies {
```
Baccata, line 5: moved this from build.sbt to be able to re-use in Smithy4sBuildPlugin

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -174,7 +299,7 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
       output = os.Path(outputPath),
       resourceOutput = os.Path(resourceOutputPath),
       skip = skipSet,
-      discoverModels = true, // we need protocol here
+      discoverModels = false, // we need protocol here
```
lewisjkl, line 302:
```suggestion
      discoverModels = false,
```

I guess that comment can be removed now

modules/docs/markdown/01-overview/05-sharing-specs.md
```diff
@@ -151,17 +151,61 @@ libraryDependencies += "organisation" % "artifact" % "version" % "smithy4s,compi
...
+
+Using this artifact in a downstream project, for instance with :
+
+```scala
+lazy val downstream = (project in file("foo"))
+  .enablePlugins(Smithy4sCodegenPlugin)
+  .settings(
+    libraryDependencies ++= Seq(
+      // compile/runtime dependency that contains Smithy4s-generated code but doesn't contain smithy files
+      "foobar" % "upstream" % "0.0.1"
```
lewisjkl, line 198:
```suggestion
      "foobar" % "upstream" % "0.0.1" % Smithy4s
```

Not sure, but seems like the smithy4s would be needed here?

lewisjkl (approved):

modules/docs/markdown/01-overview/05-sharing-specs.md
Baccata: Nope, that's the whole point of this PR. The library itself doesn't contain smithy code, but was generated using smithy4s, and therefore has a reference to the artifacts that it pulled for its own code-generation.

kubukoz: if it contains generated code it should be considered a Scala dependency, right? so probably `%%` to account for the suffix. Seeing this I'm wondering if the whole thing will work on non-JVM platforms...

kubukoz:
something is not yes (this test wasn't running earlier because there was an `.only` in the spec):

image

modules/docs/markdown/01-overview/05-sharing-specs.md
Baccata: non-JVM Scala libraries are still packaged in jars, and still carry manifests

Baccata: I do wonder if we shouldn't just use the whole (upstream) classpath for code-generation. I think this may be more intuitive and closer to users' mental models. WDYT ?

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
kubukoz: dunno, if it was a structured parsing result we could probably do that - if there's no way then the answer to my question is "no" ;)

Baccata: Let's postpone this : right now the problem is more about "can we you render a dependency as a coursier-parsable string in both SBT and mill ?" Because I couldn't find a way to do that easily, I have the control over the format and parsing with coursier is not necessary as we're falling in the simplest possible cases (normal java dep, normal scala dep). 

If it proves to be limitative we'll revise.

PR #613, opened 2022-11-18
Implement HttpProtocolCompliance w/o IO

daddykotex:
This branch targets `refine-compliance-tests-dx` and I assumed it was based on `series/0.17` so I merged that in. 
The only relevant commit in that PR is: https://github.com/disneystreaming/smithy4s/commit/ee22af5a51e1744a19a172f6e923b31a5f2173f9

### DIFF
--- a/modules/compliance-tests/src-ce2/smithy4s/compliancetests/Compat.scala
+++ b/modules/compliance-tests/src-ce2/smithy4s/compliancetests/Compat.scala
@@ -16,16 +16,19 @@
 
 package smithy4s.compliancetests
 
-import cats.effect.IO
+import cats.effect._
 import cats.effect.concurrent.Deferred
-import cats.effect.ContextShift
-import cats.effect.Timer
-import cats.effect.Concurrent
+import cats.effect.implicits._
+import cats.Monad
+import scala.concurrent.duration.FiniteDuration
 
 private[compliancetests] class CompatEffect[F[_]](implicit
-    val F: Concurrent[F]
-) {
+    val _Concurrent: Concurrent[F],
+    val _Time: Timer[F],
+    val _Monad: Monad[F]
+) extends CompatUtils[F] {
   def deferred[A]: F[Deferred[F, A]] = Deferred[F, A]
+  def timeout[A](f: F[A], delay: FiniteDuration): F[A] = f.timeout(delay)
 
   val utf8Encode: fs2.Pipe[F, String, Byte] = fs2.text.utf8Encode[F]
   val utf8Decode: fs2.Pipe[F, Byte, String] = fs2.text.utf8Decode[F]
--- a/modules/compliance-tests/src-ce2/smithy4s/compliancetests/Compat.scala
+++ b/modules/compliance-tests/src-ce2/smithy4s/compliancetests/Compat.scala
@@ -18,18 +18,19 @@ package smithy4s.compliancetests
 
 import cats.effect._
 import cats.effect.concurrent.Deferred
-import cats.effect.implicits._
-import cats.Monad
 import scala.concurrent.duration.FiniteDuration
 
 private[compliancetests] class CompatEffect[F[_]](implicit
     val _Concurrent: Concurrent[F],
-    val _Time: Timer[F],
-    val _Monad: Monad[F]
+    val _Timer: Timer[F]
 ) extends CompatUtils[F] {
+  // CE2 Deferred is in a cats.effect.concurrent.Deferred
   def deferred[A]: F[Deferred[F, A]] = Deferred[F, A]
-  def timeout[A](f: F[A], delay: FiniteDuration): F[A] = f.timeout(delay)
+  // CE2 timeout is on Concurrent
+  def timeout[A](f: F[A], delay: FiniteDuration): F[A] =
+    Concurrent.timeout[F, A](f, delay)
 
+  // utf8 encode/decode under fs2.text
   val utf8Encode: fs2.Pipe[F, String, Byte] = fs2.text.utf8Encode[F]
   val utf8Decode: fs2.Pipe[F, Byte, String] = fs2.text.utf8Decode[F]
 }
--- a/modules/compliance-tests/src-ce3/smithy4s/compliancetests/Compat.scala
+++ b/modules/compliance-tests/src-ce3/smithy4s/compliancetests/Compat.scala
@@ -18,19 +18,20 @@ package smithy4s.compliancetests
 
 import cats.effect.Async
 import cats.effect.Deferred
-import cats.Monad
 import com.comcast.ip4s.Host
 import com.comcast.ip4s.Port
 import scala.concurrent.duration.FiniteDuration
 
 private[compliancetests] class CompatEffect[F[_]](implicit
-    val _Async: Async[F],
-    val _Monad: Monad[F]
+    val _Async: Async[F]
 ) extends CompatUtils[F] {
+  // CE3 Deferred is in a cats.effect.Deferred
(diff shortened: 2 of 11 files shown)


daddykotex:
do you mind if I push that merge commit to `refine-compilaance-test-dx` @Baccata ?
that would remove the noise here

.github/workflows/ci.yml
```diff
@@ -154,7 +154,7 @@ jobs:
       - name: Setup Java and Scala
         uses: olafurpg/setup-scala@v13
```
kubukoz, line 155: I thought we weren't using setup-scala anymore

kubukoz:
@daddykotex go for it ;) the target branch seems to be using the new service encoding so it's 0.17

image

.github/workflows/ci.yml
daddykotex: I'm not sure how you stumbled upon this. It was not even in the diff, was that on a specific commit?

modules/compliance-tests/src-ce3/smithy4s/compliancetests/Compat.scala
```diff
@@ -16,13 +16,20 @@
...
 import com.comcast.ip4s.Host
 import com.comcast.ip4s.Port
-import cats.effect.Async
+import scala.concurrent.duration.FiniteDuration
 
-private[compliancetests] class CompatEffect[F[_]](implicit val F: Async[F]) {
+private[compliancetests] class CompatEffect[F[_]](implicit
+    val _Async: Async[F],
+    val _Monad: Monad[F]
+) extends CompatUtils[F] {
```
kubukoz, line 29: I would've just used context bounds tbh

modules/compliance-tests/src/smithy4s/compliancetests/CompatUtils.scala
```diff
@@ -0,0 +1,38 @@
...
+
+private object CompatUtils {
+  type ME[F[_]] = MonadError[F, Throwable]
+}
+
+abstract class CompatUtils[F[_]: CompatUtils.ME] {
+  def raiseError[A](err: Throwable): F[A] =
+    MonadError[F, Throwable].raiseError(err)
+
+  implicit def monoid[A: Monoid]: Monoid[F[A]] = new Monoid[F[A]] {
```
kubukoz, line 31: I believe this is `Applicative.monoid[F, A]`.

modules/compliance-tests/src-ce3/smithy4s/compliancetests/Compat.scala
```diff
@@ -16,13 +16,20 @@
...
+import cats.Monad
 import com.comcast.ip4s.Host
 import com.comcast.ip4s.Port
-import cats.effect.Async
+import scala.concurrent.duration.FiniteDuration
 
-private[compliancetests] class CompatEffect[F[_]](implicit val F: Async[F]) {
+private[compliancetests] class CompatEffect[F[_]](implicit
+    val _Async: Async[F],
+    val _Monad: Monad[F]
```
kubukoz, line 28: Async implies Monad so you don't need to have both

modules/compliance-tests/src-ce2/smithy4s/compliancetests/Compat.scala
```diff
@@ -16,16 +16,19 @@
...
+import cats.effect.implicits._
+import cats.Monad
+import scala.concurrent.duration.FiniteDuration
 
 private[compliancetests] class CompatEffect[F[_]](implicit
-    val F: Concurrent[F]
-) {
+    val _Concurrent: Concurrent[F],
+    val _Time: Timer[F],
+    val _Monad: Monad[F]
```
kubukoz, line 28: Concurrent implies Monad so you can drop Monad from this

modules/compliance-tests/src/smithy4s/compliancetests/CompatUtils.scala
```diff
@@ -0,0 +1,38 @@
...
+import cats.MonadError
+import cats.Monad
+
+private object CompatUtils {
+  type ME[F[_]] = MonadError[F, Throwable]
+}
+
+abstract class CompatUtils[F[_]: CompatUtils.ME] {
+  def raiseError[A](err: Throwable): F[A] =
+    MonadError[F, Throwable].raiseError(err)
```
kubukoz, line 29: not sure why we need this (i.e. why not `MonadThrow[F].raiseError`, or `.raiseError[F, A]` - I think both exist in cats-core)

modules/compliance-tests/src/smithy4s/compliancetests/internals/ClientHttpComplianceTestCase.scala
```diff
@@ -91,44 +91,45 @@ private[compliancetests] class ClientHttpComplianceTestCase[
...
-            .of[IO] { case req =>
+            .of[F] { case req =>
               req.body.compile.toVector
                 .map(fs2.Stream.emits(_))
                 .map(req.withBodyStream(_))
                 .flatMap(requestDeferred.complete(_))
-                .as(Response[IO]())
+                .as(Response[F]())
             }
             .orNotFound
```
kubukoz, line 128: `HttpRoutes.of{...}.orNotFound` is `HttpApp.apply`, or even just plain `Kleisli`

.github/workflows/ci.yml
kubukoz: no I think it was on the one big diff before you updated the base branch

modules/compliance-tests/src/smithy4s/compliancetests/internals/ClientHttpComplianceTestCase.scala
Baccata: That's not exactly accurate, but the point remains that you should indeed use `HttpApp.apply` in this instance

kubukoz: > That's not exactly accurate

Care to elaborate? I meant in this particular instance (the function passed to `HttpRoutes.of` is total).

Baccata: >  I meant in this particular instance

That addition to your initial statement makes it accurate 😄

modules/compliance-tests/src/smithy4s/compliancetests/CompatUtils.scala
daddykotex: I did not event know `MonadThrow` was part of cats, wow

modules/compliance-tests/src-ce2/smithy4s/compliancetests/Compat.scala
daddykotex: yeah, I did not know if it was expected we rely on type class hierarchy like that or if it was better to be explicit
I suppose I already rely on the hierarchy w/o knowing it so...

modules/compliance-tests/src-ce3/smithy4s/compliancetests/Compat.scala
daddykotex: I don't think I can't, at least not in the way the code is written right now
because I declare them as `val` here, I can import these implicits in the rest of the code

for example in `ClientHttpComplianceTestCase`, we receive an instance of `ce: CompatEffect` and I can `import ce._` which puts the instances in scope of that we can call things like `request.bodyText` and `thing.pure[F]`

I've gotta say that between my many attempts and rewrite to get the compat layer to work, there is most likely a simpler/better way to achieve the same goal

Baccata (approved):

PR #614, opened 2022-11-18
Endpoint-Specific Middleware

lewisjkl:
Adds the ability to provide a middleware that can provide HttpApp[F] => HttpApp[F] conversions based on the data contained in the service and endpoint (such as Hints). This will allow more idiomatic implementations of things such as authentication/authorization and other mechanisms that rely on inspecting traits on a per-endpoint basis.

More docs to come as part of the PR.

### DIFF
--- a/modules/guides/src/smithy4s/guides/Auth.scala
+++ b/modules/guides/src/smithy4s/guides/Auth.scala
@@ -26,7 +26,7 @@ import com.comcast.ip4s._
 import smithy4s.http4s.SimpleRestJsonBuilder
 import smithy4s.Hints
 import org.http4s.headers.Authorization
-import smithy4s.http4s.EndpointSpecificMiddleware
+import smithy4s.http4s.ServerEndpointMiddleware
 
 final case class ApiToken(value: String)
 
@@ -91,8 +91,8 @@ object AuthMiddleware {
 
   def apply(
       authChecker: AuthChecker
-  ): EndpointSpecificMiddleware[IO] =
-    new EndpointSpecificMiddleware.Simple[IO] {
+  ): ServerEndpointMiddleware[IO] =
+    new ServerEndpointMiddleware.Simple[IO] {
       private val mid: HttpApp[IO] => HttpApp[IO] = middleware(authChecker)
       def prepareWithHints(
           serviceHints: Hints,
--- a/modules/guides/src/smithy4s/guides/Auth.scala
+++ b/modules/guides/src/smithy4s/guides/Auth.scala
@@ -31,6 +31,7 @@ import smithy4s.kinds.{FunctorAlgebra, PolyFunction5, Kind1}
 import smithy4s.Hints
 import org.http4s.headers.Authorization
 import cats.data.OptionT
+import smithy4s.http4s.EndpointSpecificMiddleware
 
 final case class APIKey(value: String)
 
@@ -96,18 +97,16 @@ object AuthMiddleware {
 
   def smithy4sMiddleware(
       authChecker: AuthChecker
-  ): smithy4s.http4s.EndpointSpecificMiddleware[HelloWorldAuthServiceGen, IO] =
-    new smithy4s.http4s.EndpointSpecificMiddleware[
-      HelloWorldAuthServiceGen,
-      IO
-    ] {
-      def prepare(service: smithy4s.Service[HelloWorldAuthServiceGen])(
-          endpoint: smithy4s.Endpoint[service.Operation, _, _, _, _, _]
+  ): EndpointSpecificMiddleware.Simple[HelloWorldAuthServiceGen, IO] =
+    new EndpointSpecificMiddleware.Simple[HelloWorldAuthServiceGen, IO] {
+      def prepareUsingHints(
+          serviceHints: Hints,
+          endpointHints: Hints
       ): HttpApp[IO] => HttpApp[IO] = {
-        service.hints.get[smithy.api.HttpBearerAuth] match {
+        serviceHints.get[smithy.api.HttpBearerAuth] match {
           case Some(_) =>
             val mid = middleware(authChecker)
-            endpoint.hints.get[smithy.api.Auth] match {
+            endpointHints.get[smithy.api.Auth] match {
               case Some(auths) if auths.value.isEmpty => identity
               case _                                  => mid
             }
--- a/modules/guides/src/smithy4s/guides/Auth.scala
+++ b/modules/guides/src/smithy4s/guides/Auth.scala
@@ -94,7 +94,7 @@ object AuthMiddleware {
   ): EndpointSpecificMiddleware[IO] =
     new EndpointSpecificMiddleware.Simple[IO] {
       private val mid: HttpApp[IO] => HttpApp[IO] = middleware(authChecker)
-      def prepareUsingHints(
+      def prepareWithHints(
           serviceHints: Hints,
           endpointHints: Hints
       ): HttpApp[IO] => HttpApp[IO] = {
--- a/modules/guides/src/smithy4s/guides/Auth.scala
+++ b/modules/guides/src/smithy4s/guides/Auth.scala
@@ -24,16 +24,11 @@ import org.http4s.ember.server._
 import org.http4s._
 import com.comcast.ip4s._
 import smithy4s.http4s.SimpleRestJsonBuilder
-import org.http4s.server.Middleware
-import org.typelevel.ci.CIString
-import scala.concurrent.duration.Duration
-import smithy4s.kinds.{FunctorAlgebra, PolyFunction5, Kind1}
 import smithy4s.Hints
 import org.http4s.headers.Authorization
-import cats.data.OptionT
 import smithy4s.http4s.EndpointSpecificMiddleware
 
-final case class APIKey(value: String)
+final case class ApiToken(value: String)
 
 object HelloWorldAuthImpl extends HelloWorldAuthService[IO] {
   def sayWorld(): IO[World] = World().pure[IO]
@@ -41,11 +36,11 @@ object HelloWorldAuthImpl extends HelloWorldAuthService[IO] {
 }
 
 trait AuthChecker {
-  def isAuthorized(token: APIKey): IO[Boolean]
+  def isAuthorized(token: ApiToken): IO[Boolean]
 }
 
 object AuthChecker extends AuthChecker {
-  def isAuthorized(token: APIKey): IO[Boolean] = {
+  def isAuthorized(token: ApiToken): IO[Boolean] = {
     IO.pure(
       token.value.nonEmpty
     ) // put your logic here, currently just makes sure the token is not empty
@@ -58,7 +53,7 @@ object AuthExampleRoutes {
   private val helloRoutes: Resource[IO, HttpRoutes[IO]] =
     SimpleRestJsonBuilder
       .routes(HelloWorldAuthImpl)
-      .middleware(AuthMiddleware.smithy4sMiddleware(AuthChecker))
+      .middleware(AuthMiddleware(AuthChecker))
       .resource
 
   val all: Resource[IO, HttpRoutes[IO]] =
@@ -79,7 +74,7 @@ object AuthMiddleware {
               ) =>
             value
         }
-        .map { APIKey.apply }
+        .map { ApiToken.apply }
 
       val isAuthorized = maybeKey
         .map { key =>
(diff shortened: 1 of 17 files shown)


lewisjkl:
@Baccata if you are happy with the contents of this PR, then I will move forward with adding tests in the project and docs for the microsite. Want to make sure the interface is solid before I move on to that though. Let me know!

modules/http4s/src-ce3/Compat.scala
```diff
@@ -18,7 +18,7 @@ package smithy4s.http4s
 
 private[smithy4s] object Compat {
   trait Package {
-    private[smithy4s] type EffectCompat[F[_]] = cats.effect.Concurrent[F]
-    private[smithy4s] val EffectCompat = cats.effect.Concurrent
+    private[smithy4s] type EffectCompat[F[_]] = cats.effect.Async[F]
+    private[smithy4s] val EffectCompat = cats.effect.Async
```
lewisjkl, line 22: Apparently needed for http4s client.fromHttpApp.

lewisjkl:
Ah yeah... `cats.effect.kernel.Unique` wasn't a thing until CE3 right? I am going to need to investigate how we may be able to implement this with the older version of http4s + CE2. Maybe just some compat classes that handle the differences.. hopefully.

modules/http4s/src/smithy4s/http4s/SmithyHttp4sRouter.scala
```diff
@@ -23,23 +23,38 @@ import cats.implicits._
...
     service: smithy4s.Service.Aux[Alg, Op],
     impl: FunctorInterpreter[Op, F],
     errorTransformation: PartialFunction[Throwable, F[Throwable]],
-    entityCompiler: EntityCompiler[F]
+    entityCompiler: EntityCompiler[F],
+    middleware: EndpointSpecificMiddleware[F]
 )(implicit effect: EffectCompat[F]) {
 
+  private val pathParamsKey = {
+    implicit val cheatUnique: Unique[Id] = new Unique[Id] {
```
kubukoz, line 41: lol

modules/http4s/src/smithy4s/http4s/EndpointSpecificMiddleware.scala
```diff
@@ -0,0 +1,38 @@
...
+  def prepare[Alg[_[_, _, _, _, _]]](service: Service[Alg])(
+      endpoint: Endpoint[service.Operation, _, _, _, _, _]
+  ): HttpApp[F] => HttpApp[F]
+}
+// format: on
+
+object EndpointSpecificMiddleware {
+
+  trait Simple[F[_]] extends EndpointSpecificMiddleware[F] {
+    def prepareUsingHints(
```
kubukoz, line 17:
```suggestion
    def prepareWithHints(
```

modules/http4s/src/smithy4s/http4s/EndpointSpecificMiddleware.scala
```diff
@@ -0,0 +1,38 @@
+package smithy4s
+package http4s
+
+import org.http4s.HttpApp
+
+// format: off
+trait EndpointSpecificMiddleware[F[_]] {
+  def prepare[Alg[_[_, _, _, _, _]]](service: Service[Alg])(
+      endpoint: Endpoint[service.Operation, _, _, _, _, _]
```
kubukoz, line 9: I like this so much.

also, is there an argument for why Endpoint couldn't have IEOSESO as type members?

kubukoz (changes requested):
This is really cool! I have a couple nitpicks as always.

modules/http4s/src/smithy4s/http4s/SmithyHttp4sRouter.scala
```diff
@@ -23,23 +23,38 @@ import cats.implicits._
...
+      def applicative: Applicative[Id] = Applicative[Id]
+
+      def unique: Id[Unique.Token] = new Unique.Token()
+
+    }
+    Key.newKey[Id, smithy4s.http.PathParams]
```
kubukoz, line 47: http4s creates keys this way

```scala
val ConnectionInfo: Key[Connection] = Key.newKey[SyncIO, Connection].unsafeRunSync()
```

it's `SyncIO`, so it's fine to `.unsafeRunSync()` on any platform. I'd say let's do the same

modules/http4s/src/smithy4s/http4s/EndpointSpecificMiddleware.scala
lewisjkl: Hmm I'm not sure it'd be possible because the same middleware implementation will be used across multiple endpoints in a service so the IOESISO will change depending on the endpoint

kubukoz: no I mean they'd be type members _in_ `Endpoint`. I think it might just be an inconvenience because it makes `Transformation[Endpoint[Op, ...]]` impossible to express directly

lewisjkl: Oh I see, sorry I misread your comment. Yeah I'm not sure about that one tbh

modules/http4s/src/smithy4s/http4s/SmithyHttp4sRouter.scala
Baccata: Meh, that's an abuse imho. My point of view is that http4s should not depend on cats-effect-core, but cats-effect-std (or even -kernel). I wish there was just an `unsafeCreate` on `Key`.

modules/http4s/src/smithy4s/http4s/EndpointSpecificMiddleware.scala
Baccata: It may be achievable, but it should be its own PR. Want to have a go at it @kubukoz ?

modules/http4s/src/smithy4s/http4s/EndpointSpecificMiddleware.scala
Baccata: Why not

modules/http4s/src/smithy4s/http4s/SmithyHttp4sRouter.scala
kubukoz: yeah, it should not, but it does ;)

(I think it's because fs2 has an explicit SyncIO compiler)

> I wish there was just an `unsafeCreate` on Key.

you know what to do...

modules/http4s/src/smithy4s/http4s/EndpointSpecificMiddleware.scala
kubukoz: perhaps

modules/http4s/src/smithy4s/http4s/SmithyHttp4sRouter.scala
Baccata: > you know what to do... 

I do : tell @lewisjkl to ignore your PR comment and keep the current implementation that is totally fine  :trollface:

kubukoz: Just know that the `new Unique` gives me the heebiejeebies :P

Baccata: Well, it may well be that we need to use `SyncIO` in CE2 :/

modules/guides/src/smithy4s/guides/Auth.scala
```diff
@@ -0,0 +1,131 @@
...
+        }
+        .map { APIKey.apply }
+
+      val isAuthorized = maybeKey
+        .map { key =>
+          authChecker.isAuthorized(key)
+        }
+        .getOrElse(IO.pure(false))
+
+      isAuthorized.flatMap {
```
daddykotex, line 90: I'm not sure which one I prefer, but there is `isAuthorized.ifM(inputApp(request), IO.raiseError(new NotAuthorizedError("Not authorized!")))` that would work here

kubukoz: I would just recommend that if you use `ifM`, you always use named arguments

kubukoz: alternatively you can do  `isAuthorized.ensure(new NotAuthorizedError(...))(identity) *> inputApp(request)` or something like that

lewisjkl: Yeah I was trying to stay lighter on the cats syntax since this is a walkthrough targeted at maybe more beginner users, but I guess maybe being more idiomatic is more important.

kubukoz: at least you didn't do: `isAuthorized.flatMap(if(_) ... else ...)` :)

lewisjkl:
Alright, I addressed comments, added tests, and added a guide to the docs. Give me your feedback on what needs to improve/change/etc so we can get this merged 🚀

modules/http4s/test/src/smithy4s/http4s/EndpointSpecificMiddlewareSpec.scala
```diff
@@ -0,0 +1,160 @@
...
+
+  private final class TestMiddleware(shouldFail: Boolean)
+      extends EndpointSpecificMiddleware.Simple[IO] {
+    def prepareUsingHints(
+        serviceHints: Hints,
+        endpointHints: Hints
+    ): HttpApp[IO] => HttpApp[IO] = { inputApp =>
+      HttpApp[IO] { request =>
+        val hasTag: (Hints, String) => Boolean = (hints, tagName) =>
+          hints.get[smithy.api.Tags].exists(_.value.contains(tagName))
```
kubukoz, line 142: first time I'm seeing anyone use this hint for anything in Scala

kubukoz (approved):

kubukoz:
amazing stuff :)

modules/http4s/test/src/smithy4s/http4s/EndpointSpecificMiddlewareSpec.scala
lewisjkl: Hah yeah it was just a really easy way I could think to test that the hints were being sent through properly. I don't have any real use cases for it atm. Though I guess it could come in handy for some arbitrary middleware thing 🤷

modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
```diff
@@ -79,12 +82,16 @@ private[http4s] class SmithyHttp4sClientEndpointImpl[F[_], Op[_, _, _, _, _], I,
...
   endpoint: Endpoint[Op, I, E, O, SI, SO],
   httpEndpoint: HttpEndpoint[I],
-  compilerContext: CompilerContext[F]
+  compilerContext: CompilerContext[F],
+  middleware: HttpApp[F] => HttpApp[F]
 )(implicit effect: EffectCompat[F]) extends SmithyHttp4sClientEndpoint[F, Op, I, E, O, SI, SO] {
 // format: on
 
+  private val transformedClient: Client[F] =
+    Client.fromHttpApp[F](middleware(client.toHttpApp))
```
Baccata, line 91: Right, so unfortunately, I don't think we should rely on `fromHttpApp` in non-test code, because it's a [non-trivial transformation](https://github.com/http4s/http4s/blob/series/0.23/client/shared/src/main/scala/org/http4s/client/Client.scala#L278) and I'm somewhat scared that we'd be giving a footgun to the users, as it's easy to mess up the request lifecycle in client-side middleware. 

Essentially the main problem is that the HttpApp abstraction should work against `Resource[F, *]` instead of `F`. I believe there's an issue somewhere in http4s, but the impact will be quite heavy on the ecosystem so I think they're waiting for further stabilisation before pulling the trigger on that. 

So, we need to require the middleware to be provided as a `Client[F] => Client[F]`, which unfortunately implies creating another construct to echo the `EndpointSpecificMiddleware`, which may mean that we'd have to rename that middleware to `EndpointSpecificServerMiddleware which is a bit of a mouthful. 

How about `ServerEndpointMiddleware` and `ClientEndpointMiddleware` ?

Baccata (changes requested):

modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
lewisjkl: Added in https://github.com/disneystreaming/smithy4s/pull/614/commits/75313f5a407bb6021c304029c64b9a6439e14b0f

Baccata (approved):

PR #619, opened 2022-11-21
AWS: expose detailed errors in AwsClient.prepare

### LINKED ISSUE #601: AWS: expose list of supported protocols
This is pretty much a nice-to-have, but anyway...

In smithy-playground, I currently maintain a list of supported AWS protocols. When another one is added in smithy4s, I'll get support for it "for free", but I'll need to remember to update [that list](https://github.com/kubukoz/smithy-playground/blob/984552c3b3293ae839386ae8fc03fb51c2c5333f/modules/core/src/main/scala/playground/OperationRunner.scala#L228-L229).

It would be cool if `AwsClient.prepare` failed with a list of known protocols (in case the service doesn't have either of them), or if there was a public list of them exposed in a val somewhere. It's also fine if this doesn't happen, it's not like we change the list of protocols every week.

kubukoz:
Closes #601


modules/aws/src/smithy4s/aws/AwsClient.scala
```diff
@@ -21,30 +21,34 @@ import cats.effect.Resource
...
-    prepare(service).map(_.build(awsEnv)).liftTo[Resource[F, *]]
+    prepare(service)
+      .leftWiden[Throwable]
+      .map(_.build(awsEnv))
+      .liftTo[Resource[F, *]]
 
   def prepare[Alg[_[_, _, _, _, _]]](
       service: smithy4s.Service[Alg]
-  ): Either[Throwable, AWSInterpreterBuilder[Alg]] =
+  ): Either[InitialisationError, AWSInterpreterBuilder[Alg]] =
```
kubukoz, line 40: This whole change to typed errors may be overkill, let me know what you think. I'd say we should expose the public list of protocols anyway, but having them in the error type gives callers the ability to show better errors on their side as well.

Baccata: I'd rather not, on the account that it makes it harder to retain bincompat

kubukoz: In dynamic land it's not completely certain that it happens at bootstrapping 😄

Baccata: sure, but at the same time in dynamic land you're likely to lift the `Either` into an `F` and let it propagate, aren't you ?

kubukoz: not really, I have preparation as a separate step and I might want to show a different message depending on whether:

- the service is not an AWS service at all
- the service is AWS but we don't support the protocol yet

kubukoz: @Baccata just reminding about this one - it's breaking territory so we should figure it out before 0.17

Baccata: Okay. The point that convinces me is that the SimpleRestBuilder does have a "UnsupportedProtocolError" typed error on the left. 

Rename the ADT `AwsClientInitialisationError` and move it to its own file and we'll merge

Baccata (approved):

PR #622, opened 2022-11-22
Catch metadata errors in operations without an Errorable

### LINKED ISSUE #621: http4s server regression: Metadata errors aren't caught in errorless operations
```smithy
$version: "2"
namespace demo

@smithy4s.api#simpleRestJson
service InitService {
  operations: [Initiate]
}

@http(method: "GET", uri: "/initiate/{input}", code: 204)
operation Initiate {
    input := {
        @required
        @httpLabel
        @length(min: 5)
        input: String
    }
}
```

If you make a server with `SimpleRestJsonBuilder.routes`, and send any request that fails the refinement (e.g. `GET /initiate/a`), you'll either:

- have an exception thrown (`Client.fromHttpApp`)
- get a 500 (using a real client/server pair)

This started in `0.16.5` (still present on `0.16.8` and `0.17.0-127-b1246cc`). I can also reproduce the same behavior with refinements in the endpoint body, however it's a different exception:

- in URL: `MetadataError$FailedConstraint: Field input, found in Path parameter input, failed constraint checks with message: length required to be >= 5, but was 4`
- in body: `PayloadError: length required to be >= 5, but was 4 (path: .)`

The issue may be related to #594. I'm actually kinda surprised it's not fixed by #596, at least in the case of bodies.

I'm going to look into fixing this asap, just wanted to have an issue as a reference first.

kubukoz:
Closes #621.


sampleSpecs/noerrors.smithy
```diff
@@ -0,0 +1,31 @@
+$version: "2.0"
+
+namespace smithy4s.example
+
+@smithy4s.api#simpleRestJson
+service EchoService {
```
kubukoz, line 6: I wanted to add a new operation to `pizza.smithy` and make a pizza test for this, but unfortunately PizzaSpec is pretty much hard-wired to a single service, with a list of errors in the service itself - so I thought it would be impossible to reproduce the erroneous behavior without changing how PizzaSpecs are implemented.

One thing came to mind, though: smithy4s (the runtime) doesn't care whether an error comes from the service or the operation, so maybe we could inline these service-wide errors into the operations that need them? (we most likely cover the support of service errors in many other tests).

That way we could merge this into Pizza and have some guarantees that users of PizzaSpec will also implement the right behavior.

sampleSpecs/pizza.smithy
```diff
@@ -288,4 +288,4 @@ structure CustomCodeInput {
 structure CustomCodeOutput {
   @httpResponseCode
   code: Integer
-}
\ No newline at end of file
+}
```
kubukoz, line 291: shamelessly adding a trailing newline because it's the POSIX thing to do.

modules/http4s/test/src/smithy4s/http4s/Http4sNoErrorsServerSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+      .map(assert.eql(_, Status.BadRequest))
+  }
+
+  test("body failing refinement results in a BadRequest") {
+    client
+      .status(
+        Request[IO](
+          method = Method.POST,
+          uri = uri"/echo/test-long"
+        ).withEntity("test".asJson)
```
kubukoz, line 73: call me lazy because I didn't want to deal with bytes/content-type by hand. I can change it if desired.

daddykotex (approved):

Issue #571, opened 2022-11-03
How to generate open api file using the cli tool


ShahOdin:
I'm looking at: https://disneystreaming.github.io/smithy4s/docs/overview/cli and trying to generate the open api from the smithy files. in a custom directory taking the structure introduced in the quick start example as the starting point. 

For a service `PlaygroundService` with multiple operations `Choice,CreateCustomer,Hello`:

```bash
playground % tree -I target

.
└── src
    └── main
        └── smithy
            ├── Choice.smithy
            ├── CreateCustomer.smithy
            ├── Hello.smithy
            └── PlaygroundService.smithy
└── temp
```

- I tried the following from the top directory ^:

  ```bash
  smithy4s generate \
  --output ./temp/ \
  --resource-output ./temp \
  --discover-models
  ```

  I simply get blank files generated under `temp/META-INF`.

- I tried `cd` ing to `./src/main/smithy` and re-ran the above. again. nothing.
- I played around with the `-- repositories` and `-- dependencies` options with no luck. 

What am I missing? Would be great if you could add a minimal example in the docs as well.

kubukoz:
I think you need to pass `./src` or, to be more precise, `./src/main/smithy`, as the positional argument to `generate`. You also probably don't need `--discover-models`.

ShahOdin:
D'oh! that does it @kubukoz thank you. :) shall I still keep the issue open for the docs to make this clearer?

kubukoz:
yeah, I think that's a good idea. Would you like to make that change in the docs?

Baccata:
probably yes

kubukoz:
I think we should at least log something. An error is fine as long as it's only for CLI usage, not programmatic codegen usage

daddykotex:
I closed my PR because in the `series/0.17`, we have a default dependencies, so my check was always false :/ Maybe a warning we have the model in hand and we realize that it contains no shapes is enough

kubukoz:
Is there anything else we want to do here or can the ticket be closed?

PR #626, opened 2022-11-22
Fail CLI if all input are empty

daddykotex:
References: https://github.com/disneystreaming/smithy4s/issues/571

The CLI fails with a non-zero exit code and prints the help along with an error message, it looks like this:

```
No input for the Smithy model. Consider pointing to a folder or a Smithy file. You can also use `--dependencies` to import from maven coordinates or use `--localJars` to point to a JAR on your file system.

Usage: smithy4s generate [--output <path>] [--resource-output <path>] [--skip <string>]... [--discover-models] [--allowed-ns <string,string,...>] [--excluded-ns <string,string,...>] [--repositories <string,string,...>] [--dependencies <string,string,...>] [--transformers <string,string,...>] [--localJars <path,path,...>] [<path>...]

Generates scala code and openapi-specs from smithy specs

Options and flags:
    --help
        Display this help text.
    --output <path>, -o <path>
        Path where scala code should be generated. Defaults to pwd
    --resource-output <path>
        Path where non-scala files should be generated. Defaults to pwd
    --skip <string>
        Indicates that some files types should be skipped during generation
    --discover-models
        Indicates whether the model assembler should try to discover models in the classpath
    --allowed-ns <string,string,...>
        Comma-delimited list of namespaces that should not be processed. If unset, all namespaces are processed (except stdlib ones)
    --excluded-ns <string,string,...>
        Comma-delimited list of namespaces that should not be processed. If unset, all namespaces are processed (except stdlib ones)
    --repositories <string,string,...>
        Comma-delimited list of repositories to look in for resolving any provided dependencies
    --dependencies <string,string,...>
        Comma-delimited list of dependencies containing smithy files
    --transformers <string,string,...>
        Comma-delimited list of transformer names to apply to smithy files
    --localJars <path,path,...>
        Comma-delimited list of local JAR files containing smithy files
```

### DIFF
--- a/modules/codegen-cli/src/smithy4s/codegen/cli/CodegenCommand.scala
+++ b/modules/codegen-cli/src/smithy4s/codegen/cli/CodegenCommand.scala
@@ -125,6 +125,21 @@ object CodegenCommand {
             localJars.getOrElse(List.empty)
           )
       }
+      .mapValidated { cmd =>
+        if (
+          cmd.specs.isEmpty && cmd.dependencies.isEmpty && cmd.localJars.isEmpty
+        ) {
+          Validated.invalidNel(
+            List(
+              "No input for the Smithy model. Consider pointing to a folder or a Smithy file.",
+              "You can also use `--dependencies` to import from maven coordinates",
+              "or use `--localJars` to point to a JAR on your file system."
+            ).mkString(" ")
+          )
+        } else {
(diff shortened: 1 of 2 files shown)


daddykotex:
Oh wow, I was sure the the build was calling the `run` directly within the same jvm, it's actually calling the CLI via `java` fail

PR #627, opened 2022-11-22
Add a message when no files are generated

daddykotex:
References #571 

Also, took the initiative to rename `--localJars` to `--local-jars` so that all flags remain in the same style.

Looks like this:

```
./smithy4s-gen-test generate # built with `cs bootstrap`
Nothing was generated. Make sure your targetting Smithy files or folders that include Smithy definitions. Otherwise, you can also use --dependencies to pull external JARs or use --local-jars to use JARs located on your file system.
```

### DIFF
--- a/modules/codegen-plugin/src/smithy4s/codegen/JsonConverters.scala
+++ b/modules/codegen-plugin/src/smithy4s/codegen/JsonConverters.scala
@@ -76,7 +76,7 @@ private[smithy4s] object JsonConverters {
         ("repositories", ca.repositories) :*:
         ("dependencies", ca.dependencies) :*:
         ("transformers", ca.transformers) :*:
-        ("localJars", ca.localJars) :*:
+        ("local-jars", ca.localJars) :*:
         LNil
     },
     {
--- a/modules/codegen-plugin/src/smithy4s/codegen/JsonConverters.scala
+++ b/modules/codegen-plugin/src/smithy4s/codegen/JsonConverters.scala
@@ -76,7 +76,7 @@ private[smithy4s] object JsonConverters {
         ("repositories", ca.repositories) :*:
         ("dependencies", ca.dependencies) :*:
         ("transformers", ca.transformers) :*:
-        ("local-jars", ca.localJars) :*:
+        ("localJars", ca.localJars) :*:
         LNil
     },
     {
--- a/modules/codegen/src/smithy4s/codegen/Codegen.scala
+++ b/modules/codegen/src/smithy4s/codegen/Codegen.scala
@@ -46,7 +46,9 @@ object Codegen { self =>
         scalaFile
       }
       val generatedNamespaces = codegenResult.map(_._2.namespace).distinct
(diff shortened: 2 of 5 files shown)


daddykotex:
Aimed at series/0.17 because of the local-jars change

lewisjkl (approved):

modules/codegen/src/smithy4s/codegen/Codegen.scala
```diff
@@ -46,7 +46,9 @@ object Codegen { self =>
         scalaFile
       }
       val generatedNamespaces = codegenResult.map(_._2.namespace).distinct
-      val resources = if (!args.skipResources) {
+      val skipResource =
+        args.skipResources || (args.specs.isEmpty && generatedNamespaces.isEmpty)
```
kubukoz, line 50: Why do we need that second part?

Baccata: ☝️ I'm curious too. In particular, `generatedNamespaces` means "the namespaces that are already generated", not "the namespaces we're intending to generate"

daddykotex: otherwise we produce two files:
```
> smithy4s generate
/Users/David.Francoeur/workspace/dev/smithy4s/META-INF/smithy/manifest
/Users/David.Francoeur/workspace/dev/smithy4s/META-INF/smithy/smithy4s.tracking.smithy

> cat META-INF/smithy/manifest
smithy4s.tracking.smithy
> cat META-INF/smithy/smithy4s.tracking.smithy 
$version: "2.0"

metadata smithy4sGenerated = [{smithy4sVersion: "dev-SNAPSHOT", namespaces: []}]
```
So my warning never prints

modules/codegen-plugin/src/smithy4s/codegen/JsonConverters.scala
```diff
@@ -76,7 +76,7 @@ private[smithy4s] object JsonConverters {
         ("repositories", ca.repositories) :*:
         ("dependencies", ca.dependencies) :*:
         ("transformers", ca.transformers) :*:
-        ("localJars", ca.localJars) :*:
+        ("local-jars", ca.localJars) :*:
```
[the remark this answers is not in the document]
daddykotex: Oh you're absolutely right, I just went on with the find/replace frenzy

modules/codegen/src/smithy4s/codegen/Codegen.scala
Baccata: ooh, gotcha

kubukoz: smort

kubukoz: normally I'd like that to be in a comment as it's more findable than a PR comment, but I don't want to block this - feel free to follow up or add it in another PR

kubukoz (approved):

PR #630, opened 2022-11-23
smithy4sInputDirs more conservative default

lewisjkl:
Restores the previous default since `unmanagedSourceDirectories` is not what the average user would need and it causes some unexpected behavior depending on the build setup (as seen on some test projects).

### DIFF
--- a/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
+++ b/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
@@ -97,8 +97,9 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
 
   // Use this with any configuration to enable the codegen in it.
   def defaultSettings(config: Configuration) = Seq(
-    config / smithy4sInputDirs := (config / unmanagedSourceDirectories).value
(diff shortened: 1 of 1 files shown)


kubukoz:
curious, what's the unexpected behavior you've seen?

kubukoz (approved):

lewisjkl:
> curious, what's the unexpected behavior you've seen?

Sorry meant to go back and edit my description above with more context. One of the things is that if you have multiple source types (e.g. java and scala) then you will get two source directories added of the same location. We could of course solve this with `.distinct`, but as you'll see below, the paths are not always the same which can lead to unexpected src directories being added.

The other issue being that if you customize your source directory locations at all, then the default of using

```scala
(config / unmanagedSourceDirectories).value
      .map(_.getParentFile() / "smithy")
```

doesn't necessarily make sense any more since you may not want your smithy files at this location. An example is if you use a mill-style directory layout. This setting would then put your smithy files at the same level as `src` instead of beneath `src`.

So basically using the singular sourceDirectory / smithy is a safer default because your main source directory is always something that is clear cut whether you are using a custom directory structure, complex build, etc or not. And then if you want to customize the location of your smithy files from there, you can override the setting. Wdyt?

kubukoz:
Sounds reasonable :)

PR #631, opened 2022-11-23
0.17.0 Changelog draft [ci-skip]

Baccata:


CHANGELOG.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Smithy-level breaking changes
+
+See https://github.com/disneystreaming/smithy4s/pull/561
+
+The smithy shapes that were previously residing under `smithy4s.api` namespace have moved to the `alloy` namespace. Alloy is a standalone library containing smithy shapes and validators, defined [here](https://github.com/disneystreaming/alloy).
+
+The reason for us to bring this change is to have a language specific location to define shapes that are relevant to the protocols/runtime-behaviours we're putting forward, that could be used by tooling working with other languages than Scala. It was important for us to lose the `4s` suffix, which is short for `for Scala`.
+
+This change implies, for instance, that any use of `smithy4s.api#simpleRestJson` in your specification will have to be replaced by `alloy#simpleRestJson`.
```
kubukoz, line 19: Might be worth noting that the Scala code using `SimpleRestJson` is not affected

CHANGELOG.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Advanced usage breaking changes
+
+The abstractions that the generated code implements and that the runtime interpreters use have undergone some massive chances.
+
+#### Unification of the natural-transformations/polymorphic functions.
+
+* Smithy4s makes a lot of use of polymorphic functions of various kinds. Those are now code-generated (see the `project/Boilerplate.scala` file) to ensure the consistency of the various ways they are being used. This means that `smithy4s.PolyFunction` has moved to `smithy4s.kinds.PolyFunction`, and that the previous `smithy4s.Transformation` is now `smithy4s.kinds.PolyFunction5`. This decision ripples in the `smithy4s.Service` interface, which now sports `toPolyFunction` and `fromPolyFunction` methods, allowing to turn a finally-encoded implementation into a final one. `smithy4s.kinds.PolyFunction2` is also a thing, and may be used in bi-functor contexts.
```
kubukoz, line 58: `sports`? fancy wording here

CHANGELOG.md
```diff
@@ -0,0 +1,65 @@
...
+
+Depending on your setup, it may be a breaking change, but `@deprecated` Smithy-traits now translate to the `@deprectated` Scala annotation in the generated code.
+
+### Normal-usage source breaking changes
+
+See https://github.com/disneystreaming/smithy4s/pull/569
+
+If you use Smithy4s in the ways that were previously advertised in the documentation, you may have to perform some small adjustments.
+
+In particular, the `simpleRestJson` extension method that was added to implementations of service-interfaces generated by Smithy4s is now deprecated, in favour of the `SimpleRestJsonBuilder` construct (which now works for any `service` smithy shape that will have been annotated with `alloy#simpleRestJson`).
```
kubukoz, line 50: tbh we could deprecate it in a patch and drop it in 0.17 altogether

CHANGELOG.md
kubukoz, line 58: I would suggest listing all the renamed types related to polyfunctions, as it's not an immediately obvious change and in general would be helpful. I posted an example on the original PR https://github.com/disneystreaming/smithy4s/pull/555#issuecomment-1294249786

CHANGELOG.md
```diff
@@ -0,0 +1,74 @@
...
+
+
+### Error Response Handling Improvements
+
+Streamlines and improves how error responses are mapped to their corresponding smithy4s-generated types. It now works such that IF no `X-Error-Type` header is found AND the status code doesn't map precisely to an error annotated with @httpCode AND exactly one error happens to have @error("client") without @httpCode, that error will be selected (provided the status code is in the 4xx range). Same for @error("server") and 5xx range. See the [error handling documentation](https://disneystreaming.github.io/smithy4s/docs/protocols/simple-rest-json/client#error-handling) for more.
```
lewisjkl, line 74: Added these descriptions which I think cover all user-facing changes I made in this release. The only other changes I made I think are adding internal stuff that is not binary breaking (so probably needs no mention here?)

CHANGELOG.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Normal-usage breaking changes in the generated code
+
+See https://github.com/disneystreaming/smithy4s/pull/599
+
+Depending on your setup, it may be a breaking change, but `@deprecated` Smithy-traits now translate to the `@deprectated` Scala annotation in the generated code. For instance, if you used `@enum` heavily, you'll probably deprecation warnings in your console when compiling. Depending on your `scalacOptions`, it is possible that these warnings turn into errors. If you want to silence these particular errors while upgrading, you can do the following:
```
kubukoz, line 42: tbh we should encourage users to move away from `@enum` and use smithy2 enums

CHANGELOG.md
```diff
@@ -0,0 +1,84 @@
...
+* `kind`-specific types were created to facilitate the "lift" of constructs to the right kinds. For instance, one inspecting the internals of this library might see things like `Kind1[IO]#toKind5` where in was previously `GenLift[IO]#λ`. We're hoping to convey meaning better, although this code is definitely still not trivial (and never will).
+* `smithy4s.Transformation` is now a typeclass-like construct, which expresses the fact that a construct can be applied on another like a function. This construct is used by the `transform` method that is generated on service interfaces, which allows to apply custom behaviour generically on all method invocations in these interfaces.
+* The `Service` interface takes a single `Alg` type parameter, the `Op` parameter has moved to type-member position, facilitating implicit search in some contexts (as well as the writing of some logic).
+* A bunch of path-dependant type-aliases were created in the `Service` interface.
```
kubukoz, line 72:
```suggestion
* A bunch of path-dependent type-aliases were created in the `Service` interface.
```

CHANGELOG.md
```diff
@@ -0,0 +1,84 @@
...
+* The `compliancetest` module has changed drastically in UX. For those not aware, this module allows to run tests written in Smithy against your own implementation of protocols. This will be useful for third-party libraries that implement `simpleRestJson` (or any other http/rest like protocol), to virtually get tests for free. We don't think this module had any users so far, but we'll slowly be porting some of our tests away from the `smithy4s` repository and into the `alloy` repository.
```
kubukoz, line 73: I assume the tests will be ported only as test traits?

Baccata: Yes.

CHANGELOG.md
Baccata: This will have this side effect.

CHANGELOG.md
```diff
@@ -0,0 +1,142 @@
...
+* `kind`-specific types were created to facilitate the "lift" of constructs to the right kinds. For instance, one inspecting the internals of this library might see things like `Kind1[IO]#toKind5` where in was previously `GenLift[IO]#λ`. We're hoping to convey meaning better, although this code is definitely still not trivial (and never will).
+* `smithy4s.Transformation` is now a typeclass-like construct, which expresses the fact that a construct can be applied on another like a function. This construct is used by the `transform` method that is generated on service interfaces, which allows to apply custom behaviour generically on all method invocations in these interfaces.
```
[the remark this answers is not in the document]
kubukoz: I think Oli meant that the construct is "like a function", as in "it can be applied to something"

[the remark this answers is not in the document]
daddykotex: then I guess, this is better: `that a construct can be applied like a function`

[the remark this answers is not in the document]
daddykotex: if you agree I can fix it

PR #639, opened 2022-11-25
Revise some type aliases / add path-dependent ones

Baccata:
* Adds a path-dependant `Impl` type-alias under the Service interface, acting as a short hand for `FunctorAlgebra`. This makes some of the code a little easier to read imho. 
* Change `WithError` to `ErrorAware` and make it a path-dependant type-alias. I was quite unhappy with `WithError`
* Add a `def apply` invocation method under the generated `ErrorAware` objects


daddykotex (approved):
I love the end result. The tricks to get there still escape me a little bit but that should not stop us

kubukoz (approved):

PR #643, opened 2022-11-28
Define source/resource generator together

daddykotex:
I did not do it for the `Test` configuration, but if you want me to, I can

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -837,8 +837,8 @@ lazy val benchmark = projectMatrix
 
 def genSmithy(config: Configuration) = Def.settings(
   Seq(
-    Compile / sourceGenerators := Seq(genSmithyScala(Compile).taskValue),
-    Compile / resourceGenerators := Seq(genSmithyResources(Compile).taskValue)
+    config / sourceGenerators := Seq(genSmithyScala(config).taskValue),
+    config / resourceGenerators := Seq(genSmithyResources(config).taskValue)
   )
 )
 def genSmithyScala(config: Configuration) = genSmithyImpl(config).map(_._1)
--- a/build.sbt
+++ b/build.sbt
@@ -164,7 +164,7 @@ lazy val core = projectMatrix
     smithy4sDependencies ++= Seq(
       Dependencies.Smithy.waiters
     ),
-    Compile / sourceGenerators := Seq(genSmithyScala(Compile).taskValue),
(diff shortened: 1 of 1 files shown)


lewisjkl (approved):

build.sbt
```diff
@@ -833,11 +830,17 @@ lazy val benchmark = projectMatrix
...
   )
   .jvmPlatform(List(Scala213), jvmDimSettings)
   .settings(Smithy4sBuildPlugin.doNotPublishArtifact)
 
+def genSmithy(config: Configuration) = Def.settings(
+  Seq(
+    Compile / sourceGenerators := Seq(genSmithyScala(Compile).taskValue),
+    Compile / resourceGenerators := Seq(genSmithyResources(Compile).taskValue)
+  )
+)
```
lewisjkl, line 843: i don't think this is what is causing the CI failure, but should use `config` here in the function rather than `Compile`

lewisjkl (changes requested):

build.sbt
daddykotex: you're right, big mishap lol

build.sbt
```diff
@@ -553,7 +553,7 @@ lazy val dynamic = projectMatrix
     Compile / smithySpecs := Seq(
       (ThisBuild / baseDirectory).value / "modules" / "dynamic" / "smithy" / "dynamic.smithy"
     ),
-    Compile / sourceGenerators := Seq(genSmithyScala(Compile).taskValue),
+    genSmithy(Compile),
```
daddykotex, line 556: I had a theory on why it broke down, but it doesn't appear to be true. I thought maybe having `resourceGenerators` here and in `core` (`dynamic` dependency) would put conflicting files on the class path and only one would be picked up

but it does not seem to be the case

daddykotex: I sort of worked around it by not applying it across the board, but I'm unsatisfied

we figured out the issue with `http4s`: the test loads a file directly, and, after the change, the file is picked up by the discoverModels because it's copied in the resources

but for the `dynamic` issue, I still can't figure out: you add 3 files to the class path:
```
dynamic.smithy
manifest
smithy4s.tracking.smithy
```

And then suddenly, `alloy#discriminated` is unresolvable

Baccata (approved):

Issue #644, opened 2022-11-28
(0.17 snapshots) Conflicting mixin trait


kubukoz:
Using the latest snapshots (`0.17.0-177-7156a5c`), if you use the AWS module together with the codegen module (as a library):

```scala
libraryDependencies ++= Seq(
  "com.disneystreaming.smithy4s" %% "smithy4s-aws-kernel" % smithy4sVersion.value,
  "com.disneystreaming.smithy4s" %% "smithy4s-codegen-cli" % smithy4sVersion.value
),
```

you'll get an error stating there's a conflict between smithy-aws-traits 1.26.0 and 1.26.3:

```
[error] (Compile / smithy4sCodegen) software.amazon.smithy.model.validation.ValidatedResultException: Result contained ERROR severity validation events: 
[error] [ERROR] aws.protocols#HttpConfiguration: Conflicting `smithy.api#mixin` trait found on shape `aws.protocols#HttpConfiguration`. The previous trait was defined at `jar:file:/Users/kubukoz/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/software/amazon/smithy/smithy-aws-traits/1.26.3/smithy-aws-traits-1.26.3.jar!/META-INF/smithy/aws.protocols.smithy [35, 1]`, and a conflicting trait was defined at `jar:file:/Users/kubukoz/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/software/amazon/smithy/smithy-aws-traits/1.26.0/smithy-aws-traits-1.26.0.jar!/META-INF/smithy/aws.protocols.smithy [35, 1]`. | Model jar:file:/Users/kubukoz/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/software/amazon/smithy/smithy-aws-traits/1.26.0/smithy-aws-traits-1.26.0.jar!/META-INF/smithy/aws.protocols.smithy:35:1
```

Repro: https://github.com/kubukoz/demos/tree/mixin-conflict-repro + `sbt smithy4sCodegen`

Found this in [smithy-playground](https://github.com/kubukoz/smithy-playground/pull/129), which uses the codegen module to call `ModelLoader`. I can get rid of it, but perhaps this is a symptom of an otherwise significant issue.

kubukoz:
fixed in #646

PR #646, opened 2022-11-29
Unifies resolution before codegen

Baccata:
Unifies resolution (via the build tool's resolver) to ensure that there are no conflict resolved jars that are passed to the code-generator. This is done by gather all ModuleIDs before re-running the resolver, in order for it to perform the necessary eviction.

Thus, we only concatenate the jars that come from dependency resolution with the internal jars (which should hardly ever conflict).

Fixes https://github.com/disneystreaming/smithy4s/issues/644 

- [x] SBT 
- [x] Mill

### DIFF
--- a/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
+++ b/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
@@ -62,26 +62,42 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
         "Sets whether this project should be used as a Smithy library by packaging the Smithy specs in the resulting jar"
       )
 
-    val smithy4sExternalDependenciesAsJars =
-      taskKey[Seq[File]](
+    val smithy4sExplicitCodegenOnlyDependencies =
+      taskKey[Seq[ModuleID]](
         List(
-          "List of jars for external dependencies that should be added to the classpath used by Smithy4s during code-generation",
+          "List of explicitly defined libraries for external dependencies that should be added to the classpath used by Smithy4s during code-generation",
           "The smithy files and smithy validators contained by these jars are included in the Smithy4s code-generation process",
           "Namespaces that were used for code generation in these dependencies will be excluded from code generation in this project.",
-          "By default, this includes the jars resulting from the resolution of library dependencies annotated with the `Smithy4s` configuration",
-          "as well as the `standard` library dependencies"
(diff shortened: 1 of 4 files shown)


modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -62,29 +62,45 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
...
+    val smithy4sExplicitCodegenOnlyDependencies =
+      taskKey[Seq[ModuleID]](
         List(
-          "List of jars for external dependencies that should be added to the classpath used by Smithy4s during code-generation",
+          "List of explicitly defined libraries for external dependencies that should be added to the classpath used by Smithy4s during code-generation",
           "The smithy files and smithy validators contained by these jars are included in the Smithy4s code-generation process",
           "Namespaces that were used for code generation in these dependencies will be excluded from code generation in this project.",
-          "By default, this includes the jars resulting from the resolution of library dependencies annotated with the `Smithy4s` configuration",
-          "as well as the `standard` library dependencies"
+          "By default, this includes the library dependencies of annotated with the `Smithy4s` configuration of the current project and its upstreams"
```
daddykotex, line 71:
```suggestion
          "By default, this includes the library dependencies annotated with the `Smithy4s` configuration of the current project and its upstreams"
```

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -62,29 +62,45 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
...
+        List(
+          "List of libraries that are declared normally and that will be added to the classpath of the code-generation process",
+          "By default, this includes all `libraryDependencies` matching the scope, both local and transitive"
+        ).mkString(" ")
+      )
+
+    val smithy4sAllExternalDependencies =
+      taskKey[Seq[ModuleID]](
+        List(
+          "Exhaustive list external dependencies that should be added to the classpath used by Smithy4s during code-generation",
```
daddykotex, line 98:
```suggestion
          "Exhaustive list of external dependencies that should be added to the classpath used by Smithy4s during code-generation",
```

daddykotex (commented):
found some typos in the descriptions

daddykotex (approved):

kubukoz (approved):
works like a charm

PR #648, opened 2022-11-29
Clean-up the codegen module to have a very small API surface

Baccata:
Nothing in this module but the `Codegen` object should be considered public API, so I'm moving most of the stuff to an internals package and tagging everything as private.


modules/codegen/src/smithy4s/codegen/internals/package.scala
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  def uncapitalise(s: String) = {
+    if (s == s.toUpperCase()) s.toLowerCase()
+    else {
+      val count: Int = s.segmentLength(c => c.==('_'), 0)
+      val after = count + 1
+      s.take(after).toLowerCase + s.drop(after)
+    }
+  }
```
lewisjkl, line 186: Couldn't these be package private as well?

lewisjkl: And same with the objects (S, T, N) above?

lewisjkl: Awesome, sounds great 👍

modules/codegen-cli/src/smithy4s/codegen/cli/DumpModel.scala
```diff
@@ -16,7 +16,7 @@
 
 package smithy4s.codegen.cli
 
-import smithy4s.codegen.ModelLoader
+import smithy4s.codegen.internals.ModelLoader
```
daddykotex, line 19: pushed this change, to fix the build, which highlights one caveat: the DumpModel uses the ModelLoader, so we have to import from `internals`

we can do otherwise and export a function or something like that if you prefer, just bringing this to your attention

daddykotex: lol there is also a codegen plugin tests that uses it haha

Baccata: Addressed in [f61403e](https://github.com/disneystreaming/smithy4s/pull/648/commits/f61403e42de2457cfa88fdd8c30b72e2b2c48b25)

modules/codegen/src/smithy4s/codegen/internals/package.scala
Baccata: Addressed in [f61403e](https://github.com/disneystreaming/smithy4s/pull/648/commits/f61403e42de2457cfa88fdd8c30b72e2b2c48b25)

kubukoz:
> smaller api is best

if you're not using it :trollface:

kubukoz (approved):
so in my understanding, these are now the only public ones:

- `CodegenCommand` + `Smithy4sCommand`
- `Codegen.dumpModel` / `Codegen.processSpecs`

Baccata:
> so in my understanding, these are now the only public ones: ...

I don't think I want to promise any API stability on the cli module. It's a CLI, not a library ! Do you want me to privatise stuff there too ? 

`smithy4s.codegen.Codegen` is the only one that should be considered library code.

kubukoz:
yeah the CLI could be private too, lgtm either way

PR #650, opened 2022-11-30
Document the assembly rule

daddykotex:


yisraelU (approved):
anyway to provide this as a default ? I cant imagine a user wants this file even if they dont have multiple jars

daddykotex:
> anyway to provide this as a default ? I cant imagine a user wants this file even if they dont have multiple jars

I don't know because we don't have a dependency on sbt-assembly

kubukoz (approved):

PR #653, opened 2022-12-01
Remove duplicate Transformation trait

kubukoz:
Also removes a type alias that causes an import conflict when you do

```scala
import smithy4s.kinds._
import smithy4s._
```


daddykotex (approved):
I stumbled on this once and I thought I had made a mistake

Baccata (approved):

PR #654, opened 2022-12-02
Remove the Service.Provider construct

Baccata:
I *think* the displacement of `Op` from parameter to member may have helped the tooling, and this is no longer needed

### DIFF
--- a/modules/aws-http4s/src/smithy4s/aws/http4s/package.scala
+++ b/modules/aws-http4s/src/smithy4s/aws/http4s/package.scala
@@ -24,7 +24,7 @@ import org.http4s.client.Client
 package object http4s {
 
   implicit final class ServiceOps[Alg[_[_, _, _, _, _]]](
-      private[this] val serviceProvider: smithy4s.Service.Provider[Alg]
+      private[this] val service: smithy4s.Service[Alg]
   ) {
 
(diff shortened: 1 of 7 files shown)


daddykotex (approved):
Wow, that's nice. For context, this is an important bit of why the Provider was there in the first place. From @Baccata :

> `Service.Provider` was a trick because we were generating `Gen`-less synthetic companions for monofunctor-type aliases. We were doing to help the tooling (can't whether it was IJ or metals, or even the compiler itself), but there were some inference problems that seem to have disappeared, so now instead of generating synthetic companions, we generate val that points to the actual companion object.

kubukoz:
I wonder if Scala 3 + Metals has no issues with it, it used to choke on some heavier types. Might give it a look later

kubukoz:
I tried completions on this code:

image

both Metals and IntelliJ seem fine most of the time (although I was using BSP for IntelliJ, but afaik it's still going to do its own thing for completions)

Minor problem:

- IntelliJ doesn't like `ErrorAware` (in Scala 3 only - it's fine on 2.13)

image

kubukoz (approved):

kubukoz:
update: the issue above is actually fixed in the 2022.3 version of the Scala plugin!

PR #657, opened 2022-12-05
Rendering: fix for case where StringNode contains quotes

yisraelU:


modules/codegen/src/smithy4s/codegen/Renderer.scala
```diff
@@ -978,8 +978,10 @@ private[codegen] class Renderer(compilationUnit: CompilationUnit) { self =>
             }
             line"smithy4s.Document.obj(${members.toList.intercalate(Line.comma)})"
           }
-          def stringNode(x: StringNode): Line =
-            line"""smithy4s.Document.fromString("${x.getValue}")"""
+          def stringNode(x: StringNode): Line = {
+            val escapedString = x.getValue.replace("\"", "\\\"")
```
kubukoz, line 982: I think this should use `renderStringLiteral`. It takes care of this and other cases (such as unicode characters, probably newlines and other escapes as well).

kubukoz (changes requested):

modules/codegen/src/smithy4s/codegen/Renderer.scala
yisraelU: @kubukoz lol , you had me going crazy looking for that method.
thats on 0.17 branch,ill move it to main

kubukoz: didn't notice this was targetting main :P 

are we actually going to keep using main though?

yisraelU: dunno , all open pr's target main.
I assumed we would merge 0.17 into main

Baccata: @yisraelU the new "default" branch is `series/0.17`, we'll be doing this from now on : the default branch will be tracking whatever current "major" version is (or "minor" version according to early-semver semantics)

yisraelU: got it

kubukoz:
I think  you're a couple commits behind, hence some extra changes. Please update and try to avoid unnecessary changes in unrelated files

yisraelU:
@kubukoz something got screwed up for me locally , closing and opening new PR

PR #664, opened 2022-12-08
Compliance tests jsonfix

yisraelU:
adds dependency on circe for the Compliance tests module
detects when content type is json and the comparison is between 2 strings parse them first using a json parser to remove insignificant characters

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -36,6 +36,12 @@ object Dependencies {
       Def.setting("org.typelevel" %%% "cats-core" % "2.8.0")
   }
 
+  object Circe {
+    val circeVersion = "0.14.1"
+    val parser = "io.circe" %% "circe-parser" % circeVersion
+    val generic = "io.circe" %% "circe-generic" % circeVersion
+  }
+
   object Decline {
     val declineVersion = "2.4.0"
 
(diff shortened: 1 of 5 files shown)


kubukoz:
Q: how about, instead of a json AST, we parse the json into a Document and compare these instead? That'd avoid the need to depend on Circe

Baccata:
> Q: how about, instead of a json AST, we parse the json into a Document and compare these instead? That'd avoid the need t!o depend on Circe

 We discussed it internally and this basically implies that we'd be testing smithy4s behaviour (json decoding) with smithy4s constructs, which ain't a good idea.

project/Dependencies.scala
```diff
@@ -36,6 +36,12 @@ object Dependencies {
       Def.setting("org.typelevel" %%% "cats-core" % "2.8.0")
   }
 
+  object Circe {
+    val circeVersion = "0.14.1"
+    val parser = "io.circe" %% "circe-parser" % circeVersion
```
Baccata, line 41: http://sca.la/percent

project/Dependencies.scala
```diff
@@ -37,9 +37,9 @@ object Dependencies {
   }
 
   object Circe {
-    val circeVersion = "0.14.1"
-    val parser = "io.circe" %% "circe-parser" % circeVersion
-    val generic = "io.circe" %% "circe-generic" % circeVersion
+    val circeVersion = "0.14.3"
+    val parser =  Def.setting("io.circe" %%% "circe-parser" % circeVersion)
```
daddykotex, line 41: ahh, is the `Def.setting` required when using `%%%` ?

daddykotex (approved):
👍

PR #672, opened 2022-12-13
Pass scalaModuleInfo to dynamic resolution invocation

Baccata:
The absence of this information seems to trigger (hard to reproduce) edge-cases.


daddykotex:
Interesting

PR #675, opened 2022-12-13
fix for timestamp format issue for aws tests

yisraelU:
maps transitively the Hints on a schema and adds a Epoch timestamp format hint

### DIFF
--- a/modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
+++ b/modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
@@ -1,0 +1,32 @@
+/*
+ *  Copyright 2021-2022 Disney Streaming
+ *
+ *  Licensed under the Tomorrow Open Source Technology License, Version 1.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     https://disneystreaming.github.io/TOST-1.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.compliancetests
+
+import smithy4s.Hints
+import smithy4s.schema.Schema
+
+package object internals {
+
+  // Due to AWS's usage of integer as the canonical representation of a Timestamp in smithy , we need to provide the decoder with instructions to use a Long instead.
+  // therefore the timestamp type is switched to type epochSeconds: Long
+  // This is just a workaround thats limited to testing scenarios
+  def mapAllTimestampsToEpoch[A](schema: Schema[A]): Schema[A] = {
+    schema.transformHintsTransitively(h =>
+      h.++(Hints(smithy.api.TimestampFormat.EPOCH_SECONDS.widen))
+    )
+  }
+}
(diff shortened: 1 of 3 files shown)


modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
```diff
@@ -0,0 +1,32 @@
...
+package smithy4s.compliancetests
+
+import smithy4s.Hints
+import smithy4s.schema.Schema
+
+package object internals {
+
+  // Due to AWS's usage of integer as the canonical representation of a Timestamp in smithy , we need to provide the decoder with instructions to use a Long instead.
+  // therefore the timestamp type is switched to type epochSeconds: Long
+  // This is just a workaround thats limited to testing scenarios
```
kubukoz, line 26: I still don't understand... what evil thing happens without this fix? Does it just affect certain protocols? If so, should we really be applying this transformation on all schemas regardless of the protocol in question?

daddykotex: I'll try to summarize but @yisraelU will do a better job since he stumbled on this. Basically, we're loading test from a Smithy model. Some of these tests looks like this: https://github.com/awslabs/smithy/blob/f598b87c51af5943686e38706847a5091fe718da/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy#L453

So we have a schema, from the Smithy model that says some member are timestamp from `http-date`, others are for `epoch-seconds`, others are `date-time`. But then, their `params` is encoded using only `epoch-seconds` value. So we create a DocumentDecoder for the structure (it will be looking for `String` or `JsonNumber` depending on the `timestampFormat`) but the values are all json numbers...

kubukoz: I see, yeah. The test traits use a date encoding unrelated to the actual transport encoding. I did the same in SmithyQL but with the DATE_TIME format.

The point from [the docs](https://smithy.io/2.0/additional-specs/http-protocol-compliance-tests.html?highlight=httprequesttests#httprequesttests-trait) that I was missing:

> Timestamp values must be converted to a Unix timestamp represented as an integer.

answers my question about "is this for all protocols?" - the answer is yes, apparently.

yisraelU: My understanding is that the question is incorrect
Smithy is protocol agnostic.
Smithy also happens to have 2 ways to represent a value of the type Timestamp in Smithy, one is with an integer the other a string ,AWS chooses to use integer

kubukoz: sure, it's protocol agnostic, but using ints here seems to be part of the spec of `@httpRequestTests`. If it's just an AWS-specific thing, then why are we enabling this behavior on _all_ protocols?

modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
```diff
@@ -0,0 +1,32 @@
...
+  def mapAllTimestampsToEpoch[A](schema: Schema[A]): Schema[A] = {
+    schema.transformHintsTransitively(h =>
```
kubukoz, line 28: Q: doesn't this add the hint to all shapes, regardless of whether they're timestamps? We should probably only add it on the timestamps

modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
Baccata: The piece of data you're looking at describes parameters and has little to do with protocols. It basically equates to the case-class instance in generated code, as opposed to the http encoding of that instance. 

In this particular context, they have elected one canonical representation of timestamps that happens to be epoch-based.

modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
daddykotex: we talked about it and concluded it did not matter, but I also noted that

modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
kubukoz: Yep that was my understanding after seeing the docs I linked above.

Baccata: Oh right I misread your message above

Baccata (approved):

Issue #674, opened 2022-12-13
Codegen not producing compilable code on latest EC2 spec


igor-ramazanov:
Hi and thank you for your work.

I couldn't set [the basic example](https://disneystreaming.github.io/smithy4s/docs/protocols/aws/aws) from the docs [working](https://github.com/igor-ramazanov/scala-playground).

Copied the `dynamodb.json` into the `src/main/smithy/`, added the SBT plugin and it fails with:
```
sbt:scala-playground> compile
[error] stack trace is suppressed; run last Compile / smithy4sCodegen for the full output
[error] (Compile / smithy4sCodegen) software.amazon.smithy.model.validation.ValidatedResultException: Result contained ERROR severity validation events:
[error] [ERROR] com.amazonaws.dynamodb#DynamoDB_20120810: Unable to resolve trait `smithy.rules#endpointRuleSet`. If this is a custom trait, then it must be defined before it can be used in a model. | Model /home/igor/personal/scala-playground/src/main/smithy/dynamodb.json:3401:41
[error] [ERROR] com.amazonaws.dynamodb#DynamoDB_20120810: Unable to resolve trait `smithy.rules#endpointTests`. If this is a custom trait, then it must be defined before it can be used in a model. | Model /home/igor/personal/scala-playground/src/main/smithy/dynamodb.json:3765:39
[error] Total time: 1 s, completed Dec 13, 2022, 1:40:19 PM
```

Most probably, I'm missing something fundamental, will be happy to contribute to the docs!

Baccata:
Hey @igor-ramazanov , looks like AWS has added some dependencies to the models of their SDK. We don't control these models, obviously, so it may occur that sometimes our documentation would fall out of sync as we' don't build them upstream. 

Anyway :  I've managed to generate the dynamodb code by adding these settings : 

``` 
    Compile / smithy4sAllowedNamespaces := List(
      "smithy.rules",
      "com.amazonaws.dynamodb"
    ),
    libraryDependencies ++= Seq(
      "com.disneystreaming.smithy4s" %% "smithy4s-aws-http4s" % smithy4sVersion.value,
      "software.amazon.smithy" % "smithy-rules-engine" % smithy4s.codegen.BuildInfo.smithyVersion % Smithy4s
    )
```

igor-ramazanov:
It's a numbers narrowing, not widening, so there's no way to silent this with `-Wconf`.

Baccata:
[...]

PR #676, opened 2022-12-13
Update AWS protocol docs

igor-ramazanov:
Fixing the https://github.com/disneystreaming/smithy4s/issues/674.

AWS has added some dependencies to the models of their SDK, updated the AWS docs to reflect that.

### DIFF
--- a/modules/docs/markdown/03-protocols/03-aws/01-aws.md
+++ b/modules/docs/markdown/03-protocols/03-aws/01-aws.md
@@ -6,7 +6,7 @@ sidebar_label: AWS
 
 Smithy4s provides functions to create AWS clients from generated code. At the time of writing this, smithy4s is only able to derive clients for AWS services that use the [AWS Json 1.0/1.1 protocol](https://awslabs.github.io/smithy/1.0/spec/aws/index.html?highlight=aws%20protocols#aws-protocols).
 
-The AWS smithy specs (that are written in json syntax) can be found in some of the [official SDKs](https://github.com/aws/aws-sdk-js-v3/tree/main/codegen/sdk-codegen/aws-models) published by AWS. These `.json files` can be understood by smithy4s, just like `.smithy`, and can be used to generate code. Just copy/paste them in your project.
+The AWS smithy specs (that are written in json syntax) can be found in some of the [official SDKs](https://github.com/aws/aws-sdk-js-v3/tree/main/codegen/sdk-codegen/aws-models) published by AWS. These `.json files` can be understood by smithy4s, just like `.smithy`, and can be used to generate code. **Just copy/paste them in your project.**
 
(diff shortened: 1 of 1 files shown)


daddykotex (approved):
Thanks for the doc update @igor-ramazanov

That's amazing

Baccata:
Hey @igor-ramazanov, thank you so much for the contribution, but I think rather than fixing the documentation, this is something that needs to be addressed in build of the library itself. The documentation, as it currently is, would therefore be up-to-date once again. 

I need to digest the reason why AWS has added a dependency to devise a course of action though.

modules/docs/markdown/03-protocols/03-aws/01-aws.md
```diff
@@ -17,9 +17,15 @@ In `build.sbt`
 ```scala
 import smithy4s.codegen.BuildInfo._
 
+Compile / smithy4sAllowedNamespaces := List(
```
kubukoz, line 20: Strange, I never had to set this key for AWS to work.

igor-ramazanov: Perhaps, because AWS's updated their schema definition:

https://github.com/aws/aws-sdk-js-v3/commits/41bb10dc093

https://github.com/aws/aws-sdk-js-v3/pull/4070

lewisjkl: Yeah I think we do. It should maybe be more specific like `smithy.api.*`

kubukoz: do we even need to do so? With the `smithy4sGenerated` metadata I thiiiiiiink it might no longer be necessary to disable any namespaces by default.

For smithy4s-core, it contains this:

```smithy
$version: "2.0"

metadata smithy4sGenerated = [{smithy4sVersion: "0.17.0", namespaces: ["alloy", "smithy.waiters", "smithy.api"]}]
```

so I think it'd automatically ignore `smithy.api` and friends.

Baccata: I want to discourage people from using `smithy` prefixes to their namespaces in userland (people have actually tried that at $work).

igor-ramazanov:
Since @Baccata is working on fixing the actual underlying issue, I think it's reasonable to close this PR in favor of the future one.

Baccata:
Right so I believe the correct course of action is gonna be to set up some cron-build somewhere that would pre-process the AWS specs to re-construct the smithy models without the rule-engine traits, and package them into consumable artifacts. 

This would facilitate the set-up for AWS-SDK usecases as people wouldn't have to copy/paste specs, and the specs could be updated with a cron-job on a daily basis. 

The rule-engine traits are probably edited by AWS in separate files to the actual specs using Smithy's `apply` keyword, but the json serialiser which they use to check-in the specs into the SDK repos (which are basically our source of truth, as third-party observers) inlines these trait applications in the json file.

modules/docs/markdown/03-protocols/03-aws/01-aws.md
kubukoz: fair enough

Baccata:
Closing as a work-around should be set-up for this.

PR #677, opened 2022-12-14
Various codegen fixes and improvements 

Baccata:
1. Add precision suffixes to generated number literals, which appears to help Scala Native (`val x: Float = 1.0` doesn't play well with SN, you need to write `1.0f) 
2. Reworks the way that some functions of `smithy4s.Service` are implemented in the generated code, in order to avoid a "method too large" compilation error which appears to happen with services that have lots of operations. Instead of pat-mats, we use methods generated in the `Operation` associated to the service. 
3. Sorts the imports before rendering them. This is purely aesthetics but it looks much better.

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
@@ -1003,7 +1003,7 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
           def nullNode(x: NullNode): Line =
             line"smithy4s.Document.nullDoc"
           def numberNode(x: NumberNode): Line =
-            line"smithy4s.Document.fromDouble(${x.getValue.doubleValue()})"
+            line"smithy4s.Document.fromDouble(${x.getValue.doubleValue()}d)"
           def objectNode(x: ObjectNode): Line = {
             val members = x.getMembers.asScala.map { member =>
               val key = s""""${member._1.getValue()}""""
--- a/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
@@ -72,19 +72,18 @@ private[internals] object Renderer {
         .filter(_._2.size > 1)
         .keySet
 
-      val allImports: List[String] = renderResult.list
-        .flatMap { line =>
-          line.segments.toList.collect {
-            case nameRef @ NameRef(pkg, _, _)
-                if pkg.nonEmpty && !nameCollisions.contains(
-                  nameRef.getNamePrefix
-                )
-                  && !nameRef.isAutoImported &&
-                  !pkg.mkString(".").equalsIgnoreCase(unit.namespace) =>
-              nameRef.show
-            case Import(value) => value
-          }
+      val allImports: List[String] = renderResult.list.flatMap { line =>
+        line.segments.toList.collect {
+          case nameRef @ NameRef(pkg, _, _)
+              if pkg.nonEmpty && !nameCollisions.contains(
+                nameRef.getNamePrefix
+              )
+                && !nameRef.isAutoImported &&
+                !pkg.mkString(".").equalsIgnoreCase(unit.namespace) =>
+            nameRef.show
+          case Import(value) => value
         }
+      }
 
       val code: List[String] = renderResult.list
         .map { line =>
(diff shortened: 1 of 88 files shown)


modules/example/src/smithy4s/example/StreamedObjects.scala
```diff
@@ -88,4 +89,7 @@ object StreamedObjectsGen extends Service.Mixin[StreamedObjectsGen, StreamedObje
   }
 }
 
-sealed trait StreamedObjectsOperation[Input, Err, Output, StreamedInput, StreamedOutput]
+sealed trait StreamedObjectsOperation[Input, Err, Output, StreamedInput, StreamedOutput] {
```
Baccata, line 92: The generated methods in this trait (and its implementations) allow to avoid having to generate a huge pattern-match, which seems to help the compiler.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -990,9 +984,9 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
         (bd: BigDecimal) => line"scala.math.BigDecimal($bd)"
       case Primitive.BigInteger => (bi: BigInt) => line"scala.math.BigInt($bi)"
       case Primitive.Unit       => _ => line"()"
-      case Primitive.Double     => t => line"${t.toString}"
-      case Primitive.Float      => t => line"${t.toString}"
-      case Primitive.Long       => t => line"${t.toString}"
+      case Primitive.Double     => t => line"${t.toString}d"
+      case Primitive.Float      => t => line"${t.toString}f"
+      case Primitive.Long       => t => line"${t.toString}L"
```
daddykotex, line 989: minor: any reason to use different casing on the suffix? 1L, 1.0F and 1.0D

I guess D looking alike 0 is one of them¿

modules/example/src/smithy4s/example/BrandService.scala
```diff
@@ -56,11 +55,12 @@ object BrandServiceGen extends Service.Mixin[BrandServiceGen, BrandServiceOperat
   type Default[F[+_]] = Constant[smithy4s.kinds.stubs.Kind1[F]#toKind5]
 
   def toPolyFunction[P[_, _, _, _, _]](impl : BrandServiceGen[P]): PolyFunction5[BrandServiceOperation, P] = new PolyFunction5[BrandServiceOperation, P] {
-    def apply[I, E, O, SI, SO](op : BrandServiceOperation[I, E, O, SI, SO]) : P[I, E, O, SI, SO] = op match  {
-      case AddBrands(AddBrandsInput(brands)) => impl.addBrands(brands)
-    }
+    def apply[I, E, O, SI, SO](op : BrandServiceOperation[I, E, O, SI, SO]) : P[I, E, O, SI, SO] = op.run(impl) 
+  }
+  case class AddBrands(input: AddBrandsInput) extends BrandServiceOperation[AddBrandsInput, Nothing, Unit, Nothing, Nothing] {
+    def run[F[_, _, _, _, _]](impl: BrandServiceGen[F]): F[AddBrandsInput, Nothing, Unit, Nothing, Nothing] = impl.addBrands(input.brands)
```
daddykotex, line 61: wow that's clever

daddykotex: I would never have thought of it lol, so easy when you think about it

modules/example/src/smithy4s/example/Numeric.scala
```diff
@@ -0,0 +1,33 @@
...
+import smithy4s.schema.Schema.struct
+
+case class Numeric(i: Int = 1, f: Float = 1.0f, d: Double = 1.0d, s: Short = 1, l: Long = 1L, bi: BigInt = scala.math.BigInt(1), bd: BigDecimal = scala.math.BigDecimal(1.0))
+object Numeric extends ShapeTag.Companion[Numeric] {
+  val id: ShapeId = ShapeId("smithy4s.example", "Numeric")
+
+  val hints : Hints = Hints.empty
+
+  implicit val schema: Schema[Numeric] = struct(
+    int.required[Numeric]("i", _.i).addHints(smithy.api.Default(smithy4s.Document.fromDouble(1.0))),
```
daddykotex, line 23: should these all use `1.0d` ?

Baccata: yup

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
Baccata: you actually get a compiler warning if you use lowercase `L`. The rest, the compiler doesn't care. That's the reason.

modules/example/src/smithy4s/example/Numeric.scala
Baccata: https://github.com/disneystreaming/smithy4s/pull/677/commits/87f2bf0b8fe75437bb69ae28f7d9eb5439b8014e

daddykotex (approved):

PR #680, opened 2022-12-14
Compliancetests fixes improvements

yisraelU:
a few fixes for Compliance Tests module
- query params parsing to Map[String,List[String]]
- sbt config for module name
- better error reporting and handling for json compares
-  schemas undergo some hint cleansing to be compatible with the Params from the AWS tests

There seems to be a problem with at least Scala3 Native tests failing for Compliance Tests module.
I've currently disabled the running of those tests by renaming the test/src folder to src-jvm-js


modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
```diff
@@ -29,4 +33,33 @@ package object internals {
...
+      case Array(k) => (k, "")
+    }
+  }
+
+  def parseQueryParams(
+      queryParams: Option[List[String]]
+  ): ListMap[String, List[String]] = {
+    queryParams.combineAll
+      .map(splitQuery)
+      .foldRight[ListMap[String, List[String]]](ListMap.empty) {
```
kubukoz, line 57: this seems awfully close to `foldMap` using the map monoid (but it won't keep the ordering like ListMap :( )

daddykotex: TIL about ListMap

daddykotex (approved):

modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
```diff
@@ -29,4 +33,33 @@ package object internals {
       h.++(Hints(smithy.api.TimestampFormat.EPOCH_SECONDS.widen))
     )
   }
+
+  def splitQuery(queryString: String): (String, String) = {
+    queryString.split("=", 2) match {
```
daddykotex, line 38: if the queryString is empty, this must yield `Array()` and you'd get a pattern match error, is that possible?

daddykotex: Ah I guess this does not happen because we're mapping on a list of string

modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
```diff
@@ -62,4 +63,33 @@ package object internals {
...
+  // regex for comma not between quotes
+  private val commaNotBetweenQuotes = ",(?=([^\"]*\"[^\"]*\")*[^\"]*$)"
+
+  private def parseSingleHeader(
+      kv: (String, String)
+  ): List[(String, String)] = {
+    val key = kv._1
+    val values = kv._2.split(commaNotBetweenQuotes).toList
+    @tailrec
+    def loop(
```
daddykotex, line 83: could this be replaced by a `map`: `values.map(value => key -> value)` ?

yisraelU: lol , im so tired

modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
yisraelU: I need to preserve order

modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
```diff
@@ -29,4 +33,57 @@ package object internals {
       h.++(Hints(smithy.api.TimestampFormat.EPOCH_SECONDS.widen))
     )
   }
+
+  implicit class SchemaOps[A](val schema: Schema[A]) extends AnyVal {
+    def clearHints: Schema[A] =
+      schema.transformHintsTransitively(_ => Hints.empty)
+  }
+
+  def splitQuery(queryString: String): (String, String) = {
```
Baccata: make all the new methods/constructs `private[compliancetests]` please

modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
```diff
@@ -29,4 +33,57 @@ package object internals {
...
+          acc.get(k) match {
+            case Some(value) => acc + (k -> (value :+ v))
+            case None        => acc + (k -> List(v))
+          }
+      }
+  }
+
+  def extractHeaders(
+      maybeHeaders: Option[Map[String, String]]
+  ): Option[Headers] =
```
Baccata: I think you could have this one return an empty `Headers` instead of an `Option[Headers]`

Baccata (approved):

Baccata:
@daddykotex can you give this a review too please, in case I've missed something ?

modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
```diff
@@ -14,19 +14,79 @@
...
+    }
+  }
+
+  private[compliancetests] def parseQueryParams(
+      queryParams: Option[List[String]]
+  ): ListMap[String, List[String]] = {
+    queryParams.combineAll
+      .map(splitQuery)
+      .foldLeft[ListMap[String, List[String]]](ListMap.empty) {
+        case (acc, (k, v)) =>
```
daddykotex, line 70: this looks like a `groupBy(_._1)`

yisraelU: except you cannot preserve order with groupBy , hence a Map vs ListMap

daddykotex: or groupBy messes up the ordering?

yisraelU: either one , you just cant get a ListMap out of a group by and the sequential ordering provided by List is lost

daddykotex (approved):
thanks for the refactoring & fixes

PR #681, opened 2022-12-16
Add a more convenient Transformation.PartiallyApplied

kubukoz:
I expect bikeshedding over the name.

As discussed on discord, here it is!


Baccata:
I'm good with the name. I was pondering whether it should be `from` but `of` is good too imho

kubukoz:
good to go?

daddykotex (approved):

PR #684, opened 2022-12-16
Fix links to target series/0.17 blob

lewisjkl:


modules/docs/markdown/06-guides/endpoint-middleware.md
```diff
@@ -96,11 +96,11 @@ structure NotAuthorizedError {
 
 There is nothing authentication specific about this error, this is a standard smithy http error that will have a 401 status code when returned.
 
-If you want to see the full smithy model we defined above, you can do so [here](https://github.com/disneystreaming/smithy4s/blob/main/modules/guides/smithy/auth.smithy).
+If you want to see the full smithy model we defined above, you can do so [here](https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/guides/smithy/auth.smithy).
```
Baccata, line 99: we should inject it from SBT really instead of hardcoding it

lewisjkl: Updated in https://github.com/disneystreaming/smithy4s/pull/684/commits/a51f9edd83930de9341866f87680ea697aa76cb6 to use the github actions environment variables to derive the URL which should be used.

Baccata (approved):

PR #685, opened 2022-12-20
Hiding internal default fields

daddykotex:
After reading this: https://smithy.io/2.0/spec/type-refinement-traits.html#default-value-serialization

I noticed that we were rendering the defaults when we _should not_. I open as draft because:

1. I believe this also affects the client and I'm not sure it should
2. I think `Internal` and `Default` should come from the alloy `protocolDefinition` rather than added here in the `HintMask`

### DIFF
--- a/modules/json/src/smithy4s/http/json/SchemaVisitorJCodec.scala
+++ b/modules/json/src/smithy4s/http/json/SchemaVisitorJCodec.scala
@@ -23,7 +23,9 @@ import java.util
 
 import com.github.plokhotnyuk.jsoniter_scala.core.JsonReader
 import com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter
+import smithy.api.Default
 import smithy.api.HttpPayload
+import smithy.api.Internal
 import smithy.api.JsonName
 import smithy.api.TimestampFormat
 import alloy.Discriminated
@@ -1094,7 +1096,12 @@ private[smithy4s] class SchemaVisitorJCodec(
       ): (Z, JsonWriter) => Unit = {
         val codec = apply(instance)
         val jLabel = jsonLabel(field)
-        if (jLabel.forall(JsonWriter.isNonEscapedAscii)) {
+        val hasDefault = instance.hints.has[Default]
+        val hasInternal = instance.hints.has[Internal]
+        if (hasInternal && hasDefault) { (_, _) =>
+          // we hide the field
+          ()
+        } else if (jLabel.forall(JsonWriter.isNonEscapedAscii)) {
           (z: Z, out: JsonWriter) =>
             {
               out.writeNonEscapedAsciiKey(jLabel)
--- a/sampleSpecs/pizza.smithy
+++ b/sampleSpecs/pizza.smithy
@@ -7,7 +7,7 @@ use alloy#simpleRestJson
 @simpleRestJson
 service PizzaAdminService {
   version: "1.0.0",
-  operations: [AddMenuItem, GetMenu, Version, Health, HeaderEndpoint, RoundTrip, GetEnum, GetIntEnum, CustomCode, Book, Echo]
+  operations: [AddMenuItem, GetMenu, Version, Health, HeaderEndpoint, RoundTrip, GetEnum, GetIntEnum, CustomCode, Book, Echo, RenderDefaults]
 }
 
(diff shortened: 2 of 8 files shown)


sampleSpecs/pizza.smithy
```diff
@@ -336,3 +336,29 @@ structure EchoBody {
...
+list AList {
+  member: String
+}
+
+map AMap {
+  key: String
+  value: Integer
+}
+
+structure GetDefault {
```
Baccata, line 355: there should be a test exhibiting `@internal` too.

modules/json/src/smithy4s/http/json/SchemaVisitorJCodec.scala
```diff
@@ -1094,7 +1096,12 @@ private[smithy4s] class SchemaVisitorJCodec(
       ): (Z, JsonWriter) => Unit = {
         val codec = apply(instance)
         val jLabel = jsonLabel(field)
-        if (jLabel.forall(JsonWriter.isNonEscapedAscii)) {
+        val hasDefault = instance.hints.has[Default]
+        val hasInternal = instance.hints.has[Internal]
+        if (hasInternal && hasDefault) { (_, _) =>
```
Baccata, line 1101: :+1:

sampleSpecs/pizza.smithy
daddykotex: The test [here](https://github.com/disneystreaming/smithy4s/pull/685/files#diff-350b86e9f4f0b2ebf6cbc04aec8cc0d8e3f66948796bd79e499eb7a43bd3d0d5R364-R370) does that: it renders a `GetDefault` structure which has a `someOtherInternalString` member with a default with an `Internal` trait on it

The test shows that it doesn't serialize it (server side)

daddykotex:
Marking as draft until https://github.com/disneystreaming/alloy/pull/43/commits is merged and published

modules/json/src/smithy4s/http/json/SchemaVisitorJCodec.scala
```diff
@@ -1094,7 +1096,12 @@ private[smithy4s] class SchemaVisitorJCodec(
...
+          // we hide the field
+          ()
```
kubukoz, line 1103: as far as I understand, we're supposed to hide the field only if the default value would've been used. What if the following are all true:

- has internal
- has default
- _is_ not default (i.e. a different value is being serialized)

?

daddykotex: > we're supposed to hide the field only if the default value would've been used

this sounds ridiculous but also true, should I ask for clarifications?

Baccata: probably

daddykotex: Opened https://github.com/awslabs/smithy/issues/1565

daddykotex: to quote Michael:

> Yes, a server should serialize fields annotated with internal (otherwise those members wouldn't be useful for anything other than requests). There are caveats mentioned around whether to serialize defaulted values marked as internal, but those are shoulds not musts.

kubukoz: IMO we should either update the condition in question to

`if (hasInternal && hasDefault && <value>.isEmpty)`

or drop the PR altogether, thus not implementing a "should"

Baccata: At the moment, I concur with Jakub on this. Imho, "internal" should really have been an AWS specific trait, not exposed in the smithy prelude. 

@daddykotex can you formulate a compelling usecase for internal ? It may change my mind

daddykotex: no, I can't

it's really just that I saw this and realized we did not support it and maybe user would have assumed it was supported. Maybe I can amend the documentation to add a note about internal.

daddykotex:
Updated https://github.com/disneystreaming/alloy/pull/43 to introduce `default` only

PR #689, opened 2022-12-20
fix path segment parsing when suffixed with query

yisraelU:
In the scenario of a constant query (and perhaps some other cases) parsing of path segments fail.
`http://yourhost/my/uri/{foo}?hello=world` should parse a label foo, which it currently does not

### DIFF
--- a/modules/core/src/smithy4s/http/HttpEndpoint.scala
+++ b/modules/core/src/smithy4s/http/HttpEndpoint.scala
@@ -26,6 +26,9 @@ trait HttpEndpoint[I] {
 
   // Returns a path template as a list of segments, which can be constant strings or placeholders.
   def path: List[PathSegment]
+
+  // Returns a map of static query parameters that are found in the uri of Http hint, there cannot be multiple of the same key !.
+  def staticQueryParams: Map[String, String]
   def method: HttpMethod
   def code: Int
 
@@ -48,6 +51,7 @@ object HttpEndpoint {
         .get(Http)
         .toRight(HttpEndpointError("Operation doesn't have a @http trait"))
       httpMethod = HttpMethod.fromStringOrDefault(http.method.value)
+      queryParams = internals.staticQueryParams(http.uri.value)
       httpPath <- internals
         .pathSegments(http.uri.value)
         .toRight(
@@ -55,6 +59,7 @@ object HttpEndpoint {
             s"Unable to parse HTTP path template: ${http.uri.value}"
           )
         )
+
       encoder <- SchemaVisitorPathEncoder(
         endpoint.input.addHints(http)
       ).toRight(
@@ -64,6 +69,7 @@ object HttpEndpoint {
     } yield {
       new HttpEndpoint[I] {
         def path(input: I): List[String] = encoder.encode(input)
+        val staticQueryParams: Map[String,String] = queryParams
         val path: List[PathSegment] = httpPath.toList
         val method: HttpMethod = httpMethod
         val code: Int = http.code
--- a/modules/core/src/smithy4s/http/HttpEndpoint.scala
+++ b/modules/core/src/smithy4s/http/HttpEndpoint.scala
@@ -27,8 +27,8 @@ trait HttpEndpoint[I] {
   // Returns a path template as a list of segments, which can be constant strings or placeholders.
   def path: List[PathSegment]
 
-  // Returns a map of static query parameters that are found in the uri of Http hint, there cannot be multiple of the same key !.
-  def staticQueryParams: Map[String, String]
+  // Returns a map of static query parameters that are found in the uri of Http hint.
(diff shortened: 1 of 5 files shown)


modules/core/src/smithy4s/http/internals/package.scala
```diff
@@ -63,12 +63,17 @@ package object internals {
   }
 
   private def fromToString(str: String): Option[PathSegment] = {
-    if (str.isEmpty()) None
-    else if (str.startsWith("{") && str.endsWith("+}"))
-      Some(PathSegment.greedy(str.substring(1, str.length() - 2)))
-    else if (str.startsWith("{") && str.endsWith("}"))
-      Some(PathSegment.label(str.substring(1, str.length() - 1)))
-    else Some(PathSegment.static(str))
+    Option(str).map { str =>
```
daddykotex, line 66: I think you got rid of a `.isEmpty()` here

you may want to reintroduce it as a `.filter(_.nonEmpty)` on the `Option`

yisraelU: ah yes , was thinking of Null , will update

modules/core/test/src/smithy4s/http/internals/PathSpec.scala
```diff
@@ -46,7 +46,7 @@ class PathSpec() extends munit.FunSuite {
   }
 
   test("Parse path pattern into path segments") {
-    val result = pathSegments("/{head}/foo/{tail+}")
+    val result = pathSegments("/{head}/foo/{tail+}?hello=world")
```
kubukoz, line 49: imo this deserves a separate test case

Baccata: I concur

modules/core/src/smithy4s/http/HttpEndpoint.scala
```diff
@@ -26,6 +26,9 @@ trait HttpEndpoint[I] {
 
   // Returns a path template as a list of segments, which can be constant strings or placeholders.
   def path: List[PathSegment]
+
+  // Returns a map of static query parameters that are found in the uri of Http hint, there cannot be multiple of the same key !.
```
Baccata: even though that's the case at the smithy level, I'd be more comfortable with modelling it in this  interface as a multimap, to future-proof it in case it ever changes.

modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
```diff
@@ -114,8 +114,10 @@ private[http4s] class SmithyHttp4sClientEndpointImpl[F[_], Op[_, _, _, _, _], I,
   def inputToRequest(input: I): Request[F] = {
     val metadata = inputMetadataEncoder.encode(input)
     val path = httpEndpoint.path(input)
+    val staticQueries = httpEndpoint.staticQueryParams
     val uri = baseUri
       .copy(path = baseUri.path.addSegments(path.map(Uri.Path.Segment(_))))
+      .withQueryParams(staticQueries)
```
Baccata: 👍

modules/core/src/smithy4s/http/HttpEndpoint.scala
yisraelU: fair enough , Ill revert

modules/core/src/smithy4s/http/internals/package.scala
```diff
@@ -56,19 +56,43 @@ package object internals {
       str: String
   ): Option[Vector[PathSegment]] = {
     str
-      .split('/')
-      .toVector
-      .filterNot(_.isEmpty())
-      .traverse(fromToString(_))
+      .split('?')
+      .headOption
```
Baccata: nitpick : here it's safe to call `.head` instead of `.headOption`, that'll save us an allocation

modules/core/src/smithy4s/http/internals/package.scala
```diff
@@ -56,19 +56,43 @@ package object internals {
...
   }
 
   private def fromToString(str: String): Option[PathSegment] = {
-    if (str.isEmpty()) None
-    else if (str.startsWith("{") && str.endsWith("+}"))
-      Some(PathSegment.greedy(str.substring(1, str.length() - 2)))
-    else if (str.startsWith("{") && str.endsWith("}"))
-      Some(PathSegment.label(str.substring(1, str.length() - 1)))
-    else Some(PathSegment.static(str))
+    Option(str).filter(_.nonEmpty).map { str =>
```
Baccata: nitpick : here too : I'd rather have a `if (str == null || str.isEmpty)` to save an allocation

Baccata (approved):
Good stuff !

PR #691, opened 2022-12-22
DO NOT MERGE - bug in httpQueryParams trait reproduction

yisraelU:
This PR just adds additional smithy to the compliance tests module test.smithy to showcase reproduction of bug when httpQuery trait and httpQueryParams trait overlap
see serialization rules here 
https://smithy.io/2.0/spec/http-bindings.html#httpqueryparams-trait

### DIFF
--- a/sampleSpecs/test.smithy
+++ b/sampleSpecs/test.smithy
@@ -8,7 +8,7 @@ use alloy#simpleRestJson
 
 @simpleRestJson
 service HelloService {
-    operations: [SayHello, Listen, TestPath]
+    operations: [SayHello, Listen, TestPath,QueryPrecedence]
 }
 
 @http(method: "POST", uri: "/")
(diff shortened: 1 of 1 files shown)


sampleSpecs/test.smithy
```diff
@@ -194,3 +194,63 @@ structure ErrorDetails {
...
+apply QueryPrecedence @httpRequestTests([
+    {
+        id: "QueryPrecedence",
+        documentation: "Prefer named query parameters when serializing",
+        protocol: simpleRestJson,
+        method: "POST",
+        uri: "/Precedence",
+        body: "",
+        queryParams: [
+            "bar=named",
```
daddykotex, line 207: I think this should be `bar=fromMap` (atleast one test fails because of that)

or I don't understand the nature of the test?

yisraelU: @daddykotex foo is a field in a  structure annotated with httpQuery trait mapping it to the key bar

daddykotex: ahhh ok, now I understand

PR #692, opened 2022-12-22
extract query parameter embedded in url

### LINKED ISSUE #690: Metadata does not capture Constant queries
The metadata schema visitor does not capture any constant queries params and therefore when rendering the metadata it can result in an unintended url with a different interpretation
```
url = "http://yoururl/hello?number=42"
```
when appending query parameter name=bob will result in
```
url = "http://yoururl/hello?number=42?name=bob"
```
with the last question mark parsed as part of the value for number

yisraelU:
this pr adds a method to split a constant query from the last path segment and add to list of existing query params gathered by the metadata
I did not add tests , as I am working on moving theses types of tests over to utilize compliance tests module
fixes #690

### DIFF
--- a/modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
+++ b/modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
@@ -114,9 +114,17 @@ private[http4s] class SmithyHttp4sClientEndpointImpl[F[_], Op[_, _, _, _, _], I,
   def inputToRequest(input: I): Request[F] = {
     val metadata = inputMetadataEncoder.encode(input)
     val path = httpEndpoint.path(input)
+    val (extractedPath, queries) =
+      path.lastOption
+        .fold((path, Map.empty[String, Seq[String]])) { segment =>
+          val (last, queries) = splitQueryAndPath(segment)
+          (path.dropRight(1) :+ last, queries)
+        }
     val uri = baseUri
-      .copy(path = baseUri.path.addSegments(path.map(Uri.Path.Segment(_))))
-      .withMultiValueQueryParams(metadata.query)
+      .copy(path =
+        baseUri.path.addSegments(extractedPath.map(Uri.Path.Segment(_)))
+      )
+      .withMultiValueQueryParams(queries ++ metadata.query)
     val headers = toHeaders(metadata.headers)
     val baseRequest = Request[F](method, uri, headers = headers)
     if (inputHasBody) {
--- a/modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
+++ b/modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
@@ -114,12 +114,7 @@ private[http4s] class SmithyHttp4sClientEndpointImpl[F[_], Op[_, _, _, _, _], I,
   def inputToRequest(input: I): Request[F] = {
     val metadata = inputMetadataEncoder.encode(input)
     val path = httpEndpoint.path(input)
-    val (extractedPath, queries) =
-      path.lastOption
-        .fold((path, Map.empty[String, Seq[String]])) { segment =>
-          val (last, queries) = splitQueryAndPath(segment)
-          (path.dropRight(1) :+ last, queries)
-        }
+    val (extractedPath, queries) = splitPathSegmentsAndQueryParams(path)
     val uri = baseUri
       .copy(path =
         baseUri.path.addSegments(extractedPath.map(Uri.Path.Segment(_)))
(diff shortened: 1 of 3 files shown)


kubukoz:
A test would make it easier to understand what's going on 😅

kubukoz:
Besides, compliance tests are for protocol implementations and this is a utility that can be used for multiple protocols. I'd say it can be unit tested here (and probably should).

yisraelU:
Fair enough , but there didn't seem to be existing unit tests for the existing utility methods.
As far as multiple protocols making use of this semantic , well they should all have the same test at the protocol level then

yisraelU:
@kubukoz should I add more info to the issue?

Baccata:
> Fair enough , but there didn't seem to be existing unit tests for the existing utility methods. 

Corners were cut in the earlier days to get to a working POC asap, but we should very much fill the gaps when we find ourselves fixing bugs. I'd appreciate if we had unit tests added for this particular utility 😺

kubukoz:
>  As far as multiple protocols making use of this semantic, well they should all have the same test at the protocol level then

If it's inherent to how the `@http` trait is supposed to work, then I disagree.

It's also sort of like a testing pyramid, where unit tests should be the most plentiful and you wouldn't cover all the corner cases in the integration tests. As far as this bugfix goes, I can't even tell if it actually fixes anything because there's no test (here or anywhere else).

modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
```diff
@@ -114,9 +114,12 @@ private[http4s] class SmithyHttp4sClientEndpointImpl[F[_], Op[_, _, _, _, _], I,
   def inputToRequest(input: I): Request[F] = {
     val metadata = inputMetadataEncoder.encode(input)
     val path = httpEndpoint.path(input)
+    val (extractedPath, queries) = splitPathSegmentsAndQueryParams(path)
```
kubukoz, line 117: This doesn't seem to be checked by any tests

modules/http4s/src/smithy4s/http4s/package.scala
```diff
@@ -60,4 +60,31 @@ package object http4s extends Compat.Package {
       (CaseInsensitive(k.toString), v.map(_.value))
     }
 
+  private[smithy4s] def splitPathSegmentsAndQueryParams(
```
kubukoz, line 63: Suggestion: these don't seem to depend on http4s, could we move them to smithy4s.http?

modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
yisraelU: what do you mean?
is this missing something https://github.com/disneystreaming/smithy4s/pull/692/files#diff-b475f48e311abe1d8cf9ca0c675210125637631a3ac2882afdc699794002dd4a

modules/http4s/src/smithy4s/http4s/package.scala
yisraelU: we could , we would need to make it non package private or use it internally (i.e. in the endpoints interface and metadata)

modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
kubukoz: I mean this line in `inputToRequest` - we're unit testing the function all right, but the usage isn't covered (testing the input->request conversion, or a full blown client test would help)

modules/http4s/src/smithy4s/http4s/package.scala
kubukoz: yeah, without @Baccata's opinion I'd keep it here then

Baccata: Right so I think the correct approach is the following : 

First, from a semantic point of view, we're gonna set that `path` is only relative to path segments, so anything with "path" in its name should not be consuming nor returning anything after the `?` in the URI. In particular, [this method](https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/core/src/smithy4s/http/internals/package.scala#L55-L63) 

So what we need is : 

1. Change the `pathSegment` implementation to just return the `path` bits, which will trickle down to the `path` method in "httpEndpoint". The `split('?')`  should therefore happen before the `split('/')`
2. add a `def staticQueryParams : Map[String, Seq[String]]` to the `HttpEndpoint`, 
3. use that method in the http4s client endpoint implementation

This will be cleaner and will be reusable by third parties. 

Interestingly you've already done step 1 in https://github.com/disneystreaming/smithy4s/pull/689, so it's likely your two PRs are logically conflicting with one another 😄. So maybe this PR (#692) should be closed and you should resume step 2 and 3 in #689

Baccata (changes requested):
See https://github.com/disneystreaming/smithy4s/pull/692#discussion_r1059928903

modules/http4s/src/smithy4s/http4s/package.scala
yisraelU: yeah b, thats probably the case, 
I will proceed with #689  and include these tests , if everything passes , ill close this one

PR #700, opened 2023-01-03
[WIP] Add AWS Query protocol encoders support

igor-ramazanov:


igor-ramazanov:
Not ready for a full-review yet, there are some things which need to be addressed first:
1. The [FormData](https://github.com/disneystreaming/smithy4s/pull/700/files#diff-ce5d21b72a5a8d0fd8aa6502e0ea8cc5fdc7a3324aff2d2547e2dccfd7fe4e36) data structure probably needs a better representation.
2. It's not clear to me if [primitives/structures/lists in tests](https://github.com/disneystreaming/smithy4s/pull/700/files#diff-66d1e93a40b30a3f035acbc022f2b29b323c2cc6a9ed6ab334b21ec34047fbe4R42) need to be prefixed with the structure name, e.g. `x=1` or `Foo.x=42`.

igor-ramazanov:
3. Perhaps, would be good to copy tests from https://github.com/awslabs/smithy/tree/main/smithy-aws-protocol-tests/model/awsQuery

modules/aws/src/smithy4s/query/FormData.scala
```diff
@@ -0,0 +1,55 @@
+package smithy4s.aws.query
+
+final case class FormData(
```
Baccata: That's not bad at all, but I'd rather we represented it as follows (which allows to re-use some types) 

```scala
import smithy4s.PayloadPath

sealed trait FormData {
   def render: String = ??? 
   def prepend(key: String) = ???
   def prepend(index: Int) = ??? 
} 
object FormData {
    final case class SimpleValue(str: String)  extends FormData
    final case class PathedValue(path: PayloadPath, value: String) extends FormData 
    final case class MultipleValues(values: Vector[PathedValue]) extends FormData 
} 
```

The there is no look-up in the encoding side of things, so no need to use a `Map`. You may need to add `def prepend` methods to the existing `PayloadPath` type.

modules/aws/src/smithy4s/query/FormData.scala
```diff
@@ -0,0 +1,55 @@
...
+    values: Map[List[String], String],
+    primitive: Option[String]
+) {
+  lazy val render: Option[String] =
```
Baccata: this can be kept a `def`, ~and return an empty string when it holds a primitive~ and return the string value when it holds a primitive (ie, no need for `Option`)

modules/aws/src/smithy4s/query/FormData.scala
```diff
@@ -0,0 +1,55 @@
...
+    case None =>
+      FormData(
+        values.map { case (segments, value) =>
+          (segment :: segments, value)
+        },
+        None
+      )
+  }
+
+  def childOf(parent: String): FormData =
```
Baccata: Looking at the [spec](https://smithy.io/2.0/aws/protocols/aws-query-protocol.html?highlight=query%20prot#collections), this doesn't need to exist : the id of the shape is never used in the encoding.

modules/aws/src/smithy4s/query/FormData.scala
```diff
@@ -0,0 +1,55 @@
...
+    primitive: Option[String]
+) {
+  lazy val render: Option[String] =
+    primitive.orElse(
+      if (values.isEmpty) None
+      else
+        Some(
+          values
+            .map { case (segments, value) =>
+              segments.reverse.mkString(".") + "=" + value
```
Baccata: In theory you should never have to reverse, as you're always gonna be computing the fields first and prepending the name of the structure field or the union alternative

Baccata: Also, you're gonna have to URL-encode both the key and value here. You can find an encoder in `smithy4s.internals.URIEncoderDecoder`, which is package private, so you'll have to write the package of this file as : 

``` 
package smithy4s
package aws.query
```

modules/aws/test/src/smithy4s/aws/query/AwsQueryCodecSpec.scala
```diff
@@ -0,0 +1,354 @@
...
+import smithy4s.schema.CompilationCache
+import smithy4s.schema.Schema._
+import weaver._
+import smithy.api.XmlFlattened
+import smithy.api.XmlName
+
+object AwsQueryCodecSpec extends SimpleIOSuite {
+
+  implicit class SchemaOps[A](schema: Schema[A]) {
+    def named(name: String) = schema.withId(ShapeId("default", name))
```
Baccata: you should be able to remove this in these tests:  the name of the shape is never to be taken into consideration in this encoding.

modules/aws/src/smithy4s/query/AwsSchemaVisitorAwsQueryCodec.scala
```diff
@@ -0,0 +1,161 @@
...
+        }
+      })
+
+      (s: S) => encoder(field.get(s)).childOf(fieldKey)
+    }
+
+    val codecs: Vector[AwsQueryCodec[S]] =
+      fields.map(field => fieldEncoder(field))
+
+    val structKey = getKey(hints, shapeId.name)
```
Baccata: that is un-necessary (and the `childOf` call needs to be removed)

modules/aws/src/smithy4s/query/AwsSchemaVisitorAwsQueryCodec.scala
```diff
@@ -0,0 +1,161 @@
...
+      alternatives: Vector[SchemaAlt[U, _]],
+      dispatch: Alt.Dispatcher[Schema, U]
+  ): AwsQueryCodec[U] = {
+
+    def encode[A](u: U, alt: SchemaAlt[U, A]): FormData = {
+      val key = getKey(alt.hints, alt.label)
+      dispatch
+        .projector(alt)(u)
+        .fold(FormData.empty)(a => compile(alt.instance)(a))
+        .childOf(key)
```
Baccata: this call should be replaced by a call to `prepend` (see my comment [there](https://github.com/disneystreaming/smithy4s/pull/700/files#r1060512960))

modules/aws/src/smithy4s/query/AwsSchemaVisitorAwsQueryCodec.scala
```diff
@@ -0,0 +1,161 @@
...
+    val codecs: Vector[AwsQueryCodec[S]] =
+      fields.map(field => fieldEncoder(field))
+
+    val structKey = getKey(hints, shapeId.name)
+
+    (s: S) =>
+      codecs
+        .map(codec => codec(s))
+        .reduce(_ combine _)
+        .childOf(structKey)
```
Baccata: the call to `childOf` needs to be removed here

modules/aws/src/smithy4s/query/AwsSchemaVisitorAwsQueryCodec.scala
```diff
@@ -0,0 +1,161 @@
...
+            label: String,
+            instance: Schema[AA],
+            get: S => Option[AA]
+        ): AwsQueryCodec[Option[AA]] = {
+          case Some(value) => compile(instance)(value)
+          case None        => FormData.empty
+        }
+      })
+
+      (s: S) => encoder(field.get(s)).childOf(fieldKey)
```
Baccata: the `childOf` call should be replaced by a call to `prepend` (see my comment [there](https://github.com/disneystreaming/smithy4s/pull/700/files#r1060512960))

modules/aws/src/smithy4s/query/AwsSchemaVisitorAwsQueryCodec.scala
```diff
@@ -0,0 +1,161 @@
...
+  ): AwsQueryCodec[S] = {
+    def fieldEncoder[A](field: SchemaField[S, A]): AwsQueryCodec[S] = {
+      val fieldKey = getKey(field.hints, field.label)
+
+      val encoder = field.foldK(new Field.FolderK[Schema, S, AwsQueryCodec] {
+        override def onRequired[AA](
+            label: String,
+            instance: Schema[AA],
+            get: S => AA
+        ): AwsQueryCodec[AA] = (a: AA) => compile(instance)(a)
```
Baccata: So the tricky part here is that you need to pre-compile the schema instead of doing on a per-request basis. Essentially you need to do : 

``` 
val compiled = compile(instance) 
(a: AA) => compiled(a)
```

it seems the same as what you wrote, semantically, but it'll performs much better at runtime.

modules/aws/src/smithy4s/query/AwsSchemaVisitorAwsQueryCodec.scala
```diff
@@ -0,0 +1,161 @@
...
+
+        override def onOptional[AA](
+            label: String,
+            instance: Schema[AA],
+            get: S => Option[AA]
+        ): AwsQueryCodec[Option[AA]] = {
+          case Some(value) => compile(instance)(value)
```
Baccata: [ditto](https://github.com/disneystreaming/smithy4s/pull/700/files#r1060526010)

modules/aws/test/src/smithy4s/aws/query/AwsQueryCodecSpec.scala
```diff
@@ -0,0 +1,354 @@
...
+      val value = stringValue
+      val hints = Hints.empty
+    }
+    object FooBar {
+      case object Foo extends FooBar("foo", 0)
+      case object Bar extends FooBar("bar", 1)
+      implicit val schema: Schema[FooBar] =
+        enumeration[FooBar](List(Foo, Bar)).x
+    }
+    val xmlFoo = "x=foo"
```
Baccata: this test is wrong : it should be `val xmlFoo = "foo"` (same for `xmlBar`)

modules/aws/src/smithy4s/query/AwsSchemaVisitorAwsQueryCodec.scala
```diff
@@ -0,0 +1,161 @@
...
+      List(getKey(hints, "member"))
+    }
+    println(s"KEY: ${key}")
+
+    (collection: C[A]) =>
+      tag
+        .iterator(collection)
+        .zipWithIndex
+        .map { case (member, index) =>
+          val newKey = ((index + 1).toString :: key).reverse.mkString(".")
```
Baccata: here you should just be doing : 

``` 
memberWriter(member).prepend(index) 
```

there's no need to reverse, nor to call `.mkString` prematurely. The `makeString` call will happen naturally when the `.render` is called on `FormData`

modules/aws/src/smithy4s/query/AwsSchemaVisitorAwsQueryCodec.scala
```diff
@@ -0,0 +1,161 @@
...
+    val cache: CompilationCache[AwsQueryCodec]
+) extends SchemaVisitor.Cached[AwsQueryCodec] { compile =>
+
+  override def primitive[P](
+      shapeId: ShapeId,
+      hints: Hints,
+      tag: Primitive[P]
+  ): AwsQueryCodec[P] =
+    (p: P) => {
+      Primitive.stringWriter(tag, hints) match {
```
Baccata, line 34: 👍

modules/aws/src/smithy4s/query/AwsSchemaVisitorAwsQueryCodec.scala
```diff
@@ -0,0 +1,161 @@
...
+      bijection: Bijection[A, B]
+  ): AwsQueryCodec[B] = (b: B) => compile(schema)(bijection.from(b))
+
+  override def refine[A, B](
+      schema: Schema[A],
+      refinement: Refinement[A, B]
+  ): AwsQueryCodec[B] = (b: B) => compile(schema)(refinement.from(b))
+
+  override def lazily[A](suspend: Lazy[Schema[A]]): AwsQueryCodec[A] = (a: A) =>
+    suspend.map(schema => compile(schema)).value(a)
```
Baccata: So this one is tricky. For starters : [ditto](https://github.com/disneystreaming/smithy4s/pull/700/files#r1060526010) 

Then, unfortunately, it's likely not gonna work like you'd think (recursion is tricky). You can't rely on SAM-conversions here, you're gonna need need to be explicit : 

```
new AwsQueryCodec[A] {
   // we want the compilation to happen in a deferred manner, but only once. 
   lazy val underlying = compile(schema).value
   def apply(a: A) : FormData = underlying(a)  
} 
```

modules/aws/src/smithy4s/query/FormData.scala
igor-ramazanov: Done

modules/aws/src/smithy4s/query/FormData.scala
igor-ramazanov: Done

modules/aws/src/smithy4s/query/FormData.scala
igor-ramazanov: Done

modules/aws/src/smithy4s/query/FormData.scala
igor-ramazanov: Done

modules/aws/test/src/smithy4s/aws/query/AwsQueryCodecSpec.scala
igor-ramazanov: Done

modules/aws/src/smithy4s/query/AwsSchemaVisitorAwsQueryCodec.scala
igor-ramazanov: Done

modules/aws/src/smithy4s/query/AwsSchemaVisitorAwsQueryCodec.scala
igor-ramazanov: Done

modules/aws/src/smithy4s/query/AwsSchemaVisitorAwsQueryCodec.scala
igor-ramazanov: Done

modules/aws/src/smithy4s/query/AwsSchemaVisitorAwsQueryCodec.scala
igor-ramazanov: Done

modules/aws/src/smithy4s/query/AwsSchemaVisitorAwsQueryCodec.scala
igor-ramazanov: Done

modules/aws/src/smithy4s/query/AwsSchemaVisitorAwsQueryCodec.scala
igor-ramazanov: Done

modules/aws/test/src/smithy4s/aws/query/AwsQueryCodecSpec.scala
igor-ramazanov: Done

modules/aws/src/smithy4s/query/AwsSchemaVisitorAwsQueryCodec.scala
igor-ramazanov: Done

modules/aws/src/smithy4s/query/AwsSchemaVisitorAwsQueryCodec.scala
igor-ramazanov: Done

igor-ramazanov:
@Baccata Hey, thanks for the thorough review! I've addressed the comments.
Few things:
1. For some reason I have to manually strip the initial `.` for the `PayloadPath => String`: https://github.com/disneystreaming/smithy4s/pull/700/files#diff-ce5d21b72a5a8d0fd8aa6502e0ea8cc5fdc7a3324aff2d2547e2dccfd7fe4e36R38
2. Pay closer attention to the [AWS Query traits](https://smithy.io/2.0/aws/protocols/aws-query-protocol.html#supported-traits). For now, only the `XmlName` and `XmlFlattened` supported.
3. Fix the `recursive` test. I guess [this test from AWS Query spec](https://github.com/awslabs/smithy/blob/main/smithy-aws-protocol-tests/model/awsQuery/input.smithy#L287) should help me.
4. Copy tests from [the AWS spec](https://github.com/awslabs/smithy/tree/main/smithy-aws-protocol-tests/model/awsQuery)?

modules/aws/src/smithy4s/query/FormData.scala
```diff
@@ -0,0 +1,57 @@
...
+      PathedValue(PayloadPath.fromString(index.toString), str)
+  }
+  final case class PathedValue(path: PayloadPath, value: String)
+      extends FormData {
+
+    /**
+     * @todo Understand the root reason and have a better solution for a workaround removing the '.' prefix.
+     */
+    override def render: String =
+      URIEncoderDecoder.encode(path.toString.stripPrefix(".")) + "=" +
```
Baccata: the `.` exist because this Path construct is used in json decoding and I wanted to make it look more like `jq`. Really you shouldn't rely on `.toString` there, but rather deconstruct the path and fold the segments, which would prevent instantiating a string only to strip it of one char.

modules/aws/src/smithy4s/query/FormData.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  final case class SimpleValue(str: String) extends FormData {
+    override def render: String = URIEncoderDecoder.encode(str)
+
+    override def prepend(key: String): FormData = {
+      PathedValue(PayloadPath.fromString(key), str)
+    }
+
+    override def prepend(index: Int): FormData =
+      PathedValue(PayloadPath.fromString(index.toString), str)
```
Baccata: Use `PayloadPath(index)` instead

modules/aws/src/smithy4s/query/FormData.scala
```diff
@@ -0,0 +1,57 @@
...
+    override def prepend(key: String): FormData = this
+
+    override def prepend(index: Int): FormData = this
+  }
+
+  final case class SimpleValue(str: String) extends FormData {
+    override def render: String = URIEncoderDecoder.encode(str)
+
+    override def prepend(key: String): FormData = {
+      PathedValue(PayloadPath.fromString(key), str)
```
Baccata: Use `PayloadPath(key)` instead

modules/aws/test/src/smithy4s/aws/query/AwsQueryCodecSpec.scala
```diff
@@ -0,0 +1,350 @@
...
+        struct(foos)(Foo.apply)
+      }
+    }
+    val expected = "x.1=1&x.2=2&x.3=3"
+
+    checkContent(expected, Foo(List(1, 2, 3)))
+  }
+
+  test("recursion") {
+    ignore("Not ready yet")
```
Baccata: the implementation of `lazily` looks good to me. I think we need to amend the datatype used in the test to carry more information, because the protocol is lossy and the current test wouldn't show much. 

If you had a required field, it will make it naturally easier to test :  

```scala 
case class  IntList(head: Int, tail: Option[IntList]) 

val input = IntList(1, Some(IntList(2, None)))
val expected = "head=1&tail.head=2" 
```

Baccata:
> Pay closer attention to the [AWS Query traits](https://smithy.io/2.0/aws/protocols/aws-query-protocol.html#supported-traits). For now, only the XmlName and XmlFlattened supported. 

In the scope of this PR, this is absolutely fine. 

So I think what I'm gonna do is prep an `aws-query-protocol` protected feature branch that we could target in this PR and my own xml-based PRs. We'll make some progress there until we have the protocol tests wired correctly, and until some manual testing showcase that the aws signature algorithm works properly, at which point we'll perform a release. 

That way, I'll deem this PR ready to go when my latest feedback is addressed

modules/aws-kernel/src/smithy4s/aws/AwsProtocol.scala
```diff
@@ -16,16 +16,15 @@
...
+import aws.protocols.{AwsJson1_0, AwsJson1_1, AwsQuery}
 import smithy4s.Hints
 import smithy4s.ShapeTag
 
 private[aws] sealed trait AwsProtocol extends Product with Serializable {}
 
 private[aws] object AwsProtocol {
   val supportedProtocols: List[ShapeTag[_]] =
-    List(AwsJson1_0, AwsJson1_1)
+    List(AwsJson1_0, AwsJson1_1, AwsQuery)
```
kubukoz, line 27: **praise:** glad you found this :D

modules/aws/src/smithy4s/query/FormData.scala
igor-ramazanov: Done

modules/aws/src/smithy4s/query/FormData.scala
igor-ramazanov: Done

modules/aws/test/src/smithy4s/aws/query/AwsQueryCodecSpec.scala
igor-ramazanov: Worked well indeed. Thanks, done

igor-ramazanov:
@Baccata Sounds reasonable. Pushed commits addressing the last comments.

modules/aws/src/smithy4s/query/FormData.scala
igor-ramazanov: Done, perhaps not ideal.

Baccata:
@igor-ramazanov, you'll need to run `headerCreateAll` in SBT to fix the current issue.

igor-ramazanov:
@Baccata Pushed

Baccata:
awesome, I'm merging into [disneystreaming:feature/aws-query-protocol](https://github.com/disneystreaming/smithy4s/tree/feature/aws-query-protocol)

Baccata (approved):

igor-ramazanov:
@Baccata Would be nice to open a draft/wip PR from the https://github.com/disneystreaming/smithy4s/tree/feature/aws-query-protocol?rgh-link-date=2023-01-12T10%3A30%3A44Z to the release branch to track the overall visibility of the progress.

Also, what are the next steps will be? I guess, wiring up all the stuff together and polishing the last details and corner cases?

Baccata:
> Also, what are the next steps will be? I guess, wiring up all the stuff together and polishing the last details and corner cases? 

Sorry I've been busy today, but I'll create some tickets asap for you to get a good idea of what's left and get to the point of a working client asap. 

> @Baccata Would be nice to open a draft/wip PR from the https://github.com/disneystreaming/smithy4s/tree/feature/aws-query-protocol?rgh-link-date=2023-01-12T10%3A30%3A44Z to the release branch to track the overall visibility of the progress.

Yup, I'm literally signing off now but I'll do that later or tomorrow. Thanks for the patience.

PR #704, opened 2023-01-04
DO NOT MERGE-reproduction of jsonName failing in compliance tests

yisraelU:
uses sampleSpecs/test.smithy
for original test see https://github.com/awslabs/smithy/blob/51b8bafba4d8d76ff028df8d7714978a25d68d24/smithy-aws-protocol-tests/model/restJson1/json-structs.smithy#L303

### DIFF
--- a/sampleSpecs/test.smithy
+++ b/sampleSpecs/test.smithy
@@ -8,7 +8,7 @@ use alloy#simpleRestJson
 
 @simpleRestJson
 service HelloService {
-    operations: [SayHello, Listen, TestPath]
+    operations: [SayHello, Listen, TestPath,JsonNameTest]
 }
 
 @http(method: "POST", uri: "/")
@@ -74,7 +74,6 @@ structure SayHelloPayload {
     result: String
 }
 
-
 @http(method: "GET", uri: "/listen")
 @readonly
 @httpRequestTests([
@@ -109,8 +108,8 @@ operation TestPath {
 // The following shapes are used by the documentation
 @simpleRestJson
 service HelloWorldService {
-  version: "1.0.0",
-  operations: [Hello]
+    version: "1.0.0",
+    operations: [Hello]
 }
 @httpRequestTests([
     {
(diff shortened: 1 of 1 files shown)


sampleSpecs/test.smithy
```diff
@@ -194,3 +193,49 @@ structure ErrorDetails {
...
+                }""",
+        bodyMediaType: "application/json",
+        params: {
+            bar: "test"
+        }
+    }
+])
+@httpResponseTests([
+    {
+        id: "JsonNameTest",
```
Baccata: This reproduction is not helped by the fact that both tests are named the same : it's unclear which error belongs to which test, cause the implementation of compliance-tests might change the ordering accidentally.

yisraelU: fixed

sampleSpecs/test.smithy
```diff
@@ -194,3 +193,49 @@ structure ErrorDetails {
...
+        documentation: "This is a test of the jsonName trait",
+        code: 200,
+        body:
+        """
+              {
+               "bar": "test"
+                }""",
+        bodyMediaType: "application/json",
+        params: {
+            foo: "test"
```
Baccata, line 230: Alright, so what's happening is that the DocumentDecoder/Encoder actually do support `@jsonName`, but these traits should NOT be taken into consideration when parsing the `params`. 

The first test doesn't pass because the `bar` parameter ends up being overlooked, because the DocumentDecoder is looking for `foo` when it shouldn't. Therefore the value associated for `bar` is set to None, which means that the body that gets received is an empty json object, which doesn't match the expected body set by the test. 

The second test doesn't pass because it successfully decodes the input when it shouldn't, associating the `foo` param to the `bar` field of the `JsonNameTestInputOutput`, which ends up being `Some("test")`. Then, the jsoniter encoder does the right thing and encodes the `bar` field using the `foo` key, as per indicated by the `@jsonName("foo")`, but that's inconsistent with the expectation written in the test. 

So essentially, the fix would probably be to remove all hints from the schema before compiling the document decoder that is used to decode the params (but keep the timestamp-related hint that you already applied everywhere) .

Baccata: To summarise : the first test is correct but fails because of a bug in the implementation, the second test is incorrect

kubukoz: I think it's weird that `DocumentDecoder` looks at `jsonName` tbh

yisraelU: Thanks for explaining

Baccata: > I think it's weird that DocumentDecoder looks at jsonName tbh

It would break least surprise if it didn't. Moreover Document is massively useful in a lot of usecases that involve existing json payloads that couldn't be expressed without it.

PR #708, opened 2023-01-06
restructured timeout call and attemptNarrow

yisraelU:
moved ce.timeout into the input.flatmap as I suspect the app is never run when there is a payload error from decoding the input and therefore a timeout is reached without displaying the cause of the error.
There are currently no timeouts with this restructure, however i am not sure if this is the correct approach


Baccata (approved):

daddykotex (commented):
This definitely will avoid catching unwanted `PayloadError`, good catch

But it's unclear to me that it's related to the timeout, especially because we don't have a test to validate that assertion

PR #712, opened 2023-01-09
[compliance tests] addresses timeouts on the server side 

yisraelU:
which are caused by payload or metadata decoding errors which are translated into an http4s response and deferred.complete is never called.
This causes the real error to remain hidden

### DIFF
--- a/modules/compliance-tests/src/smithy4s/compliancetests/internals/ServerHttpComplianceTestCase.scala
+++ b/modules/compliance-tests/src/smithy4s/compliancetests/internals/ServerHttpComplianceTestCase.scala
@@ -102,7 +102,7 @@ private[compliancetests] class ServerHttpComplianceTestCase[
 
                   if (endpointInternal.id == endpoint.id)
                     inputDeferred.complete(in.asInstanceOf[I]) *>
-                      raiseError(new NotImplementedError)
+                      raiseError(new IntendedShortCircuit)
                   else raiseError(new Throwable("Wrong endpoint called"))
                 }
               }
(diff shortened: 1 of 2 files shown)


modules/compliance-tests/src/smithy4s/compliancetests/internals/ServerHttpComplianceTestCase.scala
```diff
@@ -110,19 +112,32 @@ private[compliancetests] class ServerHttpComplianceTestCase[
...
               server.orNotFound
                 .run(makeRequest(baseUri, testCase))
-                .attemptNarrow[NotImplementedError] *>
-                ce.timeout(inputDeferred.get, 1.second).flatMap { foundInput =>
-                  inputFromDocument
-                    .decode(testCase.params.getOrElse(Document.obj()))
-                    .liftTo[F]
-                    .map { decodedInput =>
-                      assert.eql(foundInput, decodedInput)
+                .attemptNarrow[NotImplementedError]
```
Baccata: While you're at it : it'd be good to change this to an error we control, something like 

```
case class IntendedShortCircuit() extends scala.util.control.NoStackTrace 
```

 or something. I opened your PR again this morning and was like "why are we doing this again ?" before I remembered.

Baccata:
Additionally : when you create a PR, please be mindful of the public nature of the project : people who aren't familiar with the project might open the pull-request list. 

You should be as precise in the titles of your PR as possible, and try to describe the problem in a way that random people reading may have a chance of understanding it.

modules/compliance-tests/src/smithy4s/compliancetests/internals/ServerHttpComplianceTestCase.scala
yisraelU: good idea

daddykotex (approved):
Good job

For future self, this PR is related to https://github.com/disneystreaming/smithy4s/pull/708

lewisjkl (approved):

PR #714, opened 2023-01-09
Fix ShapeId.parse not working for valid shapes

kubukoz:


modules/core/src/smithy4s/ShapeId.scala
```diff
@@ -28,8 +28,13 @@ object ShapeId extends ShapeTag.Has[ShapeId] {
...
       val segments = string.split("#")
-      if (segments.length > 1) None
-      else Some(ShapeId(segments(0), segments(1)))
+      if (
+        segments.length == 2 &&
+        segments(0).nonEmpty &&
+        segments(1).nonEmpty
+      )
+        Some(ShapeId(segments(0), segments(1)))
+      else None
```
kubukoz, line 37: **thought:** It's still a pretty simplistic implementation, long-term we might want to do something like https://github.com/disneystreaming/smithy-translate/blob/659c266a1c301440fcefdb1b90467fb0e12e3bea/modules/formatter/src/parsers/ShapeIdParser.scala, but for now I just wanted to address the immediate issue - valid shapes not parsing.

modules/core/test/src/smithy4s/ShapeIdParseSpec.scala
```diff
@@ -0,0 +1,29 @@
+package smithy4s
+import munit._
+
+class ShapeIdParseSpec extends FunSuite {
+  test("ShapeId can be parsed from valid shape IDs") {
+    val shapeId = ShapeId.parse("smithy4s#ShapeId")
+    expect.same(shapeId, Some(ShapeId("smithy4s", "ShapeId")))
+  }
+
```
kubukoz, line 9: this was failing before

Baccata (approved):

PR #715, opened 2023-01-10
Adds protocol compliance tests to the Http4s module

yisraelU:
This replaces the pizza spec with the protocol compliance tests using alloy-protocol-tests
In addition, the tests for ComplianceTests module were removed as they were not testing the module , merely using it
Will remove additional components once I confirm what is no longer necessary


modules/http4s/test/src/smithy4s/http4s/Http4sPizzaSpec.scala
```diff
@@ -1,43 +0,0 @@
...
-package smithy4s.http4s
-
-import cats.effect.IO
-import cats.effect.Resource
-import org.http4s.Uri
-import org.http4s.client.Client
-import org.http4s.implicits._
-import smithy4s.example.PizzaAdminService
-
-object Http4sPizzaSpec extends smithy4s.tests.PizzaSpec {
```
Baccata: Unless I'm mistaken, I believe there might be tests in PizzaSpec that cannot be checked using compliance tests, that have to do with the `mapError` combinator. So I'd expect this suite to exist, but I'd expect a lot of tests to be eliminated from the `PizzaSpec` trait.

yisraelU: Ok Then I will go through one by one and prune

yisraelU: @Baccata I think most of the pizza spec will still be required , not just error transformation as the tests are testing specifically the http4s implementation not the protocol, i.e. a not found path returns a 404 or certain requests produce certain responses. Perhaps we should just leave it as is rather then micro pruning it

build.sbt
```diff
@@ -645,11 +646,13 @@ lazy val http4s = projectMatrix
       else moduleName.value
     },
     Test / allowedNamespaces := Seq(
-      "smithy4s.hello"
+      "smithy4s.hello","alloy.test"
     ),
     Test / smithySpecs := Seq(
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "hello.smithy"
     ),
+     Test / smithy4sSkip := Seq("openapi"),
```
kubukoz: **issue:** this file needs formatting

project/Dependencies.scala
```diff
@@ -26,9 +26,10 @@ object Dependencies {
 
   val Alloy = new {
     val org = "com.disneystreaming.alloy"
-    val version = "0.1.2"
+    val version = "0.1.8"
     val core = org % "alloy-core" % version
     val openapi = org %% "alloy-openapi" % version
+    val `protocol-tests` = org % "alloy-protocol-tests" % version
```
kubukoz: **suggestion:** I'd just call it `protocolTests` to avoid the backticks.

modules/http4s/test/src/smithy4s/http4s/Http4sPizzaSpec.scala
kubukoz: > testing specifically the http4s implementation not the protocol

isn't that what all the protocol tests do? Test the interpreter for conforming to the protocol.

yisraelU: no, specifically aws protocol tests , test Smithy-Model -> HTTP and vise versa , not HTTP -> HTTP

Baccata: Here are the tests I think can be removed in favour of smithy-defined tests : 

* https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/tests/src/smithy4s/tests/PizzaSpec.scala#L405-L433
* https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/tests/src/smithy4s/tests/PizzaSpec.scala#L70-L83
* https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/tests/src/smithy4s/tests/PizzaSpec.scala#L85-L117
* https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/tests/src/smithy4s/tests/PizzaSpec.scala#L141-L161
* https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/tests/src/smithy4s/tests/PizzaSpec.scala#L335-L357
* https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/tests/src/smithy4s/tests/PizzaSpec.scala#L384-L403
* https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/tests/src/smithy4s/tests/PizzaSpec.scala#L510-L543

As they test the implementation of the protocol. The fact that they might be testing it differently to what we'd be doing in smithy-written tests obviously doesn't matter that much.

yisraelU:
Current failure is due to scala 3 native failing when running the Protocol Compliance Tests , which is as of yet an unsolved problem see #680

Baccata:
@yisraelU the problem is that the regex you used [there](https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala#L85-L86) is probably not working in native c/c++.

I suck at regexes, but I suggest you either go to the scala discord, channel scala-native, and ask for help to port the regex in a way that will work on JVM and Native alike, or use an online tool like https://regex101.com/ to experiment in both java and go to find a regex that will work in both. If it works in go chances are it'll work in scala-native too.

Baccata:
You can also experiment with the following scala-cli script to test regexes on native : 

```scala
//> using platform "native"

@main
def main() = {
  val commaNotBetweenQuotes = ",(?=([^\"]*\"[^\"]*\")*[^\"]*$)"
  """foo""".split(commaNotBetweenQuotes)
}
```

you'll see that in the current state it throws

yisraelU:
@Baccata great , will do

modules/http4s/test/src/compliancetests/ProtocolComplianceTest.scala
```diff
@@ -22,11 +22,11 @@ import org.http4s._
 import org.http4s.client.Client
 import smithy4s.kinds.FunctorAlgebra
 import smithy4s.Service
-import smithy4s.example.test._
```
kubukoz, line 25: idk about everyone else, but my github gave up trying to syntax highlight this file.

modules/tests/src/smithy4s/tests/PizzaSpec.scala
```diff
@@ -67,55 +61,6 @@ abstract class PizzaSpec
     "price" -> Json.fromFloatOrNull(9.0f)
   )
 
-  val positiveEmptyLabel = """|Happy path:
```
kubukoz, line 70: Are there any tests in PizzaClientSpec that we could similarly remove?

yisraelU:
replaced the regex with a foldLeft to parse multi headers

modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
```diff
@@ -82,15 +82,33 @@ package object internals {
...
+            (acc, begin + 1, end + 1, quote)
+          case _ =>
+            if (s.length == end + 1)
+              (acc :+ s.substring(begin, end + 1), end + 1, end + 1, quote)
+            else {
+              (acc, begin, end + 1, quote)
+            }
+        }
+    }._1
+      .toList
```
lewisjkl: It's a bit unclear to me what some of these cases are doing.. I might just be dense, but I think some comment  explaining more what edge cases you are handling here would be helpful

yisraelU: you are definitely not dense.
Ill add clear (hopefully) explanations

lewisjkl: Looks good, thanks for adding 👍

yisraelU:
Will add tests for parseList function tomorrow

modules/compliance-tests/test/src/ComplianceUtilsSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.compliancetests
+
+import smithy4s.compliancetests.internals.parseList
+import weaver._
+object ComplianceUtilsSpec extends FunSuite {
```
Baccata, line 21: this is great, thank you :)

Baccata (approved):

PR #716, opened 2023-01-11
Remove unsupported resource note

kubukoz:


Baccata:
I think it'd be good to have a description of how resources are handled, rather than just removing it altogether

kubukoz:
Fair point. Will amend

zetashift:
How about something like:

```
Smithy has a `resource` type of shape, that represents CRUD specialized services. Smithy4s only transitively transfers the operations defined in a `resource` to the service that lists that `resource`.
```

Baccata:
I like it 😄

Baccata:
Closing in favour of https://github.com/disneystreaming/smithy4s/pull/825. Thanks @zetashift !

PR #719, opened 2023-01-12
Add documentation post about protocols

lewisjkl:
This is a blog post that is intended to discuss what protocols are and build up a general understanding for what `alloy#simpleRestJson` and the like are.

I did my best to research my points here for accuracy and precise language, but please leave feedback if you see inaccuracies or things that need general clarification.


modules/docs/markdown/03-protocols/02-definition.md
```diff
@@ -0,0 +1,106 @@
...
+
+Now that we have covered the basis of what HTTP defines, it is important to call out a few things that it does not define.
+
+The most major thing left out of the HTTP specification is the format for data in the body of requests and responses. As far as HTTP is concerned, these bodies are just Bytes. Defining what these bodies look like is left to higher-level protocols (as we will discuss below).
+
+HTTP also does not define what types of resources requests should be able to work with. This means requests can be used to get or modify files stored in a system, in memory state, a third party API, or anything else one could imagine.
+
+Additionally, although HTTP provides some mechanisms for basic authentication, it does not define anything around authorization. Other protocols, such as OAuth 2.0 are often used with HTTP applications.
+
+#### REST vs HTTP
```
Baccata, line 66: I'm putting this comment here in a semi-arbitrary fashion, but I think we're missing a transition section after `things not defined in http` that we'd call `protocols in the level of smithy`, where we'd exemplify it concretely : 

"a protocol is a set of concrete rules that define how data modelled in smithy get transcribed into lower-level semantics. It can be accompanied by a set of annotations that allow to capture the details of that transcription that cannot be inferred from the shape themselves, or to customise that transcription" 

then give a concrete example of the @http annotation for instance.

modules/docs/markdown/03-protocols/02-definition.md
```diff
@@ -0,0 +1,106 @@
...
+
+Finally, at the top layer we have the protocol that we are all familiar with: http. Other examples of application-layer protocols are FTP, SSH, and SMTP. All of these protocols define their own procedures for interactions between host applications. Essentially, they are adding more structure to interactions that use TCP or UDP underneath. HTTP, for example, uses TCP to facilitate connections.
+
+#### Summary
+
+In summary, TCP/IP is a collection of protocols, grouped into layers, that make up the basis for interactions between computers.
+
+## HTTP
+
+Before we move on, let's make sure we have a common understanding of what HTTP actually is. HTTP stands for HyperText Transfer Protocol. As mentioned above, it builds upon TCP to send messages between two hosts. The most important thing for us to understand here is what HTTP defines and what is does not define.
```
daddykotex:
```suggestion
Before we move on, let's make sure we have a common understanding of what HTTP actually is. HTTP stands for HyperText Transfer Protocol. As mentioned above, it builds upon TCP to send messages between two hosts. The most important thing for us to understand here is what HTTP defines and what it does not define.
```

modules/docs/markdown/03-protocols/02-definition.md
```diff
@@ -0,0 +1,106 @@
...
+
+#### alloy#simpleRestJson
+
+The main protocol that we have defined for use in smithy4s is `alloy#simpleRestJson`. This protocol uses RESTful HTTP semantics and has HTTP bodies that are encoded as JSON. Essentially, this protocol is building on the various protocols we defined above, and just adding some constraints around how they are used. There are more constraints around how the JSON is encoded and a few other things which you can read about in full in the [alloy documentation](https://github.com/disneystreaming/alloy#alloysimplerestjson).
+
+#### gRPC
+
+gRPC is a framework for remote procedure calls. At the same time, we can consider it to be another protocol that is defined at the same level that `alloy#simpleRestJson` is defined. GRPC, in its current implementation, uses HTTP version 2 under the hood. It has additional definition around what the HTTP interactions look like such as the bodies being encoded with protocol buffers (protobuf). Additionally, all gRPC messages are sent with the HTTP POST method and use a specific header to specify which remote procedure is being invoked.
```
daddykotex:
```suggestion
gRPC is a framework for remote procedure calls. At the same time, we can consider it to be another protocol that is defined at the same level that `alloy#simpleRestJson` is defined. gRPC, in its current implementation, uses HTTP version 2 under the hood. It has additional definition around what the HTTP interactions look like such as the bodies being encoded with protocol buffers (protobuf). Additionally, all gRPC messages are sent with the HTTP POST method and use a specific header to specify which remote procedure is being invoked.
```
minor case inconsistency

modules/docs/markdown/03-protocols/02-definition.md
```diff
@@ -0,0 +1,106 @@
...
+@protocolDefinition()
+@trait(selector: "service")
+structure simpleRestJson {}
+```
+
+This then allows you to annotate service shapes with the protocol. Although the protocol is defined in Smithy like this, the exact meaning of the protocol is defined elsewhere in documentation and more importantly in interpreters that implement the protocol (such as the interpreters that smithy4s uses to create http4s applications). Additionally, protocol definitions typically come with test cases, also [defined in Smithy](https://smithy.io/2.0/additional-specs/http-protocol-compliance-tests.html), which help to define the specific semantics of the protocol.
+
+## Conclusion
+
+Hopefully this article has helped to outline what a protocol is and more specifically what layers of protocols we are building upon to create something like the `alloy#simpleRestJson` protocol. If you have any more questions about protocols, especially the protocols we have defined and use in Smithy4s, don't hesitate to reach out.
```
daddykotex: Do we need a comma here: `Hopefully,` ?

daddykotex (commented):
Great job, it is really accessible

modules/docs/markdown/03-protocols/02-definition.md
lewisjkl: Addressed in https://github.com/disneystreaming/smithy4s/pull/719/commits/7116e1fa736765210ed008ec027a9109ed6d61a7

Baccata (approved):
Awesome stuff !

Issue #720, opened 2023-01-13
[AWS] Create `AwsQueryInterpreter` 


Baccata:
**This work should target the `feature/aws-query-protocol` branch**

 Same conceptual level as [here](https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/aws/src/smithy4s/aws/internals/AwsJsonRPCInterpreter.scala)
 
The difference is that it'll amend the `Endpoint` (by creating a new instance of one) before passing it to the `AwsUnaryEndpoint`. The amendment will transform the schemas to add hints that will let the CodecAPI access additional information that's required during encoding.

#### Amendments to the input schema of the endpoint : 

The Input schema of the endpoint should be amended by the name the operation (accessible from its shapeId)  and the version property of the service (accessible from the smithy4s.Service interface). 

Hints carrying this information will have to be created, which can be done directly in Scala code (no need to create them in Smithy). Have a look at `InputOutput` to see a manually created hint. 

#### Amendments necessary to the AwsQueryCodecAPI

The `AwsQueryCodecAPI` will have to distinguish between when it's called for the input or for the output, in order to create a form-data encoder, or an xml decoder. This distinction can be achieved by inspect the hints of the Schema it receives for the `InputOutput` enumeration. 

This does imply that the `Codec` type member will have to change to carry something like `Either[FormDataEncoder, XmlDecoder]`. It'll be okay to throw an error if we're asked to decode something when we have a Left, or asked to encode something with a Right : this codecAPI should never be accessed by users directly and should very much move to an internal package and become package-private. 

When serialising inputs, this information will need to be added to the FormData before rendering : See https://smithy.io/2.0/aws/protocols/aws-query-protocol.html#request-serialization

#### Regarding the wiring the the Xml decoding 

Regarding the invocation of the XmlDecoder in the `CodecAPI`, you can have a look at [this](https://github.com/disneystreaming/smithy4s/blob/feature/aws-query-protocol/modules/xml/test/src/smithy4s/xml/XmlCodecSpec.scala#L565-L574). You'll have to create a stream from a `Chunk[Byte]` wrapping the `Array[Byte]` you receive from the CodecAPI

Baccata:
@igor-ramazanov, do you think you'll be able to pick this ?

igor-ramazanov:
@Baccata Hi, yes.

igor-ramazanov:
@Baccata 

[...]

PR #723, opened 2023-01-13
[WIP] Wiring up parts of AWS Query protocol

igor-ramazanov:
#720 

A WIP PR to combine different parts of AWS Query protocol support into a fully working implementation.

### DIFF
--- a/modules/aws/src/smithy4s/aws/internals/AwsQueryRPCInterpreter.scala
+++ b/modules/aws/src/smithy4s/aws/internals/AwsQueryRPCInterpreter.scala
@@ -1,0 +1,84 @@
+/*
+ *  Copyright 2021-2022 Disney Streaming
+ *
+ *  Licensed under the Tomorrow Open Source Technology License, Version 1.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     https://disneystreaming.github.io/TOST-1.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.aws
+package internals
+
+import cats.MonadThrow
+import smithy4s.{Endpoint, Hints, Schema, ShapeId, StreamingSchema}
+import smithy4s.http.CodecAPI
+import smithy4s.kinds._
+
+// format: off
+/**
+ * An interpreter for unary operations in the AWS_QUERY protocol
+ */
+private[aws] class AwsQueryRPCInterpreter[Alg[_[_, _, _, _, _]], Op[_,_,_,_,_], F[_]](
+    service: smithy4s.Service.Aux[Alg, Op],
+    endpointPrefix: String,
+    awsEnv: AwsEnvironment[F],
+    contentType: String,
+    codecAPI: CodecAPI
+)(implicit F: MonadThrow[F])
+    extends PolyFunction5[Op, AwsCall[F, *, *, *, *, *]] {
+// format: on
+
+  def apply[I, E, O, SI, SO](
+      op: Op[I, E, O, SI, SO]
+  ): AwsCall[F, I, E, O, SI, SO] = {
+    val (input, endpoint) = service.endpoint(op)
+    awsEndpoints(amendEndpoint(endpoint)).toAwsCall(input)
+  }
+
+  private def amendEndpoint[I, E, O, SI, SO](
+      endpoint: Endpoint[Op, I, E, O, SI, SO]
+  ): Endpoint[Op, I, E, O, SI, SO] = {
+    new Endpoint[Op, I, E, O, SI, SO] {
+      def id: ShapeId = endpoint.id
+      def input: Schema[I] =
+        // TODO: Pass a service id and version
+        endpoint.input.addHints(smithy4s.internals.InputOutput.Input.widen)
+      def output: Schema[O] =
+        endpoint.output.addHints(smithy4s.internals.InputOutput.Output.widen)
+      def streamedInput: StreamingSchema[SI] = endpoint.streamedInput
+      def streamedOutput: StreamingSchema[SO] = endpoint.streamedOutput
+      def hints: Hints = endpoint.hints
+      def wrap(input: I): Op[I, E, O, SI, SO] = endpoint.wrap(input)
+    }
+  }
+
+  private val signer: AwsSigner[F] = AwsSigner.rpcSigner[F](
+    service.id,
+    endpointPrefix,
+    awsEnv,
+    contentType
+  )
+
+  private val awsEndpoints =
+    new PolyFunction5[
+      Endpoint[Op, *, *, *, *, *],
+      AwsUnaryEndpoint[F, Op, *, *, *, *, *]
+    ] {
+      def apply[I, E, O, SI, SO](
+          endpoint: Endpoint[Op, I, E, O, SI, SO]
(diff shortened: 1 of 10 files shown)


modules/aws/src/smithy4s/aws/internals/AwsQueryRPCInterpreter.scala
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  private def amendEndpoint[I, E, O, SI, SO](
+      endpoint: Endpoint[Op, I, E, O, SI, SO]
+  ): Endpoint[Op, I, E, O, SI, SO] = {
+    new Endpoint[Op, I, E, O, SI, SO] {
+      def id: ShapeId = endpoint.id
+      def input: Schema[I] =
+        // TODO: Pass a service id and version
+        endpoint.input.addHints(smithy4s.internals.InputOutput.Input.widen)
```
igor-ramazanov: I'll have to pass a service id and version over here.
Will need to create a new hint very similar to `InputOutput`, except that `Input` part will be like a `case class Input(serviceId: String, version: String)`.

modules/aws/src/smithy4s/query/AwsQueryCodecAPI.scala
```diff
@@ -16,47 +16,91 @@
...
+  override def createCache(): Cache = CompilationCache.make[AwsQueryCodec]
 
   override def compileCodec[A](
       schema: Schema[A],
       cache: Cache
-  ): Codec[A] = schema.compile(new AwsSchemaVisitorAwsQueryCodec(cache))
+  ): Codec[A] =
+    schema.hints match {
+      case InputOutput.hint(InputOutput.Input) =>
+        Left(schema.compile(new AwsSchemaVisitorAwsQueryCodec(cache)))
```
igor-ramazanov: Here I'll have to use the `InputOutput.Input(serviceId, version)` somehow.

modules/aws/src/smithy4s/query/AwsQueryCodecAPI.scala
```diff
@@ -16,47 +16,91 @@
...
+        .through(documents[SyncIO, XmlDocument])
+        .map(doc => XmlCursor.fromDocument(doc))
+        .map(cursor => xmlDecoder.decode(cursor))
+        .rethrow
+        .head
+        .compile
+        .lastOrError
+        .map(a =>
+          // TODO: have no idea if it's correct or not
+          BodyPartial(Map.empty.withDefaultValue(a))
```
igor-ramazanov: I don't understand this part. What kind of a `Map[String, A]` should be created here for an `a: A`?

modules/aws/src/smithy4s/query/AwsQueryCodecAPI.scala
```diff
@@ -16,47 +16,91 @@
...
+          )
+        )
+  }
 
   override def decodeFromByteBufferPartial[A](
       codec: Codec[A],
       bytes: ByteBuffer
-  ): Either[PayloadError, BodyPartial[A]] = decodingError
+  ): Either[PayloadError, BodyPartial[A]] =
+    decodeFromByteArrayPartial(codec, bytes.array())
```
igor-ramazanov: Is it safe to reuse the another method?

modules/aws/src/smithy4s/query/AwsQueryCodecAPI.scala
```diff
@@ -16,47 +16,91 @@
...
-  override def writeToArray[A](
-      codec: Codec[A],
-      value: A
-  ): Array[Byte] =
-    codec(value).render.getBytes("UTF-8")
+  override def writeToArray[A](codec: Codec[A], value: A): Array[Byte] =
+    codec match {
+      case Left(encoder) => encoder(value).render.getBytes("UTF-8")
+      case Right(_) =>
+        throw new RuntimeException("Boom! TODO: define a proper error")
```
igor-ramazanov: Is it safe to simply throw out an error over here?

igor-ramazanov (commented):
@Baccata Hi, here's an initial sketch of wiring up the parts together. Could you please take a look and review it?

modules/aws/src/smithy4s/aws/internals/AwsQueryRPCInterpreter.scala
igor-ramazanov: I'm not 100% sure if I'm on the right track with the `Input(serviceId: String, version: String` implementation.

Baccata: It'd be more like : 

```scala
QueryOperation(operationName: String, version: String) // operationName as opposed to serviceId
```

modules/aws/src/smithy4s/query/AwsQueryCodecAPI.scala
Baccata: Right, so the `partial` stuff is there because in some protocols, only parts of the data come from the http body. The rest may come from elsewhere,  http headers, etc. So the `BodyPartial` aims at capturing this partial value. 

Here, we're decoding the totality of the of the body from the payload, so you can return `BodyPartial(_ => a)`, which means that data that comes from elsewhere is not needed for the decoding here.

modules/aws/src/smithy4s/query/AwsQueryCodecAPI.scala
Baccata: yeah. You'll have to modify the QueryCodec to add the static information you get from the hint you created.

modules/aws/src/smithy4s/query/AwsQueryCodecAPI.scala
Baccata: yeah that will do ... byte-buffer methods are there to help other backends. For the AWS stuff, we're only supporting http4s anyway, and we're getting byte-arrays from fs2 chunks. You could return even do a `sys.error("Coding error")` there and it'd be fine.

modules/aws/src/smithy4s/query/AwsQueryCodecAPI.scala
Baccata: yeah.

modules/aws/src/smithy4s/query/AwsQueryCodecAPI.scala
igor-ramazanov: Done

modules/aws/src/smithy4s/query/AwsQueryCodecAPI.scala
igor-ramazanov: Done

modules/aws/src/smithy4s/query/AwsQueryCodecAPI.scala
igor-ramazanov: Done

modules/aws/src/smithy4s/aws/internals/AwsQueryRPCInterpreter.scala
igor-ramazanov: Done

modules/aws/src/smithy4s/query/AwsQueryCodecAPI.scala
igor-ramazanov: Done

igor-ramazanov:
@Baccata I've addressed the comments, please, take a look.

igor-ramazanov:
FYI: this a good one to test: https://raw.githubusercontent.com/aws/aws-sdk-js-v3/main/codegen/sdk-codegen/aws-models/cloudsearch.json

igor-ramazanov:
Tested with the AWS CloudWatch and it works!

igor-ramazanov:
Ok, apparently, there's a problem either with the wiring up code or decoding, because I can see the outgoing requests and incoming responses, but it decodes the responses into empty case classes, i.e. with all fields set to `None`.

Baccata:
Closing in favour of https://github.com/disneystreaming/smithy4s/pull/729

PR #725, opened 2023-01-13
Fix decoding an optional `Document.DNull` field

### LINKED ISSUE #697: Decoding an optional field fails when receiving a Document.DNull
I noticed that the DocumentDecoderSchemaVisitor will produce a decoder that always fails on DNull,
As null is both a valid json value and smithy value, shouldnt this be handled ?

zetashift:
Fixes: https://github.com/disneystreaming/smithy4s/issues/697

Test was made from kubukoz's reproduction example:
```scala
//> using scala "3.2.1"
//> using lib "com.disneystreaming.smithy4s::smithy4s-core:0.17.2"
import smithy4s.schema.Schema
import smithy4s.Document

@main def demo = println {
  val optionalFieldSchema =
    Schema
      .struct[Option[String]](Schema.string.optional[Option[String]]("test", identity))(identity)

  Document
    .Decoder
    .fromSchema(optionalFieldSchema)
    .decode(Document.obj("test" -> Document.DNull))
}
```

### DIFF
--- a/modules/core/src/smithy4s/internals/DocumentDecoderSchemaVisitor.scala
+++ b/modules/core/src/smithy4s/internals/DocumentDecoderSchemaVisitor.scala
@@ -311,7 +311,6 @@ class DocumentDecoderSchemaVisitor(
         Map[String, Document]
     ) => Unit = {
       val jLabel = jsonLabel(field)
-
       val maybeDefault = field.instance.getDefault
 
       if (field.isOptional) {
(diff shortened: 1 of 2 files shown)


modules/core/src/smithy4s/internals/DocumentDecoderSchemaVisitor.scala
```diff
@@ -323,6 +322,7 @@ class DocumentDecoderSchemaVisitor(
           val path = PayloadPath.Segment(jLabel) :: pp
           fields
             .get(jLabel) match {
+            case Some(DNull) => buffer(None)
```
kubukoz, line 325: **issue:** this is exactly what I was thinking, how dare you steal my ideas :P `/s`

kubukoz (approved):

kubukoz:
Thanks for the contribution @zetashift!

modules/core/src/smithy4s/internals/DocumentDecoderSchemaVisitor.scala
zetashift: Haha, great minds think alike? 👀

PR #727, opened 2023-01-13
Add mqtt module

alexcardell:
Expose mqtt traits for code generation

My plan is to implement a rabbitmq backend with https://github.com/itv/bucky, likely published as a separate module

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -69,7 +69,8 @@ lazy val allModules = Seq(
   dynamic,
   testUtils,
   guides,
-  complianceTests
+  complianceTests,
+  mqtt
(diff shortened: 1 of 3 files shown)


kubukoz:
context: https://github.com/disneystreaming/smithy4s/discussions/726

kubukoz:
hey @alexcardell, I was a bit too quick to suggest adding this here, I think for the time being the better thing to do will be to help you generate these traits in your project. Sorry for the confusion.

We can consider doing it here later, if there's a concrete reason why it can't be done in userland. Let's work in #726 to solve your problem.

alexcardell:
@kubukoz No worries!

Issue #496, opened 2022-10-06
Adding -source:future causes compilation errors due to deprecated wildcard syntax


scala-impala:
On Scala 3.2.0, adding the "-source:future" compiler option causes compilation to fail with the following error message:
`_ is deprecated for wildcard arguments of types: use ? instead`

Baccata:
As a workaround, you could probably use a multi-module build and code-generate in a separate module, so that your application code could still use `-source:future`. 

We'll probably provide a pathway to opt-in `?` syntax eventually, but truth be told, it's gonna be pretty low on the priority list. If you want to contribute, we could possibly offer some guidance.

kubukoz:
You can disable warnings in generated sources. Might be a good idea because at some point smithy4s might start generating `@deprecated` annotations on methods/types that have the `@deprecated` trait, for example.

I'm using something like this in one of the projects:

`scalacOptions += "-Wconf:src=src_managed/.*:silent"`

scala-impala:
@Baccata I would love to contribute (smithy4s is awesome!) but would definitely need some guidance as I don't have any OSS experience.

kubukoz:
The biggest problem right now would probably be to figure out how the users would configure this... but as long as you can parameterize the rendere with _something_, we can probably figure that out later - you could start a PR that hardcodes the renderer to use the Scala 3 syntax, and we could pick it up later to pass that all the way from whatever is going to set that. I'm thinking codegen-cli / sbt-plugin / mill-plugin settings.

That being unclear, here are some things that are clear atm and should be useful if you want to give this a shot:

---

1. Where's the code

In general, you're gonna work in the `codegen` module. 

Generating code roughly starts in the `Codegen` type (you shouldn't have to deal with it much). It runs in basically two phases:

- Smithy model -> IR (intermediate representation), called SmithyToIR and
- IR -> String, called `Renderer`.
 
The exact look of the generated code is only a concern of the latter, which lives in `object Renderer` (`codegen` module).

You can look around and draw some inspiration from https://github.com/disneystreaming/smithy4s/pull/599 which touches a couple areas in the renderer.

2. How to check that it works

The testing ability is quite limited as of now, as testing is mostly done by either:

- we have some sources in `modules/example/src` that are generated (from sources in `sampleSpecs/*.smithy`) and checked-in. The CI setup checks if these are up-to-date after everything has been generated, compiled and tested
[...]

PR #731, opened 2023-01-14
Generate scaladoc for `@documentation` trait

zetashift:
Attempt at resolving: https://github.com/disneystreaming/smithy4s/issues/256
@kubukoz's feedback loop tips in https://github.com/disneystreaming/smithy4s/issues/496 were very very helpful!
Still a draft, because I am not sure if I am following the logic of `Renderer.scala` correct everywhere.
However I was quite happy seeing this:
image

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/internals/IR.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/IR.scala
@@ -250,7 +250,7 @@ private[internals] object Hint {
   case class Constraint(tr: Type.Ref, native: Native) extends Hint
   case class Protocol(traits: List[Type.Ref]) extends Hint
   case class Default(typedNode: Fix[TypedNode]) extends Hint
-  case class Documentation(docString: String) extends Hint
+  case class Documentation(docString: String, memberDocs: Map[String, String]) extends Hint
   case class Deprecated(message: Option[String], since: Option[String])
       extends Hint
   // traits that get rendered generically
--- a/modules/codegen/src/smithy4s/codegen/internals/IR.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/IR.scala
@@ -250,8 +250,10 @@ private[internals] object Hint {
   case class Constraint(tr: Type.Ref, native: Native) extends Hint
   case class Protocol(traits: List[Type.Ref]) extends Hint
   case class Default(typedNode: Fix[TypedNode]) extends Hint
-  case class Documentation(docString: String, memberDocs: Map[String, String])
-      extends Hint
+  case class Documentation(
+      docString: Option[String],
+      memberDocs: Option[Map[String, String]]
+  ) extends Hint
   case class Deprecated(message: Option[String], since: Option[String])
       extends Hint
   // traits that get rendered generically
--- a/modules/codegen/src/smithy4s/codegen/internals/IR.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/IR.scala
@@ -250,7 +250,8 @@ private[internals] object Hint {
   case class Constraint(tr: Type.Ref, native: Native) extends Hint
   case class Protocol(traits: List[Type.Ref]) extends Hint
   case class Default(typedNode: Fix[TypedNode]) extends Hint
-  case class Documentation(docString: String, memberDocs: Map[String, String]) extends Hint
+  case class Documentation(docString: String, memberDocs: Map[String, String])
+      extends Hint
   case class Deprecated(message: Option[String], since: Option[String])
       extends Hint
   // traits that get rendered generically
--- a/modules/codegen/src/smithy4s/codegen/internals/IR.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/IR.scala
@@ -250,6 +250,7 @@ private[internals] object Hint {
   case class Constraint(tr: Type.Ref, native: Native) extends Hint
   case class Protocol(traits: List[Type.Ref]) extends Hint
   case class Default(typedNode: Fix[TypedNode]) extends Hint
+  case class Documentation(docString: String) extends Hint
   case class Deprecated(message: Option[String], since: Option[String])
       extends Hint
   // traits that get rendered generically
--- a/modules/codegen/src/smithy4s/codegen/internals/IR.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/IR.scala
@@ -251,8 +251,8 @@ private[internals] object Hint {
   case class Protocol(traits: List[Type.Ref]) extends Hint
   case class Default(typedNode: Fix[TypedNode]) extends Hint
   case class Documentation(
-      docString: Option[String],
-      memberDocs: Option[Map[String, String]]
+      docLines: List[String],
+      memberDocLines: Map[String, List[String]]
   ) extends Hint
   case class Deprecated(message: Option[String], since: Option[String])
       extends Hint
--- a/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
@@ -32,6 +32,7 @@ import scala.jdk.CollectionConverters._
 import Line._
 import LineSyntax.LineInterpolator
 import ToLines.lineToLines
+import scala.annotation.tailrec
 
 private[internals] object Renderer {
 
@@ -70,7 +71,6 @@ private[internals] object Renderer {
         .flatMap(_.segments.toList)
         .groupBy {
           // we need to compare NameRefs as they would be imported in order to avoid unnecessarily qualifying types in the same package
-          //
           case ref: NameRef => ref.asImport
           case other        => other
         }
@@ -164,11 +164,33 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
       }
   }
 
-  private def documentationAnnotation(hints: List[Hint]): Line = {
+  /**
+    * Returns the given list of Smithy documentation strings formatted as Scaladoc comments
+    *
+    * @param l splitted docstring from the IR
+    * @return formatted list of scaladoc lines
+    */
+  private def makeDocLines(l: List[String]): List[Line] = {
+    @tailrec
+    def loop(l: List[String], acc: List[Line]): List[Line] = {
+      l match {
+        case Nil                  => acc
+        case _ :: tl if tl == Nil => loop(tl, acc :+ line"  */")
+        case hd :: tl             => loop(tl, acc :+ line"  * $hd")
+      }
+    }
+    loop(l.tail, List(line"/** ${l.head}"))
+  }
+
+  private def documentationAnnotation(hints: List[Hint]): Lines = {
     hints
       .collectFirst { case h: Hint.Documentation => h }
       .foldMap { doc =>
-        line"/** ${doc.docString} */"
+        val lines = doc.docString.split(System.lineSeparator()).toList
+        val scalaDoc =
+          if (lines.length > 1) makeDocLines(lines)
+          else List(line"/** ${lines.head} */")
+        Lines(scalaDoc)
       }
   }
 
--- a/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
@@ -164,11 +164,20 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
       }
   }
 
+  private def documentationAnnotation(hints: List[Hint]): Line = {
+    hints
+      .collectFirst { case h: Hint.Documentation => h }
+      .foldMap { doc =>
+        line"/** ${doc.docString} */"
+      }
+  }
+
   def renderPackageContents: Lines = {
     val typeAliases = compilationUnit.declarations.collect {
       case TypeAlias(_, name, _, _, _, hints) =>
         lines(
           deprecationAnnotation(hints),
+          documentationAnnotation(hints),
           line"type $name = ${compilationUnit.namespace}.${name}.Type"
         )
     }
@@ -220,12 +229,14 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
 
     lines(
       deprecationAnnotation(hints),
+      documentationAnnotation(hints),
       block(line"trait $genName[F[_, _, _, _, _]]")(
         line"self =>",
         newline,
         ops.map { op =>
           lines(
             deprecationAnnotation(op.hints),
+            documentationAnnotation(op.hints),
             line"def ${op.methodName}(${op.renderArgs}): F[${op
               .renderAlgParams(genNameRef.name)}]"
           )
@@ -555,6 +566,7 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
 
     lines(
       deprecationAnnotation(product.hints),
+      documentationAnnotation(product.hints),
       base
     )
   }
@@ -618,6 +630,7 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     def altVal(altName: String) = line"${uncapitalise(altName)}Alt"
     lines(
       deprecationAnnotation(hints),
+      documentationAnnotation(hints),
       line"type ${NameDef(name.name)} = ${members
         .map { case (_, tpe) => line"$tpe" }
         .intercalate(line" | ")}",
@@ -665,6 +678,7 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
 
     lines(
       deprecationAnnotation(hints),
+      documentationAnnotation(hints),
       block(
         line"sealed trait ${NameDef(name.name)} extends scala.Product with scala.Serializable"
       )(
@@ -681,6 +695,7 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
             // format: off
             lines(
               deprecationAnnotation(altHints),
+              documentationAnnotation(altHints),
               line"case object $cn extends $name",
               line"""private val ${cn}Alt = $Schema_.constant($cn)${renderConstraintValidation(altHints)}.oneOf[$name]("$realName").addHints(hints)""",
               line"private val ${cn}AltWithValue = ${cn}Alt($cn)"
@@ -690,6 +705,7 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
             val cn = caseName(a)
             lines(
               deprecationAnnotation(altHints),
+              documentationAnnotation(altHints),
               line"case class $cn(${uncapitalise(altName)}: $tpe) extends $name"
             )
           case Alt(_, realName, UnionMember.ProductCase(struct), altHints) =>
@@ -792,6 +808,7 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
       hints: List[Hint]
   ): Lines = lines(
     deprecationAnnotation(hints),
+    documentationAnnotation(hints),
     block(
       line"sealed abstract class ${name.name}(_value: String, _name: String, _intValue: Int, _hints: $Hints_) extends $Enumeration_.Value"
     )(
@@ -812,6 +829,7 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
 
         lines(
           deprecationAnnotation(hints),
+          documentationAnnotation(hints),
           line"""case object $valueName extends $name("$value", "${e.name}", $intValue, $valueHints)"""
         )
       },
@@ -838,6 +856,7 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     val closing = if (recursive) ")" else ""
     lines(
       deprecationAnnotation(hints),
+      documentationAnnotation(hints),
       obj(name, line"$Newtype_[$tpe]")(
         renderId(shapeId),
         renderHintsVal(hints),
--- a/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
@@ -170,29 +170,31 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     * @param l splitted docstring from the IR
     * @return formatted list of scaladoc lines
     */
-  private def makeDocLines(l: List[String]): List[Line] = {
+  private def makeDocLines(nel: NonEmptyList[String]): List[Line] = {
     @tailrec
     def loop(l: List[String], acc: List[Line]): List[Line] = {
       l match {
         case Nil => acc
-        case hd :: tl if tl == Nil =>
-          loop(tl, acc :+ line"  * $hd" :+ line"  */")
+        case hd :: Nil =>
+          loop(Nil, acc :+ line"  * $hd" :+ line"  */")
         case hd :: tl => loop(tl, acc :+ line"  * $hd")
       }
     }
-    loop(l.tail, List(line"/** ${l.head}"))
+    loop(nel.tail, List(line"/** ${nel.head}"))
   }
 
   private def documentationAnnotation(hints: List[Hint]): Lines = {
     hints
       .collectFirst { case h: Hint.Documentation => h }
-      .foldMap { doc =>
-        val lines = doc.docString.split(System.lineSeparator()).toList
-        val scalaDoc =
-          if (lines.length > 1) makeDocLines(lines)
-          else List(line"/** ${lines.head} */")
-        Lines(scalaDoc)
+      .flatMap { doc =>
+        NonEmptyList.fromList(doc.docString.linesIterator.toList)
       }
+      .foldMap(nel => {
+        val lines =
+          if (nel.size > 1) makeDocLines(nel)
+          else List(line"/** ${nel.head} */")
+        Lines(lines)
+      })
   }
 
   def renderPackageContents: Lines = {
--- a/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
@@ -174,10 +174,9 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     @tailrec
     def loop(l: List[String], acc: List[Line]): List[Line] = {
       l match {
-        case Nil => acc
-        case hd :: Nil =>
-          loop(Nil, acc :+ line"  * $hd" :+ line"  */")
-        case hd :: tl => loop(tl, acc :+ line"  * $hd")
+        case hd :: Nil => acc :+ line"  * $hd" :+ line"  */"
+        case hd :: tl  => loop(tl, acc :+ line"  * $hd")
+        case Nil       => acc
       }
     }
     loop(nel.tail, List(line"/** ${nel.head}"))
--- a/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
@@ -174,9 +174,10 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     @tailrec
     def loop(l: List[String], acc: List[Line]): List[Line] = {
       l match {
-        case Nil                  => acc
-        case _ :: tl if tl == Nil => loop(tl, acc :+ line"  */")
-        case hd :: tl             => loop(tl, acc :+ line"  * $hd")
+        case Nil => acc
+        case hd :: tl if tl == Nil =>
+          loop(tl, acc :+ line"  * $hd" :+ line"  */")
+        case hd :: tl => loop(tl, acc :+ line"  * $hd")
       }
     }
     loop(l.tail, List(line"/** ${l.head}"))
--- a/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
@@ -32,7 +32,6 @@ import scala.jdk.CollectionConverters._
 import Line._
 import LineSyntax.LineInterpolator
 import ToLines.lineToLines
-import scala.annotation.tailrec
 
 private[internals] object Renderer {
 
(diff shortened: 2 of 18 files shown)


modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -164,11 +164,20 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
       }
   }
 
+  private def documentationAnnotation(hints: List[Hint]): Line = {
+    hints
+      .collectFirst { case h: Hint.Documentation => h }
+      .foldMap { doc =>
+        line"/** ${doc.docString} */"
```
zetashift: Is there another combinator that I can use to go from `Option[Hint] => Line` than `foldMap`? 
And another question, for `@documentation` traits, do I have to have to take into consideration multiline docstrings?

Baccata: foldmap is very much adequate. 

And yes, you'll have to take consideration ~multimap~ multiline. I've implemented something similar just yesterday for another project and it was pretty easy. Pro-tip : it's actually easier with mutations, see [here](https://github.com/build-server-protocol/build-server-protocol/pull/426/files#diff-ebe23f8ef14b3b5be88e609b4cb30ec8f88d3dd5b9817ea63cc3d0f7b4622f20R111-R116)

zetashift: Thank you! What do you mean by multimap? Is it datastructure the IR emits for things that span multiple lines?

Baccata: sorry, it's late here and I'm pretty beat, it was a typo and I meant "multiline" 😅

modules/example/src/smithy4s/example/package.scala
```diff
@@ -16,6 +16,7 @@ package object example {
   val DeprecatedService = DeprecatedServiceGen
 
   type StreamedBlob = smithy4s.example.StreamedBlob.Type
+  /** This is a simple example of a "quoted string" */
```
kubukoz, line 19: We should have an example (used in the compilation pipeline, like these files) that contains the string `*/`. Ideally it would be escaped somehow, but I don't have a concrete proposal for _how_ atm.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -792,6 +808,7 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
       hints: List[Hint]
   ): Lines = lines(
     deprecationAnnotation(hints),
+    documentationAnnotation(hints),
```
kubukoz, line 835: Looks like I accidentally made it easier for you by putting all these `deprecationAnnotation` calls in - happy if it saved you some time :)

kubukoz (dismissed):
looking good so far, would love to see multi-line examples and a solution for escaping `*/`. :)

modules/example/src/smithy4s/example/package.scala
kubukoz: random idea to just do `\*\/`.

zetashift: I was thinking about how to escape `*/` as well, but I didn't get any genius ideas so far. is it just going to be a `docString.replace("*/", "\*\/")`? 

Tho for now I'll focus on getting multi-line to work

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
zetashift: It served as an awesome example, and I was like "is it really this easy to add the annotation"? It also helped clarify the structure of these `renderToX` functions.

zetashift:
So another random question popped up.

Do we expect things like `@constructor` `@tparam` for certain members/shapes to also be generated, or does that fall out of scope?

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
zetashift: I'm also realizing that `documentationAnnotation` should be _above_ `deprecationAnnotation`, because you'll want it to be like:
```scala
/** Now deprecated, used to do something useful */
@deprecated
def destroyEverything()
```

kubukoz: good point, yeah

kubukoz:
I think that's not necessary at this stage... but `@param` might be easy to add in the future, and for methods corresponding to operations might be quite useful.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -838,6 +856,7 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     val closing = if (recursive) ")" else ""
     lines(
       deprecationAnnotation(hints),
+      documentationAnnotation(hints),
       obj(name, line"$Newtype_[$tpe]")(
         renderId(shapeId),
         renderHintsVal(hints),
```
kubukoz: (not related to this line in particular)

I think we're missing an example (and possibly an impl) for documenting structure members...

I guess this could look better if we do `@param` inside the structure's Scaladoc, as the lines would get quite lengthy if we just put it on the case class fields themselves. Maybe let's add this in a future PR.

In the meantime, I'd like to see an example of a documented `alt` case in a union as well. Ideally, every usage of `documentationAnnotation(hints)` would be checked with an example.

zetashift: So should I make a `DocumentationService.smithy` file that has all these variants of docs on members/shapes/unions?

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
zetashift: So if we ignore the requirement that every line between `/**` and `*/` should start with an asterisk `*`, this also seems to print multilines fine:

```scala
  private def documentationAnnotation(hints: List[Hint]): Lines = {
    hints
      .collectFirst { case h: Hint.Documentation => h }
      .foldMap { doc =>
        Lines(line"/** ${doc.docString} */")
      }
  }
```

Result
```kotlin
/// This is the first line
/// Second line of information
/// This is a very interesting string
string AString
```

```scala
/** This is the first line
Second line of information
This is a very interesting string */
object AString extends Newtype[String] {

```

Your POC code for typescript also only renders the startin `/**` and `*/` but in smithy4s case I think it's easier to format it according to the scaladoc guidelines

zetashift:
RE: multiline.

I got something working with :
```scala
  private def makeDocLines(l: List[String]): List[Line] = {
    def loop(l: List[String], acc: List[Line]): List[Line] = {
      l match {
        case Nil                  => acc
        case _ :: tl if tl == Nil => loop(tl, acc :+ line"  */")
        case hd :: tl             => loop(tl, acc :+ line"  * $hd")
      }
    }
    if (l.length > 1) loop(l.tail, List(line"/** ${l.head}")) else List(line"/** ${l.head} */")

  }
  private def documentationAnnotation(hints: List[Hint]): Lines = {
    hints
      .collectFirst { case h: Hint.Documentation => h }
      .foldMap { doc =>
        val lines = doc.docString.split(System.lineSeparator()).toList
        Lines(makeDocLines(lines))
      }
  }

```

However, as Baccata said, the number of lines/easyness could be better with a mutable solution, what fits better here? (or is my solution overthinking it?)

EDIT: renders the following: 
```scala
/** This is the first line
  * Second line of information
  */
object AString extends Newtype[String] {
```

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
kubukoz: That works, but I meant that we need these cases to be test covered - not necessarily as part of a new spec, reusing existing ones is fine

Baccata:
@zetashift great work so far! I'm fine with the logic being implemented in terms of tailrec, it's fine. 

Before we merge this in, I have a few asks : 

* I'd like for some multiline documentation to be added to some example specs (see the samples files and the examples project in sbt) so that we can have a look. The "example" project has checked-in code-generation results so that we can have a quick look at what happens when a PR changes the rendering logic. 
* I don't think `@tparam` or `@constructor` need to be generated, I do however think that `@param` should be generated for structures / operations (by aggregating the `member` documentations). So it'd probably be a good idea to change the `Hint.Documentation` to hold both the shape's docs and its fields when it has some. 
* Union members documentation should be generated above the `Case`-suffixed ADT members

zetashift:
@Baccata sounds good!(I've been using the "example" project to check out my changes locally as well, thanks to kubukoz's tips).

As for point 2 and 3, I'll try to get that working.

kubukoz:
Fields have their own hints, is it really necessary to change the parent's?

Baccata:
> Fields have their own hints, is it really necessary to change the parent's?

not 100% necessary, but it means `def documentationAnnotation` needs to be amended with a `memberHints: List[List[Hint]]`  additional parameter and that's not looking too great

kubukoz:
_or_ we make an extra function that calls this after doing a local transformation. Only some callsites will actually need this

Baccata:
Right. I think my point should be more that the purpose of the IR is to facilitate the rendering, by pre-digesting the raw data coming from the model. If the Renderer needs to apply logic to gather some data from several layers of the IR, the IR is not doing his job very well

kubukoz:
Fair enough.

modules/example/src/smithy4s/example/GetObjectInput.scala
```diff
@@ -6,15 +6,21 @@ import smithy4s.ShapeId
...
-  val hints: Hints = Hints.empty
+  val hints: Hints = Hints(
+    smithy.api.Documentation("Input for getting an Object\nall fields are required\nand are given through HTTP labels\nSee https://smithy.io/2.0/spec/http-bindings.html?highlight=httppayload#http-uri-label"),
+  )
 
   implicit val schema: Schema[GetObjectInput] = struct(
-    ObjectKey.schema.required[GetObjectInput]("key", _.key).addHints(smithy.api.HttpLabel(), smithy.api.Required()),
-    BucketName.schema.required[GetObjectInput]("bucketName", _.bucketName).addHints(smithy.api.HttpLabel(), smithy.api.Required()),
+    ObjectKey.schema.required[GetObjectInput]("key", _.key).addHints(smithy.api.Required(), smithy.api.Documentation("Sent in the URI label named \"key\".\nKey can also be seen as the filename\nIt is always required for a GET operation"), smithy.api.HttpLabel()),
+    BucketName.schema.required[GetObjectInput]("bucketName", _.bucketName).addHints(smithy.api.Required(), smithy.api.Documentation("Sent in the URI label named \"bucketName\"."), smithy.api.HttpLabel()),
```
zetashift, line 30: This doesn't seem to get rendered as a scaladoc

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -164,10 +164,42 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
+    * @param l splitted docstring from the IR
+    * @return formatted list of scaladoc lines
+    */
+  private def makeDocLines(l: List[String]): List[Line] = {
+    @tailrec
+    def loop(l: List[String], acc: List[Line]): List[Line] = {
+      l match {
+        case Nil => acc
+        case hd :: tl if tl == Nil =>
+          loop(tl, acc :+ line"  * $hd" :+ line"  */")
```
kubukoz:
```suggestion
        case hd :: Nil =>
          loop(Nil, acc :+ line"  * $hd" :+ line"  */")
```

also, I guess you can inline the `loop` call to be just the second parameter (because `acc` would be returned):

```suggestion
        case hd :: Nil =>
          acc :+ line"  * $hd" :+ line"  */"
```

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -164,10 +164,42 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
+        case hd :: tl => loop(tl, acc :+ line"  * $hd")
+      }
+    }
+    loop(l.tail, List(line"/** ${l.head}"))
```
kubukoz: we assume the list is non-empty, shall we encode this into types? Cats is available in the codegen module AFAIK.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -164,10 +164,42 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
+  }
+
+  private def documentationAnnotation(hints: List[Hint]): Lines = {
+    hints
+      .collectFirst { case h: Hint.Documentation => h }
+      .foldMap { doc =>
+        val lines = doc.docString.split(System.lineSeparator()).toList
```
kubukoz: Splitting by line separator will flatten repeated newlines into one:

`foo\n\nbar` -> `Array("foo", "", "bar")`.

I would recommend Scala's `.linesIterator` instead.

modules/example/src/smithy4s/example/Foo.scala
```diff
@@ -11,15 +11,24 @@ import smithy4s.schema.Schema.int
...
 
-  val hints: Hints = Hints.empty
+  val hints: Hints = Hints(
+    smithy.api.Documentation("Helpful information for Foo\nint, bigInt and bDec are useful number constructs\nThe string case is there because."),
+  )
 
   case class IntCase(int: Int) extends Foo
+  /** this is a comment saying you should be careful for this case
+    * you never know what lies ahead with Strings like this
+    */
```
kubukoz, line 31: I'm cracking up :joy:

modules/example/src/smithy4s/example/FooService.scala
```diff
@@ -11,9 +11,16 @@ import smithy4s.kinds.PolyFunction5
 import smithy4s.kinds.toPolyFunction5.const5
 import smithy4s.schema.Schema.unit
 
+/** The most basics of services
+  * GetFoo is it's only operation
```
kubukoz: nitpick: `its only operation`

modules/example/src/smithy4s/example/package.scala
kubukoz: that `replace` idea sounds good yeah

modules/example/src/smithy4s/example/GetObjectInput.scala
kubukoz: that's the `aggregating the member documentations` part, right? I don't see this having been implemented yet.

zetashift: ah yes this should be rendered as `@constructor` arguments to `GetObjectInput` I believe, and I have not started yet with rendering those type of things

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
zetashift: I have very very superficial experience with Cats:
```scala
  private def documentationAnnotation(hints: List[Hint]): Lines = {
    hints
      .collectFirst { case h: Hint.Documentation => h }
      .flatMap { doc =>
        NonEmptyList.fromList(doc.docString.linesIterator.toList)
      }
      .foldMap(nel => {
        val lines =
          if (nel.size > 1) makeDocLines(nel)
          else List(line"/** ${nel.head} */")
        Lines(lines)
      })
  }
```
Something like this?

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
zetashift: I did this, but the `case Nil` is still there because else I get an exhaustion warning

modules/example/src/smithy4s/example/GetObjectInput.scala
Baccata: Yup, so the correct way to go at it would be for `Hint.Documentation` type to change to : 

``` 
case class Documentation(docString: String, memberDocs: ListMap[String, String]) extends Hint
```

so that you'd be capturing the documentation of the members during the creation of the intermediate representation, and using it where it makes sense (ie, documentation on structures for now)

zetashift: Thank you for the hint(pun not intended)! I understand that we want to capture the doc of the members (if there is any) in the IR. But how do I fill `memberDocs` in the `SmithyToIR.scala` file?

```scala
    case doc: DocumentationTrait =>
      Hint.Documentation(doc.getValue(), )
```

I've looked around a bit, and I cannot see where `DocumentationTrait` provides a function to get the member shapes. Do I have to make that logic myself?

kubukoz: you wouldn't be using `DocumentationTrait` but <checks code> you'd need access to the shape whose hint you're computing.

For example:

```diff
diff --git a/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala b/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
index d2ac2bc2..affdff51 100644
--- a/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
@@ -105,7 +105,7 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
     new ShapeVisitor[Option[Decl]] {
 
       private def getDefault(shape: Shape): Option[Decl] = {
-        val hints = traitsToHints(shape.getAllTraits().asScala.values.toList)
+        val hints = shapeHints(shape)
 
         val recursive = hints.exists {
           case Hint.Trait => true
@@ -218,7 +218,7 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
       }
 
       override def structureShape(shape: StructureShape): Option[Decl] = {
-        val hints = traitsToHints(shape.getAllTraits().asScala.values.toList)
+        val hints = shapeHints(shape)
         val isTrait = hints.exists {
           case Hint.Trait => true
           case _          => false
@@ -251,7 +251,7 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
       override def unionShape(shape: UnionShape): Option[Decl] = {
         val rec = isRecursive(shape.getId())
 
-        val hints = traitsToHints(shape.getAllTraits().asScala.values.toList)
+        val hints = shapeHints(shape)
         val isTrait = hints.exists {
           case Hint.Trait => true
           case _          => false
@@ -681,8 +681,8 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
 
     }
 
-  private def hints(shape: Shape): List[Hint] = traitsToHints(
-    shape.getAllTraits().asScala.values.toList
+  private def hints(shape: Shape): List[Hint] = shapeHints(
+    shape
   )
 
   def toTypeRef(id: ToShapeId): Type.Ref = {
@@ -730,7 +730,7 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
   @annotation.nowarn(
     "msg=class UniqueItemsTrait in package traits is deprecated"
   )
-  private val traitToHint: PartialFunction[Trait, Hint] = {
+  private def traitToHint(shape: Shape): PartialFunction[Trait, Hint] = {
     case _: ErrorTrait => Hint.Error
     case t: ProtocolDefinitionTrait =>
       val shapeIds = t.getTraits()
@@ -757,7 +757,9 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
     case ConstraintTrait(tr) => Hint.Constraint(toTypeRef(tr), unfoldTrait(tr))
   }
 
-  private def traitsToHints(traits: List[Trait]): List[Hint] = {
+  private def shapeHints(shape: Shape): List[Hint] = {
+    val traits = shape.getAllTraits().asScala.values.toList
+
     val nonMetaTraits =
       traits
         .filterNot(_.toShapeId().getNamespace() == "smithy4s.meta")
@@ -770,7 +772,9 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
     val nonConstraintNonMetaTraits = nonMetaTraits.collect {
       case t if ConstraintTrait.unapply(t).isEmpty => t
     }
-    traits.collect(traitToHint) ++ nonConstraintNonMetaTraits.map(unfoldTrait)
+    traits.collect(traitToHint(shape)) ++ nonConstraintNonMetaTraits.map(
+      unfoldTrait
+    )
   }
 
   case class AltInfo(name: String, tpe: Type, isAdtMember: Boolean)

```

then you can find the member docs like this (roughly):

```scala
val yourShapeMemberDocs = shape
  .members()
  .asScala
  .map(_.getTarget())
  .map(model.expectShape(_))
  .map(_.getTrait(classOf[DocumentationTrait]))
```

kubukoz: that `private def hints(shape: Shape)` is kinda useless now. I only left it in because it has way too many usages... I guess you could remove it, rename `shapeHints` to `hints` and call it as `this.hints` wherever `hints` is taken.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
kubukoz: yep, I guess that'll have to stay

modules/example/src/smithy4s/example/GetObjectInput.scala
zetashift: Ah I see! I'll get to this weekend! :D, that clears a few things up.
After that I'll work on the escaping of docstrings.
And hopefully then I can undraft this PR.

kubukoz: sounds great. No pressure ;)

zetashift: > rename shapeHints to hints and call it as this.hints wherever hints is taken.

Do I rename it `this.hints`? The function `hints` isn't part of `ShapeVisitor` but `SmithyIR`
So (from other functions) this seems to work:

```
      private def getDefault(shape: Shape): Option[Decl] = {
        val _hints = hints(shape)

```

kubukoz: if I'm understanding it right, and `ShapeVisitor` is nested in `SmithyIR`, you could still call it as `SmithyIR.this.hints`... but I'm kinda too lazy to try it out in this given I'd have to repeat that diff above :joy:

just push what you have, we can iterate on that :)

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -742,6 +738,19 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+      val memberDocs = shape
+        .members()
+        .asScala
+        .map(_.getTarget())
+        .map(shapeId => (shapeId.name, model.expectShape(shapeId)))
+        .map({ case (name, shape) =>
+          (name, shape.getTrait(classOf[DocumentationTrait]).asScala)
+        })
+        .collect({ case (name, Some(v)) => (name, v.getValue()) })
+        .toMap
```
zetashift: memberDocs seems to always come up as an empty `Map`, and I don't know why exactly :S

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -743,7 +739,18 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
     case d: DeprecatedTrait =>
       Hint.Deprecated(d.getMessage.asScala, d.getSince.asScala)
     case doc: DocumentationTrait =>
-      Hint.Documentation(doc.getValue())
+      val memberDocs = shape
+        .members()
+        .asScala
+        .map(_.getTarget())
+        .map(shapeId => (shapeId.name, model.expectShape(shapeId)))
```
kubukoz: I think this `shapeId.name` should be the member name instead. Can't see what this outputs as, but I feel like that's going to be a different thing (e.g. in `foo: String` you'll see `String` instead of `foo`).

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
kubukoz: I'll take a look

kubukoz: yeah so I think it's just that we were looking at the target type's docs, see https://github.com/disneystreaming/smithy4s/pull/731/files#r1083577898

I pushed a commit to fix that, also managed to squeeze in a couple cleanups (e.g. got rid of the tailrec function in favor of `mkString`): d3234c7ba8827f2f89a4b421167680a67287e1c4

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -165,56 +164,46 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
-    @tailrec
-    def loop(
-        l: List[String],
-        acc: List[Line],
-        memberDocs: Map[String, String]
-    ): List[Line] = {
-      l match {
-        case hd :: Nil => acc :+ line"  * $hd" :+ line"  */"
-        case hd :: tl  => loop(tl, acc :+ line"  * $hd", memberDocs)
-        case Nil       =>
```
kubukoz: this case never actually matched AFAICT

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -165,56 +164,46 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
-          )
-        val memberDocs = doc.memberDocs
-        (shapeDocs, memberDocs)
+        split(doc.docString) ++
+          doc.memberDocs.flatMap { case (memberName, memberDoc) =>
+            split(memberDoc) match {
+              case head :: rest =>
+                s"@param $memberName $head" :: rest
+
+              case Nil => sys.error("impossible: no lines")
```
kubukoz, line 195: this is just to satisfy the pattern match exhaustiveness check - `split` should never return an empty list, so we can assume it's not one.

I guess we could do an `NonEmptyList.unsafeFrom` wrap inside `split` itself - feel free to do so

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
kubukoz: realized the linesIterator can be empty, so another one... 0b3bf439ccd802d67151df1b768c1eebe12c3e23

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -738,19 +738,6 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
       Hint.PackedInputs
     case d: DeprecatedTrait =>
       Hint.Deprecated(d.getMessage.asScala, d.getSince.asScala)
-    case doc: DocumentationTrait =>
```
kubukoz: Moved this out to a separate function to make it easier to get member docs even if the shape itself has no doc trait.

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -764,6 +751,31 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+
+        (
+          member.getMemberName(),
+          memberDocs.orElse(targetDocs)
+        )
+      }
+      .collect { case (name, Some(v)) => (name, v.getValue()) }
+      .toMap
+
+    // todo: what if the shape in question doesn't have docs, but members do?
```
kubukoz: I think we'll have to change the `Hint.Documentation` class to account for the shape not having docs. Just having two `Option`s in it might be enough, or a `Ior[ParentDocs, MemberDocs]`

kubukoz:
@zetashift I may've made a little mess in these commits, if it gets out of hand let me know and I can get this over the finish line. I guess at this point we'll need to:

- make sure we have [the things Olivier mentioned](https://github.com/disneystreaming/smithy4s/pull/731#issuecomment-1385067758)
- add support for `@param` docs where the struct itself doesn't have its own docs (moved to #771)
- support member docs on operation methods (moved to #772)

zetashift:
@kubukoz I'll definitely try to get this over the finish line myself!

Two `Options` for `Hint.Documentation` sounds good to me, I'm not familiar with the `Ior` datatype.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -186,18 +186,18 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
-        val memberLinesOpt = doc.memberDocs.flatMap {
+      .flatMap { doc =>
+        val shapeLinesOpt = doc.docString.map(split(_)).map(_.toList)
+        val memberLinesOpt = doc.memberDocs.map(m => m.flatMap {
           case (memberName, memberDoc) =>
             split(memberDoc) match {
               case Some(NonEmptyList(head, rest)) =>
                 s"@param $memberName $head" :: rest
               case None => Nil
             }
```
zetashift: help, I'm losing the game of aligning types! :P

No but seriously, I'm having trouble getting the docs in a `List` as `List(shapeLinesOpt, memberLinesOpt).flatten.flatten`, as in I don't know how to get unnest these options(shapeDocs and memberDocs)  using `map` or `flatMap` correctly.

Baccata: pro-tip: you shouldn't have to : the `lines()` dsl method is there to flatten stuff for you, no matter the level of nesting. The code should roughly look like  : 

```scala
hints.collectFirst { .... }.foldMap { doc => 
   lines(
      doc.docString.flatMap(split), 
      doc.memberDocs.combineAll.foldMap {
           case (memberName, memberDoc) => //... 
      } 
   ) 
}
```

modules/codegen/src/smithy4s/codegen/internals/IR.scala
```diff
@@ -250,8 +250,10 @@ private[internals] object Hint {
   case class Constraint(tr: Type.Ref, native: Native) extends Hint
   case class Protocol(traits: List[Type.Ref]) extends Hint
   case class Default(typedNode: Fix[TypedNode]) extends Hint
-  case class Documentation(docString: String, memberDocs: Map[String, String])
-      extends Hint
+  case class Documentation(
```
Baccata, line 253: I think it may be easier to pre-split : 

``` 
case class Documentation(docLines: List[String], memberDocLines: Map[String, List[String]]) extends Hint 
```

zetashift: With the optionality stuff (docLines or memberDocLines might be missing/empty), do I wrap both properties in an Option?

`docsLines: Option[List[String]], memberDocLines: Option[Map[String, List[String]]`

kubukoz: probably not, `Nil`/`Map.empty` would be the equivalent. I'm thinking to either do options and change the types inside to be `NonEmpty(...)`, or keep it like in @Baccata's suggestion.

zetashift:
I'm trying to apply @Baccata suggestions of pre-splitting, and it was working out fine, but `memberDocLines` is a `Map[String,List[String]]` and but I'm still having trouble type checking it.

Ignoring the fact that my fp-fu is failing, I do think it might better to have `Option`s(or something equivalent) in `Hint.Documentation` for indicating that shape docs or member docs might be absent

Baccata:
PR'ing against your PR : https://github.com/zetashift/smithy4s/pull/1

zetashift:
So I believe that only leaves this point:
> - Union members documentation should be generated above the Case-suffixed ADT members

Which looks good to me, anything I'm missing here :O?

Currently union shapes are rendered like this:
image

EDIT: nope nvm the `StrCase` doc is duplicated into the `Foo` shapes docs

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -701,8 +701,13 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     val caseNamesAndIsUnit =
       caseNames.zip(alts.map(_.member == UnionMember.UnitCase))
 
+    val shapeDocs = hints
```
kubukoz: **question:** wait, how does this change actually help? I don't see anything union-related in the diff.

zetashift: The `Foo.scala` union gets its `@param str` doc comment removed. It should be removed because the doc comment should be placed on `case class StrCase`
image

zetashift: (sorry for the screenshot, I don't know how to do a fancy diff block :P)

modules/example/src/smithy4s/example/FooService.scala
```diff
@@ -11,9 +11,16 @@ import smithy4s.kinds.PolyFunction5
...
+/** The most basics of services
+  * GetFoo is its only operation
+  */
 trait FooServiceGen[F[_, _, _, _, _]] {
   self =>
 
+  /** Returns a useful Foo
+    * No input necessary to find our Foo
+    * The path for this operation is "/foo"
+    */
```
kubukoz, line 23: **question:** would this also include parameter docs? I don't think we have an example of that

zetashift: With parameter docs here you mean an `input` on an operation for example? I could chuck a doc comment on `PutObject`in `example.smithy`

Baccata: Let's defer this to a later PR, as there is a little bit of nuance to it with regards to packed inputs.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
Baccata: I agree with the reasoning, but transformation is something that should happen in `SmithyToIR`, let's move it over there.

zetashift: I've been pondering over how I would do that, without it getting a bit funky.
I think that would require that `Documentation#shapeDocs` would be a `Hint[Documentation]` rather than a `List[String]` no?

Baccata: Indeed

zetashift: I actually can't figure out how to transform a `Option[DocumentationTrait]` into a `Hint.Documentation` where `shapeDocs` is a `Hint.Documentation` as well:
```
    shape.getTrait(classOf[DocumentationTrait]).asScala.map { doc =>
      val shapeDocs = Hint.Documentation(split(doc.getValue()), Map()) // Can't chuck in a List[String] in shapeDocs anymore
      Hint.Documentation(shapeDocs, memberDocs)
```

zetashift: Do I just introduce a new type, like `Hint.ShapeDocumentation`?

Baccata: I've given it some thoughts, and I think the most straightforward way is to keep the types as they are, but change [this line](https://github.com/disneystreaming/smithy4s/pull/731/files#diff-3d87f30f2b03db7c8e012d2c726b1e1a0f1227fe70d2e5c24166a9046c4428a5R758) to add a condition like `if(shape.isUnion) Map.empty else ...`, that'll be sufficient

zetashift: So simple, love it! :P

Almost makes me think that it's too simple to work.

Baccata (approved):
Awesome work @zetashift, thank you so much for this contribution !

zetashift:
Thank you and @kubukoz very much for all the guidance and help! Some of this stuff I would never come up with on my own

PR #733, opened 2023-01-16
Tweak background highlight color

daddykotex:
Turns out the original style is not applied. The dark one is, but the other one isn't. This solution: overriding the var works for both:

Dark:

Screen Shot 2023-01-16 at 10 05 27

Light:

Screen Shot 2023-01-16 at 10 02 20


Baccata (approved):
Disclaimer : I trust you blindly on this

kubukoz (approved):
Before this change, this is what I saw on the page:

image

the PR does indeed fix it

PR #734, opened 2023-01-16
Split out "default rendering" page

kubukoz:
For some reason, it was previously in [unwrapping](https://disneystreaming.github.io/smithy4s/docs/codegen/customisation/unwrapping) (titled `New types (and unwrapping)`):

image

Now it's a separate page. Also, the category was renamed from `customization` to `Customization` for consistency with others:

image

The new page looks like this:

image


Baccata (approved):
good catch

PR #736, opened 2023-01-17
Add renderer option to change wildcard type arg

### LINKED ISSUE #496: Adding -source:future causes compilation errors due to deprecated wildcard syntax

albertpchen:
fixes https://github.com/disneystreaming/smithy4s/issues/496

This adds a `wildcardArgument: String` argument to the renderer config. This can be configured in the smithy files using `metadata smithy4sWildcardArgument = "?"`. Also updates the mill and sbt plugins to automatically generate a smithy file with the appropriate `smithy4sWildcardArgument` value.

The rules the plugins follow to determine the wildcard value are as follows
1. if Scala version is 3.1 or above, use `?` (according to https://dotty.epfl.ch/docs/reference/changed-features/wildcards.html `_` is supposed to be deprecated at this version, but it looks like that's been pushed to a later version?)
2. if Scala major version is 3 and `-source future` or `-source:future` are in the scalac options, use `?`
3. otherwise use `_`

adds the following new settings/tasks/methods to the sbt plugin and mill plugins
- `smithy4sGeneratedSmithyFilesDir` setting for the directory to write generated smithy files to
- `smithy4sWildcardArgument` task that determines the wildcard string according to above rules
- `smithy4sGeneratedSmithyFiles` task that generates the smithy file with wildcard metadata setting


Baccata:
@albertpchen impressive work, the approach is great 👍 . 

Do you mind signing the [CLA](https://github.com/disneystreaming/smithy4s/pull/736#issuecomment-1384788429) ? Due to company policy, we will not be able to accept the PR until you do so.

daddykotex:
Here me out @Baccata but I think we'd need to align with [SmithyResources](https://github.com/disneystreaming/smithy4s/blob/2db7cea8e6d9fec9906db6d4d9a493fcd96b9059/modules/codegen/src/smithy4s/codegen/internals/SmithyResources.scala) at least on the location to write those Smithy files to.

For context @albertpchen , we use this to track the dependencies of an artifact produced with Smithy4s. We do a similar thing to you where we write a Smithy file that has some metadata in it.

As far as I understand the code, we use the `resource-output` arg of the codegen implementation for that. In that folder, we write the files under `META-INF/smithy/*.smithy`. Here are different values:

* on the cli, its the `--resource-output` arg, defaults to `pwd`
* via sbt, it's `smithy4sResourceDir` which defaults to the module's `resourceManaged` setting
* via mill, its `smithy4sResourceOutputDir` which defaults to `T.ctx().dest / "resources"`

If I understand your PR correctly, you use different directories:
* for sbt: `(config / sourceManaged).value / "smithy" / "generated-metadata.smithy"`
* for mill: `PathRef(T.ctx().dest / "smithy")`

I'm not 100%, but I'd think it's preferable to align. Also SmithyResources produces a `manifest` which I think is required if your smithy file is to end up in a jar that's going to be loaded by the Smithy model loader

daddykotex:
Or maybe we want to do different things, because your PR affects codegen, so writing the file happens _before_ codegen. Where SmithyResources writes stuff during codegen for further processing

albertpchen:
@daddykotex sorry if I'm misunderstanding, but I think I the generated files do end up in the smithy resource dir. the output of `smithy4sGeneratedSmithyFiles` is fed as part of the input smithy files to the codegen step, which copies the input files to the resource directory. if I add a `$ pause` in the `wildcard-config` sbt-test, I see the following layout in the resource managed directory
```
target/scala-3.2.1/resource_managed/
└── main
    └── META-INF
        └── smithy
            ├── generated-metadata.smithy
            ├── manifest
            ├── service.smithy
            └── smithy4s.tracking.smithy
```

I also see `generated-metadata.smithy` appear in the `manifest`. I'm not sure why it appears twice though.
```
smithy4s.tracking.smithy
service.smithy
generated-metadata.smithy
generated-metadata.smithy
```

daddykotex:
> @daddykotex sorry if I'm misunderstanding, but I think I the generated files do end up in the smithy resource dir. the output of `smithy4sGeneratedSmithyFiles` is fed as part of the input smithy files to the codegen step, which copies the input files to the resource directory. if I add a `$ pause` in the `wildcard-config` sbt-test, I see the following layout in the resource managed directory
> 
> ```
> target/scala-3.2.1/resource_managed/
> └── main
>     └── META-INF
>         └── smithy
>             ├── generated-metadata.smithy
>             ├── manifest
>             ├── service.smithy
>             └── smithy4s.tracking.smithy
> ```
> 

Ok that'd explain why you write it in:
```
config / smithy4sGeneratedSmithyFilesDir := {
      (config / sourceManaged).value / "smithy" / "generated-metadata.smithy"
},
```
but it also shows up in the resource folder. Not sure this is a good name: `smithy4sGeneratedSmithyFilesDir`. it feels like we're talking about a directory when it truth it's a file.

> I also see `generated-metadata.smithy` appear in the `manifest`. I'm not sure why it appears twice though.
> 
> ```
> smithy4s.tracking.smithy
> service.smithy
> generated-metadata.smithy
> generated-metadata.smithy
> ```

That's a good question

Baccata:
> Or maybe we want to do different things, because your PR affects codegen, so writing the file happens before codegen. 

That is indeed what's going on : the file is not meant for consumption of downstream artifacts, it's there only for the configuration of the current codegen run. We use smithy metadata as a mean to configure things because it's a build-tool agnostic mechanism, so I'm quite happy with the approach taken by @albertpchen there. As a matter of fact the best course of action would be to delete the file altogether after run, but it's inconsequential since, during the model loading of downstream projects that run smithy4s, we purposefully discard any "smithy4s"-prefixed metadata, so that the decisions of upstream regarding rendering do not affect downstream. 

> I also see generated-metadata.smithy appear in the manifest. I'm not sure why it appears twice though.

That's probably something that needs to be figured out before we sign on on this.

albertpchen:
the manifest issue is because both `src_managed` and the generated file – which defaults to `src_managed/smithy/generated-metadata.smithy` – are in the input file list. when creating the manifest it doesn't deduplicate files it already saw when traversing the parent directory. I fixed this by adding a `.distinct` to the manifest file list.

`ModelAssembler` already takes care of duplicates with an internal hashmap https://github.com/awslabs/smithy/blob/main/smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java#L270

I also changed `smithy4sGeneratedSmithyFilesDir` to `smithy4sGeneratedSmithyMetadataFile`

Baccata (approved):

Baccata:
LGTM. @daddykotex, please give it another review

build.sbt
```diff
@@ -442,7 +442,9 @@ lazy val codegenPlugin = (projectMatrix in file("modules/codegen-plugin"))
         // for the code being built
         (`aws-kernel`.jvm(Scala213) / publishLocal).value,
         (core.jvm(Scala213) / publishLocal).value,
+        (core.jvm(Scala3) / publishLocal).value,
         (dynamic.jvm(Scala213) / publishLocal).value,
+        (dynamic.jvm(Scala3) / publishLocal).value,
```
daddykotex: is that a mistake or you do need dynamic to be published locally for sbt-codegen to be testable?

daddykotex (commented):
I have added one last comment other than that, this is ready to go and as soon as we have an answer from your side, it's going in.

build.sbt
albertpchen: sorry, yes this is a mistake

daddykotex:
I added a final touch on top of your amazing work. Just some validation to keep users from using invalid setting value (especially because only `_` and `?` make sense): https://github.com/disneystreaming/smithy4s/pull/736/commits/3154335f697564cd72e1e46d8aaa8d05580cdfcb

daddykotex:
If the build passes, I will merge 👍 

Kudos and, on behalf of everybody working on Smithy4s

Thank you very much

modules/codegen-plugin/src/sbt-test/codegen-plugin/wildcard-config/build.sbt
```diff
@@ -0,0 +1,50 @@
...
+    scalacOptions ++= {
+      if (scalaVersion.value == scala3) {
+        Seq("-Xfatal-warnings", "-source:future")
+      } else {
+        Seq("-Xfatal-warnings")
+      }
+    },
+    libraryDependencies ++= Seq(
+      "com.disneystreaming.smithy4s" %% "smithy4s-core" % smithy4sVersion.value,
+      "com.disneystreaming.smithy4s" %% "smithy4s-dynamic" % smithy4sVersion.value
```
kubukoz: **issue:** I think this is why your test fails. Seems to be unused, so:

```suggestion
```

daddykotex:
The error looks like:
```
> cs bootstrap 'com.disneystreaming.smithy4s:smithy4s-codegen-cli_2.13:dev-SNAPSHOT' -o smithy4s-gen-test -f
Wrote /Users/David.Francoeur/workspace/dev/smithy4s/smithy4s-gen-test

> ./smithy4s-gen-test generate in.smithy
java.lang.IllegalArgumentException: `smithy4sWildcardArgument` possible values are: `?` or `_`. found `invalid`.
        at smithy4s.codegen.internals.Renderer$Config$.load(Renderer.scala:56)
        at smithy4s.codegen.internals.SmithyToIR.<init>(SmithyToIR.scala:72)
        at smithy4s.codegen.internals.SmithyToIR$.apply(SmithyToIR.scala:44)
        ....
```

modules/codegen-plugin/src/sbt-test/codegen-plugin/wildcard-config/build.sbt
daddykotex: 👍 pushed a fix

PR #741, opened 2023-01-18
add yourkit credit

lewisjkl:
CleanShot 2023-01-17 at 17 22 38

Not sure if this is sufficient or not. Happy to relocate elsewhere if need be.


Baccata (approved):
I've added a bit to the footer too

PR #743, opened 2023-01-19
Smithy resources for dynamic module

### LINKED ISSUE #687: Model resources in dynamic module + handle them the same way codegen does
See https://github.com/disneystreaming/smithy4s/pull/686

When a service reference a resource, the operations of the resource (and all its sub-resources) should be added to the dynamic service. 

This implies that resources should be fully modelled in our dynamic model.

msosnicki:
Closes #687

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -204,24 +204,25 @@ lazy val core = projectMatrix
       "smithy4s.example.collision"
     ),
     Test / smithySpecs := Seq(
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "metadata.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "recursive.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "adtMember.smithy",
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "bodies.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "misc.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "product.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "weather.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "defaults.smithy",
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "discriminated.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "untagged.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "packedInputs.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "enums.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "errorHandling.smithy",
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "errors.smithy",
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "example.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "adtMember.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "namecollision.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "reservednames.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "enums.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "defaults.smithy",
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "kvstore.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "errorHandling.smithy"
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "metadata.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "misc.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "namecollision.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "packedInputs.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "product.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "recursive.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "reservednames.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "resources.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "untagged.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "weather.smithy"
     ),
     (Test / sourceGenerators) := Seq(genSmithyScala(Test).taskValue),
     Compile / packageSrc / mappings ++= {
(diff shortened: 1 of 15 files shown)


build.sbt
```diff
@@ -204,24 +204,25 @@ lazy val core = projectMatrix
...
     Test / smithySpecs := Seq(
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "metadata.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "recursive.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "adtMember.smithy",
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "bodies.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "misc.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "product.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "weather.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "defaults.smithy",
       (ThisBuild / baseDirectory).value / "sampleSpecs" / "discriminated.smithy",
```
msosnicki, line 210: Just sorted and added `resources.smithy`

modules/dynamic/test/src-jvm/smithy4s/dynamic/OperationSpec.scala
```diff
@@ -18,31 +18,12 @@ package smithy4s
...
 
-class OperationSpec() extends munit.FunSuite {
-
-  // This is not ideal, but it does the job.
-  val cwd = System.getProperty("user.dir");
-  val pizzaSpec = Paths.get(cwd + "/sampleSpecs/pizza.smithy").toAbsolutePath()
+class OperationSpec() extends DummyIO.Suite {
 
   test("Decode operation") {
-    IO(
```
msosnicki, line 35: `check` no longer required (due to `DummyIO.Suite`). Here it was not called at all, but test was silently failing (missing `alloy` dependency).

modules/dynamic/test/src-jvm/smithy4s/dynamic/package.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import smithy4s.dynamic.DummyIO._
+import java.nio.file.Paths
+import software.amazon.smithy.model.{Model => SModel}
+import software.amazon.smithy.model.shapes.ModelSerializer
+import cats.syntax.all._
+
+package object dynamic {
+  // This is not ideal, but it does the job.
+  private val cwd = System.getProperty("user.dir");
```
Baccata: make this lazy : running effects during instantiation of static objects (including package objects) is a big no-no

modules/dynamic/test/src-jvm/smithy4s/dynamic/ResourceSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  test("Compile resource operations as endpoints") {
+    loadDynamicModel("resources.smithy").map { model =>
+      val compiled = DynamicSchemaIndex.load(model)
+
+      val endpoints = compiled.allServices.head.service.endpoints
+      expect.same(
+        endpoints.map(_.id).toSet,
+        List("ListPublishers", "GetBook", "BuyBook")
```
msosnicki: What would be better ways to test it? Throw it at some algebra, like in `DynamicJsonServerSpec`?

build.sbt
```diff
@@ -204,24 +204,25 @@ lazy val core = projectMatrix
       "smithy4s.example.collision"
     ),
     Test / smithySpecs := Seq(
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "metadata.smithy",
-      (ThisBuild / baseDirectory).value / "sampleSpecs" / "recursive.smithy",
+      (ThisBuild / baseDirectory).value / "sampleSpecs" / "adtMember.smithy",
```
Baccata, line 207: I appreciate the sorting, thanks

modules/dynamic/smithy/dynamic.smithy
```diff
@@ -184,11 +184,20 @@ structure ServiceShape {
   version: String,
   errors: MemberList,
   operations: MemberList,
+  resources: MemberList,
   traits: TraitMap
 }
 
-/// TODO
 structure ResourceShape {
+  /// ignored: identifiers, properties, collectionOperations
```
msosnicki, line 192: Couldn't get `collectionOperations` to work. Same applied for `list` member shape. Looks like they have some specific requirements, as I get following error: 
```
smithy4s.example#ListPublishers: This operation is bound as a collection operation on the `smithy4s.example#Publisher` resource, but it improperly binds all of the identifiers of the resource to members of the operation input. 
```
Will try to solve that and push a fix for it

modules/dynamic/test/src-jvm/smithy4s/dynamic/package.scala
msosnicki: Good catch, will do!

modules/dynamic/test/src/smithy4s/dynamic/DummyIO.scala
```diff
@@ -25,15 +26,20 @@ object DummyIO {
...
-    def check(): Unit = io match {
-      case Left(e)  => throw e
-      case Right(_) => ()
-    }
+
+  trait Suite extends munit.FunSuite {
+    override def munitValueTransforms: List[ValueTransform] =
+      ioValueTransform :: super.munitValueTransforms
+
+    private val ioValueTransform: ValueTransform =
```
Baccata, line 34: 👍

modules/example/src/smithy4s/example/GetObjectOutput.scala
```diff
@@ -14,7 +14,7 @@ object GetObjectOutput extends ShapeTag.Companion[GetObjectOutput] {
   val hints: Hints = Hints.empty
 
   implicit val schema: Schema[GetObjectOutput] = struct(
-    ObjectSize.schema.required[GetObjectOutput]("size", _.size).addHints(smithy.api.Default(smithy4s.Document.fromDouble(0.0d)), smithy.api.Required(), smithy.api.HttpHeader("X-Size")),
+    ObjectSize.schema.required[GetObjectOutput]("size", _.size).addHints(smithy.api.Default(smithy4s.Document.fromDouble(0.0d)), smithy.api.HttpHeader("X-Size"), smithy.api.Required()),
```
Baccata: mmm, I wonder whether the CI is gonna scream about this : the ordering might be subject to the JVM you run. We'll see

modules/dynamic/test/src-jvm/smithy4s/dynamic/ResourceSpec.scala
Baccata: I'm happy enough with this tbh, the bit that was missing was the transitive gathering of operations, the fact that operations are correctly encoded dynamically is already tested.

modules/dynamic/test/src-jvm/smithy4s/dynamic/OperationSpec.scala
Baccata: good catch

modules/example/src/smithy4s/example/GetObjectOutput.scala
msosnicki: Failed, when running `sbt test` from nix shell, it produces old output. Reverting

Baccata (commented):
LGTM. @kubukoz, you wanna double check ?

Baccata (approved):

kubukoz:
will do

build.sbt
kubukoz: not saying to do it immediately, but any reason not to pull in the whole directory?

modules/dynamic/smithy/dynamic.smithy
kubukoz: guess we can follow up with that

kubukoz (approved):

PR #744, opened 2023-01-19
[AWS] Awscall to F[_] Transformation #730

### LINKED ISSUE #730: [AWS] Write a transformation that turns `AwsCall[F, I, E, O, SI, SO]` into `F[O]`
[AwsCall](https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/aws/src/smithy4s/aws/AwsCall.scala#L20) is a construct that captures all type parameters involved in the interaction with AWS SDKs. This construct is required as it allows to handle the various kinds of interactions and combinations of unary/streamed requests/responses. 

The way it works is that an implicit evidence is required when calling one of the methods held by this construct, to prove that the interaction is handle-able in a unary-fashion, or to prove that it has a stream component to it. 

This way of doing things is explained by the fact that the generated interfaces are homogenous in context types : we cannot have some methods that return an `F` and others that return something else. 

However, once we have an instance of an algebra that works against `AwsCall`, it would be nice to easily turn it into an algebra that works against `F`. 

Smithy4s defines several higher-kinded functor typeclasses, in particular [FunctorK5](https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/core/src-3/generated/kinds/functorK.scala#L44-L46) , which is implemented by the companion objects of [all generated services](https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/example/src/smithy4s/example/FooService.scala#L47).

So what the implementor of this ticket needs to write is something like : 

```scala
import smithy4s.kinds._ 

def forceUnary[Alg[_[_,_,_,_,_]], F[_]](service: Service[Alg]) : PolyFunction5[AwsCall[F, *,*,*,*,*], kind1[F]#toKind5]
``` 

Which uses the schemas of the endpoint associated to the call to produce the evidence required to call `.run` on the `AwsCall`, obtaining an `F`. If no such endpoint can be provided, an error should be raised in `F`. 

For inspiration, you can look at [this](https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/core/src/smithy4s/Transformation.scala#L85-L97) implementation, which is used to turn an algebra that works against a monofunctor into an algebra that works against a bifunctor, utilising the information of the endpoints to know which errors are associated to which operation.

yisraelU:
closes #730
Currently throwing an exception as F has no structure.
should I add a context bound ?

### DIFF
--- a/modules/aws/src/smithy4s/aws/package.scala
+++ b/modules/aws/src/smithy4s/aws/package.scala
@@ -18,7 +18,7 @@ package smithy4s
 
 import cats.MonadThrow
 import smithy4s.kinds._
-
+import smithy4s.aws.AwsOperationKind.Unary.unary
 package object aws {
 
   type Timestamp = smithy4s.Timestamp
--- a/modules/aws/src/smithy4s/aws/package.scala
+++ b/modules/aws/src/smithy4s/aws/package.scala
@@ -35,7 +35,7 @@ package object aws {
     MonadThrow[F].catchNonFatal(new String(bytes, "UTF-8"))
 
   // format: off
- private[aws] def simplify[Alg[_[_, _, _, _, _]], F[_]:ApplicativeThrow](service: Service[Alg]): service.Interpreter[AwsCall[F, *, *, *, *, *]] => service.FunctorInterpreter[F] = {
+  private[aws] def simplify[Alg[_[_, _, _, _, _]], F[_]:ApplicativeThrow](service: Service[Alg]): service.Interpreter[AwsCall[F, *, *, *, *, *]] => service.FunctorInterpreter[F] = {
     interpreter =>
      new PolyFunction5[service.Operation, Kind1[F]#toKind5] {
       override def apply[I, E, O, SI, SO ](op: service.Operation[I, E, O, SI, SO ]): F[O] = {
--- a/modules/aws/src/smithy4s/aws/package.scala
+++ b/modules/aws/src/smithy4s/aws/package.scala
@@ -19,6 +19,7 @@ package smithy4s
 import cats.{ApplicativeThrow, MonadThrow}
 import smithy4s.kinds._
 import smithy4s.aws.AwsOperationKind.Unary.unary
+
 package object aws {
 
   type Timestamp = smithy4s.Timestamp
@@ -34,10 +35,10 @@ package object aws {
     MonadThrow[F].catchNonFatal(new String(bytes, "UTF-8"))
 
   // format: off
-  def simplify[Alg[_[_, _, _, _, _]], F[_]:ApplicativeThrow](service: Service[Alg]): service.Interpreter[AwsCall[F, *, *, *, *, *]] => service.FunctorInterpreter[F] = {
+ private[aws] def simplify[Alg[_[_, _, _, _, _]], F[_]:ApplicativeThrow](service: Service[Alg]): service.Interpreter[AwsCall[F, *, *, *, *, *]] => service.FunctorInterpreter[F] = {
     interpreter =>
      new PolyFunction5[service.Operation, Kind1[F]#toKind5] {
-      override def apply[A0, A1, A2, A3, A4](op: service.Operation[A0, A1, A2, A3, A4]): F[A2] = {
+      override def apply[I, E, O, SI, SO ](op: service.Operation[I, E, O, SI, SO ]): F[O] = {
   // format: on
           val endpoint = service.opToEndpoint(op)
           (endpoint.streamedInput, endpoint.streamedOutput) match {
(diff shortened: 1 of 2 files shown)


kubukoz:
I'm honestly surprised this compiles on all Scala versions

kubukoz:
> Currently throwing an exception as F has no structure.
> should I add a context bound ?

I think that'd be a safer approach - if someone calls this transformation on a service with streaming- and non-streaming operations, we won't break the whole thing if they're only going to call some of these methods...

modules/aws/src/smithy4s/aws/package.scala
```diff
@@ -32,4 +33,22 @@ package object aws {
   private[aws] def utf8String[F[_]: MonadThrow](bytes: Array[Byte]): F[String] =
     MonadThrow[F].catchNonFatal(new String(bytes, "UTF-8"))
 
+  // format: off
+  def simplify[Alg[_[_, _, _, _, _]], F[_]](service: Service[Alg]): service.Interpreter[AwsCall[F, *, *, *, *, *]] => service.FunctorInterpreter[F] = {
```
kubukoz: I'm wondering if it'd be more convenient for the callers if `service` was implicit and `interpreter` was a normal function parameter.

yisraelU:
> I'm honestly surprised this compiles on all Scala versions

Are you referring to the implicit ev requirement?

yisraelU:
> > Currently throwing an exception as F has no structure.
> > should I add a context bound ?
> 
> I think that'd be a safer approach - if someone calls this transformation on a service with streaming- and non-streaming operations, we won't break the whole thing if they're only going to call some of these methods...

Yeah of course.
I just wasn't sure what @Baccata had in mind as he provided the signature

kubukoz:
> Are you referring to the implicit ev requirement?

No, referring to how the streaming types are unified after the pattern match so that you can call `.run` without extra trickery. I've seen plenty of GADT issues back in the day, so it's good to see it working right

yisraelU:
> > Are you referring to the implicit ev requirement?
> 
> No, referring to how the streaming types are unified after the pattern match so that you can call `.run` without extra trickery. I've seen plenty of GADT issues back in the day, so it's good to see it working right

yeah thats what I meant by the evidence requirement , the fact that it recognizes the proof of the streaming types via the pattern match

Baccata:
Cool, I too wasn't sure it'd compile on all Scala versions. 

> should I add a context bound ? 

Yup, good idea, it eluded me when I wrote the sig. 

Alright, let's: 

1. ~add a `MonadThrow` constraint in there so that the exception is raised in `F`~ already done, gj !
2. change the exception type to `IllegalAccessException`
3. make that simplify method package-private 
4. add a `buildSimple` method new to [this one](https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/aws/src/smithy4s/aws/AwsClient.scala#L81-L83) so that users can have a functor client easily

modules/aws/src/smithy4s/aws/package.scala
```diff
@@ -32,4 +33,24 @@ package object aws {
   private[aws] def utf8String[F[_]: MonadThrow](bytes: Array[Byte]): F[String] =
     MonadThrow[F].catchNonFatal(new String(bytes, "UTF-8"))
 
+  // format: off
+  def simplify[Alg[_[_, _, _, _, _]], F[_]:ApplicativeThrow](service: Service[Alg]): service.Interpreter[AwsCall[F, *, *, *, *, *]] => service.FunctorInterpreter[F] = {
+    interpreter =>
+     new PolyFunction5[service.Operation, Kind1[F]#toKind5] {
+      override def apply[A0, A1, A2, A3, A4](op: service.Operation[A0, A1, A2, A3, A4]): F[A2] = {
```
Baccata: let's call those `I, E, O, SI, SO` as usual

modules/aws/src/smithy4s/aws/package.scala
```diff
@@ -16,8 +16,9 @@
 
 package smithy4s
 
-import cats.MonadThrow
-
+import cats.{ApplicativeThrow, MonadThrow}
+import smithy4s.kinds._
+import smithy4s.aws.AwsOperationKind.Unary.unary
```
Baccata, line 21: let's add a blank line after this (weird that scalafmt wouldn't say anything)

yisraelU:
@Baccata I thought ApplicativeThrow would be better for this function as principle of least power

modules/aws/src/smithy4s/aws/package.scala
yisraelU: yeah I should have though of that

Baccata (approved):
awesome !

PR #745, opened 2023-01-20
Make decline cli's `service` parameter implicit

kubukoz:


kubukoz:
This _is_ clearly breaking, right?

```
// old
public smithy4s.decline.Smithy4sCli(com.monovore.decline.Opts<smithy4s.decline.Entrypoint<Alg, F>>, smithy4s.Service<Alg>, cats.MonadError<F, java.lang.Throwable>);
public smithy4s.decline.Smithy4sCli(com.monovore.decline.Opts<smithy4s.decline.Entrypoint<Alg, F>>, cats.MonadError<F, java.lang.Throwable>, smithy4s.Service<Alg>);
// new
```

Apparently MiMa is off in the decline module:

```
[info] [decline] decline: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
```

Baccata:
It is breaking but you could make a smart constructor in the companion object instead of breaking ?

Baccata:
> Apparently MiMa is off in the decline module:

let's turn it on

kubukoz:
> It is breaking but you could make a smart constructor in the companion object instead of breaking ?

won't help my usecase as I'm mostly doing `extends Smithy4sCli`... but I can stop doing that, too. Or wait for 0.18.

Baccata:
> won't help my usecase as I'm mostly doing extends Smithy4sCli... but I can stop doing that, too. Or wait for 0.18. 

mmm you add an additional constructor ?

kubukoz:
or that yeah

Baccata (approved):

PR #747, opened 2023-01-20
Dynamic compliance tests

yisraelU:
[Baccata's summary of this PR] 

This adds the necessary machinery allowing for compliance tests to be executed against dynamically loaded indexes. 

In particular : 

* Adds a schema-based Eq implementation allowing for the structural comparison of dynamic, array-based data 
* Removes all pieces of logic that rely on assumption regarding how the data is stored in memory (ie, in the context of dynamic services, the data is not encoded as case classes that abide by structural equality)

TODOs   

- [x] Find a way to make it work in JS/Native.


modules/compliance-tests/src/smithy4s/compliancetests/internals/Assertions.scala
```diff
@@ -50,13 +50,26 @@ private[internals] object assert {
...
-    } else {
+  def arrEql(expected: Array[Any], actual: Array[Any]): ComplianceResult = {
+    if (java.util.Arrays.deepEquals(expected, actual)) success
+    else
       fail(
-        s"Actual value: ${pprint.apply(actual)} was not equal to ${pprint.apply(expected)}."
+        s"Arrays are not equal  ${pprint.apply(actual)} \n expected:  ${pprint.apply(expected)}"
       )
+  }
+  Tuple2
```
kubukoz: **issue:** I don't think you meant this to end up here :D

yisraelU: no there is a bunch of garbage here , was just opened to share code with @Baccata

modules/compliance-tests/src/smithy4s/compliancetests/internals/ErrorResponseTest.scala
```diff
@@ -17,11 +17,34 @@
...
 
 private[compliancetests] final case class ErrorResponseTest[A, E](
     schema: Schema[A],
-    inject : A => E,
+    inject: A => E,
     errorable: Errorable[E]
-)
+) {
+
+  def kleisliFy[F[_]: ApplicativeThrow]: Document => F[Throwable] = {
```
kubukoz, line 60: **suggestion:** `asFunction`?

yisraelU:
rough implementation , not cleaned up yet.
1. Added EqSchemaVisitor
2. To handle Client errors, we need to be able to compare 2 throwables (because thats how the service will send us an Error)
So, I am storing a E => Option[A] in the ErrorResponseTest class. https://github.com/disneystreaming/smithy4s/blob/504a338a840a8c1da0f91ab1e6bd41f958901fce/modules/compliance-tests/src/smithy4s/compliancetests/internals/ErrorResponseTest.scala#L33
This works to my knowledge and most tests are passing now, (however I didn't yet test for false positives)

@Baccata The issue I am currently having is implementing union for Eq , as the new implementation requires an EncoderK and I cant find a way to make that work for a Function2

Baccata:
> The issue I am currently having is implementing union for Eq , as the new implementation requires an EncoderK and I cant find a way to make that work for a Function2

Done. The trick is currying + fixing the type of the RHS of the comparison to `U` (the union). See in [6bddfe9](https://github.com/disneystreaming/smithy4s/pull/747/commits/6bddfe9b6507922ca000193b349db382f7514b62)

yisraelU:
I see,
I actually went down the curried route , but I was attempting A => A => Boolean

modules/compliance-tests/src/smithy4s/compliancetests/internals/eq/EqSchemaVisitor.scala
```diff
@@ -0,0 +1,198 @@
...
+import smithy4s.compliancetests.internals.eq.Smithy4sEqInstances._
+import cats.implicits.catsSyntaxEq
+import cats.kernel.Eq
+import smithy4s._
+import smithy4s.schema.{Schema, _}
+
+import java.util.UUID
+import smithy4s.capability.EncoderK
+
+object EqSchemaVisitor extends SchemaVisitor[Eq] { self =>
```
kubukoz, line 27: as a user, it would be nice to have it as public. I know, API surface...

Baccata: End users don't want that, it's wildly inefficient. I however think we should tweak the dynamic model to respect structural equality.

modules/compliance-tests/src/smithy4s/compliancetests/internals/eq/EqSchemaVisitor.scala
```diff
@@ -0,0 +1,198 @@
...
+      shapeId: ShapeId,
+      hints: Hints,
+      values: List[EnumValue[E]],
+      total: E => EnumValue[E]
+  ): Eq[E] =
+    (x: E, y: E) => {
+      val enumX = total(x)
+      val enumY = total(y)
+      enumX.intValue === enumY.intValue && enumX.stringValue === enumY.stringValue
+    }
```
kubukoz: **suggestion:** 

```suggestion
    Eq.by { x =>
      val enumX = total(x)
      (enumX.intValue, enumX.stringValue)
    }
```

modules/http4s/test/src/ProtocolComplianceTest.scala
```diff
@@ -59,32 +58,41 @@ object ProtocolComplianceTest extends SimpleIOSuite {
...
+            .map {
+              case Right(expectations) => expectations
+              case Left(e)             => failure(e.getMessage)
+            }
         }
-    }
-  )
+      )
+    )
+    .unsafeRunAndForget()
```
Baccata: Hah, interesting ... this is a sign we shouldn't use `SimpleIOSuite` but rather a higher-level suite-type, like : 

https://github.com/disneystreaming/weaver-test/blob/main/modules/core/src/weaver/suites.scala#L29 

Feel free to park this work until we can sync on this (or attempt to wire it, totally up to you)

yisraelU: @Baccata yeah , this is just a placeholder to get it working end to end
the file IO cross platform is the current blocker

modules/compliance-tests/src/smithy4s/compliancetests/internals/eq/EqSchemaVisitor.scala
yisraelU: @kubukoz Can I lend a hand here ? do you know what has to be done?

yisraelU:
added the missing piece to load env variables when running scalajs
Last component to fix is the IO

modules/http4s/test/src-js/smithy4s/compliancetests/EnvCompat.scala
```diff
@@ -0,0 +1,23 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.compliancetests
+
+import scala.scalajs.js
+trait EnvCompat {
```
Baccata: Good to see you building an intuition for stuff like this 👍  (ie: what's gonna be platform-specific and how to implement such things), but CE already [provides it](https://github.com/typelevel/cats-effect/blob/9aade3c348b847d9c250b0e2028c691d37cffa09/std/shared/src/main/scala/cats/effect/std/Env.scala#L51)

build.sbt
```diff
@@ -646,16 +648,24 @@ lazy val http4s = projectMatrix
...
+    (Test / sourceGenerators) := Seq(genSmithyScala(Test).taskValue),
+    (Test / resourceGenerators) := Seq(dumpModel(Test).taskValue),
+    (Test / envVars) ++=  {
+      val files: Seq[File] =
+        (Test / resourceGenerators) {
+          _.join.map(_.flatten)
+        }.value
+
+      files.foldLeft(Map.empty[String, String]) { case (acc, file) =>
+        acc.updated("MODEL_DUMP", file.getAbsolutePath)
```
Baccata: 👍

build.sbt
```diff
@@ -869,6 +879,27 @@ def genSmithy(config: Configuration) = Def.settings(
 def genSmithyScala(config: Configuration) = genSmithyImpl(config).map(_._1)
 def genSmithyResources(config: Configuration) = genSmithyImpl(config).map(_._2)
 
+// SBT setting to specify artifacts to be included in the Smithy model for compliance testing
+val complianceTestDependencies     = SettingKey[Seq[ModuleID]]("complianceTestDependencies")
+
+// writes out a json representation of the smithy model pulled from Smithy4s dependencies config
+def dumpModel(config: Configuration): Def.Initialize[Task[Seq[File]]] = Def.task{
```
Baccata: 👍 awesome, you're probably gonna need to cache-it (see how the codegen task works). It's greatly annoying that SBT doesn't do that for you, but 🤷

Baccata:
Great work so far. From where I stand, all that's needed left is : 

* the removal of the `unsafeRunAndForget` clause by relying on a higher kind of suite than `SimpleIOSuite`, cause it could create very annoying race conditions and hard-to-debug. 
* the caching of the file dump so that it only gets re-dumped upon changing of the inputs of the task.

yisraelU:
@Baccata yeah unsafeRunAndForget doesnt work with js, the test ends before it runs, when debugging there was also an issue with adding a test to the TestSuite after initialization (or something like that)

Baccata:
See [fc625c1](https://github.com/disneystreaming/smithy4s/pull/747/commits/fc625c11624431c2b05061a734d89f659ebb2410) for what I meant. It occurred to me that maybe what was needed wasn't too clear.

yisraelU:
Thanks , I tried a few variations , but was unclear about which one to use

Baccata (approved):
@daddykotex, could you give this a review as well, please ? And @kubukoz if you can find the time.

Baccata:
Great work on this @yisraelU !

build.sbt
```diff
@@ -635,6 +635,8 @@ lazy val http4s = projectMatrix
         Dependencies.Http4s.dsl.value,
         Dependencies.Http4s.client.value,
         Dependencies.Alloy.core % Test,
+        Dependencies.Alloy.`protocol-tests` % Test,
```
daddykotex: do you still need this one if you define it [on L653](https://github.com/disneystreaming/smithy4s/pull/747/files#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91R653)

build.sbt
```diff
@@ -646,16 +648,26 @@ lazy val http4s = projectMatrix
...
-    (Test / sourceGenerators) := Seq(genSmithyScala(Test).taskValue)
+    (Test / sourceGenerators) := Seq(genSmithyScala(Test).taskValue),
+    (Test / resourceGenerators) := Seq(dumpModel(Test).taskValue),
+    (Test / envVars) ++= {
+      val files: Seq[File] =
+        (Test / resourceGenerators) {
+          _.join.map(_.flatten)
+        }.value
+
+      files.foldLeft(Map.empty[String, String]) { case (acc, file) =>
```
daddykotex: basically, if I understand correctly, this means that if `files` contains multiple files, only the last value will be in `MODEL_DUMP`

if that's correct, maybe a `files.lastOption.map(...).getOrElse(Map.empty)` will me more explicit?

build.sbt
```diff
@@ -869,6 +881,88 @@ def genSmithy(config: Configuration) = Def.settings(
...
+
+// writes out a json representation of the smithy model pulled from Smithy4s dependencies config
+// result is cached using the dependency list as the cache key
+def dumpModel(config: Configuration): Def.Initialize[Task[Seq[File]]] =
+  Def.task {
+    val dumpModelCp = (`codegen-cli`.jvm(
+      Smithy4sBuildPlugin.Scala213
+    ) / Compile / fullClasspath).value
+      .map(_.data)
+    val mc = "smithy4s.codegen.cli.Main"
```
daddykotex: you can probably use the `mainClass` task

```
sbt:sauce> inspect mainClass
[info] Task: scala.Option[java.lang.String]
[info] Description:
[info]  Defines the main class for packaging or running.
```

modules/compliance-tests/src/smithy4s/compliancetests/internals/Assertions.scala
```diff
@@ -29,7 +30,7 @@ private[internals] object assert {
   def fail(msg: String): ComplianceResult = Left(msg)
 
   private def isJson(bodyMediaType: Option[String]) =
-    bodyMediaType.exists(_.equalsIgnoreCase("application/json"))
+    bodyMediaType.forall(_.equalsIgnoreCase("application/json"))
```
daddykotex, line 33: to be sure: this means that if no `bodyMediaType` is defined, we consider the payload to be json

modules/compliance-tests/src/smithy4s/compliancetests/internals/Assertions.scala
```diff
@@ -50,8 +51,19 @@ private[internals] object assert {
     }
   }
 
-  def eql[A](expected: A, actual: A): ComplianceResult = {
-    if (expected == actual) {
+  def neq[A: Eq](expected: A, actual: A): ComplianceResult = {
```
daddykotex: two minor things, feel free to ignore

1. I think `eq/neq` or `eql/neql` would make better pairs rather than `eql/neq`
2. Is it possible to implement using `eql` and `swap` on `Either`?

modules/compliance-tests/test/src/EqVisitorSpec.scala
```diff
@@ -0,0 +1,321 @@
...
+package smithy4s.compliancetests
+
+import cats.kernel.Eq
+import smithy4s.compliancetests.internals.eq.EqSchemaVisitor
+import smithy4s.schema.{Schema, SchemaVisitor}
+import smithy4s.schema.Schema._
+import smithy4s.{ByteArray, Hints, ShapeId, Timestamp}
+import weaver.{Expectations, FunSuite}
+
+object EqVisitorSpec extends FunSuite {
```
daddykotex, line 26: good job on the spec

modules/http4s/test/src-ce3/smithy4s/compliancetests/ProtocolComplianceTest.scala
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  private val pizzaSpec = generateTests(
+    ShapeId("alloy.test", "PizzaAdminService")
+  )
+
+  private val path = Env
+    .make[IO]
+    .get("MODEL_DUMP")
+    .map(_.fold(sys.error(""))(fs2.io.file.Path(_)))
```
daddykotex: fill in the blank here

daddykotex (changes requested):
Good job, mostly minor comments

There is one `sys.error("")` that needs to be addressed, the rest I'm pretty flexible on

modules/compliance-tests/src/smithy4s/compliancetests/internals/Assertions.scala
yisraelU: correct

modules/compliance-tests/src/smithy4s/compliancetests/internals/Assertions.scala
yisraelU: 1. agreed
2. 2. i wouldnt be able to provide the specific error message for expected failure

modules/http4s/test/src-ce3/smithy4s/compliancetests/ProtocolComplianceTest.scala
Baccata: NB : you can use `liftTo[IO](someThrowable)` on `Option`, provided you have `import cats.syntax.all._` in scope

yisraelU: done

build.sbt
yisraelU: whats the difference?

build.sbt
yisraelU: good point

build.sbt
yisraelU: good catch

build.sbt
daddykotex: one is hardcoded and the other one is found by the build tool

the latter is more _future proof_ (if we ever change the name)

kubukoz: I like the idea of `mainClass` here

yisraelU: great

daddykotex (approved):
Good work

PR #748, opened 2023-01-20
testing smithy4s apis - docs

lewisjkl:
I've put down my thoughts on testing Smithy4s apis in general since we have received a lot of questions around this in the past. Would love feedback on:

1. If you agree with the points I made
2. If I need to go into more depth on any part
3. If I lacked precision in my descriptions anywhere

Thanks all 🙂

### DIFF
--- a/modules/docs/markdown/06-guides/testing.md
+++ b/modules/docs/markdown/06-guides/testing.md
@@ -7,7 +7,7 @@ In this guide, we will give you some guiding principles and other things to cons
 
 ## What Smithy4s Does
 
-For starters, let's briefly cover what it is that Smithy4s does. At a high level, Smithy4s takes your Smithy model and generates Scala code based on that model. The model is essentially just a Scala representation of the Smithy model. The generated code _does not_ contain any logic for any specific library such as http4s or jsoniter. Instead, it contains abstractions that Smithy4s itself defines that interpreters can be based on. This means that the generated code is not coupled to any library (other than Smithy4s itself). The interpreters that we have created in Smithy4s will then take the generated code and use it to power, for example, and HTTP service that uses http4s and jsoniter.
+For starters, let's briefly cover what it is that Smithy4s does. At a high level, Smithy4s takes your Smithy model and generates Scala code based on that model. The model is essentially just a Scala representation of the Smithy model. The generated code _does not_ contain any logic for any specific library such as http4s or jsoniter. Instead, it contains abstractions that Smithy4s itself defines that interpreters can be based on. This means that the generated code is not coupled to any library (other than Smithy4s itself). The interpreters that we have created in Smithy4s will then take the generated code and use it to power, for example, an HTTP service that uses http4s and jsoniter.
 
 ## How we Test Smithy4s
 
--- a/modules/docs/markdown/06-guides/testing.md
+++ b/modules/docs/markdown/06-guides/testing.md
@@ -1,0 +1,41 @@
+---
+sidebar_label: Testing
+title: Testing Smithy4s Applications
+---
+
+In this guide, we will give you some guiding principles and other things to consider when testing Smithy4s applications.
+
+## What Smithy4s Does
+
+For starters, let's briefly cover what it is that Smithy4s does. At a high level, Smithy4s takes your Smithy model and generates Scala code based on that model. The model is essentially just a Scala representation of the Smithy model. The generated code _does not_ contain any logic for any specific library such as http4s or jsoniter. Instead, it contains abstractions that Smithy4s itself defines that interpreters can be based on. This means that the generated code is not coupled to any library (other than Smithy4s itself). The interpreters that we have created in Smithy4s will then take the generated code and use it to power, for example, and HTTP service that uses http4s and jsoniter.
+
+## How we Test Smithy4s
+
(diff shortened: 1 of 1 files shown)


modules/docs/markdown/06-guides/testing.md
```diff
@@ -0,0 +1,41 @@
+---
+sidebar_label: Testing
+title: Testing Smithy4s Applications
+---
+
+In this guide, we will give you some guiding principles and other things to consider when testing Smithy4s applications.
+
+## What Smithy4s Does
+
+For starters, let's briefly cover what it is that Smithy4s does. At a high level, Smithy4s takes your Smithy model and generates Scala code based on that model. The model is essentially just a Scala representation of the Smithy model. The generated code _does not_ contain any logic for any specific library such as http4s or jsoniter. Instead, it contains abstractions that Smithy4s itself defines that interpreters can be based on. This means that the generated code is not coupled to any library (other than Smithy4s itself). The interpreters that we have created in Smithy4s will then take the generated code and use it to power, for example, and HTTP service that uses http4s and jsoniter.
```
kubukoz: **issue:** `and HTTP service` -> `an HTTP service`

we all need more Grammarly in our lives

modules/docs/markdown/06-guides/testing.md
```diff
@@ -0,0 +1,41 @@
...
+
+Another reason you may think of testing the Smithy4s interpreters is to make sure that their implementation, such as the way they encode something in JSON, lines up with some requirements that you have. This is fair enough, but realize that the thing you are really testing here is your understanding of how Smithy4s works. That may be worth your while, or it may not be. The decision is ultimately up to you. If this is your reason for testing, you may want to explore contract-based testing as outlined below.
```
kubukoz, line 25: **thought:** I think there's another reason why people do this: 

They don't trust they wrote the right Smithy specs (or that they didn't make a mistake in refactoring them).

Let me explain:

imagine I decided with my colleagues that we're going to return a payload with a top-level string.

I write the spec:

```smithy
@http(...)
operation Foo {
  output := {
    sth: String
  }
}
```

I add a smithy4s implementation of this operation and don't write route-level tests for it. Turns out I messed up, the response actually looks like this:

```json
{
  "sth": "foo"
}
```

instead of what I expected to be `"foo"`.

---

so this is why the users would write a test on the routes, hitting the right URL and checking if the response is 

```scala
val response = Json.fromString("foo")
```

. Now, I don't think these are good tests because they run way too late after introducing the mistake, and a better solution (on a similar level of setup complexity) would be writing compliance tests. Then, if they want to be hardcore about it, contract testing...

kubukoz: I'd say this sounds like what you described as schema testing, but I've seen it also involve testing error responses based on the request.

modules/docs/markdown/06-guides/testing.md
```diff
@@ -0,0 +1,41 @@
...
+
+For your specific use case, you may have a need to employ schema-based and/or contract-based testing. Schema based testing would be one way of checking that schema changes over time do not impact existing clients. Contract-based testing could accomplish this same thing and more. However, contract based testing is usually more difficult to implement than schema-based testing so you may find it not to be worth it for your use case. In either case, remember to make sure you know _what_ it is you are testing and what you are not.
+
+## TL;DR
+
+Ultimately our recommendation is that you unit test all of your business logic as well as any custom implementations you have created on top of Smithy4s. You should be able to leave the rest of the unit testing up to us. However, your requirements, concerns, and experience may vary and we encourage you to do whatever will help you write the best software.
+
+Depending on your use case, you may find value in Schema or Contract testing. These methods can help you make sure your API is stable for consumers over time.
```
kubukoz, line 41: **suggestion:** can we recommend some resources on how to do schema and contract testing?

kubukoz:
If you agree with the points I made
    If I need to go into more depth on any part
    If I lacked precision in my descriptions anywhere

I think you did a great job on describing the problem, explaining what doesn't and what might need to be tested. It's probably as in-depth as it needs to be, and I agree for the most part - only thing I'm concerned about is in the comments above.

modules/docs/markdown/06-guides/testing.md
lewisjkl: Hah for real!

modules/docs/markdown/06-guides/testing.md
lewisjkl: Yeah I guess this falls into the category that I talked about of "testing your own understanding of Smithy/Smithy4s" which I think there is value in.. It just _feels_ like it shouldn't be necessary. Though I do agree that there is a certain sort of regression that it could prevent.

I think if we were to recommend that people do this kind of testing, we may want to provide something that makes it easier out of the box. Like for example we could have a validator that checks that the `examples` trait (or a new trait) values match what a request/response would actually look like for the operation given the protocol at hand.  Kind of like compliance tests, but here we are really checking that the encoding is what the user expects, not that smithy4s will produce the expected encoding. If any of that makes sense 😆

modules/docs/markdown/06-guides/testing.md
lewisjkl: Sure, do you have any in mind? I know pact is one, but I think that is a commercial thing (not positive) and I don't really love pushing people toward stuff that isn't OSS 🤷

Baccata (approved):

Baccata:
Good stuff, I'm happy to merge it as is and for it to be iterated over.

PR #749, opened 2023-01-20
Document smithy4sWildcardArgument

daddykotex:
cc @albertpchen

localhost_3000_smithy4s_docs_codegen_customisation_wildcard

Screen Shot 2023-01-20 at 16 29 14

### DIFF
--- a/modules/docs/markdown/04-codegen/01-customisation/08-wildcard.md
+++ b/modules/docs/markdown/04-codegen/01-customisation/08-wildcard.md
@@ -1,0 +1,19 @@
+---
+sidebar_label: Wildcard imports
+title: Adjust Scala wildcard
+---
+
+Scala has a specific syntax for wildcard argument in types. In Scala 2, that was the underscore: `_`. But with Scala 3, this is changing. See this [page](https://docs.scala-lang.org/scala3/reference/changed-features/wildcards.html) for more information.
+
+Smithy4s now has a way for you to control that and the good thing is that you probably don't have to worry about it. If you're using Smithy4s via `mill` or `sbt`, then it's taken care of you. It can be overriden via the following keys:
+
+* in mill, task: `def smithy4sWildcardArgument = "?" // or "_"`
+* in sbt, setting: `smithy4sWildcardArgument := "?" // or "_"`
+
+If you are using Smithy4s via the CLI, then they way to utilize this feature is through your Smithy specifications. The simplest approach is too add a file with the following content to you CLI invocation:
+
+```kotlin
+$version: "2"
+
+metadata smithy4sWildcardArgument = "?" // can also be `_`
+```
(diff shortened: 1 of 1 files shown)


modules/docs/markdown/04-codegen/01-customisation/08-wildcard.md
```diff
@@ -0,0 +1,19 @@
+---
+sidebar_label: Wildcard imports
```
kubukoz: **issue:** types, not imports

modules/docs/markdown/04-codegen/01-customisation/08-wildcard.md
```diff
@@ -0,0 +1,19 @@
...
+title: Adjust Scala wildcard
+---
+
+Scala has a specific syntax for wildcard argument in types. In Scala 2, that was the underscore: `_`. But with Scala 3, this is changing. See this [page](https://docs.scala-lang.org/scala3/reference/changed-features/wildcards.html) for more information.
```
kubukoz: **nitpick:** I've heard links with labels like `this`, `page`, etc. are bad for accessibility - IIRC the issue is that screen readers won't give any useful label, and users will be forced to listen to the entire URL and figure out themselves what the target is.

Also, these labels tend to be kinda useless for everyone else as well...

I would suggest being more descriptive about the target:

```suggestion
Scala has a specific syntax for wildcard argument in types. In Scala 2, that was the underscore: `_`. But with Scala 3, this is changing. See [the language reference page](https://docs.scala-lang.org/scala3/reference/changed-features/wildcards.html) for more information.
```

modules/docs/markdown/04-codegen/01-customisation/08-wildcard.md
```diff
@@ -0,0 +1,19 @@
...
+
+Smithy4s now has a way for you to control that and the good thing is that you probably don't have to worry about it. If you're using Smithy4s via `mill` or `sbt`, then it's taken care of you. It can be overriden via the following keys:
```
kubukoz:
```suggestion
Smithy4s now has a way for you to control that, and the good thing is that you probably don't have to worry about it. If you're using Smithy4s via `mill` or `sbt`, then it's taken care of you. It can be overriden via the following keys:
```

modules/docs/markdown/04-codegen/01-customisation/08-wildcard.md
```diff
@@ -0,0 +1,19 @@
...
+
+* in mill, task: `def smithy4sWildcardArgument = "?" // or "_"`
+* in sbt, setting: `smithy4sWildcardArgument := "?" // or "_"`
+
+If you are using Smithy4s via the CLI, then they way to utilize this feature is through your Smithy specifications. The simplest approach is too add a file with the following content to you CLI invocation:
```
kubukoz: **issue:** 
```suggestion
If you are using Smithy4s via the CLI, then they way to utilize this feature is through your Smithy specifications. The simplest approach is to add a file with the following content to your CLI invocation:
```

kubukoz (changes requested):

daddykotex:
Thanks for the thorough review

kubukoz (approved):

Issue #722, opened 2023-01-13
[AWS] Devise a way to test signature parity between the official AWS Java SDK and the smithy4s AWS SDK


Baccata:
Something that's really hard to test is the assessment that the signature algorithm works correctly. 

What we should do is run our AWS requests and requests performed by the AWS Java SDK (reconfigured to call upon an http4s client) against a mock http4s backend that'll intercept the requests metadata and communicate to the test via some Ref or Deferred. 

This should give us some confidence that the signature algorithm is correctly implemented. We can target specific methods of specific services, and add to them as we go.

daddykotex:
Isn't there a way to generate the signature that'd be use w/o actually making http calls? If so it could be setup in unit tests and be easier to debug/fix, etc.

Baccata:
It's hard to do, because the signature depends on the fact that the inputs are correctly serialised in http payloads and headers. They also depend on us retrieving the correct metadata from the service specification, which is part of the signature.

kubukoz:
> the AWS Java SDK (reconfigured to call upon an http4s client)

how do you even do this?? Seems like there'd be some work required to translate their request model into http4s, which sounds... like too much. Maybe we can just override the AWS endpoint to hit a local http4s server and capture requests this way?

Baccata:
> how do you even do this 

in the sdk v2 you can override the http client, just need to implement some interface

kubukoz:
right, so you have to translate. Maybe I'm overestimating how much effort that is.

daddykotex:
Maybe you don't have to overwrite or re-implement the client but just point it to a mock service you'd spin up on a given port?

kubukoz:
> Maybe we can just override the AWS endpoint to hit a local http4s server and capture requests this way?

that's what I meant here

Baccata:
customising the client is lighter-weight cause it can run fully in memory without exercising the tcp layer

Baccata:
plus from what I remember it's not that complicated

daddykotex:
Ok so I got a pretty good testing ground. When I run the code, I can easily retrieve the headers added by the AWS SDK as well as the version built on Smithy4s generated code. For example, when I run for a `listTable` dynamodb operation, I get the following `Authorization` header (added newline for clarity):

For AWS:
```
AWS4-HMAC-SHA256
Credential=ASIA*************/20230123/us-east-1/dynamodb/aws4_request
[...]

PR #756, opened 2023-01-23
Load file from credentials file (~/.aws/credentials)

daddykotex:
While working on #722 I realized we could not pull the info from `~/.aws/credentials` so I added it.

what do you think?

### DIFF
--- a/modules/aws-http4s/src/smithy4s/aws/http4s/package.scala
+++ b/modules/aws-http4s/src/smithy4s/aws/http4s/package.scala
@@ -17,7 +17,7 @@
 package smithy4s
 package aws
 
-import cats.effect.Temporal
+import cats.effect.Async
 import cats.effect.Resource
 import org.http4s.client.Client
 
@@ -27,7 +27,7 @@ package object http4s {
       private[this] val service: smithy4s.Service[Alg]
   ) {
 
-    def awsClient[F[_]: Temporal](
+    def awsClient[F[_]: Async](
         client: Client[F],
         awsRegion: AwsRegion
     ): Resource[F, AwsClient[Alg, F]] = for {
--- a/modules/aws/src/smithy4s/aws/AwsCredentialsFile.scala
+++ b/modules/aws/src/smithy4s/aws/AwsCredentialsFile.scala
@@ -16,15 +16,14 @@
 
 package smithy4s.aws
 
+import cats.effect.Resource
+import cats.effect.Sync
 import cats.syntax.all._
+import java.nio.file.Paths
 import smithy4s.aws.kernel.AWS_ACCESS_KEY_ID
 import smithy4s.aws.kernel.AWS_SECRET_ACCESS_KEY
 import smithy4s.aws.kernel.AWS_SESSION_TOKEN
 import smithy4s.aws.kernel.AwsCredentials
-
-import java.nio.file.Paths
-import cats.effect.Sync
-import cats.effect.Resource
 
 final case class AwsCredentialsFile(
     default: Option[AwsCredentials],
--- a/modules/aws/src/smithy4s/aws/AwsCredentialsFile.scala
+++ b/modules/aws/src/smithy4s/aws/AwsCredentialsFile.scala
@@ -16,14 +16,18 @@
 
 package smithy4s.aws
 
-import cats.effect.Resource
-import cats.effect.Sync
+import cats.effect.Concurrent
 import cats.syntax.all._
-import java.nio.file.Paths
+import fs2.io.file.Files
 import smithy4s.aws.kernel.AWS_ACCESS_KEY_ID
 import smithy4s.aws.kernel.AWS_SECRET_ACCESS_KEY
 import smithy4s.aws.kernel.AWS_SESSION_TOKEN
 import smithy4s.aws.kernel.AwsCredentials
+
+final case class AwsCredentialsFileException(
+    message: String,
+    cause: Throwable = null
+) extends RuntimeException(message, cause)
 
 final case class AwsCredentialsFile(
     default: Option[AwsCredentials],
@@ -31,36 +35,43 @@ final case class AwsCredentialsFile(
 )
 
 object AwsCredentialsFile {
-  def loadFromDisk[F[_]](home: String, profile: Option[String])(implicit
-      F: Sync[F]
+  private type LoadCredentials[A] = Either[AwsCredentialsFileException, A]
+
+  def fromDisk[F[_]](path: fs2.io.file.Path, profile: Option[String])(implicit
+      F: Concurrent[F],
+      Files: Files[F]
   ): F[AwsCredentials] = {
-    loadFile(home)
-      .use(lines => processFileLines(lines).pure[F])
-      .flatMap(creds =>
-        profile
-          .map(p =>
-            creds.profiles
-              .get(p)
-              .toRight(
-                new RuntimeException(
-                  s"Profile `$p` was not found in the credentials file."
-                )
-              )
-              .liftTo[F]
+    for {
+      lines <- Files.readUtf8Lines(path).compile.toList
+      creds <- processFileLines(lines).liftTo[F]
+
+      defaultProfile = creds.default
+      requestedProfile <- profile.traverse { p =>
+        creds.profiles
+          .get(p)
+          .toRight(
+            AwsCredentialsFileException(
+              s"Profile `$p` was not found in the credentials file."
+            )
           )
-          .getOrElse(
-            creds.default
-              .toRight(
-                new RuntimeException(
-                  s"No default profile is available in the credentials file."
-                )
-              )
-              .liftTo[F]
+          .liftTo[F]
+      }
+
+      finalProfile <- requestedProfile
+        .orElse(defaultProfile)
+        .toRight(
+          AwsCredentialsFileException(
+            "No default profile is available in the credentials file."
           )
-      )
+        )
+        .liftTo[F]
+
+    } yield finalProfile
   }
 
-  private[aws] def processFileLines(lines: List[String]): AwsCredentialsFile = {
+  private[aws] def processFileLines(
+      lines: List[String]
+  ): LoadCredentials[AwsCredentialsFile] = {
     def inProfile(
         rest: List[String],
         currentProfile: String,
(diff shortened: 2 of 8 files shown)


modules/aws/src/smithy4s/aws/AwsCredentialsProvider.scala
```diff
@@ -36,24 +39,33 @@ object AwsCredentialsProvider {
...
-  )(implicit F: Temporal[F]): Resource[F, F[AwsCredentials]] = {
-    Resource
-      .eval(fromEnv[F])
-      .map(F.pure)
-      .orElse(refreshing[F](fromECS(httpClient)))
+  )(implicit F: Async[F]): Resource[F, F[AwsCredentials]] = {
+    val env = Resource.eval(fromEnv[F]).map(F.pure)
+    val fromDisk = Resource.eval(loadFromDisk[F]).map(F.pure)
+    env
+      .orElse(fromDisk)
```
Baccata: `fromDisk` should be lowest priority. On local setups, solutions like [aws-vault](https://github.com/99designs/aws-vault) or [finto](https://github.com/threadwaste/finto) exist to emulate the AWS runtimes and should take precedence over the config filec.

modules/aws/src/smithy4s/aws/AwsCredentialsFile.scala
```diff
@@ -0,0 +1,158 @@
...
+
+    val data = lookingForProfile(lines.filter(_.trim.nonEmpty), Map.empty)
+    AwsCredentialsFile.profilesFromMap(data)
+  }
+
+  private def loadFile[F[_]: Sync](home: String): Resource[F, List[String]] =
+    Resource
+      .fromAutoCloseable(
+        Sync[F].delay(
+          scala.io.Source
```
Baccata: Unfortunately, this is not gonna be working on node environments. Let's use [fs2.io](https://fs2.io/#/io?id=files) instead, that'll work on the three environments.

modules/aws/src/smithy4s/aws/AwsCredentialsFile.scala
```diff
@@ -0,0 +1,158 @@
...
+            .fromFile(Paths.get(home, ".aws", "credentials").toUri())
```
Baccata: let's not hardcode this here, but inject the `fs2.io.files.Path` indicating the config file all the way here

modules/aws-http4s/src/smithy4s/aws/http4s/package.scala
```diff
@@ -27,7 +27,7 @@ package object http4s {
       private[this] val service: smithy4s.Service[Alg]
   ) {
 
-    def awsClient[F[_]: Temporal](
+    def awsClient[F[_]: Async](
```
Baccata, line 30: this is not binary compatible but we're not making a strong promise regarding the AWS modules at this point, moreover I'm pretty confident it's unlikely to break any user code .

modules/aws/src/smithy4s/aws/AwsCredentialsFile.scala
```diff
@@ -0,0 +1,158 @@
...
+    val others = dataPerProfile
+      .filterNot(_._1 == "default")
+      .map { case (profile, data) =>
+        profile -> credentialsFromMap(data)
+      }
+    AwsCredentialsFile(default, others)
+  }
+}
+
+object Profile {
```
Baccata: this object and its contents should be private

modules/aws/src/smithy4s/aws/AwsCredentialsFile.scala
```diff
@@ -0,0 +1,158 @@
...
+import smithy4s.aws.kernel.AWS_SESSION_TOKEN
+import smithy4s.aws.kernel.AwsCredentials
+
+final case class AwsCredentialsFile(
+    default: Option[AwsCredentials],
+    profiles: Map[String, AwsCredentials]
+)
+
+object AwsCredentialsFile {
+  def loadFromDisk[F[_]](home: String, profile: Option[String])(implicit
```
Baccata: don't put `home` there, but rather the full location of the file (using fs2.io.files.Path).

modules/aws/src/smithy4s/aws/AwsCredentialsFile.scala
```diff
@@ -0,0 +1,158 @@
...
+  ): F[AwsCredentials] = {
+    loadFile(home)
+      .use(lines => processFileLines(lines).pure[F])
+      .flatMap(creds =>
+        profile
+          .map(p =>
+            creds.profiles
+              .get(p)
+              .toRight(
+                new RuntimeException(
```
Baccata: let's create actual exceptions for this

modules/aws/src/smithy4s/aws/AwsCredentialsFile.scala
```diff
@@ -0,0 +1,158 @@
...
+                  s"No default profile is available in the credentials file."
+                )
+              )
+              .liftTo[F]
+          )
+      )
+  }
+
+  private[aws] def processFileLines(lines: List[String]): AwsCredentialsFile = {
```
Baccata: I'm pretty certain that this can fail, the type should reflect that

modules/aws/src/smithy4s/aws/AwsCredentialsFile.scala
```diff
@@ -0,0 +1,158 @@
...
+    default: Option[AwsCredentials],
+    profiles: Map[String, AwsCredentials]
+)
+
+object AwsCredentialsFile {
+  def loadFromDisk[F[_]](home: String, profile: Option[String])(implicit
+      F: Sync[F]
+  ): F[AwsCredentials] = {
+    loadFile(home)
+      .use(lines => processFileLines(lines).pure[F])
```
Baccata: It's unlikely you can safely `pure` it. It should be `F.delay()` to have a chance to catch the exceptions. Alternatively, the `processFileLines` should return `F` or `Either`

modules/aws/src/smithy4s/aws/AwsCredentialsFile.scala
```diff
@@ -0,0 +1,158 @@
...
+                new RuntimeException(
+                  s"Profile `$p` was not found in the credentials file."
+                )
+              )
+              .liftTo[F]
+          )
+          .getOrElse(
+            creds.default
+              .toRight(
+                new RuntimeException(
```
Baccata: ditto https://github.com/disneystreaming/smithy4s/pull/756/files#r1084974649

modules/aws/src/smithy4s/aws/AwsCredentialsFile.scala
```diff
@@ -0,0 +1,158 @@
...
+  def loadFromDisk[F[_]](home: String, profile: Option[String])(implicit
+      F: Sync[F]
+  ): F[AwsCredentials] = {
+    loadFile(home)
+      .use(lines => processFileLines(lines).pure[F])
+      .flatMap(creds =>
+        profile
+          .map(p =>
+            creds.profiles
+              .get(p)
```
Baccata: you should be able to call `getOrElse(default)` there. In any case, it's too much [hadouken](https://tse3.mm.bing.net/th?id=OIP.CEQDgPo1GeSaTXpZXudjtwHaFQ&pid=Api) for my taste

modules/aws/src/smithy4s/aws/AwsCredentialsFile.scala
Baccata: Also, let's add a test for this with a temp file (created with fs2.io) to ensure that it works on the 3 platforms

modules/aws/src/smithy4s/aws/AwsCredentialsProvider.scala
```diff
@@ -36,24 +39,33 @@ object AwsCredentialsProvider {
...
       session = SysEnv.envValue(AWS_SESSION_TOKEN).toOption
     } yield AwsCredentials.Default(keyId, accessKey, session)
-    either match {
-      case Right(value) => F.pure(value)
-      case Left(t)      => F.raiseError(t)
-    }
+    either.liftTo[F]
+  }
+
+  def loadFromDisk[F[_]](implicit F: Sync[F]): F[AwsCredentials] = {
```
Baccata: let's just call this `fromDisk` (drop the `load`, it's implied), taking a `Profile` for the profile as a parameter. Then create a `getProfileFromEnv[F[_]]` method (profiles may be fetched from elsewhere, a config file or whatnot).  Then wire these methods in the `default` method. 


Also, let's make a newtype for `Profile` (look at `AwsRegion`) for inspiration.

Baccata:
The initiative is very welcome, however it needs polishing, in particular for cross-platform usage. 

Generally speaking: and in particular in native/sjs, when it comes to using the java stdlib (or the scala stdlib that delegates to the java one for I/O stuff), it's likely that things will compile, but it may happen that they don't link. 

In this case, scala-native does implement a subset of java.nio (using pure scala that complies with the same signatures as the java implementation) for files, but it's not the case for sjs.

daddykotex:
> The initiative is very welcome, however it needs polishing, in particular for cross-platform usage.
> 
> Generally speaking: and in particular in native/sjs, when it comes to using the java stdlib (or the scala stdlib that delegates to the java one for I/O stuff), it's likely that things will compile, but it may happen that they don't link.
> 
> In this case, scala-native does implement a subset of java.nio (using pure scala that complies with the same signatures as the java implementation) for files, but it's not the case for sjs.

100%, I opened the PR as soon as I got something working to avoid spending too much time on the implementation in case we decided to drop it. Good to know, it could make it's way in!

I'll make the changes that you highlighted, thanks for the thorough review.

At first, I implemented it with fs2-io but I had to add a dependency and so I decided to roll it back. But I did not think that it would buy me the multi-platform ticket!

Baccata:
> add a dependency and so I decided to roll it back 

yeah the line is very gray there, but I think the reality is that I'm actually contemplating merging `aws` and `aws-http4s` in a single module because we're gonna need `fs2.Stream` in `aws` and it's silly to not acknowledge that http4s is likely gonna be the only backend we ever support for that particular stuff.

modules/aws/src/smithy4s/aws/AwsCredentialsFile.scala
kubukoz: `use` will catch it, but yeah it's better to  `delay` for extra sanity

daddykotex: wrapped it all up in `Either`s

modules/aws/src/smithy4s/aws/AwsCredentialsFile.scala
daddykotex: IMO, if you ask for profile X and it's not found, it should fail

not fallback to the default without you knowing

modules/aws/src/smithy4s/aws/AwsCredentialsProvider.scala
daddykotex: I encoded it as `def getProfileFromEnv: Option[String]` because `F` is not required. Unless we change `SysEnv.envValue` to return a `F`

modules/aws/src/smithy4s/aws/AwsCredentialsFile.scala
daddykotex: just let me know what you prefer and I'll get to it

but as an example, the AWS sdk does the following if you have a `aws_profile` env var, but not a corresponding profile in the credentials file:
```
ProfileCredentialsProvider(profileName=MY_PROFILE_SSO, profileFile=ProfileFile(profilesAndSectionsMap=[{default=Profile(name=default, properties=[region])}, {}])): Profile file contained no credentials for profile 'MY_PROFILE_SSO': ProfileFile(profilesAndSectionsMap=[{default=Profile(name=default, properties=[region])}, {}])
```

modules/aws/src/smithy4s/aws/AwsCredentialsProvider.scala
kubukoz: CE went with `F` for env reading, YMMV though

kubukoz:
> yeah the line is very gray there, but I think the reality is that I'm actually contemplating merging aws and aws-http4s in a single module because we're gonna need fs2.Stream in aws and it's silly to not acknowledge that http4s is likely gonna be the only backend we ever support for that particular stuff.

I'm of two minds on this, on one hand that's true, but having a split allows users to reuse some of our work in case they want to support another backend. I guess we can always split it again, if someone asks for that.

modules/aws/src/smithy4s/aws/AwsCredentialsFile.scala
Baccata: > IMO, if you ask for profile X and it's not found, it should fail

Agreed.

Baccata:
> but having a split allows users to reuse some of our work in case they want to support another backend 

I was thinking that at some point, but now I don't really want to optimise for those hypothetical users. If they disagree with http4s they'll disagree with fs2 most likely, and abstracting over streaming libraries is not a thought I want to entertain.

Baccata (approved):

PR #757, opened 2023-01-25
Fixes a bug that prevent the inclusion of upstream libraries

Baccata:
Fixes a bug reported by @ahjohannessen, that leads normal dependencies to not be included in the classpath that Smithy4s receives to load the specs from.

### DIFF
--- a/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
+++ b/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
@@ -329,7 +329,9 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
           .map(_.allFiles)
           .fold(uw => throw uw.resolveException, identity)
       }
-      getJars(dependenciesTask.value)
+      // Forcing configurations to None as the dynamic fetcher seems to emit
+      // every moduleID that has a configuration.
+      getJars(dependenciesTask.value.map(_.withConfigurations(None)))
     }
 
   def transitiveLibraryDependencies: Def.Initialize[Task[Seq[ModuleID]]] = {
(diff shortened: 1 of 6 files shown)


modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -329,7 +329,9 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
           .map(_.allFiles)
           .fold(uw => throw uw.resolveException, identity)
       }
-      getJars(dependenciesTask.value)
+      // Forcing configurations to None as the dynamic fetcher seems to emit
```
daddykotex: did you mean `omit` instead of `emit`?

kubukoz: well now I _actually_ understand what's happening :D

kubukoz:
It'd be very nice to sbt-test this.

ahjohannessen:
I have tested this against my repro setup: 
 
- https://github.com/ahjohannessen/smithy4s-repro-app
 - https://github.com/ahjohannessen/smithy4s-repro-lib
 - https://github.com/ahjohannessen/smithy4s-repro-specs
 
and in the originial private work repo that surfaced this bug.

All seems good and fixed 👍

Baccata:
> It'd be very nice to sbt-test this.

Right, I'm gonna be honest : I'm not managing to inline a repro : I don't understand what combination of settings led to the problematic situation in the first place, SBT is just too opaque to me. 

Considering the fact that it's not a runtime-library problem, and that I have way more important fish to fry, I'm just gonna merge the fix. If people want to attempt at creating a repro that actually fails the tests when the fix is removed, they are more than welcome.

PR #764, opened 2023-01-27
More efficient parsing of json objects and arrays

plokhotnyuk:
It can speed up parsing of collections of simple values.


Baccata (approved):
Impressive work, as usual. Thank you so much Andriy !

PR #769, opened 2023-01-30
Generate hints for collection members

### LINKED ISSUE #600: Hints on collection members aren't generated
Spawned by https://github.com/disneystreaming/smithy4s/pull/599#discussion_r1020671261.

You can apply some hints on collection members, e.g.:

```smithy
list Strings {
  @deprecated @documentation("foo") member: String
}
```

and they don't end up in the generated schema. This should be fixed, so that the collection's schema includes the hints.

Also: check if the dynamic model reads these hints.

albertpchen:
fixes https://github.com/disneystreaming/smithy4s/issues/600

Updates renderer to emit code to add hints for `list` members and `map` keys/values.

Also add a test to check that the dynamic model reads these hints.

### DIFF
--- a/modules/codegen/test/src/smithy4s/codegen/internals/RendererSpec.scala
+++ b/modules/codegen/test/src/smithy4s/codegen/internals/RendererSpec.scala
@@ -1,0 +1,88 @@
+/*
+ *  Copyright 2021-2022 Disney Streaming
+ *
+ *  Licensed under the Tomorrow Open Source Technology License, Version 1.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     https://disneystreaming.github.io/TOST-1.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


Baccata (approved):
Great work, thanks a lot !

modules/codegen/test/src/smithy4s/codegen/internals/RendererSpec.scala
```diff
@@ -0,0 +1,88 @@
...
+      contents.find(_.contains("object ListWithMemberHints")) match {
+        case None =>
+          fail(
+            "No generated scala file contains valid ListWithMemberHints definition"
+          )
+        case Some(code) => code
+      }
+
+    val memberSchemaString =
+      """string.addHints(smithy.api.documentation("listFoo"), smithy.api.deprecated(message = None, since = None))"""
```
kubukoz, line 47: **issue:** This is wrong: it should be `Documentation`, not `documentation`. Somehow we don't hit that in other shapes... 🤔

kubukoz: moved to https://github.com/disneystreaming/smithy4s/issues/774

PR #776, opened 2023-01-31
Fix missing struct member docs bug

### LINKED ISSUE #771: Generate member Scaladoc for structs that don't have their own documentation
After #731, if you try to generate code from

```smithy
$version: "2"

namespace demo

structure Hello {
  /// the field
  field: String
}
```

then you get this:

```scala
package demo

/* imports omitted */
case class Hello(field: Option[String] = None)
object Hello extends ShapeTag.Companion[Hello] {
/* omitted */
}
```

However, if you add docs to `Hello` itself:

```diff
namespace demo
 
+/// the struct
 structure Hello {
   /// the field
   field: String

```

then you get full-blown docs on the struct:

```scala
/** the struct
  * @param field
  *   the field
  */
case class Hello(field: Option[String] = None)
```

## Expectation

I think the behavior with the original Smithy snippet should be that the struct gets this Scaladoc:

```scala
/** @param field
  *   the field
  */
```

msosnicki:
closes #771

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
@@ -757,6 +757,10 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
 
     def split(s: String) =
       s.replace("*/", "\\*\\/").linesIterator.toList
+    val shapeDocs = shape
+      .getTrait(classOf[DocumentationTrait])
+      .asScala
+      .fold(List.empty[String])(doc => split(doc.getValue()))
     val memberDocs: Map[String, List[String]] =
       if (shape.isUnionShape()) Map.empty
       else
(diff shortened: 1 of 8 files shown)


msosnicki:
closes #771

daddykotex (approved):
it's nice that the smithy specs were already documented and the examples just get updated :D

kubukoz:
fyi @msosnicki you need to write the "closes" thing in the description of the PR itself, otherwise GH won't catch it. Moved :)

modules/example/src/smithy4s/example/SomeVector.scala
```diff
@@ -10,7 +10,7 @@ import smithy4s.schema.Schema.vector
 
 object SomeVector extends Newtype[Vector[String]] {
   val id: ShapeId = ShapeId("smithy4s.example", "SomeVector")
-  val hints: Hints = Hints()
+  val hints: Hints = Hints.empty
```
kubukoz, line 13: **praise:** nice one

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -757,6 +757,10 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
 
     def split(s: String) =
       s.replace("*/", "\\*\\/").linesIterator.toList
+    val shapeDocs = shape
+      .getTrait(classOf[DocumentationTrait])
+      .asScala
+      .fold(List.empty[String])(doc => split(doc.getValue()))
```
kubukoz, line 763: **nitpick:** 

```suggestion
      .foldMap(doc => split(doc.getValue()))
```

kubukoz (approved):

PR #778, opened 2023-01-31
Tweak refinements

daddykotex:
I'm trying to use refinements on a type that's imported like:

```scala
import thingy.types.`ISO 3166-1 alpha-2`
```

If I try, I get the following error:

```
[error] [ERROR] alloy.common#countryCodeFormat: Error validating trait `smithy4s.meta#refinement`.targetType: String value provided for `smithy4s.meta#Classpath` must match regular expression: ^(?:_root_\.)?(?:[a-zA-Z][\w]*\.?)*$ | TraitValue N/A:0:0
```

The first fix address the strict pattern. The second fix deals with the fact that you can't use a refinement if your trait is used on member shape.

See added test for an example.


lewisjkl (approved):
LGTM, good find on the member shapes 👍

modules/protocol/resources/META-INF/smithy/smithy4s.meta.smithy
```diff
@@ -85,11 +85,13 @@ structure refinement {
...
 /// e.g. com.test_out.v2.Something
-@pattern("^(?:_root_\\.)?(?:[a-zA-Z][\\w]*\\.?)*$")
+/// e.g. com.test_out.v2.`Something`
+@pattern("^(?:_root_\\.)?(?:[a-zA-Z`][\\w]*\\.?)*$")
 string Classpath
 
 /// e.g. com.test_out.v2.Something._
-@pattern("^(?:_root_\\.)?(?:[a-zA-Z][\\w]*\\.?)*\\.(?:_|given)$")
+/// e.g. com.test_out.v2.`Something`._
+@pattern("^(?:_root_\\.)?(?:[a-zA-Z`][\\w]*\\.?)*\\.(?:_|given)$")
```
kubukoz, line 94: **thought:** I'm _this_ close to saying we should just accept anything.

Baccata:
> The second fix deals with the fact that you can't use a refinement if your trait is used on member shape.

I'm not sure I like this, it was a rather conscious decision, the rationale being that it makes sense from a domain modelling perspective that constraints should equate new types (even though AWS does not do that). 

If you're gonna enable that, at the very least we need an example (as in, the example project) of refinement usage involving a member, to verify that it generate sensible and compilable code.

daddykotex:
> I'm not sure I like this, it was a rather conscious decision, the rationale being that it makes sense from a domain modelling perspective that constraints should equate new types (even though AWS does not do that).

I agree with this statement. In the sense that at the end of the day, we want the refinements to point to a new type. What I'm proposing is using a type that's define outside of the specification. 

> If you're gonna enable that, at the very least we need an example (as in, the example project) of refinement usage involving a member, to verify that it generate sensible and compilable code.

I'll add an example in an extra commit and if it is not satisfying, happy to close it and find a different encoding for my issue

daddykotex:
I've updated the `Age` refinement example. I removed the arbitraty `Age` and `PersonAge` initially defined (for no real reason, maybe?) that are not used and instead added the `ageFormat` annotation directly on the members concerned. This remove `Age` and `PersonAge` and directly go to `refined.Age` where the validation is defined. See [this commit](https://github.com/disneystreaming/smithy4s/pull/778/commits/36093d1962c3210172d6290f75747d436fc4339d)

modules/example/src/smithy4s/example/StructureWithRefinedTypes.scala
```diff
@@ -4,17 +4,20 @@ import smithy4s.Hints
 import smithy4s.Schema
 import smithy4s.ShapeId
 import smithy4s.ShapeTag
+import smithy4s.example.refined.Age
+import smithy4s.example.refined.Age.provider._
+import smithy4s.schema.Schema.int
 import smithy4s.schema.Schema.struct
 
-case class StructureWithRefinedTypes(age: Age, personAge: PersonAge, fancyList: Option[smithy4s.example.FancyList] = None, unwrappedFancyList: Option[smithy4s.example.refined.FancyList] = None, name: Option[smithy4s.example.Name] = None, dogName: Option[smithy4s.example.refined.Name] = None)
+case class StructureWithRefinedTypes(age: Option[Age] = None, personAge: Option[Age] = None, fancyList: Option[smithy4s.example.FancyList] = None, unwrappedFancyList: Option[smithy4s.example.refined.FancyList] = None, name: Option[smithy4s.example.Name] = None, dogName: Option[smithy4s.example.refined.Name] = None)
```
daddykotex: interesting: there was no option before, even though it was not `required`

Baccata: That's most likely because the `$version` clause is not defined in the smithy file, and the smithy 2.0 parser/assembler automatically adds `default` traits in some conditions when parsing smithy 1.0 files. One condition might be the absence of other traits.

daddykotex: I decided to try a few different things to test, but my observations are as follow:

Screenshot 2023-02-02 at 10 15 55

All cases have either a default or are using `Option` except the one where `ageFormat` is applied to a custom simple shape like `integer Age`.

Baccata: tbh, I'm having difficulties digesting this table without seeing the smithy code. Also, the information that's missing is what the model tells us once it's been loaded : what is the condition for having an automatically applied `@default` trait ?

daddykotex: So your assumptions when you said:
> That's most likely because the $version clause is not defined in the smithy file, and the smithy 2.0 parser/assembler automatically adds default traits

Is absolutely true, but it does not seem to be related to traits but instead is just because the type is `integer`: https://smithy.io/2.0/guides/migrating-idl-1-to-2.html?highlight=boxed#converting-primitive-root-level-shapes-from-1-0 or in [code](https://github.com/awslabs/smithy/blob/56263d64885688118225acfba565d5eea6d1bc3c/smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelInteropTransformer.java#L111)

And the [full list of types that have default is here](https://github.com/awslabs/smithy/blob/56263d64885688118225acfba565d5eea6d1bc3c/smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelInteropTransformer.java#L62).

But I feel like the `default` value supplied does not play well with the refinement type. For all we know, the refinement provider could consider `0` to be an invalid age.

modules/example/src/smithy4s/example/StructureWithRefinedMember.scala
```diff
@@ -0,0 +1,23 @@
...
+import smithy4s.Hints
+import smithy4s.Schema
+import smithy4s.ShapeId
+import smithy4s.ShapeTag
+import smithy4s.example.refined.Age
+import smithy4s.example.refined.Age.provider._
+import smithy4s.schema.Schema.int
+import smithy4s.schema.Schema.struct
+
+case class StructureWithRefinedMember(otherAge: Option[Age] = None)
```
daddykotex, line 12: added another structure instead of messing with the other one, but I opened an issue to track what I've saw in [issue #791](https://github.com/disneystreaming/smithy4s/issues/791)

daddykotex:
IMO, this is reviewable/mergeable as the main issue here is unrelated to the changes here. So if everyone is on-board that this has value, this can be reviewed / merged. Otherwise this could be closed.

modules/example/src/smithy4s/example/StructureWithRefinedTypes.scala
Baccata: Agreed. I wonder what happens if you put something like `@range(min: 1)` on top of the integer... does the default get generated still ?

daddykotex: yes, haha: https://github.com/daddykotex/scala-scripts/blob/main/smithy-playground.scala#L38

Baccata (approved):

PR #781, opened 2023-02-01
Adding operation @param documentation

### LINKED ISSUE #772: Generate method documentation for operations
Another follow-up to #731.

## Reproduction

Generate code for this spec:

```smithy
$version: "2"

namespace demo

service HelloService {
  operations: [GetHello]
}

/// op
operation GetHello {
  input: Hello
}

/// the struct
structure Hello {
  /// the field
  field: String
}
```

## Current behavior

The generated operation looks like this

```scala
  /** op */
  def getHello(field: Option[String] = None): F[Hello, Nothing, Unit, Nothing, Nothing]
```

## Expectation

Operations would have their inputs documented using `@param` annotations (just like structures that have these members):

```scala
  /** op
    * @param field
    *   the field
    */
  def getHello(field: Option[String] = None): F[Hello, Nothing, Unit, Nothing, Nothing]
```

this should also avoid the issue from #771, and generate `@param` docs even if the operation itself isn't documented.

msosnicki:
closes #772

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
@@ -195,15 +195,20 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     )
   }
 
-  private def documentationAnnotation(hints: List[Hint]): Lines = {
+  private def documentationAnnotation(
+      hints: List[Hint],
+      skipMemberDocs: Boolean = false
+  ): Lines = {
     hints
       .collectFirst { case h: Hint.Documentation => h }
       .foldMap { doc =>
         val shapeDocs: List[String] = doc.docLines
-        val memberDocs: List[String] = doc.memberDocLines.flatMap {
-          case (memberName, text) =>
-            s"@param $memberName" :: text.map("  " + _)
-        }.toList
+        val memberDocs: List[String] =
+          if (skipMemberDocs) List.empty
+          else
+            doc.memberDocLines.flatMap { case (memberName, text) =>
+              s"@param $memberName" :: text.map("  " + _)
+            }.toList
 
         val maybeNewline =
           if (shapeDocs.nonEmpty && memberDocs.nonEmpty) List("", "") else Nil
@@ -276,7 +281,10 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
         newline,
         ops.map { op =>
           lines(
-            documentationAnnotation(op.hints),
+            documentationAnnotation(
+              op.hints,
+              op.hints.contains(Hint.PackedInputs)
+            ),
             deprecationAnnotation(op.hints),
             line"def ${op.methodName}(${op.renderArgs}): F[${op
               .renderAlgParams(genNameRef.name)}]"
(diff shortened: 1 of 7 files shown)


modules/codegen/test/src/smithy4s/codegen/internals/ScaladocSpec.scala
```diff
@@ -0,0 +1,109 @@
...
+        |
+        |/// Struct docs
+        |structure DocumentedStruct {
+        |  /// int docs
+        |  @required
+        |  int: Integer
+        |  /// struct docs
+        |  struct: UndocumentedStruct
+        |  
+        |  @required
```
msosnicki, line 21: Added this to highlight this problem (maybe it's just smithy synax rules in play?).
If the ordering of annotation and comment is like here (first annotation and then docs), then in AWS model, the documentation trait is missing, only required trait is there. I think the same applies for other annotations, like `@packedInputs` before the comment on an operation/service etc.

modules/codegen/test/src/smithy4s/codegen/internals/ScaladocSpec.scala
```diff
@@ -0,0 +1,109 @@
...
+        |  /// required before comment  
+        |  string: String
```
msosnicki, line 23: Added this to highlight this problem (maybe it's just smithy synax rules in play?).
If the ordering of annotation and comment is like here (first annotation and then docs), then in AWS model, the documentation trait is missing, only `@required` trait is there. I think the same applies for other annotations, like `@packedInputs` before the comment on an operation/service etc.

For example, in the @packedInput test, changing spec to
```
@packedInputs
/// packed operation docs
operation PackedHelloOp {
  input: Hello
}
```
will make test fail

msosnicki: For example, in the `@packedInput` test, changing spec to 
```
@packedInputs
/// packed operation docs
operation PackedHelloOp {
  input: Hello
}
```
will make test fail

kubukoz: > AWS model,

I think I know what you mean, but to avoid confusion I'd call it the Smithy model (as opposed to the Smithy4s model) :)

that seems odd, perhaps worth an issue in awslabs/smithy

msosnicki: Cool, I will replicate with minimal example, and report there. To me too it seemed like a bug, cause I couldn't find the documentation  trait anywhere else, it seemed to be just lost.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -276,7 +281,10 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
         newline,
         ops.map { op =>
           lines(
-            documentationAnnotation(op.hints),
+            documentationAnnotation(
+              op.hints,
+              op.hints.contains(Hint.PackedInputs)
```
Baccata, line 286: 👍

modules/codegen/test/src/smithy4s/codegen/internals/ScaladocSpec.scala
Baccata: Awesome, thanks for the heads-up, that's good to know 👍

Baccata (approved):

Baccata:
@kubukoz, let us know if you're happy with it

msosnicki:
For reference, the issue reported upstream: https://github.com/awslabs/smithy/issues/1600

kubukoz (approved):

PR #782, opened 2023-02-01
handle null values in default trait

lewisjkl:
So the fix here comes down to when someone uses an empty default trait (null value implied) like the following:

```smithy
structure MyStruct {
  @default
  a: String
}
```

The [javaDoc](https://github.com/awslabs/smithy/blob/9ae748ccc2b67d3c3acffdd027db084c568b3e6b/smithy-model/src/main/java/software/amazon/smithy/model/traits/DefaultTrait.java) on the `DefaultTrait` implies that this should be handled as a "zero value" for String (empty string). They read:

> Indicates that a shape is defaulted, meaning a zero value is used when one is not provided.

However, I can't find any other areas where this is documented in depth. I can reach out to the smithy maintainers to ask for clarification, but I wanted to post this before I do so in case I am missing something.


lewisjkl:
Another thing to remember is that if we did want to take the approach outlined in this PR, I need to do more testing with simple shapes which have defaults e.g.

```smithy
@default
string MyString
```

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -683,6 +683,28 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
 
     }
 
+  private def imputeZeroValuesOnDefaultTraits(
```
lewisjkl, line 686: Without this, all of the values in the Default Hint would be Document.null. If that were the case, then the codecs and such would need to be aware of what to do with a null default. Imputing the implied value seems preferable

yisraelU:
> 

This might have some useful info
https://github.com/awslabs/smithy/blob/main/designs/defaults-and-model-evolution.md

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -683,6 +683,28 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+      shape: Shape
+  ): List[Trait] = shape.getAllTraits().asScala.values.toList.map {
+    case default: DefaultTrait if default.toNode == Node.nullNode =>
```
kubukoz, line 689: got to ask the obvious question, does the dynamic module handle this? or would things just blow up?

Baccata: It won't blow up, because [this](https://github.com/disneystreaming/smithy4s/blob/ba331d19fc45db1aba3331a9997776393a00bb05/modules/core/src/smithy4s/schema/Schema.scala#L84), but it won't be giving the same behaviour either. 

However, @kubukoz does have a point. This should also be handled at runtime in that `getDefaultValue` method.

lewisjkl: Just to clarify, you mean handled at runtime instead of changing the generated code? Or in addition to?

lewisjkl:
I ended up creating an issue to try to figure out if I am going at this the right way: https://github.com/awslabs/smithy/issues/1601

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
Baccata: In addition to. The generated code bit is required for default parameters of scala constructors and methods. The runtime bit is required for interpreters applied on dynamic service instances.

lewisjkl: For implementing the `getDefaultValue` method, I will need to know the type of schema I am working with so I know which value to transform the null default to. However, I don't think a simple pattern match on the current schema will work due to lazy schema.. do I need to implement a simple schema visitor to accomplish this? Or do you have another recommendation?

Baccata: Yeah I think a simple schema visitor might actually be a good idea.

modules/core/src/smithy4s/schema/DefaultValueSchemaVisitor.scala
```diff
@@ -0,0 +1,135 @@
...
+
+  def refine[A, B](
+      schema: Schema[A],
+      refinement: Refinement[A, B]
+  ): GetDefaultValue[B] = {
+    val res = apply(schema)
+    new GetDefaultValue[B] {
+      def getDefault: Option[Document] = res.getDefault
+    }
+  }
```
lewisjkl: Not really positive I handled these correctly

modules/core/src/smithy4s/schema/Schema.scala
```diff
@@ -79,7 +79,7 @@ sealed trait Schema[A]{
   final def refined[B]: PartiallyAppliedRefinement[A, B] = new PartiallyAppliedRefinement[A, B](this)
 
   final def getDefault: Option[Document] =
-      this.hints.get(smithy.api.Default).map(_.value)
+      DefaultValueSchemaVisitor(this).getDefault
```
lewisjkl: Should the `DefaultValueSchemaVisitor` be held in a lazy val or something like that?

modules/core/test/src/smithy4s/schema/DefaultValueSpec.scala
```diff
@@ -0,0 +1,96 @@
...
+      val name = stringValue
+      val value = stringValue
+      val hints = Hints.empty
+    }
+    case object Foo extends FooBar("foo", 0)
+    val e: Schema[FooBar] = Schema.enumeration[FooBar](List(Foo))
+    testCaseOpt(e, None)
+  }
+
+  test("bijection") {
```
daddykotex, line 107: I think this test should be `refinement`

daddykotex: so you'd be missing a test for `bijection`

lewisjkl: Good catch, will add 👍

daddykotex: I think these tests essentially confirm that we dont need to worry about: https://github.com/disneystreaming/smithy4s/pull/782/files#r1099073507

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
yisraelU: TIL 'impute'

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -683,6 +683,28 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+      val tpe = shape.asMemberShape().asScala match {
+        case Some(memShape) => model.getShape(memShape.getTarget).get.getType
+        case None           => shape.getType
+      }
+      val newNode = tpe match {
+        case ShapeType.STRING  => Node.from("")
+        case ShapeType.MAP     => Node.objectNode()
```
yisraelU: what about ShapeType.LIST to  Node.arrayNode() ?

modules/core/src/smithy4s/schema/DefaultValueSchemaVisitor.scala
```diff
@@ -0,0 +1,135 @@
...
+      hints: Hints,
+      tag: Primitive[P]
+  ): GetDefaultValue[P] = from(
+    hints,
+    tag match {
+      case PShort      => Some(Document.fromInt(0))
+      case PString     => Some(Document.fromString(""))
+      case PFloat      => Some(Document.fromDouble(0))
+      case PDouble     => Some(Document.fromDouble(0))
+      case PTimestamp  => None
```
yisraelU: I'm not sure if Timestamp also has a default, it definitely sounds like from here that its coherent.
 ```timestamp: 0 seconds since the Unix epoch```
https://github.com/awslabs/smithy/blob/ee0b4ff90daaaa5101f32da936c25af8c91cc6e9/designs/defaults-and-model-evolution.md

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
lewisjkl: Good catch, I will add tests for all types and fix that

modules/core/src/smithy4s/schema/Schema.scala
Baccata: I think you have the wrong way to go at it. When you talked about using a SchemaVisitor, I was thinking of producing the `A`, not the Document. The visitor should be used in `getDefaultValue`, not `getDefault` (unless there's something I've not thought of).

Baccata: https://github.com/disneystreaming/smithy4s/pull/792

Baccata (changes requested):
See https://github.com/disneystreaming/smithy4s/pull/792

lewisjkl:
I am going to add some docs explaining how we handle the empty defaults (I'm fine if that happens in another PR though too)

Baccata (approved):

Baccata:
Good work !

modules/core/src/smithy4s/schema/Schema.scala
```diff
@@ -79,9 +79,13 @@ sealed trait Schema[A]{
   final def refined[B]: PartiallyAppliedRefinement[A, B] = new PartiallyAppliedRefinement[A, B](this)
 
   final def getDefault: Option[Document] =
-      this.hints.get(smithy.api.Default).map(_.value)
+    this.hints.get(smithy.api.Default).map(_.value)
+
+  final def getDefaultValue: Option[A] = getDefault.flatMap {
+    case Document.DNull => this.compile(DefaultValueSchemaVisitor)
```
kubukoz, line 85: this is sick asl, thank you!

PR #787, opened 2023-02-03
adt trait for sealed hierarchies with common members

lewisjkl:
Adds a new trait `smithy4s.meta#adt` that when applied to Union shapes will lead to different rendering. Namely, the rendering will take common mixins from members of the ADT and move them to the level of the sealed trait. More documentation and test case examples to come as the PR progresses.

TODO:
- [x] Add validator that will check for proper usage of the adt trait
- [x] More example usages in sampleSpecs/examples to check proper handling of edge cases
- [x] Add documentation


modules/example/src/smithy4s/example/TestAdt.scala
```diff
@@ -0,0 +1,66 @@
...
+
+sealed trait TestAdt extends AdtMixinOne with AdtMixinTwo with scala.Product with scala.Serializable {
+  @inline final def widen: TestAdt = this
+}
+object TestAdt extends ShapeTag.Companion[TestAdt] {
+  val id: ShapeId = ShapeId("smithy4s.example", "TestAdt")
+
+  val hints: Hints = Hints.empty
+
+  case class AdtOne(lng: Option[Long] = None, sht: Option[Short] = None, blb: Option[ByteArray] = None, str: Option[String] = None) extends TestAdt with AdtMixinTwo with AdtMixinOne with AdtMixinThree
```
lewisjkl: Will update so that AdtMixinTwo and AdtMixinOne don't render here. AdtMixinThree still should since it is not a part of the `TestAdt` level.

modules/example/src/smithy4s/example/OrderNumber.scala
```diff
@@ -10,7 +10,7 @@ import smithy4s.schema.Schema.int
 object OrderNumber extends Newtype[Int] {
   val id: ShapeId = ShapeId("smithy4s.example", "OrderNumber")
   val hints: Hints = Hints(
-    smithy.api.Default(smithy4s.Document.fromDouble(0.0d)),
+    smithy.api.Box(),
```
Baccata, line 13: wut ?

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -217,6 +218,10 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
         }
       }
 
+      private def isPartOfAdt(shape: Shape): Boolean = {
+        getAdtParent(shape).isDefined
```
Baccata: might as well bake the `hasTrait(classOf[AdtMemberTrait])` check in there

modules/example/src/smithy4s/example/OrderNumber.scala
lewisjkl: I believe this is from changing the smithy file in sampleSpecs over to use the smithy version 2 syntax directly. Prior it was a version 1 file being loaded which would have added these other traits.

Baccata: I thought the "box" trait had been removed :/

lewisjkl: Yeah.. this has really confused me before and I honestly don't remember what is going on here since it seems to conflict their documentation. But, here is an example using smithy-model version 1.27.2 (the latest)
CleanShot 2023-02-06 at 09 37 29

lewisjkl:
Let me know if there are other example usages that you think would be helpful to add.

daddykotex (commented):
Great work, for me, the only missing piece for this PR is a bit of documentation to complement what we have for `@adtMember` in [ADTs section of the docs](https://disneystreaming.github.io/smithy4s/docs/codegen/customisation/adts)

lewisjkl:
Added some documentation. As an aside, should we now deprecate the `adtMember` trait (with the deprecated smithy trait)? I am thinking we should since it does basically the same thing. Thoughts?

modules/docs/markdown/04-codegen/01-customisation/02-adts.md
```diff
@@ -38,7 +38,111 @@ case class InStoreOrder(id: OrderNumber, locationId: Option[String] = None)
...
+
+Using the `adt` trait does come with some restrictions. First are requirements for the union which is annotated with the `adt` trait:
+
+1. The union must contain at least one member
+2. The union's members must only target structure shapes
+
+Additionally, there are some requirements that are added onto the structure shapes that the union targets:
+
+1. The structures must NOT be the target of any other union, structure, etc. They can only be the target in the ONE union that is annotated with the `adt` trait.
+2. The structures must contain at least one member.
```
Baccata, line 76: I'm not sure that one is necessary, but happy to merge the PR as is and revise later

Baccata (approved):

PR #789, opened 2023-02-06
Fixes Scala code being copied in resources folder

Baccata:
Under the right condition, before this fix, generated scala code would be copied to the resources folder, which would lead to some compilation errors due to duplication, because the compiler picks up resources as sources for some reason.

The bug was introduced in https://github.com/disneystreaming/smithy4s/pulls?page=2&q=is%3Apr+is%3Aclosed, which generate some smithy files under the `sourcesManaged / "smithy"` directory, which is then used as an input for the codegen. The codegen has a resource generator to it, which copies local files under input directories to the META-INF/smithy resource folder.

However, when generating a namespace that starts with "smithy", the Scala code gets generated in the same sourcesManaged / "smithy" directory. 

~Arguably, the problem is that it should be generated in sourcesManaged / "scala" / "smithy", but at this point I'm unsure what the implications for doing that are. So instead, I'm ensuring that whatever gets copied in META-INF/smithy are actual smithy files.~ 

After a quick test, it looks like Intellij is cool with the sources being generated under `sourcesManaged / "scala"`, so I've changed the behaviour of the SBT plugin to do that and have adjusted the tests accordingly.

### DIFF
--- a/modules/codegen-plugin/src/sbt-test/codegen-plugin/smithy-rules/build.sbt
+++ b/modules/codegen-plugin/src/sbt-test/codegen-plugin/smithy-rules/build.sbt
@@ -1,0 +1,10 @@
+lazy val root = (project in file("."))
+  .enablePlugins(Smithy4sCodegenPlugin)
+  .settings(
+    scalaVersion := "2.13.10",
+    libraryDependencies ++= Seq(
+      "com.disneystreaming.smithy4s" %% "smithy4s-core" % smithy4sVersion.value,
+      "software.amazon.smithy" % "smithy-rules-engine" % smithy4s.codegen.BuildInfo.smithyVersion % Smithy4s
+    ),
+    Compile / smithy4sAllowedNamespaces := List("smithy.rules")
+  )
--- a/modules/codegen/src/smithy4s/codegen/internals/SmithyResources.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/SmithyResources.scala
@@ -29,9 +29,15 @@ private[smithy4s] object SmithyResources {
 
   def produce(
       resourceOutputFolder: os.Path,
-      localSmithyFiles: List[os.Path],
+      specs: List[os.Path],
       namespaces: List[String]
   ): List[os.Path] = {
+
+    val localSmithyFiles = specs.flatMap { spec =>
+      if (os.isDir(spec)) os.walk(spec).filter(_.ext == "smithy")
+      else if (spec.ext == "smithy") List(spec)
+      else Nil
+    }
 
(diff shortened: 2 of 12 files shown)


modules/codegen-plugin/src/sbt-test/codegen-plugin/smithy-rules/build.sbt
```diff
@@ -0,0 +1,10 @@
+lazy val root = (project in file("."))
```
Baccata, line 1: new scripted test, verifying that generating code from `smithy-rules-engine` does not lead to compilation error.

daddykotex (approved):
Good catch wow,

I like the addition of `scala` because it gets us closer to what the sources directory look like under a typical jvm project. I'm surprised I never noticed it was missing

kubukoz:
> The bug was introduced in [disneystreaming/smithy4s/pulls?page=2 (is:closed)](https://github.com/disneystreaming/smithy4s/pulls?page=2&q=is%3Apr+is%3Aclosed+sort%3Aupdated-desc),

broken link, can you update it? I don't know which PR this was about

modules/codegen/src/smithy4s/codegen/internals/SmithyResources.scala
```diff
@@ -29,10 +29,16 @@ private[smithy4s] object SmithyResources {
 
   def produce(
       resourceOutputFolder: os.Path,
-      localSmithyFiles: List[os.Path],
+      specs: List[os.Path],
       namespaces: List[String]
   ): List[os.Path] = {
 
+    val localSmithyFiles = specs.flatMap { spec =>
+      if (os.isDir(spec)) os.walk(spec).filter(_.ext == "smithy")
```
kubukoz: What about json files?

kubukoz: oh wait, is this about the files generated in https://github.com/disneystreaming/smithy4s/pull/736 ?

kubukoz (approved):
approving with the assumption that this is, indeed, about https://github.com/disneystreaming/smithy4s/pull/736

modules/codegen/src/smithy4s/codegen/internals/SmithyResources.scala
Baccata: actually you have a point

Baccata: It is indeed about #736, but you are correct about the fact that this filter right here needs to allow for json files. This is addressed in [f25c1fa](https://github.com/disneystreaming/smithy4s/pull/789/commits/f25c1fa1fbf9ae108f59a9b8a4e242d42e5ad80e)

Issue #791, opened 2023-02-06
Issue with refinement on structure member with a model v1


daddykotex:
When parsing a V1 model, Smithy will add `DefaultTrait` under some conditions. One of the refinements examples we have in this repo is affected by this:
1. a root shape is introduced: `integer Age`
2. it is used in V1 model

The member shape gets a `DefaultTrait` added to it, and so when we render the refinement type, we don't wrap in `Option`. But we also don't set a default value.

Documented here: https://smithy.io/2.0/guides/migrating-idl-1-to-2.html?highlight=boxed#converting-primitive-root-level-shapes-from-1-0

Source is [in smithy-model](https://github.com/awslabs/smithy/blob/56263d64885688118225acfba565d5eea6d1bc3c/smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelInteropTransformer.java)

What should we do?

-------

I've setup a small scala-cli script just to confirm that this is what is actually going on:

```scala
//> using scala "2.13.10"
//> using lib "software.amazon.smithy:smithy-model:1.27.0"

import software.amazon.smithy.model.Model
import scala.jdk.CollectionConverters._
import software.amazon.smithy.model.shapes.ShapeId

object Main extends App {
  val model = Model
    .assembler()
    .addUnparsedModel(
      "defaults.smithy",
      s"""|namespace atest
          |
          |structure DirectInt {
          |  member: Integer
          |}
          |
          |integer MyInt
          |structure IndirectInt {
          |  member: MyInt
          |}
          |""".stripMargin
    )
    .assemble()
    .unwrap()
  println(
    model.expectShape(ShapeId.from("atest#DirectInt$member")).getAllTraits()
  )
  println(
    model.expectShape(ShapeId.from("atest#IndirectInt$member")).getAllTraits()
  )
// {}
// {smithy.api#default=Trait `smithy.api#default`, defined at defaults.smithy [9, 3]}
}
```

lewisjkl:
Kind of related (inverse?) discussion here: https://github.com/disneystreaming/smithy4s/pull/787#discussion_r1097633603

daddykotex:
yes, I used your example to generate mine :D

Baccata:
> What should we do?

I believe we should wrap it in Option, in this case. Most of the time, it hardly makes sense for a refined type to have a default value.

daddykotex:
To be clear:  we found the issue because the model v1 loading adds a default trait.

[...]

PR #792, opened 2023-02-08
Revise default value implementation

Baccata:
* Default for things that have constraints (ie refinements) do not make sense
* It's more straightforward and performant to generate a value of the target type directly


daddykotex (approved):
Wow that's the kind of PR that makes me realize I overcomplicate most of my work in Smithy4s because I still don't have the full picture in my mind!

modules/core/src/smithy4s/schema/DefaultValueSchemaVisitor.scala
```diff
@@ -18,128 +18,79 @@ package smithy4s
...
   def refine[A, B](
       schema: Schema[A],
       refinement: Refinement[A, B]
-  ): GetDefaultValue[B] = {
-    val res = apply(schema)
-    new GetDefaultValue[B] {
-      def getDefault: Option[Document] = res.getDefault
-    }
-  }
+  ): Option[B] = None
```
daddykotex, line 92: At first, I was hopeful this would solve: https://github.com/disneystreaming/smithy4s/issues/791, but because the `SmithyToIR` works directly with Smithy `Shape`, I don't think it will... right?

Relevant decision happens in this piece of code: https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala#L845

Baccata: You are correct, it won't solve that issue, unfortunately

lewisjkl (approved):
This is awesome, thanks for taking the time 👍

PR #794, opened 2023-02-08
Add an enumeration member to Enumeration.Value

daddykotex:
Also expose `hints` and `id` so that we can use them when working with on `Enumeration.Value`.

This allows me to write code like this:

```scala
   // assume this hint exist and it can be used on `enum`
   //@trait(selector: "enum")
   //structure SomeHint { name: String }
    type SomeHint = ???
    
    def extract(tag: Enumeration.Value) = {
      val name = tag.enumeration.hints // can go up to the enum, and retrieve the hints
        .get[SomeHint]
        .map(_.name)
        .getOrElse(tag.enumeration.id.name) // can go up to the enum, and retrieve the id
      name -> tag.value

    }
    def process(value: Long, enumValues: Enumeration.Value*) = {
      val nameAndValues: Seq[(String, String)] = enumValues.map { extract }

      // implement
      ???
    }
```


modules/json/test/src/smithy4s/http/json/FaceCard.scala
```diff
@@ -20,10 +20,12 @@ import smithy4s.schema.Schema._
 
 sealed abstract class FaceCard(_value: String, _name: String, _intValue: Int)
```
daddykotex: there is already a generated `FaceCard` (from a smithy spec) in the example module, shouldn't we use that instead?

lewisjkl: Imo you can definitely update it to be the case. I think I duplicated it at the time because the PRs were in flight at roughly the same time.. don't remember to be honest 😅

Baccata (approved):

Baccata:
Good work !

PR #799, opened 2023-02-09
Emit warning on default trait on refined types

daddykotex:
Following #795 , it suggested that the validator emits a warning if a user adds a `@default` annotation on a shape that's acting as a refinement trait. That's what I added in this PR.

I grabbed the synthetic default trick from the smithy-model source code and the tests show that it works.


Baccata (approved):

Baccata:
That's great, thank you !

PR #800, opened 2023-02-09
add smithy model transformation docs

lewisjkl:
Adding docs about how to add transformations which change the smithy model prior to codegeneration.


modules/docs/markdown/06-guides/smithy-transformations.md
```diff
@@ -0,0 +1,124 @@
...
+    "RemoveBeforeCodegenTransformation"
+  }
+
+  def transform(x: TransformContext) = {
+    val toRemove = x
+      .getModel()
+      .getShapesWithTrait(new ToShapeId() {
+        override def toShapeId(): ShapeId =
+          ShapeId.from("preprocessed#removeBeforeCodegen")
+      })
```
kubukoz: ShapeId implements ToShapeId, you can just do 

```suggestion
      .getShapesWithTrait(ShapeId.from("preprocessed#removeBeforeCodegen"))
```

modules/docs/markdown/06-guides/smithy-transformations.md
```diff
@@ -0,0 +1,124 @@
...
+
+Now we will walk through the steps of actually implementing the Smithy model transformer. The `smithy-build` Java library provides an interface for these called `ProjectionTransformer`. Here is an example of a transformer that will remove the members marked with the `removeBeforeCodegen` trait as discussed above.
+
+```scala
+final class RemoveBeforeCodegenTransformation extends ProjectionTransformer {
+  def getName() = {
+    "RemoveBeforeCodegenTransformation"
+  }
+
+  def transform(x: TransformContext) = {
```
kubukoz: suggestion: x -> ctx?

kubukoz: also add a type annotation, it'd be nice to show what the output is

modules/docs/markdown/06-guides/smithy-transformations.md
```diff
@@ -0,0 +1,124 @@
...
+    x.getTransformer().removeShapes(x.getModel(), toRemove)
+  }
+}
+```
+
+Here we create a `ProjectionTransformer` named `RemoveBeforeCodegenTransformation`. Inside the `transform` method we remove all shapes that are marked with the `removeBeforeCodegen` trait and return the resultant model.
+
+## Register the Transformer
+
+We need to register the Transformer so that the Smithy tooling will be able to find it when it does a lookup for all `ProjectionTransformer` implementations. We do this by creating a file at `src/main/resources/META-INF/services/` called `software.amazon.smithy.build.ProjectionTransformer`. Inside of this file, we need to place the fully qualified classpath that leads to our transformer (e.g. `preprocessed.transformations.RemoveBeforeCodegenTransformation`).
```
kubukoz: Let's add a note saying this is the SPI mechanism from the java stdlib.

Also I'd drop that last sentence, show the actual file contents, and maybe comment on it later with something like

> this file will be used by the [SPI mechanism](https://docs.oracle.com/javase/tutorial/sound/SPI-intro.html) used by the Smithy tooling to "register" the class as a transformer that can be used.

modules/docs/markdown/06-guides/smithy-transformations.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Wire up in SBT
+
+In our `build.sbt` file we will create a new project called `transformers` that looks as follows:
+
+```scala
+lazy val transformers = (project in file("transformers"))
```
kubukoz: more than meets the eyeeeeee

modules/docs/markdown/06-guides/smithy-transformations.md
```diff
@@ -0,0 +1,124 @@
...
+      .getShapesWithTrait(new ToShapeId() {
+        override def toShapeId(): ShapeId =
+          ShapeId.from("preprocessed#removeBeforeCodegen")
+      })
+    x.getTransformer().removeShapes(x.getModel(), toRemove)
+  }
+}
+```
+
+Here we create a `ProjectionTransformer` named `RemoveBeforeCodegenTransformation`. Inside the `transform` method we remove all shapes that are marked with the `removeBeforeCodegen` trait and return the resultant model.
```
kubukoz: nitpick: I'm not sure this line is that helpful as it says only _what_ we are doing, some of these should be pretty clear from the code itself.

kubukoz: also if we're gonna keep it, I'd mention that the "name" of the transformer doesn't have to match the class name and is defined by the method we implement

modules/docs/markdown/06-guides/smithy-transformations.md
```diff
@@ -0,0 +1,124 @@
...
+This is the model that will ultimately be fed into the Smithy4s code generation tooling.
+
+## Implement the Transformer
+
+Now we will walk through the steps of actually implementing the Smithy model transformer. The `smithy-build` Java library provides an interface for these called `ProjectionTransformer`. Here is an example of a transformer that will remove the members marked with the `removeBeforeCodegen` trait as discussed above.
+
+```scala
+final class RemoveBeforeCodegenTransformation extends ProjectionTransformer {
+  def getName() = {
+    "RemoveBeforeCodegenTransformation"
```
kubukoz: worth mentioning that this name is significant

modules/docs/markdown/06-guides/smithy-transformations.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Wire up in SBT
+
+In our `build.sbt` file we will create a new project called `transformers` that looks as follows:
+
+```scala
+lazy val transformers = (project in file("transformers"))
+  .settings(
+    name := "transformers",
+    libraryDependencies += "software.amazon.smithy" % "smithy-build" % smithy4s.codegen.BuildInfo.smithyVersion
```
kubukoz: we should probably not encourage users to rely on this BuildInfo value. I'll defer to @Baccata for an opinion on this though

modules/docs/markdown/06-guides/smithy-transformations.md
```diff
@@ -0,0 +1,124 @@
...
+  )
+```
+
+This project is where our `RemoveBeforeCodegenTransformation` will live. As such, it requires the `smithy-build` dependency so it can access the `ProjectionTransformer` interface.
```
kubukoz: "where our transformation will live" would be fine I think

modules/docs/markdown/06-guides/smithy-transformations.md
```diff
@@ -0,0 +1,124 @@
...
+
+Now, in our project that is using the Smithy4s SBT plugin (`.enablePlugins(Smithy4sCodegenPlugin)`) we need to add the following settings:
```
kubukoz: "sbt" is written in lowercase :P

modules/docs/markdown/06-guides/smithy-transformations.md
```diff
@@ -0,0 +1,124 @@
...
+
+```scala
+...
```
kubukoz: nitpick level 1000:

```suggestion
// ...
```

modules/docs/markdown/06-guides/smithy-transformations.md
```diff
@@ -0,0 +1,124 @@
...
+.enablePlugins(Smithy4sCodegenPlugin)
+.settings(
+  ...
+  // transformers
+  Compile / smithy4sModelTransformers += "RemoveBeforeCodegenTransformation",
```
kubukoz: Worth mentioning again that this is the name returned by the method

kubukoz: maybe we can also change this name so that it's clearly different from the class name? sth like `remove-before-codegen`. I'm not sure if there's any convention for this.

modules/docs/markdown/06-guides/smithy-transformations.md
```diff
@@ -0,0 +1,124 @@
...
+  Compile / smithy4sAllDependenciesAsJars += (transformers / Compile / packageBin).value
+)
+```
+
+Here we are telling the smithy4s SBT plugin which transformers should be applied prior to code generation. In this case, "RemoveBeforeCodegenTransformation" is the only one. From there, we are telling the plugin that it should use the transformers SBT project we created above as a dependency. Adding it this way will make sure that your transformer is NOT added as a dependency to anything other than the Smithy4s plugin.
```
kubukoz: > make sure that your transformer is NOT added as a dependency to anything other than the Smithy4s plugin.

how about

> make sure that your transformer is NOT added as a compile-time or run-time dependency of your Scala code, just as a codegen-time dependency of the Smithy4s plugin.

modules/docs/markdown/06-guides/smithy-transformations.md
```diff
@@ -0,0 +1,124 @@
...
+├── build.sbt
+├── core
+│   └── src
+│       └── main
+│           ├── scala
+│           │   └── com
+│           │       └── example
+│           │           └── Main.scala
+│           └── smithy
+│               └── preproccessed.smithy // The first smithy snippet shown above
```
kubukoz:
```suggestion
│               └── preproccessed.smithy // The Smithy models that will be transformed and used for codegen
```

modules/docs/markdown/06-guides/smithy-transformations.md
```diff
@@ -0,0 +1,124 @@
...
+├── project
+│   ├── build.properties
+│   └── plugins.sbt
+└── transformers
+    └── src
+        └── main
+            ├── resources
+            │   └── META-INF
+            │       └── services
+            │           └── software.amazon.smithy.build.ProjectionTransformer // The file which registers our ProjectionTransformer
```
kubukoz: I'd try flattening the directories to save the poor readers some horizontal scrolling

```suggestion
            ├── resources/META-INF/services
            │   └── software.amazon.smithy.build.ProjectionTransformer // The file which registers our ProjectionTransformer
```

kubukoz: same below

modules/docs/markdown/06-guides/smithy-transformations.md
```diff
@@ -0,0 +1,124 @@
...
+
+We need to register the Transformer so that the Smithy tooling will be able to find it when it does a lookup for all `ProjectionTransformer` implementations. We do this by creating a file at `src/main/resources/META-INF/services/` called `software.amazon.smithy.build.ProjectionTransformer`. Inside of this file, we need to place the fully qualified classpath that leads to our transformer (e.g. `preprocessed.transformations.RemoveBeforeCodegenTransformation`).
+
+## Wire up in SBT
+
+In our `build.sbt` file we will create a new project called `transformers` that looks as follows:
+
+```scala
+lazy val transformers = (project in file("transformers"))
+  .settings(
```
kubukoz: Q: shouldn't this module use scala 2.12? or, more specifically, we should mention that it has to be the Scala version used by your sbt, which in sbt 1.x is a 2.12.

modules/docs/markdown/06-guides/smithy-transformations.md
```diff
@@ -0,0 +1,124 @@
...
+  ...
+  // transformers
+  Compile / smithy4sModelTransformers += "RemoveBeforeCodegenTransformation",
+  Compile / smithy4sAllDependenciesAsJars += (transformers / Compile / packageBin).value
+)
+```
+
+Here we are telling the smithy4s SBT plugin which transformers should be applied prior to code generation. In this case, "RemoveBeforeCodegenTransformation" is the only one. From there, we are telling the plugin that it should use the transformers SBT project we created above as a dependency. Adding it this way will make sure that your transformer is NOT added as a dependency to anything other than the Smithy4s plugin.
+
+## Directory Structure
```
kubukoz: before this, as a user I'd like to show the actual outcome of the transformation. Maybe we can show a trimmed snippet of the generated case class?

the snippet would be sth like

```scala
// note the lack of the `id` field
case class MyStruct(name: String)
//...
```

modules/docs/markdown/06-guides/smithy-transformations.md
```diff
@@ -0,0 +1,124 @@
...
+
+This project is where our `RemoveBeforeCodegenTransformation` will live. As such, it requires the `smithy-build` dependency so it can access the `ProjectionTransformer` interface.
+
+Now, in our project that is using the Smithy4s SBT plugin (`.enablePlugins(Smithy4sCodegenPlugin)`) we need to add the following settings:
+
+```scala
+...
+.enablePlugins(Smithy4sCodegenPlugin)
+.settings(
+  ...
```
kubukoz: I was thinking about possibly adding smithy4s-core as a dependency, but existing users already know to do that, and we have these dots to say it's not all 😅

kubukoz (changes requested):
I like the example, it's simple and easy to follow, just got some nitpicks and a couple improvement ideas

modules/docs/markdown/06-guides/smithy-transformations.md
Baccata: We should very much encourage users to rely on this BuildInfo value.

Baccata (approved):

PR #813, opened 2023-02-15
Dynamic compliance tests

yisraelU:
[@Baccata's notes]

* Changes the default timestamp format used in Json/Documents to be `EPOCH_SECOND`, in an effort to align with the semantics of the AWS protocols. This is **behaviour breaking**, but is for the better as it aligns the behaviour of what Smithy4s does with the behaviour of the AWS-owned smithy libraries/code-generators, and reduces the hoops we have to jump through so that Smithy4s-provided tooling can talk to AWS-provided tooling. It is a hard decision to make, as we've tried to avoid breaking behaviour in the past, but is very much the correct decision for the sanity of the maintainers, and the users who read the Smithy documentation. 

* Amends the compliance tests module to make it easier to run tests from a dynamically loaded module. This is gonna let us pull new tests as they come by virtue of bumping versions, without having to inspect the test specifications to know what we need to add.  

* Wires a bunch of cherry-picked tests from the Aws Rest Json protocol. This allows us to assert how much work there is and what are the differences between our "SimpleRestJson" protocol and the official "AWS Rest Json" protocol. Eventually, we'd like to support the latter out of the box, but devil's in the detail and it's a slow process.


modules/aws-http4s/test/src-ce3/smithy4s/compliancetests/ProtocolComplianceTest.scala
```diff
@@ -0,0 +1,154 @@
...
+                           ): DynamicSchemaIndex => List[ComplianceTest[IO]] = { dynamicSchemaIndex =>
+    dynamicSchemaIndex
+      .getService(shapeId)
+      .toList
+      .flatMap(wrapper => {
+        HttpProtocolCompliance
+          .clientAndServerTests(SimpleRestJsonIntegration, wrapper.service)
+      })
+  }
+  val whitelist = Set(
```
Baccata:
```suggestion
  val allowlist = Set(
```

modules/aws-http4s/test/src-ce3/smithy4s/compliancetests/ProtocolComplianceTest.scala
```diff
@@ -0,0 +1,154 @@
...
+    "EndpointOperation",
+    "EndpointWithHostLabelOperation",
+    "TestBodyStructure",
+    "TestPayloadStructure",
+  ).map(id => IdRef(s"aws.protocoltests.restjson#$id"))
+
+  private def loadDynamic(
+                           doc: Document
+                         ): Either[PayloadError, DynamicSchemaIndex] = {
+    Document.decode[Model](doc).map(_.filter(whitelist)).map(load)
```
Baccata:
```suggestion
    Document.decode[Model](doc).map(_.filter(allowlist)).map(load)
```

modules/transformers/src/ProtocolTransformer.scala
```diff
@@ -0,0 +1,23 @@
...
+  def transform(ctx: TransformContext): Model = {
+    val rename: JHMap[ShapeId, ShapeId] = {
+      val map = new JHMap[ShapeId, ShapeId]()
+      map.put(
+        ShapeId.from("aws.protocols#restJson1"),
+        ShapeId.from("alloy#simpleRestJson")
+      )
+      map
+    }
+    ctx.getTransformer().renameShapes(ctx.getModel(), rename)
```
Baccata: I think there's a `mapTrait` that would be better suited : you're not supposed to rename the protocol, but change the protocol the service knows about

yisraelU: ok good to know

yisraelU: actually quite annoying as the HttpRequestTestCase are not traits rather Nodes that contain the protocol

modules/http4s/test/src-ce3/smithy4s/compliancetests/ProtocolComplianceTest.scala
```diff
@@ -131,22 +150,69 @@ object ProtocolComplianceTest extends EffectSuite[IO] with BaseCatsSuite {
...
+    )
+  }
+
+  def readConfigFromJar: IO[Seq[Byte]] = {
+    Resource
+      .fromAutoCloseable(
+        IO(
+          new java.io.BufferedReader(
+            new java.io.InputStreamReader(
+              getClass.getResourceAsStream("/META-INF/json/test-config.json")
```
Baccata: That method is JVM specific : you'll have to dump it onto the file system at build time.

Alternatively, we could put this configuration into a smithy-compliant json file as metadata, so that it'd already benefit from the build logic you added to dump the model as json file. 

We'd then add a `def metadata: Map[String, Document]` to our `DynamicSchemaIndex`, that could be really useful in the future.

yisraelU: yes , I am aware , thats why its still in draft , i'm just trying to dot a couple i's before tackling this

yisraelU: I actually had a plan to move this to Alloy , so the dumping logic , decoding can live in one place , to provide reuse for the Disney addition too

yisraelU: done

modules/compliance-tests/src/smithy4s/compliancetests/internals/MappedHintsSchemaVisitor.scala
```diff
@@ -0,0 +1,144 @@
...
+  Schema,
+  SchemaAlt,
+  SchemaField,
+  SchemaVisitor
+}
+import smithy4s.{kinds, Bijection, HintMask, Hints, Lazy, Refinement, ShapeId}
+import smithy4s.compliancetests.internals.MappedHintsSchemaVisitor.hintMask
+import smithy4s.schema.Schema._
+
+object MappedHintsSchemaVisitor {
```
Baccata: I think we could do this more concisely. Remind me the problem this solves again ?

Baccata: Alright, I remember ... I'm started to think the better course of action would be to align on what AWS does, and change the behaviour for 0.18. 

The maintenance burden induced by the difference is starting to be somewhat uncomfortable. I think aligning would be preferable, for our own sake, as well as for the sake of the ecosystem, considering the difference of default violates the principle of least-surprise ... 

In order to reduce the risk based on the change of behaviour in user's applications, we could add a smithy-level validator (in alloy) that would issue a warning for any timestamp shape connected transitively to a service with `alloy#simpleRestJson` stating that it's preferable for users to be explicit about timestamp formats ... 

What do you think ?

Baccata: @yisraelU, I've created the `series/0.18`. When you have the time, could you : 

- [x] Change this PR to target that new series branch
- [x] Sync this PR with the new target 
- [x] Change the default timestamp format to EPOCH_SECONDS in the relevant Json/Document schema visitors 
- [x] Remove the then un-necessary Service/Endpoint transformation 

Thank you in advance ;)

PR #819, opened 2023-02-17
PathEncode for Byte

yisraelU:
handle a Byte schema in the Schema Visitor for PathEncode


Baccata:
What about the decoding side ? Is it handled there ?

yisraelU:
> What about the decoding side ? Is it handled there ?

what decoding side ?
theres only encoding for paths , decoding is handled by http4s

Baccata:
That's only partially true, but I've had a look and we do handle it on the decoding side, in the Metadata decoder. At the end this is the function that's being called : 

* https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/core/src/smithy4s/schema/Primitive.scala#L117

PR #820, opened 2023-02-17
[http4s] Enrich Http Endpoint Exception message

yisraelU:
persists the cause of  endpoint failure into the message of the Exception thrown

### DIFF
--- a/modules/http4s/src/smithy4s/http4s/SmithyHttp4sReverseRouter.scala
+++ b/modules/http4s/src/smithy4s/http4s/SmithyHttp4sReverseRouter.scala
@@ -62,7 +62,7 @@ class SmithyHttp4sReverseRouter[Alg[_[_, _, _, _, _]], Op[_, _, _, _, _], F[_]](
           .left
           .map { e =>
             throw new Exception(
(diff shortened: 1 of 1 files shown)


daddykotex (approved):

modules/http4s/src/smithy4s/http4s/SmithyHttp4sReverseRouter.scala
```diff
@@ -62,7 +62,7 @@ class SmithyHttp4sReverseRouter[Alg[_[_, _, _, _, _]], Op[_, _, _, _, _], F[_]](
           .left
           .map { e =>
             throw new Exception(
-              s"Operation ${endpoint.name} is not bound to http semantics",
+              s"Operation ${endpoint.name} is not bound to http semantics: ${e.message}",
               e
```
kubukoz, line 66: I'm not sure this is very helpful, `e` is propagated to the exception as a cause so it should appear in stack traces. What's the usecase?

yisraelU: The stacktrace will just contain a cause that takes you to where the either was lifted not to the actual source

PR #822, opened 2023-02-20
Add JCodecs for Double and Float that handle NaN and Infinity

### LINKED ISSUE #760: Handle NaN/Infinity/-Infinity in the AWS protocols
See https://smithy.io/2.0/aws/protocols/aws-json-1_1-protocol.html#shape-serialization

[This is where](https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/aws/src/smithy4s/json/AwsSchemaVisitorJCodec.scala) the code should be amended. 

[This is the implementation](https://github.com/plokhotnyuk/jsoniter-scala/blob/master/jsoniter-scala-macros/shared/src/test/scala/com/github/plokhotnyuk/jsoniter_scala/macros/JsonCodecMakerSpec.scala#L580-L603)

The reason we're not doing this in the "standard" implementation is performance related.

zetashift:
Closes https://github.com/disneystreaming/smithy4s/issues/760

I couldn't figure out what `expecting` does so I just wrote a small description from the smithy docs for now.
The implementation is from jsoniter so can't say I did much lol.

EDIT: I should probably add some tests that test the decoding and encoding
EDIT2: I couldn't find anything more specific than `JCodec`

### DIFF
--- a/modules/aws/src/smithy4s/json/AwsSchemaVisitorJCodec.scala
+++ b/modules/aws/src/smithy4s/json/AwsSchemaVisitorJCodec.scala
@@ -23,8 +23,8 @@ import smithy4s.Timestamp
 import smithy4s.http.json.Cursor
 import smithy4s.http.json.JCodec
 import smithy4s.http.json.SchemaVisitorJCodec
-import smithy4s.schema.Primitive
 import smithy4s.schema.CompilationCache
+import smithy4s.schema.Primitive
 
 private[aws] class AwsSchemaVisitorJCodec(cache: CompilationCache[JCodec])
     extends SchemaVisitorJCodec(maxArity = 1024, cache) {
--- a/modules/aws/src/smithy4s/json/AwsSchemaVisitorJCodec.scala
+++ b/modules/aws/src/smithy4s/json/AwsSchemaVisitorJCodec.scala
@@ -82,7 +82,7 @@ private[aws] class AwsSchemaVisitorJCodec(cache: CompilationCache[JCodec])
         if (in.isCharBufEqualsTo(len, "NaN")) Float.NaN
         else if (in.isCharBufEqualsTo(len, "Infinity")) Float.PositiveInfinity
         else if (in.isCharBufEqualsTo(len, "-Infinity")) Float.NegativeInfinity
-        else in.decodeError("illegal double")
+        else in.decodeError("illegal float")
       } else {
         in.rollbackToken()
         in.readFloat()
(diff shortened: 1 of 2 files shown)


Baccata:
A few tests would certainly be appreciated 😄 

> I couldn't find anything more specific than JCodec 

I suppose I was thinking of something else ... doesn't matter.

zetashift:
> I suppose I was thinking of something else ... doesn't matter.
It could very well will be that I looked wrong :P, I'm not that great with typeclasses and the likes.

I'll add some tests! Since I don't know anything about weaver or testing in Scala in general, I'll look into it bit.

Is it correct that I should put the `AwsSchemaVisitorJCodec` tests in `./modules/aws/test/src/smithy4s/aws`?

modules/aws/src/smithy4s/json/AwsSchemaVisitorJCodec.scala
```diff
@@ -36,10 +36,73 @@ private[aws] class AwsSchemaVisitorJCodec(cache: CompilationCache[JCodec])
...
+        else if (in.isCharBufEqualsTo(len, "Infinity")) Double.PositiveInfinity
+        else if (in.isCharBufEqualsTo(len, "-Infinity")) Double.NegativeInfinity
+        else in.decodeError("illegal double")
+      } else {
+        in.rollbackToken()
+        in.readDouble()
+      }
+
+    def encodeValue(d: Double, out: JsonWriter): Unit =
+      if (_root_.java.lang.Double.isFinite(d)) out.writeVal(d)
```
plokhotnyuk: `_root_` could be removed here. It was required in jsoniter-scala for macros to test if they work in polluted namespaces.

zetashift: Oooh, didn't know that. Thank you!

modules/aws/test/src/smithy4s/aws/AwsSchemaVisitorJCodecTest.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  private val readerConfig: ReaderConfig = ReaderConfig
+    .withThrowReaderExceptionWithStackTrace(true)
+    .withAppendHexDumpToParseException(true)
+    .withCheckForEndOfInput(false)
+  
+  def readFromString[A: JCodec](str: String): A = {
+    com.github.plokhotnyuk.jsoniter_scala.core
+      .readFromString[A](str, readerConfig)
+  }
```
zetashift: Is this the way? :P

Baccata: I think so indeed

zetashift:
Did I do something wrong with the tests? It took 11m 40s to run the tests in the CI :O

Baccata:
@zetashift nope, you did nothing bad, it's just a LOT of compilation and tests happening. That's the cost of a monorepo. When we move to a more declarative build tool option (like bleep), it'll be easier to accurately compute what needs to be tested and what doesn't based on the changes in PRs.

Baccata (approved):
Awesome, thanks a lot for that !

PR #825, opened 2023-02-23
Amend partial `resource` support note in the docs

zetashift:
I couldn't amend https://github.com/disneystreaming/smithy4s/pull/716 because I couldn't suggest for deletion parts, so I opened a new PR, just to save the time.


Baccata (approved):

Baccata:
Many thanks for this !

PR #826, opened 2023-02-23
[Http4s] - client uses withEmptyBody instead of withEntity when there is no body

yisraelU:
this solves the issue of a content-length header provided on GET requests when there is no body

### DIFF
--- a/modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
+++ b/modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala
@@ -25,7 +25,6 @@ import org.http4s.Request
 import org.http4s.Response
 import org.http4s.Uri
 import org.http4s.client.Client
-import scodec.bits.ByteVector
 import smithy4s.kinds._
 import smithy4s.http._
 import smithy4s.schema.SchemaAlt
@@ -122,7 +121,7 @@ private[http4s] class SmithyHttp4sClientEndpointImpl[F[_], Op[_, _, _, _, _], I,
     val baseRequest = Request[F](method, uri, headers = headers)
     if (inputHasBody) {
       baseRequest.withEntity(input)
-    } else baseRequest.withEntity(ByteVector.empty)
+    } else baseRequest.withEmptyBody
   }
 
   private def outputFromResponse(response: Response[F]): F[O] =


Baccata:
@daddykotex, I think you had made some chances wrt empty bodies and whatnot. Can you give it a sanity check ?

daddykotex:
yes, this is basically a revert of a commit I made a while back
see: https://github.com/disneystreaming/smithy4s/pull/522 and https://github.com/disneystreaming/smithy4s/issues/515

daddykotex:
At some point, there was a discussion we should should have it on all requests method or not. Maybe this is the time to look into it more. I had created https://github.com/http4s/http4s/pull/6762 but the spec is hard to digest and I was not clear on what should be done so I sort of dropped the ball.

The problem is that the spec is vague and, imo, a GET request can have a `Content-Length: 0`, it's not invalid. It is just because some implementations in the wild reject it

yisraelU:
The Aws restJson1 protocol stipulates that there must not be a content-length header when its a GET request and a payload is not possible , meaning there is no input modeled in smithy targeting the payload (i.e. unit or just simply only header fields etc..)
So its not really talking about if there is no body , rather if there cant be a body
according to this test https://github.com/awslabs/smithy/blob/ff9df7d9b6f8c50511c3ddf816125e8f93d24339/smithy-aws-protocol-tests/model/restJson1/http-content-type.smithy#L273

Baccata:
There's a lot of nuance there : 

* @daddykotex adding an empty BitVector as a body was motivated by a very real issue raised by a user. The problem is that this solution automatically adds the `Content-Length: 0`  header, which conflicts with the notion that GET requests should not have `Content-Length` headers, which is a requirement that AWS validates for.  
* The [issue](https://github.com/disneystreaming/smithy4s/issues/515) in question was triggered by the *absence* of body, as a opposed to the *presence* of an empty-body. 
* `withEmptyBody` might well solve both things, but saying "that's what AWS says it should do" is not good enough : at the very least, we need to verify it will not lead to a regression to that issue. Looking at the code quickly, I'm not clear whether `withEntity(BitVector.empty)` and `withEmptyBody` have the same semantics wrt bodies. The annoying bit is that the issue is exhibited during interaction with real servers, and I don't really want to pull Ember in our tests. So what I'll ask of you @yisraelU, is to publish smithy4s locally and run it against the repro from the [issue](https://github.com/disneystreaming/smithy4s/issues/515) and see whether the problem re-appears. 
* By the way @daddykotex, this PR from @yisraelU is not a revert, because we don't know whether `request.withEmptyBody` as the main meaning as a request that never had an entity set. 
* If the problem is still fixed, then it's all good, otherwise, we'll need to pat-mat on the method, with the following logic : 

``` 
if (Set(Method.GET, Method.DELETE, Method.CONNECT, Method.TRACE).contains(method)) request.withEmptyBody else request.withEntity(BitVector.empty)
```

bilki:
FWIW the problem we have seen in the wild with this header, while using smithy, is that ALB is marking those requests as `non_compliant`, as part of this mechanism: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#desync-mitigation-mode Which could lead to blocked requests in the strictest mode. I guess other servers/load balancers implementations may as well ignore the header or not, and most likely none of them ask it to be mandatory for `GET`s.

yisraelU:
@Baccata I am going to take a look at everything .
I want to make it clear ,
1. I am not asserting we should do something because AWS does it 
2. AWS does **not** insist on this simply because its a GET request, rather its schema based . 
If the schema does not define a body , then a GET request should not contain the content-length header, If this is important to us , perhaps we should add a SchemaVisitor to track this

daddykotex:
All that is fair, and I did not mean `revert` in the sense that it will just reintroduce the problem I fixed, more in the sense that yes, this is familiar to me I explicitly introduced this change because there was an issue when no body was supplied.

We can come up with a better fix that'll work in both cases 😃 

Note that I introduced a test back when I fixed the original issue, if there was a regression, I assume we'd have seen it on the CI, except if the test is gone or not running for some reason.

yisraelU:
@daddykotex yeah was just about to write that, we actually test for this
https://github.com/disneystreaming/smithy4s/actions/runs/4253923323/jobs/7399452607#step:5:263

Baccata:
Oh cool, this means that we're already using Ember in our tests. In that case, if we can just add a test to ensure that the `Content-Length` is absent from such a request, I'm good with the fix 👍

armanbilge:
> Note that I introduced a test back when I fixed the original issue, if there was a regression, I assume we'd have seen it on the CI, except if the test is gone or not running for some reason.

Also, the problem was fixed(-ish) in ember client too :) so this workaround was no longer needed.

PR #827, opened 2023-02-23
[Core] MetaEncode/MetaDecode when handling httpQueryParams trait

yisraelU:
### **This contains both MetaEncode and MetaDecode changes**
I would like to have split them up into separate PR's however the Metadata tests perform a full round trip of encoding and decoding , making this a more difficult task to separate out

This contains some changes to how metadata is encoded and decoded when a structure has the [`httpQueryParams`](https://smithy.io/2.0/spec/http-bindings.html#smithy-api-httpqueryparams-trait) trait on a member

Decoder:
1. the Reserved Queries was removed , this was create to prevent exactly what we **want** to happen , which is that all query params should end up in the structure that is annotated with the `httpQueryParams`  regardless if they are also bound to another structure member
2. we also need to handle when the `httpQueryParams` trait is applied to a `Map[String,String]` and repeated query param is found, for this instead of throwing an exception , we grab the first value 
From https://smithy.io/2.0/spec/http-bindings.html#httpqueryparams-trait

> When servers deserialize the query string into a map of string, they SHOULD take the first encountered value for each key. Since this rule applies to all future query string values, and changing from a map of string to a map of list of string is backwards-incompatible, care should be taken to use map of string only when it is certain that multiple values for any query string will never be meaningful for the operation.

Encoder:
For this we needed to require the values annotated with `httpQueryParams` never override values annotated with `httpQuery`
see previous link

> If a member with the httpQueryParams trait and a member with the [httpQuery trait](https://smithy.io/2.0/spec/http-bindings.html#httpquery-trait) conflict, clients MUST use the value set by the member with the [httpQuery trait](https://smithy.io/2.0/spec/http-bindings.html#httpquery-trait) and disregard the value set by httpQueryParams. 

We do this by inserting values into the Metadata only if the key doesnt exist, and in the SchemaVisitorMetadataWriter we ensure that the field with the `httpQueryParams` annotation is processed last

### DIFF
--- a/modules/core/src/smithy4s/http/internals/MetaDecode.scala
+++ b/modules/core/src/smithy4s/http/internals/MetaDecode.scala
@@ -97,7 +97,9 @@ private[http] sealed abstract class MetaDecode[+A] {
         (metadata, putField) =>
           val iter: Iterator[(FieldName, FieldName)] = metadata.query.iterator
             .map { case (k, values) =>
-              k -> values.head
+              if (values.nonEmpty) {
+                k -> values.head
+              } else throw MetadataError.NotFound(fieldName, QueryParamsBinding)
             }
           if (iter.nonEmpty && optional) putField.putSome(fieldName, f(iter))
           else if (iter.isEmpty && optional) putField.putNone(fieldName)
--- a/modules/core/src/smithy4s/http/internals/MetaDecode.scala
+++ b/modules/core/src/smithy4s/http/internals/MetaDecode.scala
@@ -45,7 +45,6 @@ private[http] sealed abstract class MetaDecode[+A] {
       binding: HttpBinding,
       fieldName: String,
       optional: Boolean,
-      reservedQueries: Set[String],
       maybeDefault: Option[Any]
   ): (Metadata, PutField) => Unit = {
     // format: off
@@ -94,8 +93,7 @@ private[http] sealed abstract class MetaDecode[+A] {
         }
       case (QueryParamsBinding, StringMapMetaDecode(f)) => {
         (metadata, putField) =>
-          val iter = metadata.query.iterator
-            .filterNot { case (k, _) => reservedQueries(k) }
+          val iter: Iterator[(FieldName, FieldName)] = metadata.query.iterator
             .map { case (k, values) =>
               if (values.size == 1) {
                 k -> values.head
@@ -108,7 +106,6 @@ private[http] sealed abstract class MetaDecode[+A] {
       case (QueryParamsBinding, StringListMapMetaDecode(f)) => {
         (metadata, putField) =>
           val iter = metadata.query.iterator
-            .filterNot { case (k, _) => reservedQueries(k) }
             .map { case (k, values) =>
               k -> values.iterator
             }
--- a/modules/core/src/smithy4s/http/internals/MetaDecode.scala
+++ b/modules/core/src/smithy4s/http/internals/MetaDecode.scala
@@ -95,9 +95,7 @@ private[http] sealed abstract class MetaDecode[+A] {
         (metadata, putField) =>
           val iter: Iterator[(FieldName, FieldName)] = metadata.query.iterator
             .map { case (k, values) =>
-              if (values.size == 1) {
-                k -> values.head
-              } else throw MetadataError.ArityError(fieldName, QueryBinding(k))
+              k -> values.head
             }
           if (iter.nonEmpty && optional) putField.putSome(fieldName, f(iter))
           else if (iter.isEmpty && optional) putField.putNone(fieldName)
--- a/modules/core/src/smithy4s/http/internals/MetaDecode.scala
+++ b/modules/core/src/smithy4s/http/internals/MetaDecode.scala
@@ -91,6 +91,8 @@ private[http] sealed abstract class MetaDecode[+A] {
         lookupAndProcess(_.query, q) { (values, fieldName, putField) =>
           putField(fieldName, f(values.iterator))
         }
+        // see https://smithy.io/2.0/spec/http-bindings.html#httpqueryparams-trait 
+      // when targeting Map[String,String] we take the first value encountered
       case (QueryParamsBinding, StringMapMetaDecode(f)) => {
         (metadata, putField) =>
           val iter: Iterator[(FieldName, FieldName)] = metadata.query.iterator
--- a/modules/core/src/smithy4s/http/internals/MetaDecode.scala
+++ b/modules/core/src/smithy4s/http/internals/MetaDecode.scala
@@ -91,7 +91,7 @@ private[http] sealed abstract class MetaDecode[+A] {
         lookupAndProcess(_.query, q) { (values, fieldName, putField) =>
           putField(fieldName, f(values.iterator))
         }
-        // see https://smithy.io/2.0/spec/http-bindings.html#httpqueryparams-trait 
+      // see https://smithy.io/2.0/spec/http-bindings.html#httpqueryparams-trait
       // when targeting Map[String,String] we take the first value encountered
       case (QueryParamsBinding, StringMapMetaDecode(f)) => {
         (metadata, putField) =>
--- a/modules/core/test/src/smithy4s/http/internals/MetadataSpec.scala
+++ b/modules/core/test/src/smithy4s/http/internals/MetadataSpec.scala
@@ -60,7 +60,6 @@ class MetadataSpec() extends FunSuite {
       .flatMap { partial =>
         s.compile(FromMetadataSchemaVisitor).read(partial.decoded.toMap)
       }
-    println(s"decoded: $result")
     expect.same(encoded, expectedEncoding)
     expect(result == Right(a))
     checkRoundTripTotal(a, expectedEncoding)
--- a/modules/core/test/src/smithy4s/http/internals/MetadataSpec.scala
+++ b/modules/core/test/src/smithy4s/http/internals/MetadataSpec.scala
@@ -65,6 +65,27 @@ class MetadataSpec() extends FunSuite {
     checkRoundTripTotal(a, expectedEncoding)
   }
 
+  def checkQueryRoundTrip[A](
+      initial: A,
+      expectedEncoding: Metadata,
+      finished: A
+  )(implicit
+      s: Schema[A],
+      loc: Location
+  ): Unit = {
+    val encoded = Metadata.encode(initial)
+    val result = Metadata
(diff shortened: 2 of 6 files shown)


modules/core/src/smithy4s/http/internals/MetaDecode.scala
```diff
@@ -92,14 +91,13 @@ private[http] sealed abstract class MetaDecode[+A] {
...
       case (QueryParamsBinding, StringMapMetaDecode(f)) => {
         (metadata, putField) =>
-          val iter = metadata.query.iterator
-            .filterNot { case (k, _) => reservedQueries(k) }
+          val iter: Iterator[(FieldName, FieldName)] = metadata.query.iterator
             .map { case (k, values) =>
-              if (values.size == 1) {
-                k -> values.head
-              } else throw MetadataError.ArityError(fieldName, QueryBinding(k))
+              k -> values.head
```
Baccata: I wouldn't remove the condition, but rather change it to `if (values.size >= 1) ... else throw new MetadataError.NotFound`

modules/core/test/src/smithy4s/http/internals/MetadataSpec.scala
```diff
@@ -60,6 +60,7 @@ class MetadataSpec() extends FunSuite {
       .flatMap { partial =>
         s.compile(FromMetadataSchemaVisitor).read(partial.decoded.toMap)
       }
+    println(s"decoded: $result")
```
Baccata: to remove

modules/core/test/src/smithy4s/http/internals/MetadataSpec.scala
```diff
@@ -160,7 +161,8 @@ class MetadataSpec() extends FunSuite {
   // QUERY PARAMETERS
   // ///////////////////////////////////////////////////////////
   test("String query parameter") {
-    val queries = Queries(str = Some("hello"))
+    val queries =
+      Queries(str = Some("hello"), slm = Some(Map("str" -> "hello")))
```
Baccata, line 206: I'd like for the values to differ so that the precedence can be exhibited

modules/core/test/src/smithy4s/http/internals/MetadataSpec.scala
```diff
@@ -204,41 +206,45 @@ class MetadataSpec() extends FunSuite {
   }
 
   test("Integer query parameter") {
-    val queries = Queries(int = Some(123))
+    val queries = Queries(int = Some(123), slm = Some(Map("int" -> "123")))
```
Baccata: [ditto](https://github.com/disneystreaming/smithy4s/pull/827/files#r1116710655)

modules/core/test/src/smithy4s/http/internals/MetadataSpec.scala
```diff
@@ -204,41 +206,45 @@ class MetadataSpec() extends FunSuite {
...
     val expected = Metadata(query = Map("int" -> List("123")))
     checkRoundTrip(queries, expected)
   }
 
   test("Boolean query parameter") {
-    val queries = Queries(b = Some(true))
+    val queries = Queries(b = Some(true), slm = Some(Map("b" -> "true")))
```
Baccata: [ditto](https://github.com/disneystreaming/smithy4s/pull/827/files#r1116710655)

modules/core/test/src/smithy4s/http/internals/MetadataSpec.scala
```diff
@@ -204,41 +206,45 @@ class MetadataSpec() extends FunSuite {
...
     val expected = Metadata(query = Map("b" -> List("true")))
     checkRoundTrip(queries, expected)
   }
 
   test("timestamp query parameters (no format)") {
     val ts = Timestamp(1970, 1, 1, 0, 0, 0)
-    val queries = Queries(ts1 = Some(ts))
+    val queries = Queries(ts1 = Some(ts), slm = Some(Map("ts1" -> epochString)))
```
Baccata: [ditto](https://github.com/disneystreaming/smithy4s/pull/827/files#r1116710655)

modules/core/test/src/smithy4s/http/internals/MetadataSpec.scala
```diff
@@ -204,41 +206,45 @@ class MetadataSpec() extends FunSuite {
...
     val expected = Metadata(query = Map("ts1" -> List(epochString)))
     checkRoundTrip(queries, expected)
   }
 
   test("timestamp query parameters (date-time format)") {
     val ts = Timestamp(1970, 1, 1, 0, 0, 0)
-    val queries = Queries(ts2 = Some(ts))
+    val queries = Queries(ts2 = Some(ts), slm = Some(Map("ts2" -> epochString)))
```
Baccata: [ditto](https://github.com/disneystreaming/smithy4s/pull/827/files#r1116710655)

modules/core/test/src/smithy4s/http/internals/MetadataSpec.scala
```diff
@@ -204,41 +206,45 @@ class MetadataSpec() extends FunSuite {
...
     val expected = Metadata(query = Map("ts2" -> List(epochString)))
     checkRoundTrip(queries, expected)
   }
 
   test("timestamp query parameters (epoch-seconds format)") {
     val ts = Timestamp(1234567890L, 0)
-    val queries = Queries(ts3 = Some(ts))
+    val queries =
+      Queries(ts3 = Some(ts), slm = Some(Map("ts3" -> "1234567890")))
```
Baccata, line 285: [ditto](https://github.com/disneystreaming/smithy4s/pull/827/files#r1116710655)

modules/core/test/src/smithy4s/http/internals/MetadataSpec.scala
```diff
@@ -204,41 +206,45 @@ class MetadataSpec() extends FunSuite {
...
     val expected = Metadata(query = Map("ts3" -> List("1234567890")))
     checkRoundTrip(queries, expected)
   }
 
   test("timestamp query parameters (http-date)") {
     val ts = Timestamp(1970, 1, 1, 0, 0, 0)
-    val queries = Queries(ts4 = Some(ts))
+    val queries = Queries(
+      ts4 = Some(ts),
+      slm = Some(Map("ts4" -> "Thu, 01 Jan 1970 00:00:00 GMT"))
```
Baccata, line 295: [ditto](https://github.com/disneystreaming/smithy4s/pull/827/files#r1116710655)

modules/core/test/src/smithy4s/http/internals/MetadataSpec.scala
```diff
@@ -254,13 +260,16 @@ class MetadataSpec() extends FunSuite {
 
   test("list of strings query param") {
     val list = List("hello", "world")
-    val queries = Queries(sl = Some(list))
+    val queries = Queries(sl = Some(list), slm = Some(Map("sl" -> "hello")))
```
Baccata: [ditto](https://github.com/disneystreaming/smithy4s/pull/827/files#r1116710655)

modules/core/test/src/smithy4s/http/internals/MetadataSpec.scala
```diff
@@ -254,13 +260,16 @@ class MetadataSpec() extends FunSuite {
...
     val expected = Metadata(query = Map("sl" -> list))
     checkRoundTrip(queries, expected)
   }
 
   test("Int Enum query parameter") {
-    val queries = Queries(ie = Some(smithy4s.example.Numbers.ONE))
+    val queries = Queries(
+      ie = Some(smithy4s.example.Numbers.ONE),
+      slm = Some(Map("nums" -> "1"))
```
Baccata, line 322: [ditto](https://github.com/disneystreaming/smithy4s/pull/827/files#r1116710655)

modules/core/test/src/smithy4s/http/internals/MetadataSpec.scala
Baccata: Also, rather than modifying the existing tests, I'd rather you created a new one with a name reflecting what we're actually testing for.

modules/core/test/src/smithy4s/http/internals/MetadataSpec.scala
yisraelU: whoops

modules/core/test/src/smithy4s/http/internals/MetadataSpec.scala
yisraelU: all tests **needed** to be modified to pass, I am not sure what other scenarios you would like tested ?

Baccata: Okay, in that case I'm not certain I agree with the semantics proposed by AWS (in the context of our protocol anyway), because round-trip invariants are a very important notion in serialisation/deserialisation. 

> which is that all query params should end up in the structure that is annotated with the httpQueryParams regardless if they are also bound to another structure member

Can you link to either the spec that says so or the test you inferred it from so that I can have a look and cry ?

yisraelU: Yes, this notion spun me up ,when i saw it..
Sure , please livestream the crying
Here are 2 request tests which applies to both server and client
https://github.com/awslabs/smithy/blob/8f32289b8f23bd53b1676af34f703680a79c6daa/smithy-aws-protocol-tests/model/restJson1/http-query.smithy#L33

https://github.com/awslabs/smithy/blob/8f32289b8f23bd53b1676af34f703680a79c6daa/smithy-aws-protocol-tests/model/restJson1/http-query.smithy#L145

Baccata: I think diverging with the AWS specs here wouldn't make much sense, even if round-tripping is violated. 

I agree to roll with your implementation. 

**However**, the tests need to be reworked. The tests were initially written based on the round-tripping invariant holding. Now that it doesn't hold anymore, you to change it to something like : 

```scala
def checkRoundTrip(initial: Queries, encoded: Metadata, final: Queries)  
```

because we need to test that `Queries(int = Some(123)` becomes `Queries(int = Some(123), slm = Some(Map("int" -> 123))`

yisraelU: Makes sense

yisraelU:
Updated tests to handle query roundtripping

Baccata (approved):

PR #831, opened 2023-02-24
[WIP] Partial-data and granularity reworks 

Baccata:
#### What is this monstrosity ? 

This PR aims at facilitating the implementation of protocols, by shifting the responsibility of various components at hand and create more granularity in the various components. 

At the center is the acknowledgment of the possibility that the partiality of encoding/decoding (eg http metadata vs http bodies) can be codified in a way that removes the need for codecs to be aware of the notion of partiality. It is liberating, because we can clean up a lot of coupling, but also implement http request/response encoders and decoders that will have the correct behaviour of decoding some of the data  from the headers/query/path and the rest from the bodies, no matter the serialisation format (xml, json, message-pack). 

#### More description to come


kubukoz:
If you get hit by a bus before writing a proper description, we won't know how to finish this anyway ;)

modules/core/src/smithy4s/Wedge.scala.scala
```diff
@@ -0,0 +1,30 @@
+package smithy4s
```
kubukoz, line 1: I know it's still a draft, but this file's name looks sus.

PR #832, opened 2023-02-24
add mapHints to Endpoint and provide a HintTransformation on Service

yisraelU:


modules/core/src/smithy4s/Transformation.scala
```diff
@@ -119,4 +119,30 @@ object Transformation {
...
+    (func: Hints => Hints, that: Service[Alg]) => new Service[Alg] {
+      override type Operation[I, E, O, SI, SO] = that.Operation[I, E, O, SI, SO]
+
+      override def endpoints: List[Endpoint[_, _, _, _, _]] = that.endpoints.map(_.mapHints(func))
+
+      override def endpoint[I, E, O, SI, SO](op: Operation[I, E, O, SI, SO]): (I, Endpoint[I, E, O, SI, SO]) = {
+        that.endpoint(op) match {
+          case (i, endpoint) => (i, endpoint.mapHints(func))
+        }
+      }
```
kubukoz: I feel like this might benefit from some caching of the `endpoint.mapHints` call

yisraelU: good point , i think ```endpoints``` would also benefit from that

kubukoz: You could make `endpoints` a val :)

yisraelU: Good point , forgot about that thing

modules/core/src/smithy4s/Transformation.scala
```diff
@@ -119,4 +119,30 @@ object Transformation {
       }
     }
 
+  implicit def mappedHintsServiceTransform[Alg[_[_, _, _, _, _]]]: Transformation[Hints=>Hints, Service[Alg], Service[Alg]] =
```
Baccata, line 122: This could be improved, because `Hints => Hints` is not specific to Endpoints, and the instance is provided implicitly. Instead, a `class MapEndpointHints(f: Hints => Hints)` should be created, and the instance should be provided against that instead of the bare `Hints => Hints`, as that would let us add other types of endpoint mapping later on.

modules/core/src/smithy4s/Endpoint.scala
```diff
@@ -53,6 +53,19 @@ trait Endpoint[Op[_, _, _, _, _], I, E, O, SI, SO] { outer =>
 
   def errorable: Option[Errorable[E]] = None
 
+  private[smithy4s] def mapHints(
+      f: Hints => Hints
+  ): Endpoint[Op, I, E, O, SI, SO] =
+    new Endpoint[Op, I, E, O, SI, SO] {
```
Baccata, line 65: This is missing the `errorable` override

modules/core/src/smithy4s/Transformation.scala
```diff
@@ -119,4 +119,30 @@ object Transformation {
...
+    (func: Hints => Hints, that: Service[Alg]) => new Service[Alg] {
+      override type Operation[I, E, O, SI, SO] = that.Operation[I, E, O, SI, SO]
+
+      override val endpoints: List[Endpoint[_, _, _, _, _]] = that.endpoints.map(_.mapHints(func))
+
+      override def endpoint[I, E, O, SI, SO](op: Operation[I, E, O, SI, SO]): (I, Endpoint[I, E, O, SI, SO]) = {
+        that.endpoint(op) match {
+          case (i, endpoint) => (i, endpoints.find(_.id == endpoint.id).get.asInstanceOf[Endpoint[I, E, O, SI, SO]])
```
Baccata: Nearly there. What I'd do would be back it with a map instead of list. 

So I'd have : 

```scala
private val endpointMap : Map[ShapeId, Endpoint[...]] = ??? 
def endpoints: List[Endpoint[...]] = endpointMap.values.toList 
```

and then use the map for a quicker access.

modules/core/src/smithy4s/Transformation.scala
```diff
@@ -119,4 +119,30 @@ object Transformation {
...
     }
 
+  implicit def mappedHintsServiceTransform[Alg[_[_, _, _, _, _]]]: Transformation[Hints=>Hints, Service[Alg], Service[Alg]] =
+    (func: Hints => Hints, that: Service[Alg]) => new Service[Alg] {
+      override type Operation[I, E, O, SI, SO] = that.Operation[I, E, O, SI, SO]
+
+      override val endpoints: List[Endpoint[_, _, _, _, _]] = that.endpoints.map(_.mapHints(func))
+
+      override def endpoint[I, E, O, SI, SO](op: Operation[I, E, O, SI, SO]): (I, Endpoint[I, E, O, SI, SO]) = {
+        that.endpoint(op) match {
```
Baccata, line 133: The `match` is un-necessary : 

``` 
val (i, endpoint) = that.endpoint(op) 
```

kubukoz: isn't that more of a style choice? Or does it differ in behavior?

modules/core/src/smithy4s/Endpoint.scala
kubukoz: the price of having `errorable` have a default implementation of `None` :( 

can we get rid of that?

yisraelU: yeah , I think i used the IDE implement methods feature :)

modules/core/src/smithy4s/Transformation.scala
yisraelU: I was using purely because  I like the style for extraction, is there an issue ?

modules/core/src/smithy4s/Transformation.scala
yisraelU: Would something like this make sense, and  provide a transformation using EndpointMapper
```
sealed trait EndpointMapper
object EndpointMapper{
 case class MapEndpointHints(func: Hints => Hints) extends EndpointMapper

 }

File: modules/core/src/smithy4s/StreamingSchema.scala [hunk 6]
```diff
@@ -16,7 +16,18 @@
...
+sealed trait StreamingSchema[A] { self =>
+
+  def transformHintsTransitively(f: Hints => Hints): StreamingSchema[A] =
+    self match {
+      case StreamingSchema.Streamed(fieldName, schema) =>
+        StreamingSchema.Streamed(
+          fieldName,
+          schema.transformHintsTransitively(f)
+        )
+      case _ => self
```
kubukoz: I think this should explicitly match against `NoStream` to avoid accidentally missing other cases if there are any in the future

modules/core/src/smithy4s/Transformation.scala
```diff
@@ -120,10 +120,12 @@ object Transformation {
...
 
   implicit def mappedHintsServiceTransform[Alg[_[_, _, _, _, _]]]: Transformation[Hints=>Hints, Service[Alg], Service[Alg]] =
-    (func: Hints => Hints, that: Service[Alg]) => new Service[Alg] {
+    (mapper: Hints=>Hints, that: Service[Alg]) => new Service[Alg] {
       override type Operation[I, E, O, SI, SO] = that.Operation[I, E, O, SI, SO]
 
-      private val cache = that.endpoints.map(e => e.id -> e.mapHints(func)).toMap
+      private val cache = that.endpoints.foldLeft(Map.empty[ShapeId, Endpoint[I, E, O, SI, SO]]) {
+        case (map, endpoint) => map.+(endpoint.id -> endpoint.mapHints(mapper))
+      }
```
kubukoz: nitpick: this may be pretty inefficient, creating N maps for N elements. Might want to use a builder instead, although I doubt this will be in the hot path anyway

modules/core/src/smithy4s/StreamingSchema.scala
yisraelU: I don't think we can do that because StreamingSchema is invariant in A.
I tried a few variations and got compiler errors

kubukoz: yeah, that's a bit messed up in scala 2.

This compiles for me on all Scala versions:

```suggestion
      case _: StreamingSchema.NoStream.type => self
```

modules/core/src/smithy4s/Transformation.scala
yisraelU: agreed , I didn't see this occurring on the hot path , however will switch a builder regardless

modules/core/src/smithy4s/Transformation.scala
Baccata: I guess it is indeed a stylistic thing, but I have a reasonably objective argument for suggesting that : I think `match` signals either partiality or plurality of patterns. Whenever there is an extraction that is total, I use a `val` extractor, as it makes future-me less prone to eyebrow-raising. Also it's more concise.

modules/core/src/smithy4s/Endpoint.scala
```diff
@@ -39,12 +39,18 @@ package smithy4s
   * datatypes that typically fit in memory with concerns of streaming (which can
   * be encoded a great many ways, using a greatt many libraries)
   */
-trait Endpoint[Op[_, _, _, _, _], I, E, O, SI, SO] { outer =>
+trait Endpoint[Op[_, _, _, _, _], I, E, O, SI, SO] {
+  outer =>
```
Baccata, line 43: please revert the change from line 42 to 53

modules/core/src/smithy4s/Transformation.scala
Baccata: I don't think it does : it'll be dangerous to evolve by adding cases to that sealed trait (wrt exhaustive-pat-mats in client code), so it'd be preferable to not expose the members, and then if you're not exposing the members, why would you have a sealed at all ?

Baccata:
Okay, so I gave it a fair bit of thought and I'm quite certain that's not the direction I want us to take, even though I very much appreciate the attempt. 

The reason I don't like this direction is that you cannot know what kinds of transformations could be applied : would people want to just transform the hints on the endpoints ? on the input schemas of all endpoints ? On the error schemas ? Using delegate implementations to do this thing is risky, because of how you could find yourself having an exponential number of those things. 

Instead, it'd be much better if we could treat both the `Endpoint` and the `Service` as good old data classes that were easy to deconstruct/transform : 

```scala 
case class Endpoint[I, E, O, SI, SO] private (...) 
```

following the [Scala best practice around binary compatibility](https://docs.scala-lang.org/overviews/core/binary-compatibility-for-library-authors.html#changing-a-case-class-definition-in-a-backwards-compatible-manner).  

Now, going down this road has its own set of problems, in particular around `Service`, which we can't model as a Data class due to the fact that we've moved the [Operation](https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/core/src/smithy4s/Service.scala#L38) from type parameter to type member, and we'll need it as a type-parameter in order to write the constructor of the data class as code. 

So one thing we could do is have `Builder` classes mirroring both `Endpoint` and `Service`, that'd be data classes, with a method that'd let us easily go from `Endpoint` to `Endpoint.Builder` and vice-versa, and same for `Service`. That'd let us retain the current interfaces, with the flexibility for transforming them without an explosion of delegate classes. 

Now, the thing is that it's gonna be a fair bit of work, and I'd like for it to be accompanied with proper tests. We should aim it for 0.18.0. 

So what I'm thinking is that for the time being : let's just inline the `transformation` part of this PR in whatever unit of work that motivated this addition, and defer a little the implementation of a more proper solution for Service/Endpoint transformation mechanism, which we'll assign you to, @yisraelU.

PR #847, opened 2023-03-02
[Core] Fix collision with known smithy4s construct names

yisraelU:
In particular this was inspired by  a smithy structure called Endpoint
asides for adding a NameDef for those names , we also need to ensure that Endpoint does not refer to the smithy4s.Endpoint rather the type alias Endpoint which is a type member of the Service trait.
The reason for this is that the default mechanism for solving collisions is to fully qualify them , however fully qualifying Endpoint would point it to the Endpoint trait thus requiring injecting an Operation.
the solution is to define it as a NameDef instead of NameRef


project/Dependencies.scala
```diff
@@ -134,7 +134,7 @@ object Dependencies {
   }
 
   object Webjars {
-    val swaggerUi: ModuleID = "org.webjars.npm" % "swagger-ui-dist" % "4.17.0"
+    val swaggerUi: ModuleID = "org.webjars.npm" % "swagger-ui-dist" % "4.16.1"
```
kubukoz: is this intentional?

yisraelU: no ,resolved

daddykotex (approved):

PR #848, opened 2023-03-03
[Core] Fix collision with known smithy4s construct names 

yisraelU:
reopens #847


modules/example/src/smithy4s/example/DogName.scala
```diff
@@ -4,13 +4,12 @@ import smithy4s.Hints
 import smithy4s.Newtype
 import smithy4s.Schema
 import smithy4s.ShapeId
-import smithy4s.example.refined.Name
 import smithy4s.schema.Schema.bijection
 import smithy4s.schema.Schema.string
 
-object DogName extends Newtype[Name] {
+object DogName extends Newtype[smithy4s.example.refined.Name] {
```
yisraelU, line 10: why should these types need to be fully qualified ?

modules/example/src/smithy4s/example/StructureWithRefinedMember.scala
```diff
@@ -4,19 +4,18 @@ import smithy4s.Hints
 import smithy4s.Schema
 import smithy4s.ShapeId
 import smithy4s.ShapeTag
-import smithy4s.example.refined.Age
 import smithy4s.example.refined.Age.provider._
 import smithy4s.schema.Schema.int
 import smithy4s.schema.Schema.struct
 
-case class StructureWithRefinedMember(otherAge: Option[Age] = None)
+case class StructureWithRefinedMember(otherAge: Option[smithy4s.example.refined.Age] = None)
```
lewisjkl, line 11: What caused this to no longer be imported at the top? I see a few others as well so maybe a regression of some kind

yisraelU: fixed, in https://github.com/disneystreaming/smithy4s/pull/848/commits/f507ab267e75b54e6cc7a0b08cdf64a02fafd0ca

modules/example/src/smithy4s/example/DogName.scala
yisraelU: fixed in https://github.com/disneystreaming/smithy4s/pull/848/commits/f507ab267e75b54e6cc7a0b08cdf64a02fafd0ca

modules/example/src/smithy4s/example/StructureWithRefinedMember.scala
Baccata: ... guys, **this is actually the bug that was fixed**, https://github.com/disneystreaming/smithy4s/commit/f507ab267e75b54e6cc7a0b08cdf64a02fafd0ca reintroduces it ...

Baccata: The reason why the bug was not exhibited is that the compiler option added was incorrect. See :  

[f84f208](https://github.com/disneystreaming/smithy4s/pull/848/commits/f84f20813e8ce6229aed105127b0a76ec08326ae)

Baccata: It is normal for the fully qualified name to be rendered here, because the Smithy namespace contains a shape of the [same name](https://github.com/disneystreaming/smithy4s/blob/series/0.17/sampleSpecs/refined.smithy#L65), and Scala 3 is not happy about it. 

Arguably this is a Scala 3 bug, but we can't not workaround on our side. 

Now please one of you kindly revert the commit that re-introduced the bug.

yisraelU: > The reason why the bug was not exhibited is that the compiler option added was incorrect. See :

I dont  know, I tried that first and got ` bad option: '-source:3.0-migration'`

yisraelU: I see this option only works with the Scala 3 compiler

Baccata (approved):
Good work

PR #854, opened 2023-03-08
escape @ in scaladocs

lewisjkl:
Currently, if the `@` symbol is left unescaped then it leads to warnings when generating the scaladoc and the `@` signs won't actually show up in the final html.

The following smithy:

```smithy
/// Test if an at-sign is rendered appropriately
/// @test
structure DocTest {}
```

used to lead to the following docs:

```scala
/** Test if an at-sign is rendered appropriately
  * @test
  */
case class DocTest()
```

Which gives the following warning during scaladoc generation:

```
Tag '@test' is not recognised
```

And renders without the `@test`:

CleanShot 2023-03-08 at 11 08 01

Now with the change in this PR,  we have no warning from scaladoc generation:

```scala
/** Test if an at-sign is rendered appropriately
  * {@literal @}test
  */
case class DocTest()
```

And the resultant html looks like:

CleanShot 2023-03-08 at 11 07 00

Note, to test the scaladoc for yourself, you can checkout this branch and run `sbt example/doc`.

There are likely other characters we need to worry about, but I don't exactly know which ones those are off the top of my head and with some quick tests I haven't found any others yet.

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
@@ -203,13 +203,13 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
       .collectFirst { case h: Hint.Documentation => h }
       .foldMap { doc =>
         val shapeDocs: List[String] =
-          doc.docLines.map(_.replaceAll("@", "{@literal @}"))
+          doc.docLines.map(_.replace("@", "{@literal @}"))
         val memberDocs: List[String] =
           if (skipMemberDocs) List.empty
           else
             doc.memberDocLines.flatMap { case (memberName, text) =>
               s"@param $memberName" :: text
-                .map(_.replaceAll("@", "{@literal @}"))
+                .map(_.replace("@", "{@literal @}"))
                 .map("  " + _)
             }.toList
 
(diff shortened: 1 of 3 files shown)


kubukoz (approved):

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -202,12 +202,15 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
-        val shapeDocs: List[String] = doc.docLines
+        val shapeDocs: List[String] =
+          doc.docLines.map(_.replaceAll("@", "{@literal @}"))
         val memberDocs: List[String] =
           if (skipMemberDocs) List.empty
           else
             doc.memberDocLines.flatMap { case (memberName, text) =>
-              s"@param $memberName" :: text.map("  " + _)
+              s"@param $memberName" :: text
+                .map(_.replaceAll("@", "{@literal @}"))
```
kubukoz: afaik `replaceAll` is regex-based, so it may be more efficient / less error-prone to use `.replace`.

lewisjkl: Good catch! https://github.com/disneystreaming/smithy4s/pull/854/commits/0f754c40254c3ce05fb84aeee8cdd7832ac5150c

PR #861, opened 2023-03-14
[ComplianceTests] Malformed compliance tests

yisraelU:
This pr adds the last component of the protocol compliance tests as defined by Smithy
https://smithy.io/2.0/additional-specs/http-protocol-compliance-tests.html#httpmalformedrequesttests

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -767,6 +767,7 @@ lazy val complianceTests = projectMatrix
         else Seq.empty
       ce3 ++ Seq(
         Dependencies.Circe.parser.value,
+        Dependencies.Smithy.utils ,
         Dependencies.Http4s.circe.value,
         Dependencies.Http4s.client.value,
         Dependencies.Weaver.cats.value % Test,
--- a/modules/compliance-tests/src/smithy4s/compliancetests/ComplianceTest.scala
+++ b/modules/compliance-tests/src/smithy4s/compliancetests/ComplianceTest.scala
@@ -17,9 +17,32 @@
 package smithy4s.compliancetests
 
 import ComplianceTest.ComplianceResult
+import smithy.test.NonEmptyString
+import smithy4s.ShapeId
 
-case class ComplianceTest[F[_]](name: String, run: F[ComplianceResult])
+case class ComplianceTest[F[_]](
+    id: String,
+    endpoint: ShapeId,
+    tags: List[String],
+    run: F[ComplianceResult]
+) {
+  private val showTags =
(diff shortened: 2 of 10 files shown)


yisraelU:
> [error] [http4sJS3] Referring to non-existent class software.amazon.smithy.utils.SimpleCodeWriter
@Baccata I dont know too much about scalajs , is this because of the nio utils included in this package 
https://github.com/awslabs/smithy/blob/main/smithy-utils/src/main/java/software/amazon/smithy/utils/IoUtils.java

kubukoz:
way to dig up an old issue, I know...

the problem is that you're using a Java class: `import software.amazon.smithy.utils.SimpleCodeWriter` - can't do this in a cross-compiled Scala module. We're gonna have to rewrite the implementation in terms of something written 100% in Scala.

yisraelU:
@kubukoz can you explain more.
Why would using a java class prevent cross compilation ? All Scala code calls java code.
Do you mean because it is being called directly rather through a layer of indirection. ?

kubukoz:
> All Scala code calls java code.

Sort of. The Scala standard library does call out to some Java classes/methods, but because Java code can't be transpiled to JS/Native, these classes/methods have to be **reimplemented in Scala** - you can find that e.g.

- [String in Scala Native's javalib](https://github.com/scala-native/scala-native/blob/d576a2a82315c7d0c7f8eb37aff202601d4ce583/javalib/src/main/scala/java/lang/String.scala)
- [String in Scala.js' s javalib](https://github.com/scala-js/scala-js/blob/7e9dbfcd5c9a9eb3fe065db568c4c96ef4e19a81/javalib/src/main/scala/java/lang/_String.scala)

> Why would using a java class prevent cross compilation?

just because Scala.js and Scala Native's compiler plugins can only output the appropriate SJSIR / NIR files from Scala files. These *IR files are needed for further processing, e.g. linking and generating output JS/LLVM.

to sum up, you can't use any Java symbol on JS/Native, unless a Scala implementation is also provided.

This is why we have platform-specific files in the first place - sometimes you can't / don't want to reimplement the Java, so:

- in Java-specific sources you'll use Java libraries,
- in JS-specific sources you'll use interop with JavaScript libraries,
- in Native-specific sources you'll use interop with native libraries.

kubukoz:
There was a project that provided a Java frontend for Scala.js, which would allow using Java code on the JS side here, but it seems to have been forgotten: https://github.com/arthanzel/scalajs-jfe

yisraelU:
closed in favor of #1918

PR #865, opened 2023-03-16
[0.18] Remove CE2 support

Baccata:
Removes support for cats-effect-2


daddykotex (approved):
Imagine the ratio of deleted lines if you had not added all those copyright headers!

Great PR, definitely will alleviate some of the pains in the build tool and the IDE.

Unrelated, but could we consider date less copy right headers, so we would not need to update them continuously (some are outdated here I guess)

Baccata:
> Unrelated, but could we consider date less copy right headers, so we would not need to update them continuously (some are outdated here I guess) 

I thought the date was a legal requirement . Any information suggesting it ain't ? Worst case scenario, we can update the date programmatically using the current year (in another PR of course)

daddykotex:
> > Unrelated, but could we consider date less copy right headers, so we would not need to update them continuously (some are outdated here I guess)
> 
> I thought the date was a legal requirement . Any information suggesting it ain't ? Worst case scenario, we can update the date programmatically using the current year (in another PR of course)

I was inspired by this: https://github.com/awslabs/smithy/pull/1614

kubukoz:
> Imagine the ratio of deleted lines if you had not added all those copyright headers!

why do I feel like every other PR adds these. aren't we checking for headers in CI? :D

modules/compliance-tests/src/smithy4s/compliancetests/internals/ClientHttpComplianceTestCase.scala
```diff
@@ -44,7 +46,7 @@ private[compliancetests] class ClientHttpComplianceTestCase[
 ](
     reverseRouter: ReverseRouter[F],
     serviceInstance: Service[Alg]
-)(implicit ce: CompatEffect[F]) {
+)(implicit ce: Async[F]) {
```
kubukoz: **suggestion:** this can be loosened to Temporal. Probably a couple other spots where we can loosen the constraints too.

project/Dependencies.scala
```diff
@@ -91,10 +90,10 @@ object Dependencies {
    * modules/tests/src-ce2/UUIDGen.scala
    */
   val CatsEffect3: Def.Initialize[ModuleID] =
-    Def.setting("org.typelevel" %%% "cats-effect" % "3.4.8")
+    Def.setting("org.typelevel" %%% "cats-effect" % "3.4.4")
 
   object Http4s {
-    val http4sVersion = Def.setting(if (isCE3.value) "0.23.18" else "0.22.15")
+    val http4sVersion = Def.setting("0.23.17")
```
kubukoz: You can drop the `Def.setting` now.

project/Dependencies.scala
```diff
@@ -112,7 +111,7 @@ object Dependencies {
 
   object Weaver {
 
-    val weaverVersion = Def.setting(if (isCE3.value) "0.8.1" else "0.6.15")
+    val weaverVersion = Def.setting("0.8.1")
```
kubukoz: ditto on `Def.setting`

kubukoz (approved):

kubukoz:
(assuming CI likes it)

PR #869, opened 2023-03-21
[ComplianceTests] provide ability to handle intended short circuit over tcp

yisraelU:
Problem: When running the protocol compliance tests on the server side for we need to exit the fake implementation , we do so via throwing an IntendedShortCircuit exception. 
we handle the exception before asserting the test results
This creates a problem when running the tests over tcp as the exception is thrown on a different JVM and interpreted into whatever the local https server running decides to do. Usually there is some default exception handler.
The simple stupid solution is :
Often http servers tend to use the exception class name in the body of the response . so we just add a clause to check for intended short circuit in the body .
I would prefer something more principled , but this is all I have right now.

A different solution would be to allow Exception Handling to be propagated in the restJsonBuilders, we currently only allow error transformation to be propagated

### DIFF
--- a/modules/compliance-tests/src/smithy4s/compliancetests/internals/IntendedShortCircuit.scala
+++ b/modules/compliance-tests/src/smithy4s/compliancetests/internals/IntendedShortCircuit.scala
@@ -18,4 +18,6 @@ package smithy4s.compliancetests
 package internals
 
 private[compliancetests] case class IntendedShortCircuit()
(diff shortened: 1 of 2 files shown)


yisraelU:
@Baccata When you have a moment to breathe..

Baccata:
Good effort, but I think it's too brittle to rely on strings ... 

Instead I propose this : https://github.com/disneystreaming/smithy4s/pull/922

PR #875, opened 2023-03-28
[0.18] Adds the concepts of PartialData/PartialSchema.

Baccata:
This will be useful for decoupling codecs from http specific concerns, in later PRs. 

The API does not aim at achieving perfect type safety, because it would be counter-productive to even attempt it. It does however provide something that we can build upon. 

This PR contains what is essentially a generalisation of the `BodyPartial/MetadataPartial` mechanism from previous versions, the difference being that the partiality can now be handled at the schema level instead of the codec level. This is liberating, in part because it is gonna allow us to remove the coupling that exists between the json-specific schema visitor and the http binding traits. It'll also let us implement protocols such as `restXml` (or any other kind of protocol that has awareness of the http binding traits) without introducing more of these annoying couplings.

### DIFF
--- a/modules/core/src/smithy4s/PartialData.scala
+++ b/modules/core/src/smithy4s/PartialData.scala
@@ -47,11 +47,11 @@ import scala.collection.compat.immutable.ArraySeq
   * In order to solve for this, the [[PartialData]] type allows to momentarily store a subset of the fields of a case
   * class so that it can be reconciled with other pieces of [[PartialData]] later on.
   */
-private[smithy4s] sealed trait PartialData[A] {
+sealed trait PartialData[A] {
   def map[B](f: A => B): PartialData[B]
 }
 // format: off
-private[smithy4s] object PartialData {
+object PartialData {
   final case class Total[A](a: A) extends PartialData[A] {
     def map[B](f: A => B): PartialData[B] = Total(f(a))
   }
--- a/modules/core/src/smithy4s/PartialData.scala
+++ b/modules/core/src/smithy4s/PartialData.scala
@@ -1,0 +1,87 @@
+/*
+ *  Copyright 2021-2022 Disney Streaming
+ *
+ *  Licensed under the Tomorrow Open Source Technology License, Version 1.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     https://disneystreaming.github.io/TOST-1.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s
+
+import scala.collection.immutable.ArraySeq
+
+/**
+  * Data structure that can hold either the totality or a subset of a larger piece of data.
+  *
+  * This can be used to reconcile bits of data that are coming from several locations,
+  * or to send a piece of data towards different locations.
+  *
+  * For instance :
+  *
+  * ```
+  * structure AB {
+  *   @httpPayload
+  *   @required
+  *   a: String,
+  *
+  *   @httpHeader("X-B")
+  *   @required
+  *   b: String
+  * }
+  * ```
+  *
+  * translates to this case class {{{ case class AB(a: String, b: String) }}}.
+  *
+  * However, codec derivation (performed by the SchemaVisitor mechanism in Smithy4s), does not let us easily solve for
+  * problem such as http message decoding, because of the notion of priority : the http body of a message should not
+  * be decoded if the metadata is not fully decoded first.
+  *
+  * In order to solve for this, the [[PartialData]] type allows to momentarily store a subset of the fields of a case
+  * class so that it can be reconciled with other pieces of [[PartialData]] later on.
+  */
+private[smithy4s] sealed trait PartialData[A] {
+  def map[B](f: A => B): PartialData[B]
+}
+// format: off
+private[smithy4s] object PartialData {
+  final case class Total[A](a: A) extends PartialData[A] {
+    def map[B](f: A => B): PartialData[B] = Total(f(a))
+  }
+  final case class Partial[A](indexes: IndexedSeq[Int], partialData: IndexedSeq[Any], make: IndexedSeq[Any] => A) extends PartialData[A] {
+    def map[B](f: A => B): PartialData[B] = Partial(indexes, partialData, make andThen f)
+  }
+
+  /**
+    * Reconciles bits of partial data (typically retrieved from various parts of a message)
+    * into a single piece of data. It is the responsibility of the caller to ensure that
+    * the individual pieces can be reconciled into the full data.
+    */
+  def unsafeReconcile[A](pieces: PartialData[A]*) : A = {
+    pieces.collectFirst {
+      case Total(a) => a
+    }.getOrElse {
+      val allPieces = pieces.asInstanceOf[Seq[PartialData.Partial[A]]]
+      var totalSize = 0
+      allPieces.foreach(totalSize += _.indexes.size)
+      val array = Array.fill[Any](totalSize)(null)
+      var make : IndexedSeq[Any] => A = null
+      allPieces.foreach { case PartialData.Partial(indexes, data, const) =>
+        make = const
+        var i = 0
+        while(i < data.size) {
+          array(indexes(i)) = data(i)
+          i += 1
+        }
+      }
+      make(ArraySeq.unsafeWrapArray(array))
+    }
+  }
+}
--- a/modules/core/src/smithy4s/PartialData.scala
+++ b/modules/core/src/smithy4s/PartialData.scala
@@ -16,7 +16,7 @@
 
 package smithy4s
 
-import scala.collection.immutable.ArraySeq
+import scala.collection.compat.immutable.ArraySeq
 
 /**
   * Data structure that can hold either the totality or a subset of a larger piece of data.
--- a/modules/core/src/smithy4s/PartialData.scala
+++ b/modules/core/src/smithy4s/PartialData.scala
@@ -74,6 +74,8 @@ object PartialData {
       val array = Array.fill[Any](totalSize)(null)
       var make : IndexedSeq[Any] => A = null
       allPieces.foreach { case PartialData.Partial(indexes, data, const) =>
+        // all the `const` values should be the same, therefore which one is called
+        // is an arbitrary choice.
         make = const
         var i = 0
         while(i < data.size) {
--- a/modules/core/src/smithy4s/PartialData.scala
+++ b/modules/core/src/smithy4s/PartialData.scala
@@ -64,7 +64,7 @@ object PartialData {
     * into a single piece of data. It is the responsibility of the caller to ensure that
     * the individual pieces can be reconciled into the full data.
     */
-  def unsafeReconcile[A](pieces: PartialData[A]*) : A = {
+  def unsafeReconcile[A](pieces: PartialData[A]*): A = {
     pieces.collectFirst {
       case Total(a) => a
     }.getOrElse {
--- a/modules/core/src/smithy4s/schema/PartialSchema.scala
+++ b/modules/core/src/smithy4s/schema/PartialSchema.scala
@@ -1,0 +1,188 @@
+/*
+ *  Copyright 2021-2022 Disney Streaming
+ *
+ *  Licensed under the Tomorrow Open Source Technology License, Version 1.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     https://disneystreaming.github.io/TOST-1.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.schema
+
+import smithy4s.PartialData
+import smithy4s.kinds.PolyFunction
+import Schema._
+
+/**
+  * A structure indicating the match result of a schema against a predicate run against its
+  * fields :
+  *
+  *   - if the schema is not of a structure, or if none of the fields matched, then `NoMatch` should be returned
+  *   - if the schema as
+  */
+sealed trait PartialSchema[A]
+
+object PartialSchema {
+
+  // format: off
+  final case class TotalMatch[A](schema: Schema[A])                extends PartialSchema[A]
+  final case class PartialMatch[A](schema: Schema[PartialData[A]]) extends PartialSchema[A]
+  final case class NoMatch[A]()                                    extends PartialSchema[A]
+  // format: on
+
+  /**
+  * When applied on a structure schema, creates a schema that, when compiled into
+  * a codec, will only encode/decode a subset of the data, based on the hints
+  * of each field.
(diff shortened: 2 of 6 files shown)


modules/core/src/smithy4s/schema/PartialSchema.scala
```diff
@@ -0,0 +1,169 @@
...
+  *   - if the schema is a structure and only a subset of its fields pass the predicate, then `PartialMatch` should be returned
+  *   - if the schema is a structure and all of its fields pass the predicate, then `TotalMatch` should be returned
+  */
+sealed trait PartialSchema[A]
+
+object PartialSchema {
+
+  // format: off
+  final case class TotalMatch[A](schema: Schema[A])                extends PartialSchema[A]
+  final case class PartialMatch[A](schema: Schema[PartialData[A]]) extends PartialSchema[A]
```
kubukoz: **question:** `PartialData` is `private[smithy4s]`, isn't it going to be a problem if it's part of a public type like `PartialSchema`?

modules/core/src/smithy4s/PartialData.scala
```diff
@@ -0,0 +1,87 @@
...
+    * into a single piece of data. It is the responsibility of the caller to ensure that
+    * the individual pieces can be reconciled into the full data.
+    */
+  def unsafeReconcile[A](pieces: PartialData[A]*) : A = {
+    pieces.collectFirst {
+      case Total(a) => a
+    }.getOrElse {
+      val allPieces = pieces.asInstanceOf[Seq[PartialData.Partial[A]]]
+      var totalSize = 0
+      allPieces.foreach(totalSize += _.indexes.size)
```
yisraelU, line 73: just curious is this for perf reasons, like why not just 
 val totalSize = allPieces.map(_.indexes.size).sum

modules/core/src/smithy4s/schema/PartialSchema.scala
```diff
@@ -0,0 +1,169 @@
...
+import Schema._
+
+/**
+  * A structure indicating the match result of running `Schema#partial` against a given predicate
+  *
+  *   - if the schema is not of a structure, or if none of the fields matched, then `NoMatch` should be returned
+  *   - if the schema is a structure and only a subset of its fields pass the predicate, then `PartialMatch` should be returned
+  *   - if the schema is a structure and all of its fields pass the predicate, then `TotalMatch` should be returned
+  */
+sealed trait PartialSchema[A]
```
kubukoz: **nitpick:** bikeshedding time, what do we think about "split" instead of "partial"?

kubukoz (commented):
Can you post an example of how the new Schema methods be used? I assume that'd be in the generated code

modules/core/src/smithy4s/PartialData.scala
kubukoz: I assume because of the extra traversal (`map + sum` requires 2)

yisraelU: in essence , thats what I was asking , like would this code be called on the hot path , that we prefer the mutable path

Baccata: it is meant to be called on the hot path, for instance to reconcile the data that was decoded from body and the data that was decoded from metadata, in http-specific protocols.

modules/core/src/smithy4s/schema/PartialSchema.scala
Baccata: good point

modules/core/src/smithy4s/schema/PartialSchema.scala
Baccata: oops that's a mis-hap

Baccata:
> Can you post an example of how the new Schema methods be used? I assume that'd be in the generated code 

from my mega-pr that I'm slowling splitting : https://github.com/disneystreaming/smithy4s/pull/831/files#diff-cc6da22b074926add2f4414ac70e62508dc1e8c196b5e37de56d0d7a95a49b4eR16

modules/core/src/smithy4s/schema/PartialSchema.scala
Baccata: renamed to `SchemaPartition`, and changed the `PartialMatch` ADT member to `SplittingMatch`, carrying both matching and not matching sides. See [6abc19b](https://github.com/disneystreaming/smithy4s/pull/875/commits/6abc19b307618f53cb428d8dbf1b8060abfea3d1)

modules/core/src/smithy4s/schema/PartialSchema.scala
Baccata: Fixed in [6abc19b](https://github.com/disneystreaming/smithy4s/pull/875/commits/6abc19b307618f53cb428d8dbf1b8060abfea3d1)

modules/core/src/smithy4s/PartialData.scala
```diff
@@ -0,0 +1,87 @@
...
+      val array = Array.fill[Any](totalSize)(null)
+      var make : IndexedSeq[Any] => A = null
+      allPieces.foreach { case PartialData.Partial(indexes, data, const) =>
+        make = const
```
kubukoz, line 79: **suggestion (non-blocking):** might be worth a comment to make it explicit that picking the latest `const` out of all pieces is an arbitrary choice (I assume all `const`s will be the same)

modules/core/test/src/smithy4s/schema/SchemaPartitionSpec.scala
```diff
@@ -0,0 +1,222 @@
...
+import Schema._
+import munit._
+
+final class SchemaPartitionSpec extends FunSuite {
+
+  test("Structure schemas can be divided into partial components") {
+    case class Foo(x: Int, y: Option[Int])
+    val schema = struct(
+      int.required[Foo]("x", _.x),
+      int.optional[Foo]("y", _.y)
```
kubukoz: **suggestion:** shall we make both required to make it even more obvious that neither can be decoded from one piece of partial data?

modules/core/src/smithy4s/schema/SchemaPartition.scala
```diff
@@ -0,0 +1,192 @@
...
+  *   - if the schema is a structure and all of its fields pass the predicate, then `TotalMatch` should be returned
+  */
+sealed trait SchemaPartition[A]
+
+object SchemaPartition {
+
+  // format: off
+  final case class TotalMatch[A](schema: Schema[A])                                                         extends SchemaPartition[A]
+  final case class SplittingMatch[A](matching: Schema[PartialData[A]], notMatching: Schema[PartialData[A]]) extends SchemaPartition[A]
+  final case class NoMatch[A]()                                                                             extends SchemaPartition[A]
```
kubukoz: **thought:** not sure about how/if the call sites would be interested, but shouldn't this wrap a Schema[A] too?

modules/core/src/smithy4s/schema/Schema.scala
```diff
@@ -86,6 +89,31 @@ sealed trait Schema[A]{
...
   }
 
+  /**
+    * When applied on a structure schema, creates a schema that, when compiled into
+    * a codec, will only encode/decode a subset of the data, based on the hints
+    * of each field.
+    *
+    * This can be used to only encode some fields of the data into the http body
+    *
+    * Returns a PartialSchema that indicates whether :
```
kubukoz: **issue:** old name

modules/core/test/src/smithy4s/schema/SchemaPartitionSpec.scala
```diff
@@ -0,0 +1,222 @@
...
+
+    import SchemaPartition._
+    xPartialSchema match {
+      case SplittingMatch(xSchema, ySchema) =>
+        val decoderX = Document.Decoder.fromSchema(xSchema)
+        val decoderY = Document.Decoder.fromSchema(ySchema)
+
+        val result =
+          (decoderX.decode(documentX), decoderY.decode(documentY)).mapN {
+            PartialData.unsafeReconcile(_, _)
```
kubukoz, line 86: **suggestion:** it seems like `unsafeReconcile` is order-independent (you can swap the order and it still works), can we have some sort of test for that? Also worth covering in a Scaladoc on the method itself.

modules/core/src/smithy4s/schema/Schema.scala
```diff
@@ -86,6 +89,31 @@ sealed trait Schema[A]{
...
+
+  /**
+    * Finds the first field that matches the criteria is used, and applies a bijection
+    * between the schema it holds and partial data, which ensures for the field's schema to
+    * be used as "top level" when decoding "payloads".
+    *
+    * NB : a "payload" is typically a whole set of data, without a typical field-based splitting
+    * into subparts. This can be, for instance, an http body.
+    */
+  final def payloadPartition(find: SchemaField[_, _] => Boolean): SchemaPartition[A] =
```
kubukoz: **suggestion:** maybe `payloadPartitionFirst` / `payloadFind`, to make it more like `find` than `partition`? The "first match" semantics seem pretty important here

modules/core/src/smithy4s/schema/Schema.scala
```diff
@@ -86,6 +89,31 @@ sealed trait Schema[A]{
...
+    * Returns a PartialSchema that indicates whether :
+    *   * no field match the condition
+    *   * some fields match the condition
+    *   * all fields match the condition
+    */
+  final def partition(filter: SchemaField[_, _] => Boolean): SchemaPartition[A] =
+    SchemaPartition(filter, payload = false)(this)
+
+  /**
+    * Finds the first field that matches the criteria is used, and applies a bijection
```
kubukoz: **suggestion:** 
```suggestion
    * Finds the first field that matches the criteria used, and applies a bijection
```

modules/core/test/src/smithy4s/schema/SchemaPartitionSpec.scala
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  test(
+    "Computing a partial schema from a non-struct schema returns no-schema"
+  ) {
+    val schema = list(int)
+
+    val partialSchema = schema.partition(_.label.startsWith("z"))
+    assertEquals(partialSchema, SchemaPartition.NoMatch[List[Int]]())
+  }
```
kubukoz: **suggestion:** imo these three tests should go first as they're by far the easiest to understand, and might be a smoother introduction

modules/core/test/src/smithy4s/schema/SchemaPartitionSpec.scala
```diff
@@ -0,0 +1,222 @@
...
+    val documentX = Document.array(Document.fromInt(1), Document.fromInt(2))
+
+    import SchemaPartition._
+    xPartialSchema match {
+      case (TotalMatch(xSchema)) =>
+        val decoderX = Document.Decoder.fromSchema(xSchema)
+
+        val result = decoderX.decode(documentX)
+
+        assert(result == Right(Foo(List(1, 2))))
```
kubukoz: **thought:** I would love to see this defined in a more "property" style, like

```scala
val decoderX = Document.Decoder.fromSchema(xSchema)
val decoderBase = Document.Decoder.fromSchema(schema)

assertEquals(
  decoderX.decode(documentX),
  decoderBase.decode(
    Document.obj(
      "x" -> documentX
    )
  )
)
```

then it becomes a bit clearer how the `xSchema` actually differs from the original schema we started with

modules/core/src/smithy4s/PartialData.scala
```diff
@@ -0,0 +1,87 @@
...
+  final case class Partial[A](indexes: IndexedSeq[Int], partialData: IndexedSeq[Any], make: IndexedSeq[Any] => A) extends PartialData[A] {
+    def map[B](f: A => B): PartialData[B] = Partial(indexes, partialData, make andThen f)
+  }
+
+  /**
+    * Reconciles bits of partial data (typically retrieved from various parts of a message)
+    * into a single piece of data. It is the responsibility of the caller to ensure that
+    * the individual pieces can be reconciled into the full data.
+    */
+  def unsafeReconcile[A](pieces: PartialData[A]*) : A = {
```
kubukoz: **thought:** haven't reviewed the logic here yet, but if tests say it's fine then it's probably fine

modules/core/src/smithy4s/schema/SchemaPartition.scala
```diff
@@ -0,0 +1,192 @@
...
+  // format: on
+
+  private[schema] def apply(
+      keep: SchemaField[_, _] => Boolean,
+      payload: Boolean
+  ): PolyFunction[Schema, SchemaPartition] =
+    new PolyFunction[Schema, SchemaPartition] {
+
+      def apply[S](fa: Schema[S]): SchemaPartition[S] = {
```
kubukoz, line 70: **thought:** need to review this in more detail

modules/core/test/src/smithy4s/schema/SchemaPartitionSpec.scala
Baccata: Good idea

modules/core/src/smithy4s/schema/SchemaPartition.scala
Baccata: Probably not : the reason the `Total` wraps it is that the schema it wraps might not actually be the same as the input schema that gets partitioned.

modules/core/test/src/smithy4s/schema/SchemaPartitionSpec.scala
Baccata: Addressed in https://github.com/disneystreaming/smithy4s/pull/875/commits/9cb8c6bde034946d6d0332034e385716cf15b9ee

modules/core/src/smithy4s/schema/Schema.scala
Baccata: Addressed in Addressed in https://github.com/disneystreaming/smithy4s/pull/875/commits/9cb8c6bde034946d6d0332034e385716cf15b9ee

modules/core/test/src/smithy4s/schema/SchemaPartitionSpec.scala
Baccata: Addressed in https://github.com/disneystreaming/smithy4s/pull/875/commits/9cb8c6bde034946d6d0332034e385716cf15b9ee

modules/core/src/smithy4s/schema/Schema.scala
Baccata: Addressed in https://github.com/disneystreaming/smithy4s/pull/875/commits/9cb8c6bde034946d6d0332034e385716cf15b9ee

modules/core/test/src/smithy4s/schema/SchemaPartitionSpec.scala
Baccata: Addressed in https://github.com/disneystreaming/smithy4s/pull/875/commits/9cb8c6bde034946d6d0332034e385716cf15b9ee

modules/core/src/smithy4s/PartialData.scala
Baccata: Addressed in https://github.com/disneystreaming/smithy4s/pull/875/commits/9cb8c6bde034946d6d0332034e385716cf15b9ee

modules/core/src/smithy4s/schema/SchemaPartition.scala
kubukoz: ohh right

kubukoz: maybe worth a code comment on these classes to highlight the extra difference?

Baccata: https://github.com/disneystreaming/smithy4s/pull/875/commits/5f310af64819b41a0d4c6d5b104abecceb397520

kubukoz (approved):
happy with this as-is, but had some minor suggestions in https://github.com/disneystreaming/smithy4s/pull/875

PR #877, opened 2023-03-29
[Http4s:middleware] pre and post error handling

yisraelU:
I also added utility function for constructing middleware from partial function as I have seen often enough boilerplate to construct a middleware and the default being identity function , i figure this might be a bit cleaner


modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sServerEndpoint.scala
```diff
@@ -107,10 +107,10 @@ private[http4s] class SmithyHttp4sServerEndpointImpl[F[_], Op[_, _, _, _, _], I,
     httpEndpoint.matches(path)
   }
 
-  private val applyMiddleware = middleware(endpoint)
+  private val applyMiddleware: HttpApp[F] => HttpApp[F] = middleware(endpoint)
 
   override val httpApp: HttpApp[F] =
-    applyMiddleware(HttpApp[F] { req =>
+    httpAppErrorHandle(applyMiddleware(httpAppErrorHandle(HttpApp[F] { req =>
```
Baccata: I'd rather the inner one was chained outside of the Kleisli context : 

```scala
.flatMap(successResponse)
.handleErrorWith(errorResponse)  
```

modules/http4s/src/smithy4s/http4s/ServerEndpointMiddleware.scala
```diff
@@ -51,4 +51,16 @@ object ServerEndpointMiddleware {
       ): HttpApp[F] => HttpApp[F] = identity
     }
 
+  def fromPartial[F[_]](
```
Baccata: 👍

Baccata:
Let's add additional tests too.

daddykotex:
Following some discussion around this, I've decided to wrap this pull request. I've added a test to make sure that if a user implements a middleware that can `raiseError(X)` then they can transform this error with `flatMapErrors` or `mapErrors`.

~Note that if the user uses `F.raiseError` within `flatMapErrors` or `throw` within `mapErrors`, then the transformation does not happen. This could be solved if we moved the special case into an independent call to `recoverWith` rather than having two cases in the `handleErrorWith`~

Implemented it as such, it made more sense and it match the previous behavior

Baccata (approved):

PR #878, opened 2023-03-30
[0.18] Http interpreters rework

Baccata:
Depends on https://github.com/disneystreaming/smithy4s/pull/875

The goal of this PR is to build on top of the PartialData work in order to increase the granularity of the various pieces necessary to build http (and REST-specialised) interpreters. In particular, this should now allow for easily tapping into granular pieces to create http-binding-aware interpreters without any coupling between the body codecs (Json/Xml) and the http bindings (httpLabel, httpHeader, httpQuery, httpPayload, etc). 

* Creates a `Metadata.Decoder` that will entirely replace the partial decoders in the future. They co-exist in this PR to reduce the size of the diff. The reason `Metadata.Partial` is being removed is because the concern of data-partitioning is moving away from decoders, and towards the schema.

* Creates a `http4s-kernel` module containing granular abstractions for capturing the Serde concerns that the http interpreter has to work with. The `EntityDecoder` and `EntityEncoder` abstractions provided by Http4s are not suited for our usecase because they are specialised for http-bodies.

* Reworks error selection logic to make it more granular, which will allow for usage in the AWS context in a later PR.

* Reworks the http4s module to make use of the new abstractions. This significantly reduces the coupling between the ServerEndpoint/ClientEndpoint implementations and the intricacies of reconciliating bits of data that come from different place. That particular logic is now owned by the `http4s-kernel` module.


modules/core/src/smithy4s/http/HttpRestSchema.scala
```diff
@@ -0,0 +1,72 @@
...
+import smithy4s.schema._
+import smithy4s.schema.SchemaPartition.NoMatch
+import smithy4s.schema.SchemaPartition.SplittingMatch
+import smithy4s.schema.SchemaPartition.TotalMatch
+
+/**
+ * This construct indicates how a schema is split between http metadata
+ * (ie headers, path parameters, query parameters, status code) and body.
+ *
+ * The schema is split between two different
```
daddykotex: minor: I think this comment is incomplete

modules/http4s-kernel/src/smithy4s/http4s/kernel/ResponseDecoder.scala
```diff
@@ -0,0 +1,109 @@
...
+    * Creates a response decoder that dispatches  the response
+    * to a given decoder, based on some discriminator.
+    */
+  def discriminating[F[_]: Concurrent, Discriminator, E](
+      discriminate: Response[F] => F[Option[Discriminator]],
+      select: Discriminator => Option[ResponseDecoder[F, E]]
+  ): ResponseDecoder[F, E] = {
+    new ResponseDecoder[F, E] {
+      def decodeResponse(response: Response[F]): F[E] =
+        response.toStrict(None).flatMap { strictResponse =>
```
daddykotex, line 75: should we make this timeout available via a parameter?

modules/compliance-tests/src/smithy4s/compliancetests/ReverseRouter.scala
```diff
@@ -19,17 +19,18 @@ package smithy4s.compliancetests
...
+import smithy4s.schema.Schema
 
 /* A construct encapsulating the action of turning an http4s route into
  * an an algebra
  */
 trait ReverseRouter[F[_]] {
   type Protocol
   def protocolTag: ShapeTag[Protocol]
-  def codecs: CodecAPI
+  def expectedResponseType(schema: Schema[_]): HttpMediaType
```
daddykotex, line 33: any reason why the schema is required here?

daddykotex: Ah I guess the `@mediaType` annotation?

modules/http4s-kernel/src/smithy4s/http4s/kernel/ResponseDecoder.scala
Baccata: it's not a timeout but yeah this should be configurable. It'll be done in a later PR though (but still target 0.18)

modules/compliance-tests/src/smithy4s/compliancetests/ReverseRouter.scala
Baccata: > Ah I guess the @mediaType annotation?

Indeed :) but also some protocols might have different content types based on other factors. Like the AWS Rest Json protocol interpret blob payloads as binary and string payloads as text.

build.sbt
```diff
@@ -636,16 +637,32 @@ lazy val xml = projectMatrix
   .jsPlatform(allJsScalaVersions, jsDimSettings)
   .nativePlatform(allNativeScalaVersions, nativeDimSettings)
 
+/**
+ * Module that contains an http4s-specific `EntityCompiler` construct
+ * that codifies the compilatin of smithy4s Schemas to EntityEncoders and
```
kubukoz: **suggestion:** 
```suggestion
 * that codifies the compilation of smithy4s Schemas to EntityEncoders and
```

modules/core/src/smithy4s/http/HttpErrorSelector.scala
```diff
@@ -0,0 +1,171 @@
...
+  val cachedDecoders: PolyFunction[SchemaAlt[E, *], ConstF] =
+    new PolyFunction[SchemaAlt[E, *], ConstF] {
+      def apply[A](alt: SchemaAlt[E, A]): F[E] = {
+        val schema = alt.instance
+        // TODO: apply proper memoization of error instances/
+        // In the line below, we create a new, ephemeral cache for the dynamic recompilation of the error schema.
+        // This is because the "compile entity encoder" method can trigger a transformation of hints, which
+        // lead to cache-miss and would lead to new entries in existing cache, effectively leading to a memory leak.
+        val cache = compiler.createCache()
+        val errorCodec: F[A] = compiler.fromSchema(schema, cache)
```
kubukoz, line 90: **thought:** can we just use the cache-less variant?

modules/http4s-kernel/src/smithy4s/http4s/kernel/RequestEncoder.scala
```diff
@@ -0,0 +1,39 @@
...
+    def addToRequest(request: Request[F], a: A): Request[F] = request
+  }
+
+  def combine[F[_], A](
+      left: RequestEncoder[F, A],
+      right: RequestEncoder[F, A]
+  ): RequestEncoder[F, A] = new RequestEncoder[F, A] {
+    def addToRequest(request: Request[F], a: A): Request[F] =
+      right.addToRequest(left.addToRequest(request, a), a)
+  }
```
kubukoz, line 37: **suggestion:** we don't have to, but maybe we could just make this a monoid? (and say it's lawful due to isomorphism with endo-functions)

modules/http4s/src/smithy4s/http4s/SimpleProtocolBuilder.scala
```diff
@@ -43,7 +44,7 @@ abstract class SimpleProtocolBuilder[P](val codecs: CodecAPI)(implicit
       impl: FunctorAlgebra[Alg, F]
   )(implicit
       service: smithy4s.Service[Alg],
-      F: Concurrent[F]
+      F: Async[F]
```
kubukoz: **question:** I missed the memo, why do we need Async?

kubukoz (approved):

modules/core/src/smithy4s/http/HttpErrorSelector.scala
Baccata: Using the cached variant allows to reduce the number of instantiations for primitives. The cache-less variant is literally cache less.

modules/http4s-kernel/src/smithy4s/http4s/kernel/RequestEncoder.scala
Baccata: I thought about it and elected not too because of how `MessageEncoder` inherits from both `ResponseEncoder` and `RequestEncoder`, and having monoids in all of these would likely mess up implicit. 

Not sure it's that valuable though. Let's add the monoid if we see a pattern that warrants it emerge : these constructs are not meant to be used that often that having a Monoid would give us much.

modules/http4s/src/smithy4s/http4s/SimpleProtocolBuilder.scala
Baccata: Reverted in [2e4b134](https://github.com/disneystreaming/smithy4s/pull/878/commits/2e4b134162c28801936b782a99ed7a83dd23c28f). It's a violation of pure FP in that non-local mutable state does get created, but it gets used in a way that shouldn't leak.

build.sbt
Baccata: Addressed in [2e4b134](https://github.com/disneystreaming/smithy4s/pull/878/commits/2e4b134162c28801936b782a99ed7a83dd23c28f)

modules/core/src/smithy4s/http/HttpRestSchema.scala
Baccata: Addressed in [2e4b134](https://github.com/disneystreaming/smithy4s/pull/878/commits/2e4b134162c28801936b782a99ed7a83dd23c28f)

modules/core/src/smithy4s/http/HttpErrorSelector.scala
kubukoz: ah, fair, I somehow thought it created a cache internally.

modules/http4s-kernel/src/smithy4s/http4s/kernel/RequestEncoder.scala
kubukoz: fair enough. I was only suggesting it because it's a method of grouping the two methods together, and possibly using some syntax.

PR #883, opened 2023-04-03
POC: dynamic binding with schemas

kubukoz:
Experimenting with the idea of giving dynamic Hint bindings access to the trait's schema.

This would enable uniform access to the trait schema on both kinds of hint bindings, enabling the analysis of traits like "give me this service's supported protocols":

```scala
def supportedProtocolsStatic[Alg[_[_, _, _, _, _]]](
    service: Service[Alg]
): List[ShapeId] = service.hints.all.toList
  .flatMap { binding =>
    val hints = binding match {
      case Hints.Binding.StaticBinding(key, _)        => key.schema.hints
      case Hints.Binding.DynamicBinding(_, _, schema) => schema().hints
    }

    hints.get(api.ProtocolDefinition).map(_ => binding.keyId)
  }
```

### DIFF
--- a/modules/core/src/smithy4s/Hints.scala
+++ b/modules/core/src/smithy4s/Hints.scala
@@ -16,6 +16,8 @@
 
 package smithy4s
 
+import smithy4s.Schema
+
 /**
   * A hint is an arbitrary piece of data that can be added to a schema,
   * at the struct level, or at the field/member level.
@@ -45,8 +47,8 @@ object Hints {
 
   def fromSeq[S](bindings: Seq[Hint]): Hints = {
     new Impl(bindings.map {
-      case b @ Binding.StaticBinding(k, _)  => k.id -> b
-      case b @ Binding.DynamicBinding(k, _) => k -> b
+      case b @ Binding.StaticBinding(k, _)     => k.id -> b
+      case b @ Binding.DynamicBinding(k, _, _) => k -> b
     }.toMap)
   }
 
@@ -59,7 +61,7 @@ object Hints {
       toMap.get(key.id).flatMap {
         case Binding.StaticBinding(k, value) =>
           if (key.eq(k)) Some(value.asInstanceOf[A]) else None
-        case Binding.DynamicBinding(_, value) =>
+        case Binding.DynamicBinding(_, value, _) =>
           Document.Decoder.fromSchema(key.schema).decode(value).toOption
       }
     def ++(other: Hints): Hints = {
@@ -76,6 +78,7 @@ object Hints {
 
   sealed trait Binding extends Product with Serializable {
     def keyId: ShapeId
+    def getSchema: Schema[_]
   }
 
   object Binding {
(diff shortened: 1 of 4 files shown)


modules/core/src/smithy4s/Hints.scala
```diff
@@ -76,25 +76,39 @@ object Hints {
...
+        value: Document,
+        schema: () => Schema[_]
+    ) extends Binding {
       override def toString = Document.obj(keyId.show -> value).toString()
+      def getSchema: Schema[_] = schema()
+
+      // todo: use the schema here somehow? is caching going to be affected?
+      private def forComparison = (keyId, value)
+      override def hashCode(): Int = forComparison.hashCode()
+      override def equals(obj: Any): Boolean =
```
Baccata, line 100: This makes me twitch ...

kubukoz: same

modules/dynamic/test/src-jvm/smithy4s/dynamic/EnumSpec.scala
```diff
@@ -223,24 +228,29 @@ class EnumSpec extends FunSuite {
...
+          EnumValue(
+            stringValue = "ICE",
+            intValue = 1,
+            value = 1,
+            name = "ICE",
+            hints = Hints(
+              (
+                ShapeId("smithy.api", "deprecated"),
+                Document.obj(),
+                index.getSchema(api.Deprecated.id).get
```
Baccata, line 249: I don't really like that the addition of a new field suddenly suddenly imposes an arbitrary need for some external reference, when so far things could be built in a somewhat inductive way. 

In particular, I feel like this added constraint closes the door the the opportunity to build data-description  some other way than building a bonafide smithy model. For instance, the current model would allow for parsing schemas from less-constrained IDLs, such as json-schema, and capture hints as json-schema freeform extensions that wouldn't be tied to smithy4s schemas

In other words, I believe the concept of smithy4s schema needs to remain somewhat decoupled from the concept of smithy model and the constraints that apply to smithy models should not necessarily be transposed to smithy4s schemas. 

In all frankness, I think it's a rather annoying mistake from me to have left the data-structure of `Hints.Binding` publicly inspectable, and I'd be in favour of closing it for 0.18 

I do however agree that there are some interesting usecases that would be allowed by a ref held by `Hints.Binding` that would allow to retrieve a schema when the circumstances allow for it : `def getSchema : Option[Schema[_]]`. This would ease my worry about being able to accommodate other usecases than bonafide smithy-schemas, whilst offering more flexibility to traverse schemas as if there was an actual model backing them up.

kubukoz: > In particular, I feel like this added constraint closes the door the the opportunity to build data-description some other way than building a bonafide smithy model. For instance, the current model would allow for parsing schemas from less-constrained IDLs, such as json-schema, and capture hints as json-schema freeform extensions that wouldn't be tied to smithy4s schemas

I never knew this was a principle... it makes sense, but seems a bit like a "what if" scenario rather than something we're about to start doing.

Baccata: > it makes sense, but seems a bit like a "what if" scenario rather than something we're about to start doing.

I actually may have a usecase internally 😅 

Also, it's no more "what if" than any hypothetical tool that'd leverage the API you're adding in this PR (once again : the display of protocols supported by a service is not motivating enough to enforce the constraint, in particular as it'd be solve by the presence of a `def : Option[Schema[_]]` on `Hints.Binding`

kubukoz:
I was looking at this today, but decided to put it on hold for now. If we get a "static schema index", this might not be _that_ necessary after all.

Baccata:
> If we get a "static schema index", this might not be that necessary after all. 

I'm thinking the same. I'll close this PR to reduce the noise.

PR #884, opened 2023-04-03
some minor renames/refactors

kubukoz:
Feel free to pick/discard any of these

### DIFF
--- a/modules/core/src/smithy4s/schema/SchemaPartition.scala
+++ b/modules/core/src/smithy4s/schema/SchemaPartition.scala
@@ -94,9 +94,8 @@ object SchemaPartition {
                     fields.zipWithIndex.filterNot(_._2 == index)
 
                   val maybeRemainingSchema =
-                    Option.when(remainingFields.nonEmpty)(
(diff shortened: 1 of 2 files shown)


modules/core/src/smithy4s/schema/SchemaPartition.scala
```diff
@@ -181,11 +196,13 @@ object SchemaPartition {
...
           get: S => Option[A]
-      ): SchemaPartition[S] = SchemaPartition.NoMatch()
+      ): SchemaPartition[S] =
+        // "single" only makes sense on required fields
+        SchemaPartition.NoMatch()
     }
 
   private def codingError: Nothing =
-    sys.error("Coding error: this should not happen on encoding side")
+    sys.error("Coding error: this should not happen on the encoding side")
```
kubukoz, line 204: **issue:** btw this doesn't only appear in encoding I think, `codingError` is used once on the decoding side (`from` in line 187)

Baccata (approved):
Thanks a lot for this !

PR #896, opened 2023-04-12
Fix description of date formats

kubukoz:
`DD` is "day of year", `dd` is day of month.

### DIFF
--- a/modules/core/src-jvm-native/smithy4s/Timestamp.scala
+++ b/modules/core/src-jvm-native/smithy4s/Timestamp.scala
@@ -256,7 +256,7 @@ object Timestamp extends TimestampCompanionPlatform {
 
   def showFormat(format: TimestampFormat): String = format match {
     case TimestampFormat.DATE_TIME =>
-      "date-time timestamp (YYYY-MM-DDThh:mm:ss.sssZ)"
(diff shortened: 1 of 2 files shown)


modules/core/src-jvm-native/smithy4s/Timestamp.scala
```diff
@@ -256,7 +256,7 @@ object Timestamp extends TimestampCompanionPlatform {
 
   def showFormat(format: TimestampFormat): String = format match {
     case TimestampFormat.DATE_TIME =>
-      "date-time timestamp (YYYY-MM-DDThh:mm:ss.sssZ)"
+      "date-time timestamp (YYYY-MM-ddThh:mm:ss.sssZ)"
     case TimestampFormat.EPOCH_SECONDS => "epoch-second timestamp"
     case TimestampFormat.HTTP_DATE =>
       "http-date timestamp (EEE, dd MMM yyyy HH:mm:ss.sss z)"
```
kubukoz: **thought:** I think this should also be `hh` (24-h hour) and not `HH` (12-h hour)

Baccata: agreed

Baccata (approved):

PR #898, opened 2023-04-14
Using an empty path broke the index redirect

daddykotex:


lewisjkl (approved):

modules/http4s-swagger/test/src/smithy4s/http4s/swagger/DocsSpec.scala
```diff
@@ -114,4 +114,25 @@ object DocsSpec extends SimpleIOSuite with TestCompat {
       )
     )
   }
+
+  test("redirect correctly with works with an empty path") {
```
kubukoz, line 118: **suggestion:** 
```suggestion
  test("redirect correctly works with an empty path") {
```

Issue #152, opened 2022-03-26
Document interaction with http4s Router and relative paths


keynmol:
Let's say you define your smithy ops as mounted at root: 

```scala
@http(method: "POST", uri: "/{name}", code: 200)
```

If you then build the routes using `val routes = SimpleRestJsonBuilder.routes(...).resource`, you can't just shift the API to a custom prefix:

```scala
val shifted = Router("/api" -> routes)
```

The request to `/api/bla` will fail with 404.

This approach will work with manually defined routes, e.g. 

```scala 
    val other = HttpRoutes.of[IO] {
      case r if r.pathInfo.segments.headOption.exists(_.decoded() == "hello") =>
        import org.http4s.dsl.io.*
        Ok(r.toString)
    }

val shifted = Router("/api" -> (routes <+> other))
```

The request to `/api/hello/bla` will succeed.

I don't know exactly whether the recommendation should be writing absolute paths in the smithy files, or whether this should be considered undesired behaviour. Note that it will obviously affect the generated Swagger docs as well.

Baccata:
> I don't know exactly whether the recommendation should be writing absolute paths in the smithy files 

That is a protocol-specific concern, and for the "simpleRestJson" protocol, the implicit rule is that the API spec should contain absolute paths. The rationale is as follows : 

1. It reinforces the source of truth aspect of the smithy specs 
2. It facilitates interop with other tooling (such as swagger/openapi) 

Facilitating shifting would break that paradigm. That being said, the fact that it doesn't work is somewhat accidental, and I'd like to understand why. But I'm not sure I want to fix it, since the bug is, in essence, a happy accident 😅. 

You are correct in the fact that it ought to be documented

kubukoz:
https://smithy.io/2.0/spec/http-bindings.html?highlight=http#uri

> The resolved absolute URI of an operation is formed by combining the URI of the operation with the endpoint of the service. (that is, the host and any base URL of where the service is deployed). For example, given a service endpoint of https://example.com/v1 and an operation pattern of /myresource, the resolved absolute URI of the operation is https://example.com/v1/myresource.

It's fine that `@simpleRestJson` doesn't support this, but I think `SmithyHttp4sRouter` could. For other protocols which might want to reuse that, it'd save some trouble and still be in line with the Smithy spec (_I think_)

tobiajo:
I agree this is very unexpected behavior. I'm considering to work around the issue like this:

```
httpRoutes = Router(
[...]

PR #899, opened 2023-04-14
Add a test to showcase router issue

daddykotex:
I've been trying to replicate a user issue where they build a `HttpRoutes` via the `SimpleRestJsonBuilder` facility. Then they use that router inside a http4s `Router` (like seen in the test).

When calling the server at the right url: `/hello/world` => `/api/hello/world`, they get a 404.

### DIFF
--- a/modules/http4s/test/src/smithy4s/http4s/RouterSpec.scala
+++ b/modules/http4s/test/src/smithy4s/http4s/RouterSpec.scala
@@ -1,0 +1,49 @@
+/*
+ *  Copyright 2021-2022 Disney Streaming
+ *
+ *  Licensed under the Tomorrow Open Source Technology License, Version 1.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     https://disneystreaming.github.io/TOST-1.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.http4s
+
+import cats.effect.IO
+import cats.effect.syntax.resource._
+import smithy4s.example._
+import weaver._
+import org.http4s.client.Client
+import org.http4s.server.Router
+
+object RouterSpec extends SimpleIOSuite {
+
+  private val pizzaImpl: PizzaAdminService[IO] =
+    new PizzaAdminService.Default[IO](IO.stub) {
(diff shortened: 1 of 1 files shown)


modules/http4s/test/src/smithy4s/http4s/RouterSpec.scala
```diff
@@ -0,0 +1,49 @@
...
+  test(
+    "SimpleProtocolBuilder (client) fails when the protocol is not present"
+  ) {
+    val test = for {
+      routes <- SimpleRestJsonBuilder.routes(pizzaImpl).resource
+      routed = Router("/api" -> routes).orNotFound
+      client = Client.fromHttpApp(routed)
+      pizzaClient <- SimpleRestJsonBuilder(PizzaAdminService)
+        .client(client)
+        .resource
```
kubukoz, line 43: AFAICS this can't really work because the server has that `/api` prefix, while the client doesn't use it. I'd try with `.client(client).uri(uri"/api")`.

kubukoz: If you do that, then this change seems to fix it:

```diff
diff --git a/modules/http4s/src/smithy4s/http4s/SmithyHttp4sRouter.scala b/modules/http4s/src/smithy4s/http4s/SmithyHttp4sRouter.scala
index 47cc0ec9..e8cd6bcc 100644
--- a/modules/http4s/src/smithy4s/http4s/SmithyHttp4sRouter.scala
+++ b/modules/http4s/src/smithy4s/http4s/SmithyHttp4sRouter.scala
@@ -43,7 +43,7 @@ class SmithyHttp4sRouter[Alg[_[_, _, _, _, _]], Op[_, _, _, _, _], F[_]](
   val routes: HttpRoutes[F] = Kleisli { request =>
     for {
       endpoints <- perMethodEndpoint.get(request.method).toOptionT[F]
-      path = request.uri.path.segments.map(_.decoded()).toArray
+      path = request.pathInfo.segments.map(_.decoded()).toArray
       (endpoint, pathParams) <- endpoints.collectFirstSome(_.matchTap(path)).toOptionT[F]
       response <- OptionT.liftF(endpoint.httpApp(request.withAttribute(pathParamsKey, pathParams)))
     } yield response
```

but I'm not sure if there might be other implications.

kubukoz:
related: https://github.com/disneystreaming/smithy4s/issues/152

Baccata:
I'm not sure whether to consider this a bug, because even if the behaviour breaks the user's intent, the user's intent is likely to be incorrect in the first place, as using `Router` to add a path prefix violates what the specification says. 

There are some unlikely scenarios that may justify a desire for this to work, but until someone presents a non-hypothetical usecase, I'm rather unwilling to address. It'd be great if someone documented the behaviour though.

daddykotex:
Sounds good me, that's what I had in mine

PR #912, opened 2023-04-19
non-orphan implicit typeclasses

lewisjkl:
Making it so typeclass instances can be added in companion objects in the generated code. This uses a mechanism similar to refinements to specify the location of the typeclass and the interpreter that provides it.

Will add more details as this PR comes out of draft status.


modules/example/src/smithy4s/example/MovieTheater.scala
```diff
@@ -0,0 +1,26 @@
...
+    smithy4s.example.Eq(),
+  )
+
+  implicit val schema: Schema[MovieTheater] = struct(
+    string.optional[MovieTheater]("name", _.name),
+  ){
+    MovieTheater.apply
+  }.withId(id).addHints(hints)
+
+  implicit val movieTheaterEq: cats.Eq[MovieTheater] = EqInterpreter.fromSchema(schema)
```
lewisjkl: This is the main thing to look at here. I still have a ways to go on this PR, but I wanted to put it up now to get early feedback before I get further along with it.

kubukoz:
ohh la la. I may like this.

modules/example/src/smithy4s/example/typeclass/EqInterpreter.scala
```diff
@@ -0,0 +1,212 @@
...
+        fa.eqv(a, u)
+      }
+      def absorb[A](f: A => (U => Boolean)): AltEq[A] = new AltEq[A] {
+        def eqv(a: A, u: U): Boolean = f(a)(u)
+      }
+    }
+
+    val precompiler = new Alt.Precompiler[Schema, AltEq] {
+      def apply[A](label: String, instance: Schema[A]): AltEq[A] = {
+        // Here we "cheat" to recover the `Alt` corresponding to `A`, as this information
```
kubukoz: Shall we add support for this in `Precompiler`?

modules/example/src/smithy4s/example/typeclass/EqInterpreter.scala
```diff
@@ -0,0 +1,212 @@
...
+    implicit val encoderKInstance = new EncoderK[AltEq, U => Boolean] {
+      def apply[A](fa: AltEq[A], a: A): U => Boolean = { (u: U) =>
+        fa.eqv(a, u)
+      }
+      def absorb[A](f: A => (U => Boolean)): AltEq[A] = new AltEq[A] {
+        def eqv(a: A, u: U): Boolean = f(a)(u)
+      }
+    }
+
+    val precompiler = new Alt.Precompiler[Schema, AltEq] {
```
kubukoz: nitpick: usually I just inline this

modules/example/src/smithy4s/example/typeclass/EqInterpreter.scala
```diff
@@ -0,0 +1,212 @@
...
+    (x: B, y: B) => eqA.eqv(refinement.from(x), refinement.from(y))
+  }
+
+  override def lazily[A](suspend: Lazy[Schema[A]]): Eq[A] = {
+    val eq = suspend.map(self(_))
+    (x: A, y: A) => eq.value.eqv(x, y)
+  }
+
+  def primitiveEq[P](primitive: Primitive[P]): Eq[P] = {
+    primitive match {
```
kubukoz: This smells an awful lot like `Primitive.deriving`.

modules/example/src/smithy4s/example/typeclass/EqInterpreter.scala
```diff
@@ -0,0 +1,212 @@
...
+    val aEq: Eq[A] = self(schema)
+    (x: B, y: B) => aEq.eqv(bijection.from(x), bijection.from(y))
+  }
+
+  override def refine[A, B](
+      schema: Schema[A],
+      refinement: Refinement[A, B]
+  ): Eq[B] = {
+    val eqA: Eq[A] = self(schema)
+    (x: B, y: B) => eqA.eqv(refinement.from(x), refinement.from(y))
```
kubukoz:
```suggestion
  self(schema).contramap(refinement.from)
```

same in bijections

modules/example/src/smithy4s/example/typeclass/EqInterpreter.scala
```diff
@@ -0,0 +1,212 @@
...
+
+  override def struct[S](
+      shapeId: ShapeId,
+      hints: Hints,
+      fields: Vector[SchemaField[S, _]],
+      make: IndexedSeq[Any] => S
+  ): Eq[S] = { (x: S, y: S) =>
+    {
+      def forField[A2](field: Field[Schema, S, A2]): Boolean = {
+        val eqField = field.foldK(new Field.FolderK[Schema, S, Eq]() {
```
kubukoz: issue: since this is an example we probably want to follow best practices, so I'd suggest some more "precompiled" way: make `forField` return an `S => Boolean` and move the `(x: S, y: S) =>` below so that we traverse the fields while instantiating the whole thing.

modules/example/src/smithy4s/example/typeclass/EqInterpreter.scala
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  override def map[K, V](
+      shapeId: ShapeId,
+      hints: Hints,
+      key: Schema[K],
+      value: Schema[V]
+  ): Eq[Map[K, V]] = {
+    implicit val valueEq: Eq[V] = self(value)
+    Eq[Map[K, V]]
```
kubukoz: kinda interesting how they made Maps comparable without an `Eq` for the key

modules/example/src/smithy4s/example/typeclass/EqInterpreter.scala
```diff
@@ -0,0 +1,212 @@
...
+      tag: CollectionTag[C],
+      member: Schema[A]
+  ): Eq[C[A]] = {
+    implicit val memberEq: Eq[A] = self(member)
+    tag match {
+      case CollectionTag.ListTag       => Eq[List[A]]
+      case CollectionTag.SetTag        => Eq[Set[A]]
+      case CollectionTag.VectorTag     => Eq[Vector[A]]
+      case CollectionTag.IndexedSeqTag => Eq[IndexedSeq[A]]
+    }
```
kubukoz: thought: we could have a `CollectionTag.deriving` in a similar spirit to `Primitive.deriving`

modules/example/src/smithy4s/example/MovieTheater.scala
kubukoz: Looks great to me.

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -810,6 +811,29 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+        }
+        .map { case (typeclassName, typeclassInfo) =>
+          Hint.Typeclass(
+            typeclassName.toShapeId,
+            typeclassInfo.getTargetType,
+            typeclassInfo.getInterpreter
+          )
+        }
+        .toList
+    NonEmptyList.fromList(result).map(Hint.Typeclasses)
```
kubukoz: nitpick: use `.toNel` syntax to get a single expression :)

kubukoz:
sorry for the comment spam @lewisjkl, this is looking great

modules/example/src/smithy4s/example/typeclass/EqInterpreter.scala
```diff
@@ -0,0 +1,212 @@
...
+
+import cats.implicits.{catsSyntaxEq, toContravariantOps}
+import smithy4s.{Schema, _}
+import cats.kernel.Eq
+import smithy4s.capability.EncoderK
+import smithy4s.schema._
+
+import java.util.UUID
+
+object EqInterpreter extends CachedSchemaCompiler.Impl[Eq] {
```
Baccata: @yisraelU already already has an (arguably cleaner) implementation of that [here](https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/compliance-tests/src/smithy4s/compliancetests/internals/eq/EqSchemaVisitor.scala) that could/should be inlined as a bonafide `smithy4s-cats` where a `Show`-deriving mechansim could also live. 

Taking on non-typeclass instances warrants it.

modules/protocol/resources/META-INF/smithy/smithy4s.meta.smithy
```diff
@@ -117,3 +117,30 @@ string Import
 /// by default. Adding this trait will cause the collection to become unwrapped.
 @trait(selector: ":is(simpleType, list, map, set)")
 structure unwrap {}
+
+/// Placing this trait on another trait marks the target trait as a
+/// typeclass. This means that shapes which are marked with the target
+/// trait will have an instance of the typeclass made available in their
```
daddykotex: suggestion: made available in the generated object companion.

lewisjkl:
> sorry for the comment spam @lewisjkl, this is looking great

No worries @kubukoz I actually copied that `EqInterpreter` from something that @yisraelU worked on a while back. I really just put it as a placeholder until he PRs in a smithy4s-cats module that adds that among other interpreters. Then we can remove it from this PR and just reference it instead.

modules/codegen/src/smithy4s/codegen/internals/IR.scala
```diff
@@ -280,6 +280,8 @@ private[internals] object Hint {
     case object IndexedSeq extends SpecializedList
   }
   case object UniqueItems extends Hint
+  case class Typeclass(id: ShapeId, targetType: String, interpreter: String)
```
yisraelU, line 283: A nice-to-have would be a type alias wherever a String refers to a Class (for doc purposes)
Perhaps this should extend to the Smithy definitions too

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -501,6 +501,24 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
+    val interpreter = NameRef(hint.interpreter)
+    val lowerCasedName = s"${tpe.name.head.toLower.toString}${tpe.name.tail}"
+    line"implicit val $lowerCasedName${hint.id.getName.capitalize}: $target[$tpe] = $interpreter.fromSchema(schema)"
+  }
+
+  private def renderTypeclasses(
+      hints: List[Hint],
+      tpe: NameRef
+  ): Option[Lines] = {
+    (hints.collectFirst { case h: Hint.Typeclasses =>
```
yisraelU: I find collectFirst to be  a bit misleading. There can be only one ,of type Hint.Typeclass and if there could be more , I think we would want them all

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -501,6 +501,24 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     }
   }
 
+  private def renderTypeclass(hint: Hint.Typeclass, tpe: NameRef): Line = {
+    val target = NameRef(hint.targetType)
+    val interpreter = NameRef(hint.interpreter)
+    val lowerCasedName = s"${tpe.name.head.toLower.toString}${tpe.name.tail}"
```
Baccata: I think we have a helper for this : `uncapitalise`

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
Baccata: It's not misleading, I think it's just wrong. It should be `collect` here.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -501,6 +501,24 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
+    line"implicit val $lowerCasedName${hint.id.getName.capitalize}: $target[$tpe] = $interpreter.fromSchema(schema)"
+  }
+
+  private def renderTypeclasses(
+      hints: List[Hint],
+      tpe: NameRef
+  ): Option[Lines] = {
```
Baccata: You can return `Lines`

Baccata:
Build fails because docs are currently linking a non-existing class. We'll wait for the cats-module to be merged to go back to this PR.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
yisraelU: oh of course .I was thinking this was Hints as opposed to a List of hint

lewisjkl: I think it's right. There should only ever be one Typeclasses hint which may contain multiple 'Typeclass' instances. It doesn't need to be modeled that way, but that's how I did it so we'd keep the idea that you can only have 1 of a given trait/hint per shape. Collect first is just so I could do the find and map at the same time

Baccata: > but that's how I did it so we'd keep the idea that you can only have 1 of a given trait/hint per shape

I've "violated" that assumption in my commit 😅. For non-meta hints, I think you are correct, but there is a distinction to be had between hints that get rendered and available at runtime and hints that aren't. Trying to abide by the invariant at codegen time seems to be detrimental to the readability of the code .

Baccata: Also, sorry ,should have amended my comment. It wasn't "wrong", I was just confused between the existence of a `Typeclass` and a `Typeclasses` type, which got mixed together in my brain.

lewisjkl: Yeah that totally makes sense, I think it's clearer without the extra typeclasses hint so works for me 👍🏼

lewisjkl:
Will add changelog for this and #942 on Monday

Baccata (approved):

Baccata (approved):

PR #914, opened 2023-04-20
Align `aws-http4s` module on the constructs now provided by `http4s-kernel`

Baccata:
This PRs aims at standardising the AWS support on http4s. Trying to support several clients and idioms would induce a perfomance burden that I'm not willing to carry, and hurt our ability to support streaming operations that some AWS services expose (in particular S3). 

* Removes the `aws` module. Its responsibilities move to the `aws-http4s` module.
* Overhauls most of the logic in order to benefit from the granular constructs introduced `http4s-kernel`
* In particular, signature now occurs via a `Client => Client` middleware. This drastically improves maintainability of the code at the cost of a little bit of indirection. This approach will make it easier to support other means of signature (in particular for streaming operations). 
* Changes the UX for instantiating clients. In particular, the instantiation is now biased towards monofunctors, as most of the interaction with AWS is unary. But also, implicit extensions are removed in favour of an `AwsClient(_, _)` call, which makes it easier to rely on auto-completion in editors. 

TODO : 

- [x] Add unit-tests for the signature middleware
- [ ] Polish UX around client instantiation 
- [ ] Add Scaladocs to the `AwsClient` construct, the `AwsRegion` construct and the`AwsEnvironment` construct, which


build.sbt
```diff
@@ -301,37 +300,6 @@ lazy val `aws-kernel` = projectMatrix
   )
   .nativePlatform(allNativeScalaVersions, nativeDimSettings)
 
-/**
- * cats-effect specific abstractions of AWS protocol interpreters and constructs
- */
-lazy val aws = projectMatrix
```
Baccata, line 307: removed as supporting non-http4s clients in not in the scope of the project anymore.

modules/aws-http4s/src/smithy4s/aws/AwsClient.scala
```diff
@@ -0,0 +1,101 @@
...
+  final class AWSInterpreterBuilder[Alg[_[_, _, _, _, _]]](
+      awsProtocol: AwsProtocol,
+      awsService: AwsService,
+      val service: smithy4s.Service[Alg]
+  ) {
+
+    private def interpreter[F[_]: Async](
+        awsEnv: AwsEnvironment[F]
+    ): service.FunctorInterpreter[F] = {
+      val clientCodecs: UnaryClientCodecs[F] = awsProtocol match {
```
Baccata, line 61: this is where the money-shot is, which entirely justifies everything that 0.18 is about : the difference in protocols is entirely driven by the codecs. The rest of the logic is entirely decoupled from it, be that signature / routing / etc / etc.

modules/aws-http4s/src/smithy4s/aws/internals/AwsJsonCodecs.scala
```diff
@@ -0,0 +1,46 @@
...
+
+import smithy4s.HintMask
+import smithy4s.IntEnum
+import smithy4s.http4s.kernel._
+import cats.effect.Async
+
+/**
+ * An client codec for the AWS_JSON_1.0/AWS_JSON_1.1 protocol
+ */
+private[aws] object AwsJsonCodecs {
```
Baccata, line 32: Putting aside the problem of streaming operations, this small construct centralises the protocol-specific details.

modules/aws-http4s/src/smithy4s/aws/internals/AwsJsonCodecs.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  private val hintMask =
+    aws.protocols.AwsJson1_0.protocol.hintMask ++
+      aws.protocols.AwsJson1_0.protocol.hintMask ++ HintMask(IntEnum)
+
+  def make[F[_]: Async]: UnaryClientCodecs[F] = {
```
Baccata: I think there's still a little bit of polishing to do there. I think the `UnaryClientCodecs` need to be revised to become a mere holder of the compiled codecs that are relevant to the client. Something like : 

```scala
trait UnaryClientCodecs[F[_], I, E, O, SI, SO] {
  def inputEncoder : RequestEncoder[F, I]
  def outputDecoder : ResponseDecoder[F, O]
  def errorDecoder : ResponseDecoder[F, Throwable]
} 

object UnaryClientCodecs {
  type For[F[_]] {
     type toKind5[I, E, O, SI, SO] 
  } 
} 
```

And the `make` here should be something like : 

```scala
def make[F[_] : Async] : PolyFunction5[Endpoint.Schemas, UnaryClientCodecs.For[F]#toKind5] 
```

The reason is that the `Schemas` might not be sufficient to produce the encoders : some information which lives at the Endpoint level (like the `@endpoint` trait) can be necessary to produce it.

modules/aws-http4s/src/smithy4s/aws/internals/AwsJsonCodecs.scala
```diff
@@ -0,0 +1,46 @@
...
+import cats.effect.Async
+
+/**
+ * An client codec for the AWS_JSON_1.0/AWS_JSON_1.1 protocol
+ */
+private[aws] object AwsJsonCodecs {
+
+  private val hintMask =
+    aws.protocols.AwsJson1_0.protocol.hintMask ++
+      aws.protocols.AwsJson1_0.protocol.hintMask ++ HintMask(IntEnum)
```
kubukoz: **issue:** `AwsJson1_0` shows up twice, that seems like a mistake.

Did you want to use 1_0 and 1_1? If so, should we really have that in a single mask?

modules/aws-http4s/src/smithy4s/aws/internals/AwsSignatureClient.scala
```diff
@@ -0,0 +1,133 @@
...
+    ): Binary = {
+      val kSecret = binaryFromString("AWS4" + key)
+      val kDate = hmacSha256(dateStamp, kSecret)
+      val kRegion = hmacSha256(regionName, kDate)
+      val kService = hmacSha256(serviceName, kRegion)
+      val kSigning = hmacSha256("aws4_request", kService)
+      kSigning
+    }
+
+    // scalafmt: { align.preset = most, danglingParentheses.preset = false, maxColumn = 240, align.tokens = [{code = ":"}]}
```
kubukoz, line 91: **praise:** TIL

build.sbt
```diff
@@ -342,25 +310,35 @@ lazy val aws = projectMatrix
  */
 lazy val `aws-http4s` = projectMatrix
   .in(file("modules/aws-http4s"))
-  .dependsOn(aws, complianceTests % "test->compile", dynamic % "test->compile")
+  .dependsOn(`aws-kernel`, `http4s-kernel`, json, xml, complianceTests % "test->compile", dynamic % "test->compile")
```
kubukoz: **issue:** I don't think we actually use either `dynamic` or `complianceTests` here.

build.sbt
```diff
@@ -342,25 +310,35 @@ lazy val aws = projectMatrix
...
-    Test / allowedNamespaces := Seq("com.amazonaws.dynamodb"),
-    Test / sourceGenerators := Seq(genSmithyScala(Test).taskValue)
+    Test / allowedNamespaces := Seq(
+      "com.amazonaws.dynamodb",
+      "smithy4s.example.aws"
+    ),
+    Test / sourceGenerators := Seq(genSmithyScala(Test).taskValue),
+    scalacOptions ++= Seq(
+      "-Wconf:msg=class AwsQuery in package (aws\\.)?protocols is deprecated:silent"
+    )
```
kubukoz: **praise:** nice

modules/aws-http4s/src/smithy4s/aws/internals/AwsJsonCodecs.scala
kubukoz: I guess they're supposed to be the same, let's just fix this to use both then

modules/aws-http4s/src/smithy4s/aws/AwsClient.scala
```diff
@@ -0,0 +1,101 @@
...
+        case AwsProtocol.AWS_JSON_1_0(_) =>
+          // TODO "application/x-amz-json-1.0"
+          AwsJsonCodecs.make[F]
+
+        case AwsProtocol.AWS_JSON_1_1(_) =>
+          // TODO "application/x-amz-json-1.1",
+          AwsJsonCodecs.make[F]
+        case _ => ???
```
kubukoz, line 67: **issue:** let's `sys.error` this or throw something explicitly

modules/aws-http4s/src/smithy4s/aws/AwsClient.scala
kubukoz, line 1: **suggestion:** not strictly related, but can we have comments on `RequestEncoder` / `ResponseDecoder` pointing to `MessageEncoder` and `MessageDecoder` as a place where some built-in implementations can be found? It took me a sec and it might just make it more discoverable.

modules/aws-http4s/src/smithy4s/aws/AwsClient.scala
```diff
@@ -0,0 +1,101 @@
...
+              endpoint: service.Endpoint[I, E, O, SI, SO]
+          ): I => F[O] =
+            new AwsUnaryEndpoint(
+              service.id,
+              service.hints,
+              awsService,
+              awsEnv,
+              endpoint,
+              clientCodecs
+            )
```
kubukoz, line 81: **suggestion:** named params

modules/aws-http4s/src/smithy4s/aws/internals/AwsJsonCodecs.scala
kubukoz: > The reason is that the Schemas might not be sufficient to produce the encoders : some information which lives at the Endpoint level (like the @endpoint trait) can be necessary to produce it.

Why not accept Endpoint directly, then?

kubukoz:
Playground is now happy with this: https://github.com/kubukoz/smithy-playground/pull/255/files - look at how smol it is now with the Async change reverted :D

Baccata:
I'm making the BDFL decision to merge this as is into 0.18 in order to unlock ongoing parallel work, and I'll address comments later

PR #917, opened 2023-04-20
Document `@httpResponseCode` as supported

kubukoz:
Related: https://github.com/disneystreaming/alloy/pull/78


daddykotex:
The list here is generated: https://disneystreaming.github.io/smithy4s/docs/protocols/simple-rest-json/overview#supported-traits-1
Maybe we should only keep one?

kubukoz:
agreed, didn't realize that :) guess we can just delete the list I changed, or suggest scrolling further to see the full list.

Baccata:
Agreed

Baccata (approved):

PR #918, opened 2023-04-20
Add Schema.biject

kubukoz:
Tiny utility for those of us who construct schemas manually.

Maybe we should codegen this too?


kubukoz:
right, in codegen we use the `Schema.bijection` overload that takes a `Bijection` - we'd need to add that to the instance methods too. Maybe let's just have this then.

kubukoz:
technically, this is source breaking if someone has an extension method like this.

Baccata:
I've already added [that](https://github.com/disneystreaming/smithy4s/blob/series/0.18/modules/core/src/smithy4s/schema/Schema.scala#L81-L82) to the 0.18 branch. 

Considering it's just a small convenience method, I'm closing this PR to save me a sync.

PR #919, opened 2023-04-20
Generate Schema.apply overloads as returning `Schema.StructSchema`

kubukoz:
I want to do some transformation on a schema I get from one of these, and the type is currently lost.

Some interesting things that we could get if this gets merged: `InvariantSemigroupal[Schema.StructSchema]`. This enables decomposition of a struct schema into multiple case classes, while keeping things flat (so, mostly hand-written schemas):

```scala
case class BigCaseClass(a: A, b: B, c: C)

val a: StructSchema[A]
val b: StructSchema[B]
val c: StructSchema[C]

(a, b, c).imapN(BigCaseClass.apply)(bcc => (bcc.a, bcc.b, bcc.c))
```

Targeting 0.18 because I believe this is binary breaking


Baccata (approved):

Baccata:
> `InvariantSemigroupal[Schema.StructSchema]`

A bit risky in that if you compose two schemas with fields named `a`, codecs may be borked. I suppose that's acceptable, considering schema-equality is not really that well defined and we're not really inforcing invariants with respect to field/alt names.

PR #920, opened 2023-04-20
[Aws-Http4s] Fix content header for Aws Signature Client

yisraelU:
In its current state , the AwsSignatureClient renders the **value** of the Content-Type header as 
```Content-Type(MediaType(application/json),None)```
which is the string representation of the http4s Content-Type class
This small fix uses the header instance that accompanies that class to extract the value out


kubukoz (approved):

yisraelU:
@kubukoz how did those timestamp files get in here

Baccata (approved):
👍

PR #921, opened 2023-04-21
Typelevel Cats based module added

yisraelU:
A module that provides a Schema Visitor for the Show and Hash typeclasses from Typelevel Cats
Implementation as follows
The general direction for Hash, is to provide a hashcode for data based off its its data model vs a smithy4s implementation detail.
This leads to data wrappers that are not reflective of the model , to be excluded from the hash implementation
*Wherever hashes are combined , this is done using the strategies outlined in [MurmurHash3](https://www.scala-lang.org/api/2.13.8/scala/util/hashing/MurmurHash3$.html)
-  Structs 
   - This follows closely the [scala implementation](https://github.com/scala/scala/blob/4b2151cbe4c857d49556b70c3fa0f5e236fd7754/src/library/scala/util/hashing/MurmurHash3.scala#L343) for  hashing Products . It combines productSeed, the struct ShapeId name, and the hash of all its parts
- Unions
  -  Combination of hash of label and the value
- Enumerations 
  - here an Enumeration is treated as its underlying data type , if its an IntEnum as an integer otherwise a String


modules/cats/src/smithy4s/catz/instances/ShowInstances.scala
```diff
@@ -0,0 +1,18 @@
...
+import smithy4s.{ByteArray, Document, ShapeId, Timestamp}
+import smithy4s.kinds.PolyFunction
+
+trait ShowInstances {
+
+  implicit val sId: Show[ShapeId] = Show.fromToString
+  implicit val byteArray: Show[ByteArray] = Show.fromToString
+  implicit val document: Show[Document] = Show.fromToString
+  implicit val ts: Show[Timestamp] = Show.fromToString
+  implicit val primShowPf: PolyFunction[Primitive, Show] =
```
Baccata: drop the `implicit`

modules/cats/src/smithy4s/catz/instances/HashInstances.scala
```diff
@@ -0,0 +1,26 @@
...
+
+      override def eqv(x: ByteArray, y: ByteArray): Boolean =
+        Eq[ByteArray].eqv(x, y)
+    }
+  implicit val documentHash: Hash[smithy4s.Document] =
+    Hash[String].contramap(_.toString)
+  implicit val shapeIdHash: Hash[ShapeId] = Hash[String].contramap(_.toString)
+  implicit val timeStampHash: Hash[Timestamp] =
+    Hash[Long].contramap(_.epochSecond)
+  implicit val primHashPf: ~>[Primitive, Hash] = Primitive.deriving[Hash]
```
Baccata: drop the `implicit` and also use `PolyFunction` instead of `~>`

modules/cats/src/smithy4s/catz/instances/HashInstances.scala
```diff
@@ -0,0 +1,26 @@
+package smithy4s.catz.instances
+
+import cats.{Eq, Hash}
+import cats.implicits.toContravariantOps
+import smithy4s.{~>, ByteArray, ShapeId, Timestamp}
+import smithy4s.schema.Primitive
+
+trait HashInstances {
```
Baccata: don't use `contramap` for these instances. Use `fromUniversalHashCode` instead.

modules/cats/src/smithy4s/catz/SchemaVisitorEq.scala
```diff
@@ -0,0 +1,180 @@
+package smithy4s.catz
```
Baccata: rename to `interopcats` instead.

modules/cats/src/smithy4s/catz/SchemaVisitorEq.scala
```diff
@@ -0,0 +1,180 @@
...
+
+import cats.kernel.Eq
+import smithy4s.{Schema, _}
+import smithy4s.capability.EncoderK
+import smithy4s.catz.instances.EqInstances._
+import smithy4s.schema._
+
+import java.util.UUID
+
+object EqSchemaVisitor extends SchemaVisitor[Eq] { self =>
```
Baccata: No point in having this one considering we can derive `Hash` which supersedes `Eq`

modules/cats/src/smithy4s/catz/instances/EqInstances.scala
```diff
@@ -0,0 +1,22 @@
+package smithy4s.catz.instances
+
+import cats.implicits.{catsSyntaxEq, toContravariantOps}
+import cats.kernel.Eq
+import smithy4s.{ByteArray, Document, Timestamp}
+
+trait EqInstances {
```
Baccata: No point in having this considering we can derive `Hash` which supersedes `Eq`

modules/cats/src/smithy4s/catz/SchemaVisitorHash.scala
```diff
@@ -0,0 +1,150 @@
...
+    val precomputed: Precompiler[Schema, Hash] = new Precompiler[Schema, Hash] {
+      override def apply[A](label: String, instance: Schema[A]): Hash[A] =
+        self(instance)
+    }
+    implicit val encoderKHash: EncoderK[Hash, Int] = new EncoderK[Hash, Int] {
+      override def apply[A](fa: Hash[A], a: A): Int = fa.hash(a)
+
+      override def absorb[A](f: A => Int): Hash[A] = new Hash[A] {
+        override def hash(x: A): Int = f(x)
+        override def eqv(x: A, y: A): Boolean = f(x) == f(y)
```
Baccata: That is very incorrect : you cannot decide on the equivalence of two values by comparing their hashcode. The implementation of `union` should look like the one of your current `Eq` related Schema Visitor (which will disappear in favour of this one), except that the type targeted by `EncoderK` should be enriched with the `Int`.

yisraelU: Thanks @Baccata 
To clarify , A Union of an Int or String should have the hashcode of the underlying value 
i.e 
```
sealed trait IntOrString
case class IntValue(i:Int) extends IntOrString
case class StringValue(str:String) extends IntOrString
```
the hash method derived should produce the hashCode of the internal data type

Baccata: Good question. I think the label should be added to the hash computation. If you have : 

``` 
union SomeInt {
  a: Int 
  b: Int
} 
```

then the hashcode should differ depending on whether you're in the `a` case or the `b` case, which implies that you may want to use the hashcode of the label in addition to the hashcode of the underlying value (by following best practice for computing hashcodes for composite values)

modules/cats/src/smithy4s/catz/instances/HashInstances.scala
kubukoz: why prefer `PolyFunction`?

Baccata: To keep consistent with what he used for `Show`, and also because I'm thinking more and more about deleting `~>` to avoid conflicting with the cats one. Not 100% certain.

modules/core/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=1.8.2
```
Baccata: remove this file.

modules/cats/src/smithy4s/interopcats/package.scala
```diff
@@ -0,0 +1,17 @@
...
+
+package object interopcats {
+
+  def combineHash(start: Int, hashes:Int*): Int = {
+    @tailrec
+    def loop(hashes: List[Int], acc: Int, length:Int): Int = hashes match {
+      case Nil =>   MurmurHash3.finalizeHash(acc, length)
+      case h :: t => loop(t, MurmurHash3.mix(acc, h),length + 1)
+    }
+    loop(hashes.toList, start, 1)
```
Baccata: let's use an imperative for-loop to avoid instantiating a list just to compute a hash value.

modules/cats/test/src/HashVisitorSpec.scala
```diff
@@ -0,0 +1,266 @@
...
+    object Foo {
+      val schema: Schema[Foo] = {
+        val foos = list(int)
+          .required[Foo]("foos", _.foos)
+        struct(foos)(Foo.apply)
+      }.withId(ShapeId("", "Foo"))
+    }
+    val foo = Foo(List(1, 2, 3))
+    val hashOutput = visitor(Foo.schema).hash(foo)
+    expect.eql(foo.foos.hashCode(), hashOutput)
```
Baccata: This doesn't seem correct. A more correct expectation would be something along the lines of  `expect.eql(foo.hashCode, hashOutput)`. 

EDIT : wrapping a piece of data in a structure should have an effect on the resulting hashcode. 

~Bonus points if we can get the same results as the native `hashCode` implementation~ Probs not gonna happen tbh : I've just tested it, and it looks like on equivalent empty case classes, the hash code is different. This kinda makes sense, but it means the implementation bases itself on additional information to exceeds the mere structural information. Maybe the class name, or something.

modules/cats/src/smithy4s/interopcats/SchemaVisitorHash.scala
```diff
@@ -0,0 +1,177 @@
...
+                  instance: Schema[A],
+                  get: S => Option[A]
+              ): Hash[Option[A]] = {
+                implicit val hashA: Hash[A] = self(instance)
+                Hash[Option[A]]
+              }
+            })
+          hashField.contramap(field.get)
+        }
+        fields.map(field => forField(field)).map(hash => hash.hash(x)).sum
```
Baccata: You're computing a visitor on the hot path. The `forField` should be called outside of the body of the `override def hash(x: S): Int` method.

Baccata: Moreover, the logic is off : you should call `combineHash` there.

modules/cats/src/smithy4s/interopcats/SchemaVisitorShow.scala
```diff
@@ -0,0 +1,150 @@
...
+
+object SchemaVisitorShow extends SchemaVisitor[Show] { self =>
+
+  override def primitive[P](
+      shapeId: ShapeId,
+      hints: Hints,
+      tag: Primitive[P]
+  ): Show[P] = primShowPf(tag)
+
+  override def collection[C[`2`], A](
```
Baccata: huh, looks like it actually compiles. Where is this from though ? 

Even though it works, please change to `def collection[C[_]]`

modules/cats/test/src/ShowVisitorSpec.scala
```diff
@@ -0,0 +1,258 @@
...
+    val foo = Timestamp.fromEpochSecond(now.getEpochSecond)
+    val showOutput = schemaVisitorShow(schema).show(foo)
+    expect.eql(showOutput, foo.toString)
+  }
+
+  test("struct") {
+    case class Foo(x: String, y: Option[String])
+    object Foo {
+      val schema: Schema[Foo] = {
+        StructSchema(
```
Baccata: Use `Schema.struct` to construct structure schemas, it'll give you type-safe methods for arity 1 to 22 that do not require casting.

modules/cats/test/src/ShowVisitorSpec.scala
```diff
@@ -0,0 +1,258 @@
...
+    }
+    val foo = Foo("foo", Some("bar"))
+    val showOutput = schemaVisitorShow(Foo.schema).show(foo)
+    expect.eql(showOutput, foo.toString)
+  }
+  test("struct: empty optional") {
+    case class Foo(x: String, y: Option[String])
+    object Foo {
+      val schema: Schema[Foo] = {
+        StructSchema(
```
Baccata: ditto

modules/cats/src/smithy4s/interopcats/SchemaVisitorShow.scala
```diff
@@ -0,0 +1,150 @@
...
+  ): Show[S] = {
+    def compileField[A](
+        schemaField: SchemaField[S, A]
+    ): S => String = {
+      val folder = new Field.FolderK[Schema, S, Show]() {
+        override def onRequired[AA](
+            label: String,
+            instance: Schema[AA],
+            get: S => AA
+        ): Show[AA] = self(instance)
```
Baccata, line 136: Right, so despite what I said previously, I think we could do a little bit than this, and instead of abiding by the classic `toString` we get from case classes, actually render the label in addition to the value.

modules/cats/src/smithy4s/interopcats/SchemaVisitorShow.scala
```diff
@@ -0,0 +1,150 @@
...
+  override def union[U](
+      shapeId: ShapeId,
+      hints: Hints,
+      alternatives: Vector[SchemaAlt[U, _]],
+      dispatch: Alt.Dispatcher[Schema, U]
+  ): Show[U] = {
+
+    val precomputed: Precompiler[Schema, Show] = new Precompiler[Schema, Show] {
+      override def apply[A](label: String, instance: Schema[A]): Show[A] = {
+        self(instance)
```
Baccata: Echoing my comment on the struct implementation, I think we could do better here, and have the `Show` for unions be similar to the ones of structures, by rendering the name of the `ShapeId` of the union, as well as the label. 

So the Show for `IntOrInt` would render something like : 

``` 
IntOrInt(intValue0 = 0) 
```

or 

```
IntOrInt(intValue1 = 1) 
```

modules/cats/src/smithy4s/interopcats/SchemaVisitorHash.scala
yisraelU: i didnt get to this part yet

modules/cats/test/src/HashVisitorSpec.scala
yisraelU: I think the test cases should be changed to be just a list,set,vector etc.. and remove the struct, as it would isolate what we are trying to test with these cases

modules/cats/src/smithy4s/interopcats/instances/ShowInstances.scala
```diff
@@ -0,0 +1,18 @@
...
+import smithy4s.{ByteArray, Document, ShapeId, Timestamp}
+import smithy4s.kinds.PolyFunction
+
+trait ShowInstances {
+
+  implicit  val sId: Show[ShapeId] = Show.fromToString
+   implicit val byteArray: Show[ByteArray] = Show.fromToString
+   implicit val document: Show[Document] = Show.fromToString
+   implicit val ts: Show[Timestamp] = Show.fromToString
+   val primShowPf: PolyFunction[Primitive, Show] =
```
kubukoz: **suggestion:** why make this public? I don't think people are going to use `Primitive`s directly

yisraelU: agreed

modules/cats/test/src/HashVisitorSpec.scala
yisraelU: the issue with mimicking scala case class hashcode method is that it uses the Product trait including the productPrefix method.
I can upcast all structs to Product and still use the forField method for each part
Currently I am using the productSeed from scala for combining hashes.

Baccata: > I can upcast all structs to Product 

You cannot guarantee that all structs are Scala products. Generally speaking, you must not make any assumptions on the in-memory representation of data during the implementation of Schema Visitors. 

It's fine that we cannot reproduce the native hashcode. HOWEVER, it's not fine that the hashcode of a product with one field is the same field as the hashcode of the value of the field. IE : `Foo(a).hash == a.hash` should be false, and your tests show that it's true.

yisraelU: please forget about the existing tests. thats not my focus 
as is listed in the comment of this draft PR under todo . structs is included.
I am focused on what should the hash method look like.
 Are we trying to follow/mimic anything that currently exists i.e. Kittens implementation  of HList .
Or are we rolling our own with combining all hashes (in the appropriate fashion) including field labels and ShapeId (or any other alternative you might have)

Baccata: > as is listed in the comment of this draft PR under todo . structs is included.
 
Honestly it wouldn't hurt if you put a little care in how your PR descriptions look like. Github's markdown can help.

 > I am focused on what should the hash method look like.

Well all we need to do to get inspiration that is decompile the bytecode to find out how Scala implements it. A quick search gives me a SO thread with a promising, relatively recent [answer](https://stackoverflow.com/a/45311240/3065964). 

So the idea is probably to use the MurmurHash3 API to replicate a similar algorithm : 

1. start with an accumulator initialised to a static value (using `productSeed` is fine) 
2. iterate on the fields, mutating the accumulator by calling `mix` 
3. call the finalisation method to "avalanche" the bits of the hash (that seems to be the correct terminology). 

It's unlikely we could get our implementation that would fully respect the "must produce the same value as the native hashcode" invariant, because : 
* the official `Static` construct is annotated with a comment "not for public use", and I'd rather respect that. 
* our in-memory representation is not to be considered 
* if you take into consideration bijections/refinements, that invariant would be super easy to violate. 

However, this invariant should be respected for a subset of the metamodel, namely primitives and collections of primitives. This should be encoded in the tests. For the rest of the metamodel, it's gonna be reasonably hard to describe invariants that should be respected. At the very least we should test that : 

1. Wrapping a value in another layer of data (be that structure/union/collection) leads to the wrapped value having a different hashcode to the unwrapped value. 
2. Hashcodes of the different members of a union should be different. 
3. Computing the hashcode of a recursive data type doesn't blow the stack.

build.sbt
```diff
@@ -726,6 +726,20 @@ lazy val `http4s-swagger` = projectMatrix
   )
   .http4sJvmPlatform(allJvmScalaVersions, jvmDimSettings)
 
+lazy val cats = projectMatrix
```
Baccata, line 730: don't forget to add it to the aggregate. Because the build is currently green (when it shouldn't), I thought the tests were actually run and that informed the way I approached the PR.

modules/cats/test/src/HashVisitorSpec.scala
yisraelU: > Honestly it wouldn't hurt if you put a little care in how your PR descriptions look like. Github's markdown can help.

Agreed

yisraelU:
@Baccata 
Im thinking should the Hash instance of ByteArray be implemented the way it is
```  
_.array.hashCode
```
Based upon our discussions above it should be different because its wrapped in a case class

In addition should Enumeration use the case class/object formula or the hash of the underlying values , perhaps string value combined with int value

Baccata:
> Im thinking should the Hash instance of ByteArray be implemented the way it is
> Based upon our discussions above it should be different because its wrapped in a case class

In this case the wrapping in a case class is an implementation detail, as opposed to something that's reflected in the metamodel. 

When you wrap a piece of data in a structure at the smithy level however, the hashcode driven by the schema (which reflects that smithy-level information) ought to take the wrapping into consideration. 

> In addition should Enumeration use the case class/object formula or the hash of the underlying values , perhaps string value combined with int value

Good question. For enumeration, I think we should use either the hashcode of the Int value if the enum has the `IntEnum` hint, or the hashcode of the String value otherwise.

Baccata (approved):

PR #922, opened 2023-04-21
[Compliance-tests] Account for TCP routing when testing how routers decode requests.

Baccata:
Avoid solely relying on intercepting IntendedShortCircuit exceptions when testing routers, as this method doesn't account for routers that exercise the TCP layer.

Instead, we still try to intercept that exception (which will work for routers built on top of Http4s, or routers that do an in-memory conversion to the Http4s models), but we now also try to intercept 5xx errors before waiting on a promise. A Router exercises the TCP layer will be likely to turn uncaught exceptions (like the one we raise) into a 500 http response.

Supersedes https://github.com/disneystreaming/smithy4s/pull/869


yisraelU:
Thanks for proposing this.
IIUC any 500 errors where the Input is deserialized we dont care about for the purpose of this test

Baccata:
> IIUC any 500 errors where the Input is deserialized we dont care about for the purpose of this test 

Yup, we deem that any 500 error may signal the correct routing and deserialisation of the input, which is followed by an attempt to retrieve the promise from the test.

yisraelU (approved):

PR #923, opened 2023-04-21
Add Field#contramap

kubukoz:


modules/core/src/smithy4s/schema/Field.scala
```diff
@@ -67,6 +67,7 @@ sealed abstract class Field[F[_], S, A] {
   final def foreachA(s: S)(f: A => Unit): Unit =
     f(this.get(s))
 
+  def contramap[B](f: B => S): Field[F, B, A]
```
Baccata: nitpick : I'd call it `S0` instead of `B`. If the function was modifying the `A`, then `B` would make sense. In this case it's modifying the `S`. 

Also, while we're at it, might as well add `def map[B](f: A => B): Field[F, S, B]` .... though, maybe in this very instance, `A1` would be preferable, to avoid this unfortunate sequence of letters.

kubukoz: we can't define `map` in the classical sense, because changing the last parameter requires changing the type of the `instance` type, so we'd need a `Contravariant[F]`. That obviously doesn't work for Schema and it adds an extra constraint to the signature.

Baccata: hah, yeah, goodpoint, I brainfarted.

Baccata (approved):

PR #926, opened 2023-04-24
[Core] add httpResponseCode support for int enum

### LINKED ISSUE #915: `httpResponseCode` - `intEnum` support
The `httpResponseCode` trait can be added to a member targetting an `intEnum` shape. Currently, smithy4s doesn't support this:

```smithy
$version: "2"

namespace hello

use alloy#simpleRestJson

@simpleRestJson
service WeatherService {
    operations: [GetWeather]
}

@http(method: "GET", uri: "/weather/{city}", code: 200)
@readonly
operation GetWeather {
    input := {
        @required
        @httpLabel
        city: String
    }
    output := {
        @required
        weather: String
        @required
        @httpResponseCode
        code: MyCode
    }
}

intEnum MyCode {
    NICE = 420
}
```

Hitting a service that uses this spec results in the status code not being reflected in the response:

```
[info] Status: 200 OK
[info] Body: {"weather":"bad weather in London","code":420}
```

Reproduction (`sbt run` starts a server and calls it with a client): https://github.com/kubukoz/demos/tree/smithy4s-http4s-responsecode-server

In most likelihood this is just because enums aren't handled in `HttpResponseCodeSchemaVisitor`.

### LINKED ISSUE #916: `@httpResponseCode` - client handling of 4xx/5xx codes
Related: #915

This might be intentional, but clients are unable to decode a response when a `@httpResponseCode` value falls into the 4xx/5xx range:

```smithy
$version: "2"

namespace hello

use alloy#simpleRestJson

@simpleRestJson
service WeatherService {
    operations: [GetWeather]
}

@http(method: "GET", uri: "/weather/{city}", code: 200)
@readonly
operation GetWeather {
    input := {
        @required
        @httpLabel
        city: String
    }
    output := {
        @required
        weather: String
        @required
        @httpResponseCode
        code: Integer
    }
}
```

```scala
package hello

import cats.effect.IO
import cats.effect.IOApp
import hello.WeatherService

  def run = SimpleRestJsonBuilder
    .routes(
      new WeatherService[IO] {
        def getWeather(city: String): IO[GetWeatherOutput] =
          IO.pure(GetWeatherOutput("bad weather in " + city, 420))
      }
    )
    .resource
    .flatMap { routes =>
      EmberServerBuilder
        .default[IO]
        .withHttpApp(routes.orNotFound)
        .build
    }
    .flatMap { server =>
      EmberClientBuilder
        .default[IO]
        .build
        .flatMap { c =>
          SimpleRestJsonBuilder(WeatherService)
            .client(c)
            .uri(server.baseUri)
            .resource
            .evalMap {
              _.getWeather("London").flatMap(IO.println(_))
            }
        }
    }
    .use_
}

```

```
[error] smithy4s.http.UnknownErrorResponse: status 420, headers: Map(Date -> Thu, 20 Apr 2023 , Connection -> keep-alive, Content-Length -> 46, Content-Type -> application/json), body:
[error] {"weather":"bad weather in London","code":420}
```

Is this the expected behavior, or should the client still succeed in its call, giving the caller the ability to inspect the response themselves?

Full repro: https://github.com/kubukoz/demos/tree/smithy4s-916

yisraelU:
Adds support for in the HttpResponseCodeSchemaVisitor to support targeting an IntEnum with the httpresponseCode trait
closes #915

### DIFF
--- a/modules/core/src/smithy4s/http/internals/HttpResponseCodeSchemaVisitor.scala
+++ b/modules/core/src/smithy4s/http/internals/HttpResponseCodeSchemaVisitor.scala
@@ -18,19 +18,10 @@ package smithy4s.http.internals
 
 import smithy4s.capability.Contravariant
 import smithy4s.Hints
-import smithy4s.schema.Field
-import smithy4s.schema.Schema
-import smithy4s.schema.SchemaField
-import smithy4s.schema.SchemaVisitor
+import smithy4s.schema.{EnumValue, Field, Primitive, Schema, SchemaField, SchemaVisitor}
 import smithy4s.ShapeId
-
-import smithy4s.http.internals.HttpResponseCodeSchemaVisitor.{
-  ResponseCodeExtractor,
-  NoResponseCode,
-  RequiredResponseCode,
-  OptionalResponseCode
-}
-import smithy4s.schema.Primitive
+import smithy4s.IntEnum
+import smithy4s.http.internals.HttpResponseCodeSchemaVisitor.{NoResponseCode, OptionalResponseCode, RequiredResponseCode, ResponseCodeExtractor}
 
 class HttpResponseCodeSchemaVisitor()
     extends SchemaVisitor.Default[ResponseCodeExtractor] {
@@ -48,6 +39,15 @@ class HttpResponseCodeSchemaVisitor()
       }
     case _ => NoResponseCode
   }
+
+
+  override def enumeration[E](shapeId: ShapeId, hints: Hints, values: List[EnumValue[E]], total: E => EnumValue[E]): ResponseCodeExtractor[E] =
+    if (hints.has[smithy.api.HttpResponseCode] && hints.has[IntEnum]) {
+      Contravariant[ResponseCodeExtractor].contramap(HttpResponseCodeSchemaVisitor.int)(e => total(e).intValue)
+    } else {
+      NoResponseCode
+    }
+
 
   override def struct[S](
       shapeId: ShapeId,
--- a/modules/core/src/smithy4s/http/internals/HttpResponseCodeSchemaVisitor.scala
+++ b/modules/core/src/smithy4s/http/internals/HttpResponseCodeSchemaVisitor.scala
@@ -18,10 +18,22 @@ package smithy4s.http.internals
 
 import smithy4s.capability.Contravariant
 import smithy4s.Hints
-import smithy4s.schema.{EnumValue, Field, Primitive, Schema, SchemaField, SchemaVisitor}
+import smithy4s.schema.{
+  EnumValue,
+  Field,
+  Primitive,
+  Schema,
+  SchemaField,
+  SchemaVisitor
+}
 import smithy4s.ShapeId
 import smithy4s.IntEnum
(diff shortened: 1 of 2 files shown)


modules/core/src/smithy4s/http/internals/HttpResponseCodeSchemaVisitor.scala
```diff
@@ -49,6 +40,15 @@ class HttpResponseCodeSchemaVisitor()
     case _ => NoResponseCode
   }
 
+
+  override def enumeration[E](shapeId: ShapeId, hints: Hints, values: List[EnumValue[E]], total: E => EnumValue[E]): ResponseCodeExtractor[E] =
+    if (hints.has[smithy.api.HttpResponseCode] && hints.has[IntEnum]) {
```
kubukoz, line 61: **thought:** maybe we don't need the second part of the conditional, as it should already be covered by smithy's own validation?

modules/core/src/smithy4s/http/internals/HttpResponseCodeSchemaVisitor.scala
```diff
@@ -49,6 +40,15 @@ class HttpResponseCodeSchemaVisitor()
     case _ => NoResponseCode
   }
 
+
+  override def enumeration[E](shapeId: ShapeId, hints: Hints, values: List[EnumValue[E]], total: E => EnumValue[E]): ResponseCodeExtractor[E] =
```
kubukoz: **issue:** this should ideally be covered by a test, even if the implementation is trivial it gives us safety that we won't accidentally break this in the future

kubukoz (changes requested):

kubukoz:
I'd target 0.17 with this, but arguably it can be considered "behavior changing", so "breaking". The old behavior was wrong, though, so that'd just be a bugfix.

modules/core/src/smithy4s/http/internals/HttpResponseCodeSchemaVisitor.scala
yisraelU: never sure about this type of scenario, like how defensively should I code..
Either way is fine for me

kubukoz: yeah, I'm never sure either :) it's fine to keep it as-is

modules/core/src/smithy4s/http/internals/HttpResponseCodeSchemaVisitor.scala
yisraelU: done

modules/core/test/src/smithy4s/http/internals/HttpResponseCodeSchemaVisitorSpec.scala
```diff
@@ -0,0 +1,122 @@
...
+            .required[SampleResponse1]("code", _.code)
+            .addHints(smithy.api.HttpResponseCode())
+        )(SampleResponse1.apply)
+        .withId("", "SampleResponse")
+
+    val id: ShapeId = ShapeId("", "SampleResponse")
+  }
+
+  test(
+    "applying HttpResponseCode works on an enumeration with an intENum hint"
```
kubukoz: **nitpick:** 
```suggestion
    "applying HttpResponseCode works on an enumeration with an intEnum hint"
```

kubukoz (approved):

PR #927, opened 2023-04-25
Add AWS signature tests 

Baccata:
* Adds property tests to verify that our implementation of the AWS signature algorithm behaves the same way as the one from AWS (at least for unary operations where the full body of the request can be loaded in memory). This is done only on JVM as it relies on calling the AWS Java SDK. 
* Fixes edge cases in the implementation of the signature algorithm. 
* Also moves the other test suites so that they are executed on all platforms.


modules/aws-kernel/src/smithy4s/aws/AwsSignature.scala
```diff
@@ -1,120 +0,0 @@
...
-/**
-  * Implementation of the AWS Signature Version 4 algorithm.
-  *
-  * See https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html
-  *
-  * The [[baguette.CryptoPlatformCompat]] contains all the platform-specific
-  * code that has to do with hashing/encrypting, and is provided for the
-  * three platforms (jvm/js/native.)
-  */
-private[aws] object AwsSignature {
```
Baccata, line 32: This is the former implementation I had yet to remove.

.scala-steward.conf
```diff
@@ -1,6 +1,8 @@
 updates.ignore = [
   { groupId = "org.scalameta" },
-  { groupId = "org.webjars" }
+  { groupId = "org.webjars" },
+  { groupId = "org.webjars.npm"},
+  { groupId = "software.amazon.awssdk" }
```
daddykotex, line 5: one thing we could do is:

```
dependencyOverrides = [
  {
    dependency = { groupId = "software.amazon.awssdk" },
    pullRequests = { frequency = "14 day" }
  }
]
```

Instead of disabling it completely, just reducing the frequency is probably useful

Details about the config: https://github.com/scala-steward-org/scala-steward/blob/main/docs/repo-specific-configuration.md

Baccata: Tbh I'm not sure it's even worth it : the signature algorithm ain't gonna change and that's the only thing we're really after, and it's a dependency we only use in tests. Even if it changes less often, I'll find it annoying to use resources to build version bumps for it when the bump would provide literally zero value ...

modules/aws-http4s/test/src-jvm/smithy4s/aws/internals/AwsSignatureTest.scala
```diff
@@ -0,0 +1,224 @@
...
+      )
+    }
+  }
+
+  def testSignature(
+      testInput: TestInput
+  ): IO[Expectations] = {
+    import testInput._
+
+    val hardClock = new java.time.Clock {
```
daddykotex: There's `Clock.fixed` if I'm not mistaken that would do exactly that

.scala-steward.conf
daddykotex: I agree w/ that

daddykotex (approved):
Great job, thanks for taking that on

build.sbt
```diff
@@ -343,9 +346,11 @@ lazy val `aws-http4s` = projectMatrix
     latest2ScalaVersions,
     jvmDimSettings ++ Seq(
       Test / smithySpecs ++= Seq(
-        (ThisBuild / baseDirectory).value / "sampleSpecs" / "aws_example.smithy",
         (ThisBuild / baseDirectory).value / "sampleSpecs" / "dynamodb.2012-08-10.json",
         (ThisBuild / baseDirectory).value / "sampleSpecs" / "lambda.json"
+      ),
+      libraryDependencies ++= Seq(
+        "software.amazon.awssdk" % "aws-core" % "2.20.49"
       )
```
kubukoz: **question:** should this be a test-only dependency instead?

Baccata: indeed it should !

Baccata: https://github.com/disneystreaming/smithy4s/pull/927/commits/17a3f03cda203d0cde24f317579e97dba3c81a2a

PR #952, opened 2023-05-09
Replace IntEnum hint with enumType field in enumerations

miguel-vila:


modules/core/src/smithy4s/schema/Schema.scala
```diff
@@ -121,7 +121,7 @@ object Schema {
   final case class PrimitiveSchema[P](shapeId: ShapeId, hints: Hints, tag: Primitive[P]) extends Schema[P]
   final case class CollectionSchema[C[_], A](shapeId: ShapeId, hints: Hints, tag: CollectionTag[C], member: Schema[A]) extends Schema[C[A]]
   final case class MapSchema[K, V](shapeId: ShapeId, hints: Hints, key: Schema[K], value: Schema[V]) extends Schema[Map[K, V]]
-  final case class EnumerationSchema[E](shapeId: ShapeId, hints: Hints, values: List[EnumValue[E]], total: E => EnumValue[E]) extends Schema[E]
+  final case class EnumerationSchema[E](shapeId: ShapeId, hints: Hints, values: List[EnumValue[E]], tag: EnumTag, total: E => EnumValue[E]) extends Schema[E]
```
Baccata: nitpick : I'd like the `tag` field before the `values`

modules/core/src/smithy4s/schema/SchemaVisitor.scala
```diff
@@ -24,7 +24,7 @@ trait SchemaVisitor[F[_]] extends (Schema ~> F) { self =>
   def primitive[P](shapeId: ShapeId, hints: Hints, tag: Primitive[P]): F[P]
   def collection[C[_], A](shapeId: ShapeId, hints: Hints, tag: CollectionTag[C], member: Schema[A]): F[C[A]]
   def map[K, V](shapeId: ShapeId, hints: Hints, key: Schema[K], value: Schema[V]): F[Map[K, V]]
-  def enumeration[E](shapeId: ShapeId, hints: Hints, values: List[EnumValue[E]], total: E => EnumValue[E]): F[E]
+  def enumeration[E](shapeId: ShapeId, hints: Hints, enumTag: EnumTag, values: List[EnumValue[E]], total: E => EnumValue[E]): F[E]
```
kubukoz: **nitpick:** I'd call this `tag` for consistency with collections

modules/core/src/smithy4s/schema/EnumTag.scala
```diff
@@ -0,0 +1,8 @@
+package smithy4s.schema
+
+sealed trait EnumTag
```
kubukoz, line 19: **suggestion:** I'd move this to live next to `EnumerationSchema` rather than in a new file

modules/compliance-tests/src/smithy4s/compliancetests/internals/CanonicalSmithyDecoder.scala
```diff
@@ -40,7 +39,7 @@ object CanonicalSmithyDecoder {
       schema: Schema[A]
   ): Document.Decoder[A] = {
     decoder.fromSchema(
-      schema.transformHintsTransitively(_.filter(_.keyId == IntEnum.id))
```
miguel-vila, line 43: we were not sure about this change, it didn't break any test in this module but I think there's only one test of regular string enumerations, we could add a similar one for int enumerations but not sure I get the point of these tests

Baccata:
you'll need to sync the series/0.18 branch and run `examples/compile` before checking-in the code

modules/core/src/smithy4s/schema/SchemaVisitor.scala
miguel-vila: addressed at https://github.com/disneystreaming/smithy4s/pull/952/commits/aae31ea73ebf44889eedf24bb35c1e4812d33211

Baccata:
@miguel-vila, you'll need to run `sbt headerCreateAll`

modules/core/src/smithy4s/schema/EnumTag.scala
Baccata: I disagree. I'd rather it stayed in its own file.

modules/core/src/smithy4s/schema/SchemaVisitor.scala
```diff
@@ -54,7 +54,7 @@ object SchemaVisitor {
     override def primitive[P](shapeId: ShapeId, hints: Hints, tag: Primitive[P]): F[P] = default
     override def collection[C[_], A](shapeId: ShapeId, hints: Hints, tag: CollectionTag[C], member: Schema[A]): F[C[A]] = default
     override def map[K, V](shapeId: ShapeId, hints: Hints, key: Schema[K], value: Schema[V]): F[Map[K,V]] = default
-    override def enumeration[E](shapeId: ShapeId, hints: Hints, values: List[EnumValue[E]], total: E => EnumValue[E]): F[E] = default
+    override def enumeration[E](shapeId: ShapeId, hints: Hints, tag: EnumTag, values: List[EnumValue[E]], total: E => EnumValue[E]): F[E] = default
```
kubukoz, line 57: **question:** @Baccata, WDYT about having `default` being `def default[A](schema: Schema[A]): F[A]`? It'd give the implementation the ability to display a more helpful debug message in case such a case is unexpected. Otherwise you only have the stack trace which tells you e.g. "enumeration" but doesn't provide a ShapeId or any of that

kubukoz (approved):

modules/core/src/smithy4s/schema/SchemaVisitor.scala
Baccata: Yeah sounds good. Feel free to open another PR

Baccata (approved):
Thanks for this

PR #953, opened 2023-05-09
Rework codec interfaces

Baccata:
Rework the UnaryCodecs interfaces to be a static group of the codecs necessary for the interpreters. They are now accompanied by a `Make` construct to build them from an Endpoint, as a certain amount of encoding and decoding logic depends on hints that are carried by the Endpoint.

PR #956, opened 2023-05-10
[Aws-Http4s] Add the http method to the Request when deriving a Message Encoder from HttpEndpoint

yisraelU:
when testing the restJson1 protocol, I discovered that all Requests were set as POST, which IIUC comes from the default request over here https://github.com/disneystreaming/smithy4s/blob/83fbb92d89649f238dced745a7af0190521853e7/modules/aws-http4s/src/smithy4s/aws/internals/AwsUnaryEndpoint.scala#L81
While this makes sense for certain protocols it seems it should be overwritten wherever the `@http` trait is being used


Baccata:
Closing cause [this](https://github.com/disneystreaming/smithy4s/pull/953/files) supersedes it.

Issue #943, opened 2023-05-05
[AWS SDK] aws.protocols#restJson1


Baccata:
See documentation [here](https://smithy.io/2.0/aws/protocols/aws-json-1_0-protocol.html) 

In theory, all building blocks are already present, as the protocol is somewhat similar to our own  `alloy#simpleRestJson`.

- [ ] Wire the necessary logic so that we can support clients for AWS services that abide by the `aws.protocols#restJson1`. See [https://github.com/disneystreaming/smithy4s/blob/83fbb92d89649f238dced745a7af0190521853e7/modules/aws-http4s/src/smithy4s/aws/AwsClient.scala#L67] 

- [ ] Wire the compliance tests for this specific protocol. See [here](https://github.com/disneystreaming/smithy4s/blob/83fbb92d89649f238dced745a7af0190521853e7/modules/aws-http4s/src/smithy4s/aws/AwsClient.scala#L67) for how it's done for the `aws.protocols#awsJson1_x` protocols

Baccata:
Closing, the implementation is done and will be released in 0.18

Baccata:
Supported in the 0.18.x lineage

PR #957, opened 2023-05-11
[Aws] Rest json1 protocol implementation

yisraelU:
Initiate RestJson1 implementation #943 
So I am using the compliance tests as a guide for the path forward.

  ### partial list

- ~versions of the http protocol can be specified in the restJson1 trait~
- mediaType trait needs to define content type on both blobs and strings
- content-type for blob needs to default to `application/octet-stream` when targeted by httpPayload
- httpChecksumRequired trait needs to be supported , this adds a header


modules/aws-kernel/src/smithy4s/aws/AwsProtocol.scala
kubukoz, line 1: **suggestion:** add the new protocol to `val supportedProtocols` please

yisraelU: good point

Baccata:
I think we should create issues for what needs to be handled separately in small PRs against this branch, now that we have better composable building blocks for creating requests and decoding responses. It would encourage contributions and would prevent having to wait on a single person for the whole implementation to meet compliance. 

> versions of the http protocol can be specified in the restJson1 trait

This ain't gonna be addressed in smithy4s, as it only operates at the `Client` level and that's something that is handled at the `ClientBuilder` level.

yisraelU:
> I think we should create issues for what needs to be handled separately in small PRs against this branch, now that we have better composable building blocks for creating requests and decoding responses. It would encourage contributions and would prevent having to wait on a single person for the whole implementation to meet compliance.

  will do this 

> > versions of the http protocol can be specified in the restJson1 trait
> 
> This ain't gonna be addressed in smithy4s, as it only operates at the `Client` level and that's something that is handled at the `ClientBuilder` level.

I see, wasnt sure if it could be constrained in the http4s Client interface. I see it cannot . So I will remove this from the list

kubukoz:
can we extract the HTTP version info from a client Request? that'd allow us to do some middleware magic and log a warning of some sort, if it's not matching the supported versions.

yisraelU:
> can we extract the HTTP version info from a client Request? that'd allow us to do some middleware magic and log a warning of some sort, if it's not matching the supported versions.

for Http4s yes ```request.httpVersion```

However not sure I see value in that .
There is no guarantee they are using middleware and if they do, this would come as a surprise.
This belongs at the protocol level , like in the RequestEncoder

kubukoz:
>  There is no guarantee they are using middleware and if they do, this would come as a surprise.

it could be part of the interpreter, that's what I mean - we'd wrap the client in the middleware ourselves.

RequestEncoder doesn't know about the client so it'd be too early, I think.

yisraelU:
> > There is no guarantee they are using middleware and if they do, this would come as a surprise.
> 
> it could be part of the interpreter, that's what I mean - we'd wrap the client in the middleware ourselves.
> 
> RequestEncoder doesn't know about the client so it'd be too early, I think.

@kubukoz RequestEncoder is abstract
I am referring to implementations 
i.e.  when calling [inputToRequest](https://github.com/disneystreaming/smithy4s/blob/4e678c5f89599f962dc18fb7dcdf3d5d6c0a402b/modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sClientEndpoint.scala#LL58C36-L58C36) in http4s client impl , we can have access to the protocol payload to determine which version Http to set

Baccata:
@yisraelU, if you have more recent changes, please push them, even if they're not complete. 

I appreciate your motivation on this, but some changes in the current state will probably have to be reverted or tweaked (like, the split of header values in assertions), and I'd rather not comment until the cards are on the table. 

Additionally : remember my [previous comment](https://github.com/disneystreaming/smithy4s/pull/957#issuecomment-1545842665) : 

> I think we should create issues for what needs to be handled separately in small PRs against this branch, now that we have better composable building blocks for creating requests and decoding responses. It would encourage contributions and would prevent having to wait on a single person for the whole implementation to meet compliance. 

To your credit, you have created some issues, but I'd much rather the work of fixing those issues was done in small units of work as opposed to one big PR that attempts at fixing everything.

yisraelU:
> @yisraelU, if you have more recent changes, please push them, even if they're not complete.
> 
> I appreciate your motivation on this, but some changes in the current state will probably have to be reverted or tweaked (like, the split of header values in assertions), and I'd rather not comment until the cards are on the table.
> 
> Additionally : remember my [previous comment](https://github.com/disneystreaming/smithy4s/pull/957#issuecomment-1545842665) :
> 
> > I think we should create issues for what needs to be handled separately in small PRs against this branch, now that we have better composable building blocks for creating requests and decoding responses. It would encourage contributions and would prevent having to wait on a single person for the whole implementation to meet compliance.
> 
> To your credit, you have created some issues, but I'd much rather the work of fixing those issues was done in small units of work as opposed to one big PR that attempts at fixing everything.

So it's not so simple , because it's a discovery process as to the source of the problems.
So this pr is meant to enable that and is not going to be merged it's just a place I can aggregate all the changes necessary to pass the protocol tests.
Once that's done ,, I'll figure out or discuss how exactly these changes should be open as a pr, due to conflict with the simple rest JSON protocol

Baccata:
> it's just a place I can aggregate all the changes necessary to pass the protocol tests. 

I can understand it's not "simple", but my point is that by doing this in a vacuum you're essentially forcing me to choose between : 
1. waiting on you to pass all tests and report your findings, which essentially makes you a bottleneck, and alienates myself and other contributors.
2. disregarding you're working on implementing this protocol in order to make progress towards the 0.18 release, thus competing against you/alienating you. 

Neither this options are great, obviously, which is why I'm asking you to post the latest state and explain your findings as soon as possible so that we can talk it out and split the work. 

As a gesture of good faith, I'll start with my thoughts : one thing that is now obvious is that the current implementation of the compliance test assertions is incorrect for AWS protocols, on both the series/0.18 branch and this PR branch, because AWS forgot to abide by common-sense when they implemented their header serialisation logic. This essentially means that : 

* the (de)serialisation of header values must account for AWS-specific semantics in the case of AWS protocols 
* the splitting of expected header values in assertions should not occur when AWS protocol implementations are being tested, but rather the raw map values should be compared in that case.

I've started to work on this first point in order to get ahead, as solving this correctly touches several layers of the library. However, I think the first thing that must occur is the creation of a feature branch that would address this second point, and the current state of your branch contains elements that can be reused to that end. In its initial state, the this feature branch would obviously contain a number of failing tests. The feature branch should be protected and numerous PRs should be raised in order to address the tests (including what I'm currently working on).

yisraelU:
@Baccata Of course I am going to push.
I was simply explaining why I dumped everything into one Branch vs separate branches and PRs

Baccata:
@yisraelU, thanks for this. I'm closing this PR in favour of this following feature branch, with an initial commit specifically based for just changing the implementation of assertions of compliance-tests to reduce extrapolation on our end, and to have the SimpleRestJson pass the tests. 

Because a fair bit of it is based on your work, I've taken the liberty to add you as co-author. Let me know if that's okay. 

https://github.com/disneystreaming/smithy4s/pull/1018

PR #964, opened 2023-05-16
[Core : Fix]fix string enum for StringAndBlobCodecSchemaVisitor

yisraelU:
A string enum should be treated like a string and therefore when providing a special string codec supporting media type , string enum case should be included

### DIFF
--- a/modules/core/test/src/smithy4s/http/StringAndBlobSpec.scala
+++ b/modules/core/test/src/smithy4s/http/StringAndBlobSpec.scala
@@ -75,6 +75,27 @@ class StringAndBlobSpec() extends munit.FunSuite {
     expect.same(Right(input), roundTripped)
     expect.same(HttpMediaType("text/csv"), mediaType)
   }
+  test("String Enum  ") {
+    val input = StringEnumBody(StringEnum.INTERESTING)
+    val codec = stringsAndBlobs.compileCodec(StringEnumBody.schema)
+    val result = stringsAndBlobs.writeToArray(codec, input)
+    val roundTripped = stringsAndBlobs.decodeFromByteArray(codec, result)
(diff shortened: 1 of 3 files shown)


kubukoz:
can we have a test?

modules/core/test/src/smithy4s/http/StringAndBlobSpec.scala
```diff
@@ -75,6 +75,27 @@ class StringAndBlobSpec() extends munit.FunSuite {
     expect.same(Right(input), roundTripped)
     expect.same(HttpMediaType("text/csv"), mediaType)
   }
+  test("String Enum  ") {
+    val input = StringEnumBody(StringEnum.INTERESTING)
```
kubukoz, line 79: hmm INTERESTING code :D

modules/core/test/src/smithy4s/http/StringAndBlobSpec.scala
```diff
@@ -75,6 +75,27 @@ class StringAndBlobSpec() extends munit.FunSuite {
     expect.same(Right(input), roundTripped)
     expect.same(HttpMediaType("text/csv"), mediaType)
   }
+  test("String Enum  ") {
```
kubukoz:
```suggestion
  test("String Enum") {
```

kubukoz (approved):

kubukoz:
LGTM, feel free to merge on green

modules/core/test/src/smithy4s/http/StringAndBlobSpec.scala
yisraelU: Lol , I have had one too many foo and bar

Issue #958, opened 2023-05-12
SchemaVisitor.Default: make `default` receive schema as parameter


kubukoz:
Proposed in https://github.com/disneystreaming/smithy4s/pull/952#discussion_r1191440382

miguel-vila:
happy to pick this up

Baccata:
Closing. See comment here : https://github.com/disneystreaming/smithy4s/pull/968#issuecomment-1742712646

PR #968, opened 2023-05-17
Address #958

miguel-vila:


Baccata:
Here's what I think we should do : 

1. Introduce a `smithy4s.schema.SchemaType` enumeration (just a good old ADT with only case objects will do)
2. Turn the default method into a `def default[A](id: ShapeId, hints: Hints, schemaType: SchemaType): F[A]`

kubukoz:
I like that.

modules/core/src/smithy4s/schema/SchemaVisitor.scala
```diff
@@ -29,8 +29,8 @@ trait SchemaVisitor[F[_]] extends (Schema ~> F) { self =>
   def union[U](shapeId: ShapeId, hints: Hints, alternatives: Vector[SchemaAlt[U, _]], dispatch: Alt.Dispatcher[Schema, U]): F[U]
   def biject[A, B](schema: Schema[A], bijection: Bijection[A, B]): F[B]
   def refine[A, B](schema: Schema[A], refinement: Refinement[A, B]): F[B]
-  def lazily[A](suspend: Lazy[Schema[A]]): F[A]
-  def nullable[A](schema: Schema[A]): F[Option[A]]
+  def lazily[A](shapeId: ShapeId, hints: Hints, suspend: Lazy[Schema[A]]): F[A]
```
Baccata, line 32: Nope, I'm veto-ing this change (same for nullable). If this change is required for the feature to work, I'll veto the feature.

Baccata (changes requested):

Baccata:
Closing this as well as the originating issue : considering the requirement for compiling schemas is `Schema ~> F`, which can take many forms, and considering this feature is not helping the implementation of the OOTB integrations, I don't think it's required at all. 

If you come up with a pattern in userland that you deem really useful, I'm happy to take a look and potentially integrate it, but right now this feels like a "it may come in handy" more than anything.

PR #975, opened 2023-05-25
Caliban (graphql) interop

kubukoz:


Baccata:
Whilst this is super cool, I'm opposed to accepting this feature in this repository. Here some reasons why : 

1. Limit maintenance burden by focusing on the currently supported feature-set, namely : 
    * server/client side http4s-based implementation for a rest-json protocol (ie `simpleRestJson`) 
    * http4s-based AWS SDK
    * CLI derivation 
    * dynamic instantiation of service and schema instances
2. Prevent coupling with the versioning cycles of the ZIO library (even transitively)
3. Protect Disney Engineers who work on this project from having to explain why a non-trivial module exists for a non-Disney use cases, which unavoidably increases maintenance burden and levels of responsibilities
4. Prevent already large build times for increasing even more.  

For transparency : my desire would be for this project to be transferred to its own organisation, which could contain additional integrations (such as this one). The timing for me to investigate if it'd be even possible is not right, as our company is undergoing some pretty heavy restructuring and we have much more urgent matters to deal with internally.

kubukoz:
That's fair, for the time being I'll move this to another org, and we can look into other options in the future.

> Prevent coupling with the versioning cycles of the ZIO library (even transitively)

tbf I don't think there will be a 3.0 soon so I wouldn't worry about it _that_ much, but it's still a valid point

Issue #773, opened 2023-01-30
Refine member Scaladoc logic to only include structs/operations


kubukoz:
Another follow-up to #713.

We only check for unions here:

https://github.com/disneystreaming/smithy4s/blob/series/0.17/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala#L760-L780

I believe we should be stricter and only hit `else` if `shape` is a struct OR operation (see #772).

Currently, it also matches:

## enums

```smithy
$version: "2"

namespace demo

/// This is an enum
enum AnEnum {
  /// this is a tomato
  TOMATO = "tomat"
}
```

```scala
/** This is an enum
  * @param TOMATO
  *   this is a tomato
  */
sealed abstract class AnEnum(_value: String, _name: String, _intValue: Int, _hints: Hints) extends Enumeration.Value {
```

## collections (lists/maps)

```smithy
/// this is a list
list AList {
  /// this is a member
  member: String
}

/// this is a map
map AMap {
  /// this is a key
  key: String
  /// this is a value
  value: String
}
```

```scala
/** this is a map
  * @param key
  *   this is a key
  * @param value
  *   this is a value
  */
object AMap extends Newtype[Map[String, String]] {
```

zetashift:
Wouldn't `operation`s rendering also be affected if it's only `struct`s?

kubukoz:
good catch :) that's true.

Althouuuuugh... for operations, it's not really "operation members", but rather members of the input shape, which is 100% a struct. So the `if` will probably still refer to structs

Baccata:
For collections, the content of the documentation is mostly correct, it's just that `@param` is the wrong annotation. We just need to drop it in favour of a custom header.

kubukoz:
Any suggestions for the header? I'd start with empty string.

Baccata:
Yeah that's fine. As long as Maps have got `key` and `value`, and `list` have "member" or "content", it's all good

PR #980, opened 2023-05-29
fix: sync enum trait rendering with enum shape as possible

i10416:
Previously, rendering result of [enum trait](https://smithy.io/2.0/spec/constraint-traits.html#enum-trait) was much different from that of [enum shape](https://smithy.io/2.0/spec/simple-types.html#enum) in that
1. smithy4s dropped enum variant hints
2. sealed abstract classes generated from enum trait held `smithy.api#enum` hints
3. sealed abstract classes from enum trait didn't have `@param` properties generated.

For example, the following smithy definition generated the Scala code below.

```smithy
@enum([
{
    value: "HEAD",
    name: "HEAD",
    documentation: "head"
},
{
    value: "TAIL",
    name: "TAIL",
    documentation: "tail",
    deprecated: true
}
])

string Coin
```

```scala
// When Coin is generated from enum shape and enum variant has
// documentation, here are `@param` properties for each variant with doc.
sealed abstract class Coin ...

...

object Coin extends Enumeration[Suit] with ...
  ...
  val hints: Hints = Hints(
    smithy.api.Enum(...) // <- When Coin is generated from enum shape, smithy.api.Enum does not exist here.
  )

  /** head */
  case object HEAD extends Coin("HEAD", "HEAD", 0, Hints()) // <- Hints is not empty when Coin is enum shape.
  /** tail  */
  @deprecated
  case object TAIL extends Coin("TAIL", "TAIL", 1, Hints()) // too

```

After this commits, enum variants generated from enum trait also hold hints such as documentation or deprecation. In addition, sealed abstract classes generated from enum trait no longer have redundant `smithy.api#enum hints`.

See https://github.com/disneystreaming/smithy4s/pull/980/files#diff-1374f8e9776053385d67f4bc78d81b77bd7730a72ebd38801479005fba548fa0 for instance.

Note that enum trait and enum shape are not 100% compatible.
For example, enum trait support neither `deprecated$since` nor `deprecated$message`.

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
@@ -312,6 +312,27 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
       override def stringShape(shape: StringShape): Option[Decl] =
         (shape match {
           case T.enumeration(e) => {
+            val namedEnumTrait =
+              if (e.hasNames()) e
+              else {
+                val defs = e.getValues().asScala.zipWithIndex.map {
+                  case (enumDef, idx) =>
+                    enumDef.getName().asScala match {
+                      case Some(_) => enumDef
+                      case None =>
+                        enumDef
+                          .toBuilder()
+                          .name(
+                            EnumUtil
+                              .enumValueClassName(None, enumDef.getValue, idx)
+                          )
+                          .build()
+                    }
+                }
+                val builder = e.toBuilder().clearEnums()
+                defs.foreach(builder.addEnum)
+                builder.build()
+              }
             val pseudoEnumShape = EnumShape
               .builder()
               .source(shape.getSourceLocation())
@@ -320,7 +341,7 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
               )
               .asInstanceOf[EnumShape.Builder]
               .id(shape.getId())
-              .setMembersFromEnumTrait(e)
+              .setMembersFromEnumTrait(namedEnumTrait)
               .build()
             enumShape(pseudoEnumShape)
           }
--- a/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
@@ -312,20 +312,11 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
       override def stringShape(shape: StringShape): Option[Decl] =
         (shape match {
           case T.enumeration(e) => {
-            val enumTrait = shape
-              .findTrait(ShapeId.from("smithy.api#enum"))
-              .asScala
-              .getOrElse(
-                throw new Exception(
-                  "smithy.api#enum shape is missing for enum trait"
-                )
-              )
             val pseudoEnumShape =
               EnumShape.fromStringShape(shape, true).asScala match {
                 case Some(shape) =>
                   shape
                     .toBuilder()
-                    .addTrait(enumTrait)
                     .asInstanceOf[EnumShape.Builder]
                     .build()
                 case None => {
@@ -354,7 +345,6 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
                   }
                   EnumShape
                     .builder()
-                    .addTrait(enumTrait)
                     .id(shape.getId())
                     .source(shape.getSourceLocation())
                     .addTraits(
--- a/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
@@ -312,37 +312,45 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
       override def stringShape(shape: StringShape): Option[Decl] =
         (shape match {
           case T.enumeration(e) => {
-            val namedEnumTrait =
-              if (e.hasNames()) e
-              else {
-                val defs = e.getValues().asScala.zipWithIndex.map {
-                  case (enumDef, idx) =>
-                    enumDef.getName().asScala match {
-                      case Some(_) => enumDef
-                      case None =>
-                        enumDef
-                          .toBuilder()
-                          .name(
-                            EnumUtil
-                              .enumValueClassName(None, enumDef.getValue, idx)
-                          )
-                          .build()
+            val pseudoEnumShape =
+              EnumShape.fromStringShape(shape).asScala match {
+                case Some(shape) => shape
+                case None => {
+                  val namedEnumTrait = {
+                    val defs = e.getValues().asScala.zipWithIndex.map {
+                      case (enumDef, idx) =>
+                        enumDef.getName().asScala match {
+                          case Some(_) => enumDef
+                          case None =>
+                            enumDef
+                              .toBuilder()
+                              .name(
+                                EnumUtil
+                                  .enumValueClassName(
+                                    None,
+                                    enumDef.getValue,
+                                    idx
+                                  )
+                              )
+                              .build()
+                        }
                     }
+                    val builder = e.toBuilder().clearEnums()
+                    defs.foreach(builder.addEnum)
+                    builder.build()
+                  }
+                  EnumShape
+                    .builder()
+                    .source(shape.getSourceLocation())
+                    .addTraits(
+                      shape.getAllTraits().values()
+                    )
+                    .asInstanceOf[EnumShape.Builder]
+                    .id(shape.getId())
+                    .setMembersFromEnumTrait(namedEnumTrait)
+                    .build()
                 }
-                val builder = e.toBuilder().clearEnums()
-                defs.foreach(builder.addEnum)
-                builder.build()
               }
-            val pseudoEnumShape = EnumShape
-              .builder()
-              .source(shape.getSourceLocation())
-              .addTraits(
-                shape.getAllTraits().values()
-              )
-              .asInstanceOf[EnumShape.Builder]
-              .id(shape.getId())
-              .setMembersFromEnumTrait(namedEnumTrait)
-              .build()
             enumShape(pseudoEnumShape)
           }
           case _ => this.getDefault(shape)
--- a/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
@@ -312,9 +312,22 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
       override def stringShape(shape: StringShape): Option[Decl] =
         (shape match {
           case T.enumeration(e) => {
+            val enumTrait = shape
+              .findTrait(ShapeId.from("smithy.api#enum"))
+              .asScala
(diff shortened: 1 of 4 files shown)


modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -311,31 +311,19 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
-              .map { case (value, index) =>
-                EnumValue(
-                  value.getValue(),
-                  index,
-                  EnumUtil.enumValueClassName(
-                    value.getName().asScala,
-                    value.getValue,
-                    index
-                  ),
-                  hints = Nil
```
i10416, line 328: This dropped all variants' hints even if there were documentation or deprecation although https://github.com/disneystreaming/smithy4s/pull/980/files#diff-3d87f30f2b03db7c8e012d2c726b1e1a0f1227fe70d2e5c24166a9046c4428a5L337 preserved documentation and deprecation.

modules/codegen/test/src/smithy4s/codegen/internals/RendererSpec.scala
```diff
@@ -85,6 +85,144 @@ final class RendererSpec extends munit.FunSuite {
...
+        )
+      ) match {
+        case None =>
+          fail(
+            "enum must be rendered as a sealed abstract class and its companion object"
+          )
+        case Some(code) => code
+      }
+    val classDoc = """/** this is an enum Suit
+                     |  * @param DIAMOND
```
i10416, line 137: Random thought: would it be better to generate `@param <variant>` even when there's no doc for variant?

i10416, line 137: Random thought: would it be better to generate `@param <variant>` even when there's no doc for variant?

modules/codegen/test/src/smithy4s/codegen/internals/RendererSpec.scala
```diff
@@ -85,6 +85,144 @@ final class RendererSpec extends munit.FunSuite {
...
+                     |  *   this is a DIAMOND
+                     |  * @param HAERT
+                     |  *   this is a HAERT
+                     |  * @param HEART
+                     |  *   this is a HEART
+                     |  */""".stripMargin
+    val diamondWithDocRendered =
```
i10416, line 144: confirm variants holds hints from `@enum` trait

i10416:
Hmm, I missed some corner cases.

modules/codegen/test/src/smithy4s/codegen/internals/RendererSpec.scala
i10416: Hmm, as mentioned https://github.com/disneystreaming/smithy4s/issues/773, `@param` sounds strange for enum variants.

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -311,31 +311,55 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+                      shape
+                        .getAllTraits()
+                        .values()
+                        .asScala
+                        .filterNot(
+                          _.toShapeId() == ShapeId.from("smithy.api#enum")
+                        )
+                        .asJavaCollection
+                    )
+                    .asInstanceOf[EnumShape.Builder]
```
kubukoz, line 360: **suggestion:** can you use an upcast instead? (`EnumShape.builder().(...): EnumShape.Builder`)

modules/example/src/smithy4s/example/LowHigh.scala
```diff
@@ -20,12 +25,12 @@ sealed abstract class LowHigh(_value: String, _name: String, _intValue: Int, _hi
 object LowHigh extends Enumeration[LowHigh] with ShapeTag.Companion[LowHigh] {
   val id: ShapeId = ShapeId("smithy4s.example", "LowHigh")
 
-  val hints: Hints = Hints(
-    smithy.api.Enum(List(smithy.api.EnumDefinition(value = smithy.api.NonEmptyString("Low"), name = Some(smithy.api.EnumConstantBodyName("LOW")), documentation = None, tags = None, deprecated = None), smithy.api.EnumDefinition(value = smithy.api.NonEmptyString("High"), name = Some(smithy.api.EnumConstantBodyName("HIGH")), documentation = None, tags = None, deprecated = None))),
-  )
+  val hints: Hints = Hints.empty
```
kubukoz, line 28: **issue:** this doesn't look right, IMO we should keep the traits originally present in the model, regardless if they're duplicated in the Enumeration schema

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
i10416: I think it gets compile error as `source` and `addTraits` come from `StringShape.Builder`, which is a super class of `EnumShape.Builder`.

modules/example/src/smithy4s/example/LowHigh.scala
i10416: I found `fromStringShape` in software.amazon.smithy.model.shapes.EnumShape.java drops enum trait from `StringShape`. Do we still need to keep it? 


```java
    public static Optional<EnumShape> fromStringShape(StringShape shape, boolean synthesizeNames) {
        if (shape.isEnumShape()) {
            return Optional.of((EnumShape) shape);
        }

        if (!shape.hasTrait(EnumTrait.ID)) {
            return Optional.empty();
        }
        StringShape stringWithoutEnumTrait = shape.toBuilder().removeTrait(EnumTrait.ID).build();
        Builder enumBuilder = EnumShape.builder();
        stringWithoutEnumTrait.updateBuilder(enumBuilder);
        ...
```

i10416: https://github.com/awslabs/smithy/blob/main/smithy-model/src/main/java/software/amazon/smithy/model/shapes/EnumShape.java#L131

kubukoz: I guess there's no "right thing to do", it'll be a judgement call - do we try to do what Smithy does in `upgradeEnums` (which is an optional transformation that uses the code path you mentioned), or do we try to make our representation as close to the original as possible?

I think for support of Smithy 1.x, we usually did the latter.

kubukoz:
Request: please avoid force pushes on PRs that have already been seen :) it's fine to keep "bad" commits on the branches, as we will squash them anyway.

i10416:
So sorry for that 🙇

modules/example/src/smithy4s/example/LowHigh.scala
i10416: Okey, I will keep enum trait information in generated hints.

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -311,31 +311,66 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+                throw new Exception(
+                  "smithy.api#enum shape is missing for enum trait"
                 )
+              )
+            val pseudoEnumShape =
+              EnumShape.fromStringShape(shape, true).asScala match {
+                case Some(shape) =>
+                  shape
+                    .toBuilder()
+                    .addTrait(enumTrait)
```
i10416: restore `EnumTrait` dropped by `fromStringShape`

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -311,31 +311,66 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+                              .build()
+                        }
+                    }
+                    val builder = e.toBuilder().clearEnums()
+                    defs.foreach(builder.addEnum)
+                    builder.build()
+                  }
+                  EnumShape
+                    .builder()
+                    .addTrait(enumTrait)
```
i10416: restore `EnumTrait` dropped by `fromStringShape`

modules/codegen/test/src/smithy4s/codegen/internals/RendererSpec.scala
Baccata: Good point, maybe bullet points might be better in this case.

modules/example/src/smithy4s/example/LowHigh.scala
Baccata: A little late to the party but I think it's fine to omit this trait. All semantic information is recoverable from the Schema, and this trait just bloats the generated code (and therefore bytecode) for no reason. 

If there was a reasonably good reason for advocating for its presence, I'd be inclined to keep it, but in the absence of such an argument, let's remove the generated `enum` hint. 

PS : sorry for the churn.

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
Baccata: See my comment here : https://github.com/disneystreaming/smithy4s/pull/980#discussion_r1214054302

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
Baccata: See my comment here : https://github.com/disneystreaming/smithy4s/pull/980#discussion_r1214054302

modules/codegen/test/src/smithy4s/codegen/internals/RendererSpec.scala
```diff
@@ -85,6 +85,192 @@ final class RendererSpec extends munit.FunSuite {
...
+    assert(
+      definition.contains(diamondWithDocRendered),
+      "DIAMOND variant must hold hints from `@enum` trait"
+    )
+    assert(
+      haertWithDeprecationAndDocRendered.forall(definition.contains),
+      "variants generated from `@enum` trait must hold hints, but deprecation hint cannot support message and since properties."
+    )
+    assert(
+      !definition.contains("smithy.api.Enum"),
```
i10416, line 166: updated: generated shape doesn't have redundant metadata.

modules/codegen/test/src/smithy4s/codegen/internals/RendererSpec.scala
```diff
@@ -85,6 +85,192 @@ final class RendererSpec extends munit.FunSuite {
...
+                     |  *   this is a DIAMOND
+                     |  * @param HAERT
+                     |  *   this is a HAERT
+                     |  * @param HEART
+                     |  *   this is a HEART
+                     |  */""".stripMargin
+    val diamondWithDocRendered =
+      """case object DIAMOND extends Suit("DIAMOND", "DIAMOND", 0, Hints(smithy.api.Documentation("this is a DIAMOND")))"""
+    // NOTE: enum trait and enum shape are not 100% compatible. For example, enum trait doesn't support deprecated$since and deprecated$message.
+    val haertWithDeprecationAndDocRendered = List(
```
i10416, line 147: Here, I use List instead of single String because multiline assertion somehow failed on CI but worked on my local machine.

Baccata (approved):

Baccata:
Thank you !

PR #981, opened 2023-05-29
improve: fallback unspecified members of deprecated trait to N/A

### LINKED ISSUE #948: `@deprecated` w/o argument is deprecated
We generate `@deprecated` annotation code from shapes that w/ the `smithy.api#deprecated` trait. `smithy.api#deprecated` is defined as:

```
/// Marks a shape or member as deprecated.
@trait
structure deprecated {
    /// The reason for deprecation.
    message: String,

    /// A description of when the shape was deprecated (e.g., a date or version).
    since: String,
}
```

So there are fields that we can use to render a proper `@deprecated` annotation in Scala code. I suggested that we get the `message` and `since` from the hint and fallback to `N/A` if empty.

i10416:
Scala compiler raises warning for deprecated annotation without message and/or since fields.

close https://github.com/disneystreaming/smithy4s/issues/948

Perhaps, should we warn or inform users to encourage them to use deprecated trait with message and since fields via smithy language server?

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
@@ -183,12 +183,13 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     hints
       .collectFirst { case h: Hint.Deprecated => h }
       .foldMap { dep =>
-        val messagePart = dep.message
-          .map(msg => line"message = ${renderStringLiteral(msg)}")
-        val versionPart =
-          dep.since.map(v => line"since = ${renderStringLiteral(v)}")
+        val messagePart =
+          line"message = ${renderStringLiteral(dep.message.getOrElse("N/A"))}"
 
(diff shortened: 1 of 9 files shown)


daddykotex:
Amazing, thanks for the PR

The CI is failing because part of the codebase (`examples/`) is code generated. This is because we like to have a visual on how code changes to the renderer affects the generated code. if you run `sbt test_3_0_jvm`, you'll get changes in the folder. You need to commit those.

I'll have a look but I think including a Smithy validator inside of `alloy-core` would give us the warning you're looking for.

daddykotex (approved):
Thanks again for the PR, that's amazing

kubukoz:
> Perhaps, should we warn or inform users to encourage them to use deprecated trait with message and since fields via smithy language server?

IMO that would live in the validators for the Smithy standard library, so not even a LS concern. I think we can have that discussion with the awslabs/smithy folks.

Baccata (approved):
Thanks for this 👍

PR #986, opened 2023-05-29
Header test multiple fixes

yisraelU:
This PR addresses what I believe were inaccuracies in the parsing of multiple headers from a string that is encoded as a smithy string.
The tests too were inaccurate.
This introduces a different method plus changes the tests , in addition to applying the parsing to both client and server.

Therefore the tests will fail as a few additional compliance tests are passing and we need to update the allow list

There is one additional problem that arises and that is the parsing of a date which allows for unescaped commas.
So this change now causes to tests with dates in the header to fail

Other than , running a date parser first with a fallback to the multi-value header parser , I am not sure how to solve this problem in a generic fashion

### DIFF
--- a/modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
+++ b/modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
@@ -24,7 +24,7 @@ import java.nio.charset.StandardCharsets
 import scala.collection.immutable.ListMap
 
 package object internals {
-  private def splitQuery(queryString: String): (String, String) = {
+  def splitQuery(queryString: String): (String, String) = {
     queryString.split("=", 2) match {
       case Array(k, v) =>
         (
@@ -43,8 +43,8 @@ package object internals {
   }
 
   private[compliancetests] def parseQueryParams(
-      queryParams: Option[List[String]]
-  ): ListMap[String, List[String]] = {
+                                                 queryParams: Option[List[String]]
+                                               ): ListMap[String, List[String]] = {
     queryParams.combineAll
       .map(splitQuery)
       .foldLeft[ListMap[String, List[String]]](ListMap.empty) {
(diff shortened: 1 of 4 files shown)


modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
```diff
@@ -69,33 +72,68 @@ package object internals {
...
   /*
-       This function takes a string and splits it on a comma delimiter and prunes extra whitespace which
+       This function takes a string that is encoded like source code and splits it on a comma delimiter and prunes extra whitespace which
        what makes it a bit more complicated is we need to keep track of if we are in an open quote or not
    */
-  private[compliancetests] def parseList(s: String): List[String] = {
-    s.foldLeft((Chain.empty[String], 0, 0, false)) {
-      case ((acc, begin, end, quote), elem) =>
-        elem match {
-          // we are in a quote so we negate the quote state and move on
```
Baccata, line 79: I'd like to see comments like in the previous version

Baccata:
> Other than , running a date parser first with a fallback to the multi-value header parser , I am not sure how to solve this problem in a generic fashion

Isn't solving that gonna be driven by the information as to whether the current char is in an open-quote or not ?

modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
yisraelU: Sure

yisraelU:
> > Other than , running a date parser first with a fallback to the multi-value header parser , I am not sure how to solve this problem in a generic fashion
> 
> Isn't solving that gonna be driven by the information as to whether the current char is in an open-quote or not ?

No , there are no quotes included in the actual payload . Here is an example of each payload
multi-value header payload -  a,b,c
date header payload - Mon, 16 Dec 2019 23:48:18 GMT
See - https://github.com/awslabs/smithy/blob/c619a9a4a16f8cffbf4adcc06d32ecbe73ade669/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy#LL110C14-L110C56

Baccata:
let's see what AWS says : https://github.com/awslabs/smithy/pull/1798

Baccata:
I think we're not gonna have a choice but to work around this. I believe that the correct course of action will be to turn our `Map[String, List[String]]` into a `Map[String, String]` and compare that with what AWS gives us, following the non-sensical format. 

When collapsing a `List[String]` into a `String`, the algo should be something like : 

1. if a `String` in the `List` happens to be a valid http-date timestamp, then concatenate as is
2. else if `String` has commas in it, escape the quotes inside the String and add quotes before concatenating 
3. else escape quotes inside the String concatenate it 

I don't like it but I'm not seeing any other solution. It's a "garbage in, garbage out" kind of situation.

Baccata:
Closing this as the following PR is implementing the serialisation behaviour expected by AWS : 

https://github.com/disneystreaming/smithy4s/pull/1035

PR #988, opened 2023-05-30
fix: avoid baking error metadata into status code

i10416:
This commit updates MessageEncoder to avoid copying status code from metadata into response status when status code is not successful value. This change allows service to keep response status set by the endpoint.

See https://github.com/disneystreaming/smithy4s/issues/916#issuecomment-1516512415 for reference.

### DIFF
--- a/modules/http4s-kernel/src/smithy4s/http4s/kernel/MessageEncoder.scala
+++ b/modules/http4s-kernel/src/smithy4s/http4s/kernel/MessageEncoder.scala
@@ -73,6 +73,11 @@ object MessageEncoder {
       val headers = toHeaders(metadata.headers)
       val status = metadata.statusCode
         .flatMap(Status.fromInt(_).toOption)
+        .filter(_.responseClass match {
+          case Status.Successful | Status.Redirection =>
+            true
+          case _ => false
+        })
         .getOrElse(response.status)
       response.withHeaders(response.headers ++ headers).withStatus(status)
     }
--- a/modules/http4s-kernel/src/smithy4s/http4s/kernel/MessageEncoder.scala
+++ b/modules/http4s-kernel/src/smithy4s/http4s/kernel/MessageEncoder.scala
@@ -74,6 +74,8 @@ object MessageEncoder {
       val status = metadata.statusCode
         .flatMap(Status.fromInt(_).toOption)
         .filter(_.responseClass match {
+          // This allows only Successful(2xx) and Redirection(3xx) values for response status code.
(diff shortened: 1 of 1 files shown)


modules/http4s-kernel/src/smithy4s/http4s/kernel/MessageEncoder.scala
```diff
@@ -73,6 +73,11 @@ object MessageEncoder {
       val headers = toHeaders(metadata.headers)
       val status = metadata.statusCode
         .flatMap(Status.fromInt(_).toOption)
+        .filter(_.responseClass match {
```
daddykotex: There is `isSuccess` on status: https://www.javadoc.io/doc/org.http4s/http4s-docs_2.13/latest/org/http4s/Status.html#isSuccess:Boolean

which is implemented in this way:
```
case object Informational extends ResponseClass { val isSuccess = true }
  case object Successful extends ResponseClass { val isSuccess = true }
  case object Redirection extends ResponseClass { val isSuccess = true }
  case object ClientError extends ResponseClass { val isSuccess = false }
  case object ServerError extends ResponseClass { val isSuccess = false }
  ```
  
  I don't know if you meant to exclude `Informational` or if you did not know about `isSuccess`. But if it's the former, I'd add a comment
  
  if it's the latter, I'd use `isSuccess`

daddykotex:
In any case, thanks again for the PRs

modules/http4s-kernel/src/smithy4s/http4s/kernel/MessageEncoder.scala
i10416: I intended to exclude `Informational` based on the comment https://github.com/disneystreaming/smithy4s/issues/916#issuecomment-1516512415

kubukoz: TBH I wasn't really thinking about 1xx at the time the discussion took place, I feel like it should also be supported here.

Maybe there's a compliance test we can use to decide?

Baccata: I agree with @daddykotex in that we should be using `isSuccess`

i10416: Thanks for feedbacks. I replaced it with `isSuccess`.

daddykotex: thanks to you, sorry for the back and forth

daddykotex (approved):

Baccata (approved):

PR #993, opened 2023-05-31
Add support for the notion of sparse collection

### LINKED ISSUE #990: [RestJson1] - Sparse trait implementation
[Docs](https://smithy.io/2.0/spec/type-refinement-traits.html#smithy-api-sparse-trait)
The sparse trait indicates that a list or map may have a null as a value.

Given that our target language is Scala , I am not sure how to approach implementation.
I would have liked to simply drop any nulls when deserializing a list, however, this makes the codec non-round-trippable.
Also, this would raise a question should the implementation for Map be different as a key cannot be null and therefore a key mapped to null can be used to signal some information 

Another option would be to have the hint affect the codegen (like the meta traits we have) and produce a List[Option[A]]

Baccata:
Closes https://github.com/disneystreaming/smithy4s/issues/990

This adds support for sparse collections by : 

1. modifying the `Schema` GADT to have a new concept, `sparse` that turns a `Schema[A]` into a `Schema[Option[A]]`. I may rename it, but this concept is somewhat orthogonal to the notion of optional field, and didn't want to name it in such a way that would mislead implementors of `SchemaVisitor`. I reached the conclusion that a new member needed to be added after experimenting with changing the Collection and Map members of the `Schema` GADT, which was making the implementation of Visitors more confusing than with this current version. 
2. Render sparse collections to make use of `SparseSchema` to wrap their members. 
3. Compile sparse collections to the correct schema in the dynamic module. 

This PR needs to be accompanied with a change to Alloy to tweak the list of tests, as new tests that were failing are now passing.

### DIFF
--- a/modules/cats/src/smithy4s/interopcats/SchemaVisitorShow.scala
+++ b/modules/cats/src/smithy4s/interopcats/SchemaVisitorShow.scala
@@ -172,7 +172,7 @@ final class SchemaVisitorShow(
   override def sparse[A](schema: Schema[A]): Show[Option[A]] = {
     val showA = self(schema)
     locally {
-      case None        => "null"
+      case None        => "N/A"
       case Some(value) => showA.show(value)
     }
   }
--- a/modules/cats/src/smithy4s/interopcats/SchemaVisitorShow.scala
+++ b/modules/cats/src/smithy4s/interopcats/SchemaVisitorShow.scala
@@ -169,7 +169,7 @@ final class SchemaVisitorShow(
     a => ss.value.show(a)
   }
 
-  override def sparse[A](schema: Schema[A]): Show[Option[A]] = {
+  override def nullable[A](schema: Schema[A]): Show[Option[A]] = {
     val showA = self(schema)
     locally {
       case None        => "N/A"
--- a/modules/cats/src/smithy4s/interopcats/SchemaVisitorShow.scala
+++ b/modules/cats/src/smithy4s/interopcats/SchemaVisitorShow.scala
@@ -169,4 +169,12 @@ final class SchemaVisitorShow(
     a => ss.value.show(a)
   }
 
+  override def sparse[A](schema: Schema[A]): Show[Option[A]] = {
+    val showA = self(schema)
+    locally {
+      case None        => "null"
+      case Some(value) => showA.show(value)
+    }
+  }
+
 }
--- a/modules/cats/src/smithy4s/interopcats/SchemaVisitorShow.scala
+++ b/modules/cats/src/smithy4s/interopcats/SchemaVisitorShow.scala
@@ -172,8 +172,8 @@ final class SchemaVisitorShow(
   override def nullable[A](schema: Schema[A]): Show[Option[A]] = {
     val showA = self(schema)
     locally {
-      case None        => "N/A"
-      case Some(value) => showA.show(value)
+      case None        => "None"
+      case Some(value) => s"Some(${showA.show(value)})"
     }
   }
 
--- a/modules/core/src/smithy4s/schema/Schema.scala
+++ b/modules/core/src/smithy4s/schema/Schema.scala
@@ -83,6 +83,7 @@ sealed trait Schema[A]{
 
   final def biject[B](bijection: Bijection[A, B]) : Schema[B] = Schema.bijection(this, bijection)
   final def biject[B](to: A => B, from: B => A) : Schema[B] = Schema.bijection(this, to, from)
+  final def sparse: Schema[Option[A]] = Schema.sparse(this)
 
   final def getDefault: Option[Document] =
     this.hints.get(smithy.api.Default).map(_.value)
--- a/modules/core/src/smithy4s/schema/Schema.scala
+++ b/modules/core/src/smithy4s/schema/Schema.scala
@@ -43,7 +43,7 @@ sealed trait Schema[A]{
     case BijectionSchema(schema, bijection) => BijectionSchema(schema.withId(newId), bijection)
     case RefinementSchema(schema, refinement) => RefinementSchema(schema.withId(newId), refinement)
     case LazySchema(suspend) => LazySchema(suspend.map(_.withId(newId)))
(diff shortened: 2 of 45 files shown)


modules/cats/src/smithy4s/interopcats/SchemaVisitorShow.scala
```diff
@@ -169,4 +169,12 @@ final class SchemaVisitorShow(
     a => ss.value.show(a)
   }
 
+  override def sparse[A](schema: Schema[A]): Show[Option[A]] = {
+    val showA = self(schema)
+    locally {
+      case None        => "null"
```
kubukoz: **nitpick:** I'd go for `None` and `Some` in the other variant, after all we probably want to use Scala-user-friendly output, no need to scare folks with nulls.

modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
```diff
@@ -331,8 +331,13 @@ private[dynamic] object Compiler {
       }
     }
 
-    override def listShape(id: ShapeId, shape: ListShape): Unit =
-      update(id, shape.traits, schema(shape.member.target).map(s => list(s)))
+    override def listShape(id: ShapeId, shape: ListShape): Unit = {
+      if (shape.traits.contains(IdRef("smithy.api#sparse"))) {
```
kubukoz, line 335: **question:** what about maps?

kubukoz (changes requested):
looks like some test changes are still needed to compile / pass tests, but generally I'm happy with how this looks!

modules/core/src/smithy4s/schema/Schema.scala
```diff
@@ -69,6 +71,7 @@ sealed trait Schema[A]{
     case BijectionSchema(schema, bijection) => BijectionSchema(schema.transformHintsTransitively(f), bijection)
     case RefinementSchema(schema, refinement) => RefinementSchema(schema.transformHintsTransitively(f), refinement)
     case LazySchema(suspend) => LazySchema(suspend.map(_.transformHintsTransitively(f)))
+    case s: SparseSchema[a] => SparseSchema(s.underlying.transformHintsTransitively(f)).asInstanceOf[Schema[A]]
```
daddykotex: I'm unfamiliar with this syntax, is the lowercased `a` in the bracket used for something? is it a type parameter?

Baccata: >  is the lowercased a in the bracket used for something? is it a type parameter? 

It's used to bind the type parameter to a type-variable so that you can re-use it elsewhere. Often when doing pattern matching against GADTs, several members of a case class will depend on single type, and you want to re-use this type to ascribe stuff to the right-hand-side of the arrow.  

If you do 

```
case s : SparseSchema[A] 
```

then the compiler will complain because `A` is already used and it cannot guarantee that the `A` in your case is the same as the `A` of the expression you're matching. 

If you use existentials (which, in all frankness, I could do here), then it'll be hard to reconcile expressions on the RHS.

modules/json/src/smithy4s/http/json/SchemaVisitorJCodec.scala
```diff
@@ -1080,6 +1080,26 @@ private[smithy4s] class SchemaVisitorJCodec(
       out.writeKey(total(x).intValue)
   }
 
+  override def sparse[A](schema: Schema[A]): JCodec[Option[A]] =
+    new JCodec[Option[A]] {
+      val underlying: JCodec[A] = self(schema)
+      def expecting: String = s"JsNull or ${underlying.expecting}"
+      def decodeKey(in: JsonReader): Option[A] = ???
```
daddykotex, line 1087: should we throw a more meaningful exception for those two

modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
Baccata: done

modules/json/src/smithy4s/http/json/SchemaVisitorJCodec.scala
Baccata: done

modules/cats/src/smithy4s/interopcats/SchemaVisitorShow.scala
Baccata: done

Baccata:
@kubukoz, renamed the new schema ADT member to `nullable` as `sparse` didn't really make sense (like, the collections are sparse, a value can't be sparse).

modules/core/src/smithy4s/schema/Schema.scala
kubukoz: IIRC this feature is severely half-baked in Scala 2, it can even succeed to match bogus values and result in `ClassCastException` 😓 

but if it compiles in 3.x, it should be fine on both versions

modules/cats/src/smithy4s/interopcats/SchemaVisitorShow.scala
```diff
@@ -169,4 +169,12 @@ final class SchemaVisitorShow(
     a => ss.value.show(a)
   }
 
+  override def nullable[A](schema: Schema[A]): Show[Option[A]] = {
+    val showA = self(schema)
+    locally {
+      case None        => "N/A"
+      case Some(value) => showA.show(value)
+    }
```
kubukoz: **issue:** still got a problem with this. If you're showing a String with the contents `N/A`, you won't be able to tell whether it's present or null.

I would again suggest wrapping the underlying string in `Some`.

kubukoz (approved):
LGTM but I do have a concern about the Show

modules/cats/src/smithy4s/interopcats/SchemaVisitorShow.scala
Baccata: _sigh_ fine !

Issue #991, opened 2023-05-31
How to return an empty Integer, Double, etc... as a null in json response


colomain:
Hello,
I have a business requirement that forces me to return values of type Double, Integer, etc... as null when they are empty, instead of not appearing in the response

Example:

instead of

```json
{
    "data": [
        {
            "id": 1
        }
    ]
}
```

I want to achieve:

```json
{
    "data": [
        {
            "id": 1,
            "nullableField": null
        }
    ]
}
```

How can I achieve this without having to specify `nullableField` as a `Document`?

I know how to do it with `Document` (and convert Json value to Document...), but this way, swagger specification got misleading, and `nullableField` appears as an object (without type).

Thank you!

kubukoz:
I don't think there's a baked in way to do this with `simpleRestJson`. I think you could create a new protocol, or at least a new protocol builder for `simpleRestJson`, which would use a custom `JsonCodecAPI` implementation that configures the jsoniter `WriterConfig` to write nulls. However, that's going to be a global setting for all Smithy services that use that builder, so you won't have the ability to do this on a case-by-case basis.

lewisjkl:
Hey @colomain, what @kubukoz suggested will definitely work for you, but I was just chatting with @Baccata and we have a way that we can add a configuration to the SimpleRestJson builder that would indicate that all optional fields with a `None` value should be encoded as `null` rather than omitting the field. I should be able to have a PR up later today or tomorrow that addresses this.

colomain:
Awesome!! Thx @lewisjkl !

daddykotex:
See the PR attached @colomain 

You should get support for that in the next 0.17 release.

daddykotex:
This was release as part of [0.17.8](https://github.com/disneystreaming/smithy4s/releases/tag/v0.17.8)

PR #995, opened 2023-05-31
Explicit null for series/0.17

lewisjkl:
Adds the ability to explicitly encode null values for optional fields that are set to `None`. See #991 for more context. (PR for series/0.18 incoming)

### DIFF
--- a/modules/json/src/smithy4s/http/json/codecs.scala
+++ b/modules/json/src/smithy4s/http/json/codecs.scala
@@ -32,7 +32,12 @@ final class JsonCodecs(
     maxArity: Int = codecs.defaultMaxArity,
     explicitNullEncoding: Boolean = false
 ) extends JsonCodecAPI(
-      codecs.schemaVisitorJCodec(_, maxArity, explicitNullEncoding),
+      cache =>
+        new SchemaVisitorJCodec(
+          maxArity = maxArity,
+          explicitNullEncoding = explicitNullEncoding,
+          cache = cache
+        ),
       Some(hintMask)
     )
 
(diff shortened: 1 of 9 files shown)


modules/json/test/src/smithy4s/http/json/JsonCodecApiTests.scala
```diff
@@ -53,12 +53,30 @@ class JsonCodecApiTests extends FunSuite {
...
+        Schema.string
+          .optional[Option[String]]("a", identity)
+      )(identity)
+
+    val capi = new JsonCodecs(HintMask.empty, explicitNullEncoding = true)
+
+    val codec = capi.compileCodec(schemaWithJsonName)
+    val encodedString = new String(capi.writeToArray(codec, null))
+
+    assertEquals(encodedString, """{"a":null}""")
```
kubukoz, line 80: nitpick / suggestion: should we have an extra assertion making sure that you can still decode this, with either `explicitNullEncoding=true` or `explicitNullEncoding=false`?

modules/json/src/smithy4s/http/json/codecs.scala
```diff
@@ -27,6 +27,15 @@ import smithy4s.internals.InputOutput
 import smithy4s.schema.CompilationCache
 import smithy4s.schema.SchemaVisitor
 
+final class JsonCodecs(
+    hintMask: HintMask = codecs.defaultHintMask,
+    maxArity: Int = codecs.defaultMaxArity,
+    explicitNullEncoding: Boolean = false
+) extends JsonCodecAPI(
+      codecs.schemaVisitorJCodec(_, maxArity, explicitNullEncoding),
```
kubukoz: suggestion: named parameters in `schemaVisitorJCodec`

Baccata (approved):
Approved but @kubukoz's comments are valid.

PR #996, opened 2023-05-31
add option for explicit null encoding 0.18

lewisjkl:
Repeat of #995 for series/0.18.


project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=1.8.2
```
lewisjkl: Not really sure how this got added into the diff...

Baccata (approved):

PR #1001, opened 2023-06-01
[Json] Handle empty body

### LINKED ISSUE #973: [RestJson1] - Optional Payload
In the case ,that an optional payload is not present, an implementation should prepare for either an empty json object or  the lack of an HTTP body completely
See example test case [here](https://github.com/awslabs/smithy/blob/main/smithy-aws-protocol-tests/model/restJson1/empty-input-output.smithy)

yisraelU:
JsonCodecApi defaults to empty json object when decoding when the http body is empty
closes #973


yisraelU:
Current Failure is due to expected failure now passing..
Need to add RestJsonIgnoreQueryParamsInResponseNoPayload to Alloy

Baccata:
I'm in the process of changing the configuration format, I'll take care of it

PR #1002, opened 2023-06-01
Add Monoid for http4s middleware

### LINKED ISSUE #974: http4s middleware: provide semigroups
`ServerEndpointMiddleware` and `ClientEndpointMiddleware` are glorified function wrappers, and currently the client/server builders only allow passing one of these.

In order to pass many, users have to combine them on their own, and I think we can make it simpler by providing semigroups out of the box (the types live in the http4s module which already has a cats dependency).

`Semigroup[ServerEndpointMiddleware]` would be equivalent to `(a, b) => a.andThen(b)`, same for client.

zetashift:
Resolves https://github.com/disneystreaming/smithy4s/issues/974

### DIFF
--- a/modules/http4s/src/smithy4s/http4s/ClientEndpointMiddleware.scala
+++ b/modules/http4s/src/smithy4s/http4s/ClientEndpointMiddleware.scala
@@ -19,7 +19,6 @@ package http4s
 
 import cats.Monoid
 import org.http4s.client.Client
-import org.http4s.HttpApp
 
 // format: off
 trait ClientEndpointMiddleware[F[_]] {
--- a/modules/http4s/src/smithy4s/http4s/ClientEndpointMiddleware.scala
+++ b/modules/http4s/src/smithy4s/http4s/ClientEndpointMiddleware.scala
@@ -48,4 +48,18 @@ object ClientEndpointMiddleware {
       ): Client[F] => Client[F] = identity
     }
 
+  implicit def semigroupClientEndpointMiddleware[F[_]]
+      : Semigroup[ClientEndpointMiddleware[F]] =
+    new Semigroup[ClientEndpointMiddleware[F]] {
+      def combine(
+          a: ClientEndpointMiddleware[F],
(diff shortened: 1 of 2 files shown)


kubukoz:
Looks like missing imports 😅

modules/http4s/src/smithy4s/http4s/ClientEndpointMiddleware.scala
```diff
@@ -48,4 +48,18 @@ object ClientEndpointMiddleware {
       ): Client[F] => Client[F] = identity
     }
 
+  implicit def semigroupClientEndpointMiddleware[F[_]]
```
kubukoz: **suggestion:** now that I think of it, we can do a full `Monoid`. `empty` would return a `ClientEndpointMiddleware` where `prepare` returns an identity function.

Same for servers.

zetashift: I'll implement the Monoid! Thank you

And sorry about the imports, I was a bit too eager and didn't get any diagnostics so I was like letsgoooo

zetashift: Does the `empty` look something like this?
```scala
def empty: ClientEndpointMiddleware[F] =
  new ClientEndpointMiddleware[F] {
    def prepare[Alg[_[_, _, _, _, _]]](service: Service[Alg])(
        endpoint: Endpoint[service.Operation, _, _, _, _, _]
    ): HttpApp[F] => HttpApp[F] =
      identity
  }

```

kubukoz: >  And sorry about the imports, I was a bit too eager and didn't get any diagnostics so I was like letsgoooo

me_irl

yes, empty looks about right! I think it can be a `val`

kubukoz (approved):

kubukoz (approved):

kubukoz:
PR so good I had to approve twice

PR #1003, opened 2023-06-02
misc: update endpoint-middleware.md

i10416:
use `putHeaders` in example because `withHeaders` replaces whole headers


Baccata (approved):

Baccata:
Thanks !

PR #1008, opened 2023-06-05
ECS provider fix

msosnicki:
In ECS, to obtain credentials the following has to be used:

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html

Applying the fix on the `0.18`, as it requires this change https://github.com/disneystreaming/smithy4s/commit/7c468ad99c6dff14d5b0c89c66893b63204a61c4


Baccata (approved):

daddykotex (approved):
Thanks

Issue #888, opened 2023-04-06
Parameterless algebras


kubukoz:
In 0.17, the generated code for services outputs the following kind of shape:

```scala
trait WeatherService[F[_, _, _, _, _]] {
  def getWeather(city: String): F[GetWeatherInput, GetWeatherError, GetWeatherOutput, Nothing, Nothing]
  def postWeather(city: String, newWeather: Degrees): F[PostWeatherInput, Nothing, PostWeatherOutput, Nothing, Nothing]
}

// in another file
case class GetWeatherInput(city: String)
case class PostWeatherInput(city: String, newWeather: Degrees)
```

Notably:

1. The input type is repeated in the method parameters (in an unpacked fashion) and in the `F`'s first type parameter
2. There is no way to call an operation without having **exactly one** instance of the input type, even if the effect type is something like `F[_, _, _, _, _] = Unit`.

The second point, importantly, prevents certain patterns that would be possible if the input was only represented in the effect type. However, the input parameters need to appear in the method signature as parameters, otherwise we lose the "scala idiomatic" aspect of the generated code, and it becomes more difficult to call service methods in the established style (plain methods).

The patterns I'm talking about involve, e.g.:

### a builder pattern

...based on implementing a subset of methods in a way that allows circumventing serialization on the ones not implemented, e.g. 

```scala
endpointBuilder(WeatherService)
  .getWeather { input => IO.println(input) }
  //.postWeather - this method is not implemented, so it won't be handled by the implementation
  .build
```

### non-monadic outputs

One usecase that was mentioned is a load testing setup: the ability to describe a scenario of load testing using something like Gatling means that operation inputs aren't just values, but e.g. a [feeder](https://gatling.io/docs/gatling/reference/current/core/session/feeder/).

```scala
gatlingBuilder(WeatherService)
  .withEndpoint(_.getWeather(session => session("city")))
```

---

The idea is that we should keep support for existing, monadic style of services, but also generate some sort of non-monadic mirror that would allow these new usecases.

Proposed shapes, in need of some bikeshedding:

```scala
trait Pipe[P[_[_, _, _, _, _]]] extends FunctorK5[P] {
  type Alg[_[_, _, _, _, _]]
  val service: Service[Alg]
  def endpoints: P[service.Endpoint]

  def toPolyFunction[P2[_, _, _, _, _]](algebra: P[P2]): PolyFunction5[service.Endpoint, P2]

  def toMonadic[F[_, _, _, _, _]](
[...]

PR #1011, opened 2023-06-07
Service Products

miguel-vila:
addresses #888


modules/core/src/smithy4s/Service.scala
```diff
@@ -34,7 +34,7 @@ import kinds._
   *   around makes it drastically easier to implement logic generically, without involving
   *   metaprogramming.
   */
-trait Service[Alg[_[_, _, _, _, _]]] extends FunctorK5[Alg] with HasId {
+trait Service[Alg[_[_, _, _, _, _]]] extends FunctorK5[Alg] with HasId with Mirror[Alg] {
```
Baccata: I'm not sure this coupling is desirable at this point, so let's not bake it in (the companion object of generated types can however implement both interfaces)

miguel-vila: hmm, that might solve our next question 😄 , we were not sure about how to implement this for the compliance test service that's generated with a single endpoint. I removed this `with Mirror` at e2d12aa1983e1103b64f8096bc240529998357b7 , but have kept the changes in the dynamic module, let us know if those make sense

kubukoz: @Baccata I remember this being useful for being able to find the "static" variant after the user gives you the "normal" one (e.g. `MyInterpreter(MyService)` having easy access to `MyServiceStatic`).

Baccata: @kubukoz can you formulate a non-theoretical usecase that'd be more convincing than that vague statement of "I remember this being useful" 😄  ?

modules/core/src/smithy4s/StaticService.scala
```diff
@@ -0,0 +1,37 @@
+package smithy4s
+
+import smithy4s.kinds.FunctorK5
+import smithy4s.kinds.PolyFunction5
+
+trait StaticService[P[_[_, _, _, _, _]]] extends FunctorK5[P] {
```
Baccata: After some thinking `Static` is a confusing denomination for this. I think we should call it `ServiceProduct` because it's essentially abstracting over services that are represent as a product of functions. That'd also give some sort of meaning to the `P` parameter (which might be better named `Prod`) 

```suggestion
trait ServiceProduct[Prod[_[_, _, _, _, _]]] extends FunctorK5[Prod] {
```

@kubukoz , thoughts ?

Baccata: Also, some scaladoc would be appreciated

modules/core/src/smithy4s/StaticService.scala
```diff
@@ -0,0 +1,37 @@
...
+  type Alg[_[_, _, _, _, _]]
+  val service: Service[Alg]
+  def endpoints: P[service.Endpoint]
```
Baccata, line 9: let's call that `endpointsProduct`

Baccata:
```suggestion
  def endpointsProduct: Prod[service.Endpoint]
```

modules/core/src/smithy4s/StaticService.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  def toPolyFunction[P2[_, _, _, _, _]](
+      algebra: P[P2]
+  ): PolyFunction5[service.Endpoint, P2]
```
Baccata:
```suggestion
  def toPolyFunction[Arrow[_, _, _, _, _]](
      algebra: Prod[Arrow]
  ): PolyFunction5[service.Endpoint, Arrow]
```

modules/core/src/smithy4s/StaticService.scala
```diff
@@ -0,0 +1,37 @@
...
+}
+
+object StaticService {
+
+  type Aux[Alg[_[_, _, _, _, _]], P[_[_, _, _, _, _]]] =
+    StaticService[Alg] {
+      type Alg[Op[_, _, _, _, _]] = P[Op]
+    }
```
Baccata: This needs to align with the rest : 

```suggestion
  type Aux[Prod[_[_, _, _, _, _]], Algebra[_[_, _, _, _, _]]] =
    ServiceProduct[Prod] {
      type Alg[F[_, _, _, _, _]] = Algebra[F]
    }
```

modules/core/src/smithy4s/StaticService.scala
miguel-vila: agree, there's already a use of `Dynamic` in the codebase so using Static here is probably confusing. I see where the `ServiceProduct` is going to, could it be something like `ServiceWithoutInputs` ?

Baccata: > ServiceWithoutInputs 

not really. It's just a different way of passing inputs. It's similar to the difference between methods and functions.

kubukoz: `ServiceProduct` seems valid but I don't like that we're suddenly talking about products/coproducts in the generated code. I'm fine with it and won't oppose this name if you decide to keep it, but I'd also consider `ServiceRecord`.

modules/core/src/smithy4s/StaticService.scala
kubukoz: nitpick: `Arrow` -> `Func`?

modules/core/src/smithy4s/Service.scala
kubukoz: like, either of the examples in the original issue?

```scala
endpointBuilder(WeatherService)
  .getWeather { input => IO.println(input) }
  //.postWeather - this method is not implemented, so it won't be handled by the implementation
  .build
```

```scala
gatlingBuilder(WeatherService)
  .withEndpoint(_.getWeather(session => session("city")))
```

kubukoz: if it can be done with `extends Mirror` then disregard me :P 

but I think it would be best if users of interpreters don't have to do `endpointBuilder(WeatherServiceProduct)`.

Baccata: > but I think it would be best if users of interpreters don't have to do  

I'm not opposed to the UX you're describing, what I'm opposed to constraining the `Service` interface to be aware of `ServiceProduct`.

kubukoz: I see. What other way do you see of making this UX possible, though? Moving this `extends` clause to the generated code?

Baccata: > Moving this extends clause to the generated code? 

yup

modules/core/src/smithy4s/StaticService.scala
Baccata: Okay.

modules/core/src/smithy4s/StaticService.scala
Baccata: Okay for Record

modules/core/src/smithy4s/ServiceProduct.scala
```diff
@@ -0,0 +1,53 @@
...
+
+}
+
+/**
+  * Provides a way to get the product version for a service.
+  * For info about the `Prod` type parameter, see [[ServiceProduct]].
+  * 
+  * @tparam Alg the algebra type parameter.
+  */
+trait Mirror[Alg[_[_, _, _, _, _]]] {
```
Baccata: I'd nest that in the `ServiceProduct` object.

modules/core/src/smithy4s/Service.scala
kubukoz: I can deal with that.

probably no use for `Mirror` in Dynamic world either, so it makes sense.

modules/core/src/smithy4s/ServiceProduct.scala
miguel-vila: done at 9dcfe463f1e1d7edef3e0b37913076fa958fee8a

modules/core/src/smithy4s/StaticService.scala
miguel-vila: think I messed this , fixed at [a2e82b7](https://github.com/disneystreaming/smithy4s/pull/1011/commits/a2e82b706cd31ecbd3ba9304fab6ad2dadb79f46)

modules/core/src/smithy4s/StaticService.scala
miguel-vila: renamed at [545d54e](https://github.com/disneystreaming/smithy4s/pull/1011/commits/545d54eba08b6bc98bce81d151c58b98675a86ca)

modules/core/src/smithy4s/StaticService.scala
miguel-vila: added some scaladocs at d8be76689ac789d83716dd1e18b81a83143af202 let me know if those are accurate 

let me know if you prefer `ServiceRecord` as the PR right now uses `ServiceProduct`

modules/core/src/smithy4s/Service.scala
miguel-vila: at [2cba2be](https://github.com/disneystreaming/smithy4s/pull/1011/commits/2cba2bec54ab74b4d7911aa37b4a7e6428111e09) I moved the extends to generated code. I haven't removed the extends from Dynamic , let me know if there's no use case there either and will remove that

modules/core/src/smithy4s/StaticService.scala
miguel-vila: done as part of c5e96829819f7d69180862005bad057dcde0083b

modules/core/src/smithy4s/StaticService.scala
kubukoz: I'm good with either

modules/core/src/smithy4s/Service.scala
kubukoz: > if there's no use case 

I don't have one _so far_, doesn't mean there won't be any. However, I don't think that information is even accessible to users (`DynamicService` is private, an implementation detail) so I'd say let's drop it.

modules/example/src/smithy4s/example/product/package.scala
```diff
@@ -0,0 +1,8 @@
+package smithy4s.example
+
+package object product {
+  type ObjectService[F[_]] = smithy4s.kinds.FunctorAlgebra[ObjectServiceGen, F]
+  val ObjectService = ObjectServiceGen
+
+
```
kubukoz: thought: if you can remove these extra newlines that'd be pretty cool

kubukoz: but I don't want to extend this PR any more so let's leave it for future generations :P

miguel-vila: this seems to come from past code because it seems other package files have the same issue. BTW I think recent changes have removed/moved the example module so I think these files shouldn't exist, will remove them

miguel-vila: removed at [7cab3a3](https://github.com/disneystreaming/smithy4s/pull/1011/commits/7cab3a3edc6703c16ce69fc840634a4e009f6060)

modules/core/src/smithy4s/Service.scala
miguel-vila: removed the dynamic changes at [95fc8c3](https://github.com/disneystreaming/smithy4s/pull/1011/commits/95fc8c38a17369f9f979a3849578faf2912df212)

kubukoz (approved):
Looks great. Some docs would be helpful, but I think we can deal with that in a follow-up PR.

Baccata (approved):

Baccata:
Docs would indeed be appreciated, but they can come in a later PR :)

PR #1018, opened 2023-06-09
[AWS SDK] RestJson1 feature branch

Baccata:
Implements the aws RestJson1 protocol in the `aws-http4s` branch : 

Impact to the 0.18 series branch : 

* Introduces a `smithy4s.Writer` abstraction that represents the action of writing data to a message. This allows to generalise a fair amount of logic in smithy4s-core that benefits all interpreters that want to abide by the http-bindings trait. 
* Rewrites `RequestEncoder` and `ResponseEncoder` to be type aliases of `smithy4s.Writer`
* Deduplicates from logic from `RequestEncoder` and `ResponseEncoder` 
* Introduces `smithy4s.http4s.kernel.MessageDecoder` which is quite similar to `smithy4s.Writer` but for the decoding type. At the time of writing this, it's hosted in the `http4s-kernel` module because of how smithy4s-core doesn't have an abstraction similar to `cats.Applicative`, which is necessary to do any composition of decoders. We may later add a `zipMap` method to our `Covariant` abstraction, which is somewhat similar to `ApplicativeError` anyway. This would further reduce the burden of third-parties. 
* Rewrites `RequestDecoder` and `ResponseDecoder` to be type aliases of `MessageDecoder`
* Deduplicates some logic from `RequestDecoder` and `ResponseDecoder` 
* Removes Primitive.PUnit altogether. Now the empty `smithy.api.Unit` type is schematised by a simple empty struct. 
* Rewrites Metadata encoding/decoding logic to accommodate the AWS-flavoured header encodings, which concatenates all values (in arguably weird ways) under a single key for each header. 
* Implements most of the necessary logic to comply with edge cases of the AWS RestJson1 protocol 
* In particular, adds support for request compression 
* Adds logic so that `SimpleRestJson` can also work with the AWS-specific error-discriminator header, in order to increase interop between tools even when protocols don't fully match. 
* Wires the compliance tests associated to the restJson1 in the aws-http4s module's test scope. 
* Removes the notion of Metadata `Partial/Total` Encoders/Decoders in favour of a simple `Encoder/Decoder`.


Baccata:
@yisraelU, the initial commit of this feature-branch aims at making sure that : 

#### 1. our interpretation of the compliance tests follows what AWS means.

This implies less extrapolation (in particular around headers). The headers (which we represent as a `MultiMap[String, String]`) are compared by first collapsing them into `Map[String, String]`, with sensible quoting when multiple values are involved. Because AWS does not apply sensible quoting, the AWS-specific interpreters **will be responsible for guaranteeing that there's only one value for each entry in the multimap**. 

This also means that the parsing of the values contained by `Map[String, String]` into a `MultiMap[String, String]` is now the responsibility of the interpreters too, not the one of the tests. To go further, this implies that the SimpleRestJson interpreters are not expected to pass the tests that are related to http-date headers with multiple values, but they are expected to be able to parse multiple values from single raw headers, when those are properly quoted/escaped. I'll be taking care of this in a subsequent PR. 

#### 2. the SimpleRestJson pass the tests against the new assertions. 

In particular, this means that the implementation changed to enforce that the server return an empty json object even when all the fields of the output are bound to metadata. 

#### 3. the AWS-RestJson1-related tests are enabled. 
No logic was changed to specifically fix the tests that were red.

PR #1022, opened 2023-06-12
Update smithy-model to 1.31.0, alloy to 0.2.2

kubukoz:


daddykotex (approved):
Changes are good to me. As far as I can tell, the failures happen in the compliance tests module. I can help if you need

kubukoz:
I'd appreciate that 😅 it might be a straightforward fix but I don't have the full context. Trying to make a bump of all the transitive versions of smithy-model so that there isn't any risk of incompatibilities in my build (it's happened before when modules from 1.27.x and 1.28.x were mixed)

daddykotex:
seems like 1.32.0 is available

kubukoz:
yeah, but we haven't updated the rest of the stack to it yet. We already have some alloy/smithy-translate deps, it's all on 1.31.0, but we're about to add a smithy4s dependency and we don't want to downgrade the others - hence trying to go to 1.31.0 here.

Later on, presumably, we'd do the same for 1.32.0. Now that I think about it, should we tell Scala Steward to target `series/0.17` with its upgrades too?

modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sServerEndpoint.scala
```diff
@@ -196,18 +198,23 @@ private[http4s] class SmithyHttp4sServerEndpointImpl[F[_], Op[_, _, _, _, _], I,
...
+    val baseResponse = putHeaders(Response[F](httpStatus), outputHeaders)
+    if (isEmpty(outputSchema)) baseResponse.withEmptyBody
+    else baseResponse.withEntity(output)
   }
 
-  def compileErrorable(errorable: Errorable[E]): E => Response[F] = {
+  private def isEmpty[A](schema: Schema[A]): Boolean = schema match {
+    case Schema.PrimitiveSchema(_, _, Primitive.PUnit) => true
+    case _                                             => false
+  }
```
Baccata: these changes should move to the `EntityCompiler`

Baccata:
>  should we tell Scala Steward to target series/0.17 with its upgrades too? 

Nope, unless it's limited to bumps of the `smithy-model` library (and other libs from awslabs)

modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sServerEndpoint.scala
daddykotex: did just that

kubukoz:
all good from my side, @daddykotex are you happy to merge?

PR #1019, opened 2023-06-09
Initial gzip encoder/decoder implementation

daddykotex:
Some protocol support `requestCompression` and this is the gzip implementation for this: https://smithy.io/2.0/spec/behavior-traits.html?highlight=compression#requestcompression-trait

Currently, it is not wired to anything, but I'll do that on another PR that's already open for `restJson1`

PR #1023, opened 2023-06-12
GZIP support for aws restJson1

daddykotex:
A few commits, but it's a merge from #1019  and the following:

1. 418986ce5aff131d8cae2e6d26d791f4aac0d4ec: fix to the branch (currently broken on scala3)
2. 7489b2040d5b852b2df9a7974c157d13e123c14f: wire the encoder on the restJson1 client
3. c95dd088cc5fb4224e365e36c7f723126f40803e: a small improvement to the assertion message
4. 99141309b9236d1ec97dd790b3f80e56257b5b2a: a fix for the a compliance test that was failing (let me know what you think of this one) - test here: https://github.com/awslabs/smithy/blob/780b4d9077456b5985db7537fcaebe19b67283f8/smithy-aws-protocol-tests/model/restJson1/requestCompression.smithy#L292


modules/http4s-kernel/src/smithy4s/http4s/kernel/RequestDecoder.scala
```diff
@@ -16,8 +16,133 @@
...
+import smithy4s.PartialData
+import smithy4s.http.HttpRestSchema
+import smithy4s.http.Metadata
+import smithy4s.schema._
 
 trait RequestDecoder[F[_], A] {
   def decodeRequest(request: Request[F]): F[A]
 }
+
+object RequestDecoder {
```
daddykotex, line 28: to @Baccata 

I'm not sure I'll be able to remove the deduplication between this and the `ResponseDecoder`. The duplication makes the code simple, I feel like introducing a way to abstract / compose might get in the way of readability more than anything

happy to be proven wrong though

Baccata: @daddykotex, here's some de-dup for the request/response encoders, that re-uses existing constructs to compose things : 

https://github.com/disneystreaming/smithy4s/pull/1023/commits/0545dab6c2467020deeb2295d17255d6d39e9142

It doesn't reduce the number of LOCs but simplifies things in the sense that it's much more compositional, and leads to much more concise functions. If it were a Scala3-only library, things would be even more concise as the distinction between polymorphic methods and `smithy4s.kinds.PolyFunction` would disappear

Baccata (approved):

PR #1025, opened 2023-06-12
Support for mill 0.11.0

daddykotex:
Should we cross-compile or we go with `0.11.0` along the 0.18 release ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Behavioural changes
 
+### Mill
+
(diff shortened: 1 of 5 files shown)


kubukoz:
given the tests seem to work with no changes, I think cross-building might be desirable - there's not a lot of code changes, and we could keep 0.10 alive until 0.19.

Baccata:
> I think cross-building might be desirable  

I'm not keen on cross-building across mill versions, for the sheer reason of maintenance burden. If people need mill support for 0.10, they will be able to copy/paste the module in userland.

kubukoz:
fair enough

Baccata (approved):

PR #1027, opened 2023-06-13
hints are not passed down to primitive from Collections when reading and writing headers

yisraelU:
Currently, if a header hint targets a list, the hint is not passed down in the SchemaVisitors resulting in losing the HttpBinding context when processing the primitive targeted by the list. 
The case I bumped into was a header targeting a list of timestamps. The timestamp format used was the default one as opposed to the one meant for a header.

Test case : https://github.com/awslabs/smithy/blob/ad6d20b190d739aa9f83c5ef40a1e6c3ffe53ba5/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy#L110

### DIFF
--- a/modules/core/src/smithy4s/http/internals/SchemaVisitorMetadataWriter.scala
+++ b/modules/core/src/smithy4s/http/internals/SchemaVisitorMetadataWriter.scala
@@ -50,7 +50,6 @@ class SchemaVisitorMetadataWriter(
     val cache: CompilationCache[MetaEncode]
 ) extends SchemaVisitor.Cached[MetaEncode] {
   self =>
-
   override def primitive[P](
       shapeId: ShapeId,
       hints: Hints,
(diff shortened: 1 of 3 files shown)


kubukoz:
You can put timestampFormat on lists????? I learn new Smithy things every day

yisraelU:
> You can put timestampFormat on lists????? I learn new Smithy things every day

sorry , I phrased this wrong , will edit

kubukoz:
okay, I see now. Should it forward all hints, though, or just some specific list of those (e.g. the HTTP bindings)?

It'd be good to have an extra test for this that doesn't rely on the compliance tests.

yisraelU:
> okay, I see now. Should it forward all hints, though, or just some specific list of those (e.g. the HTTP bindings)?

Good point , ill refine 
 
> It'd be good to have an extra test for this that doesn't rely on the compliance tests.

Yes , Will add.

modules/core/src/smithy4s/http/internals/SchemaVisitorMetadataWriter.scala
```diff
@@ -68,7 +68,7 @@ class SchemaVisitorMetadataWriter(
       tag: CollectionTag[C],
       member: Schema[A]
   ): MetaEncode[C[A]] = {
-    self(member) match {
+    self(member.addHints(hints)) match {
```
Baccata: You'll likely need to amend `map` as well

Baccata:
Thanks for this. Closing as I needed the fix in : https://github.com/disneystreaming/smithy4s/pull/1035

PR #1029, opened 2023-06-14
Make spec errors observable from the ServiceEndpointMiddleware

msosnicki:
Currently, the spec errors are handled before the `ServiceEndpointMiddleware` has a chance to kick in. This makes some use cases a bit hard to implement (in my case tracing, where we need to attach certain error info from middleware). 

The proposed change, converts errors to responses only once, after the whole error handling pipeline takes place. 

I also wanted to reuse `transformError` as it makes the intent clear, that the `errorTransformation` shouldn't care about the errors coming from the spec (as opposed to using `errorTransformation` directly). 

If this PR is accepted, I can try dropping in some test that would document such behavior


Baccata:
Superseded by https://github.com/disneystreaming/smithy4s/pull/1034

PR #1031, opened 2023-06-14
Adding a comment in flatMapErrors

msosnicki:
Actually, on `0.18` the comment is not fully true. This is because the errors from transformations, are not piped through the `errorTransformer`.


kubukoz:
is this just because 0.18 is not synced with 0.17, or was there a deliberate divergence?

msosnicki:
@Baccata mentioned there might be a need for pre and post middleware error handlers, not sure it is planned for 0.18 though. Is there any issue where it is discussed? I guess that two method overrides could be provided easily, where you can specify just a single handler (like currently, with behavior as documented above), and two handlers if you need flexibility.
I was wondering too if there are cases where knowledge about `Endpoint` would be useful for error handling (similarly to middlewares).

Baccata (approved):

Baccata:
Closing as superseded by https://github.com/disneystreaming/smithy4s/pull/1083/files

PR #1034, opened 2023-06-14
Revert original behavior where middleware get all errors

daddykotex:
When merging https://github.com/disneystreaming/smithy4s/pull/877, I introduced a behaviour change that I had not anticipated: the errors defined in the smithy specs were turned into http response before users had a chance to react to them in the middleware they provide

this is problematic and this fix reverts that

this PR is an alternative to this one: https://github.com/disneystreaming/smithy4s/pull/1029


Baccata:
@msosnicki okay, after a little bit of thought I'm okay with making the spec'd errors observable from the middleware. If you're happy with this implementation from @daddykotex , I'd rather merge this instead of yours, because of the added unit-test.

msosnicki:
> @msosnicki okay, after a little bit of thought I'm okay with making the spec'd errors observable from the middleware. If you're happy with this implementation from @daddykotex , I'd rather merge this instead of yours, because of the added unit-test.

Sure, that works! Thanks :)

Baccata:
@daddykotex could you port the test to the 0.18 branch ?

Issue #971, opened 2023-05-19
[SimpleRestJson] Question: Should we include empty headers 


yisraelU:
In the RestJson1 protocol headers are dropped when the value is one of `null`, empty string, or empty collection
Should this be implemented in simpleRestJson ?

Baccata:
Can you share the corresponding test ?

yisraelU:
@Baccata https://github.com/awslabs/smithy/blob/24abf76268955ea642f9cc5811f3adb3de7b2f8f/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy#L415

Issue #970, opened 2023-05-19
[RestJson1] Base64 encode header thats targeted by MediaType


yisraelU:
the restJson1 protocol dictates when a `httpHeader` is targeting a  string that is annotated with the `mediaType` trait, the header value must be Base64 encoded.

Here is one of the tests for this https://github.com/awslabs/smithy/blob/be68f3bbdfe5bf50a104b387094d40c8069f16b1/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy#L592

PR #1035, opened 2023-06-15
[AWS REST Protocols] Implement AWS header (de)serialisation

Baccata:
Closes https://github.com/disneystreaming/smithy4s/issues/971
Closes https://github.com/disneystreaming/smithy4s/issues/970

Implement AWS Header de-serialisation, which abides by the following rules:  

* For each header name, the corresponding header values are merged together in a single string, with comma-separated values. 
* Values are quoted before being concatenated when the string contains a comma. 
* UNLESS the header values are timestamp values encoded using the `HTTP_DATE` encoding, in which case the values are not quoted (even if they contain commas). 

When encoding from a list, the serialisation logic therefore concatenates values together
When decoding into a list, the deserialisation logic multiple values from a single header. 

This PR also removes the concepts of `Total` and `Partial` metadata decoder, as the partiality is now handled at the schema level rather than the codec level. 

Although this logic is implemented in `core`, it is only wired in the AWS module.


Baccata:
The number of failing tests for AWS Rest Json 1 is now down to 6 \o/

modules/core/test/src/smithy4s/http/internals/HeaderSplitSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+import munit._
+
+class HeaderSplitSpec extends FunSuite {
+
+  test("Header split: single value") {
+    checkNormal("abcd efgh")("abcd efgh")
+  }
+
+  test("Header split: multiple normal strings") {
+    checkNormal("a, b, c")("a", "b", "c")
```
daddykotex, line 28: maybe one test case with no space after the comma?

modules/core/src/smithy4s/http/Metadata.scala
```diff
@@ -151,26 +150,13 @@ object Metadata {
...
 
   /**
-    * Decodes all the fields that are bound to http metadata,
-    * shoving values in a map.
+    * If possible, decode the data from fields that are bound to http metadata.
     */
-  def decodePartial[A](metadata: Metadata)(implicit
-      decoder: PartialDecoder[A]
-  ): Either[MetadataError, MetadataPartial[A]] =
+  def decode[A](metadata: Metadata)(implicit
```
daddykotex, line 155: I'm surprised no calls to `Metadata.decode` happens outside of test files

Or am I missing something

Baccata: @daddykotex yeah, it's rare to use `Metadata.decode` in this codebase as we use `Metadata.Decoder.fromSchema` instead.

lewisjkl (approved):

daddykotex:
I see failure related to content encoding in here, I'll check this out and try to fix it

PR #1036, opened 2023-06-15
[AWS SDK] Additional aws encoding details for the Rest Json protocol 

Baccata:
Besides the MD5 test (which appears to be false-positive), this PR takes care of tweaking the last bits that need tweaking for the AWS Rest Json protocol to be green. 

Depends on https://github.com/disneystreaming/smithy4s/pull/1035

(https://github.com/disneystreaming/smithy4s/pull/1035 should be merged before this one)

PR #1037, opened 2023-06-15
fix header test

yisraelU:
The false positive tests noted over [here](https://github.com/disneystreaming/smithy4s/pull/1036#issue-1759032380) , are the result of not testing correctly.
The required headers and forbid headers is for testing keys only while allowing the value to vary
The headers is for testing **both** key and value, not **just** the value .
All header tests defined in restJson dont have a required headers field as every case has value requirements too.
Please see [docs](https://smithy.io/2.0/additional-specs/http-protocol-compliance-tests.html#httprequesttests-trait)

> requireHeaders | [string] | A list of header field names that must appear in the serialized HTTP message, but no assertion is made on the value. Headers listed in headers do not need to appear in this list.
-- | -- | --

### DIFF
--- a/modules/compliance-tests/src/smithy4s/compliancetests/internals/Assertions.scala
+++ b/modules/compliance-tests/src/smithy4s/compliancetests/internals/Assertions.scala
@@ -129,6 +129,11 @@ private[internals] object assert {
       .combineAll
   }
 
+  /**
+   * A list of header field names that must appear in the serialized HTTP message, but no assertion is made on the value.
+   * Headers listed in headers do not need to appear in this list.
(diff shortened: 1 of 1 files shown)


daddykotex:
I remember reading this and implementing that but I may have gotten it wrong, thanks for finding that

yisraelU:
> I remember reading this and implementing that but I may have gotten it wrong, thanks for finding that

@daddykotex I am pretty sure you got it right.
checksum header test used to fail,
this was a more recent change

Baccata:
fair enough :+1:

PR #1038, opened 2023-06-16
Add support for gzip in aws json

daddykotex:


modules/aws-http4s/src/smithy4s/aws/internals/package.scala
```diff
@@ -0,0 +1,46 @@
...
+import smithy4s.schema.CachedSchemaCompiler
+import smithy4s.kinds.FunctorK
+import smithy4s.Hints
+
+package object internals {
+
+  private[internals] type RequestEncoderCompiler[F[_]] =
+    CachedSchemaCompiler[RequestEncoder[F, *]]
+
+  private[internals] def applyCompression[F[_]: Compression](
```
Baccata, line 30: @daddykotex I've moved this here as I wasn't comfortable with having what this does as a public method where you had it : the shape was a bit awkward as it didn't force the applied request transformation to be gzip specific.

PR #1040, opened 2023-06-16
Addressing SimpleRestJson tests :

Baccata:
* Using the Amazon discriminator as well as the current one, in order to improve compatibility with AWS-issued code-generators that target the AWS restJson protocol
* Adding a content-type when forcing an empty json object in the body of http responses.

### DIFF
--- a/modules/http4s/src/smithy4s/http4s/internals/SimpleRestJsonCodecs.scala
+++ b/modules/http4s/src/smithy4s/http4s/internals/SimpleRestJsonCodecs.scala
@@ -34,6 +34,12 @@ private[http4s] class SimpleRestJsonCodecs(
   private val hintMask =
     alloy.SimpleRestJson.protocol.hintMask ++ HintMask(IntEnum)
   private val underlyingCodecs = smithy4s.http.json.codecs(hintMask, maxArity)
+  private val errorHeaders = List(
+    smithy4s.http.errorTypeHeader,
+    // Adding X-Amzn-Errortype as well to facilitate interop
+    // with Amazon-issued code-generators.
+    smithy4s.http.amazonErrorTypeHeader
+  )
 
   private def addEmptyJsonToResponse[F[_]](
       response: Response[F]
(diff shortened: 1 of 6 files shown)


lewisjkl (approved):

modules/http4s/src/smithy4s/http4s/internals/SimpleRestJsonCodecs.scala
```diff
@@ -33,10 +35,16 @@ private[http4s] class SimpleRestJsonCodecs(
...
-    )
+  private def addEmptyJsonToResponse[F[_]](
+      response: Response[F]
+  ): Response[F] = {
+    response
+      .withBodyStream(
+        response.body
+          .ifEmpty(fs2.Stream.chunk(fs2.Chunk.array("{}".getBytes())))
+      )
+      .withContentType(`Content-Type`(mediaType"application/json"))
```
daddykotex, line 52: TIL there is a mediatype string interpolator

kubukoz: same

daddykotex (approved):
I'll approve because the code makes sense on its own, and the tests reflects that it works, but I'm not quite sure I understand why

we specify `amazonErrorTypeHeader` in two places:

1. in the unary server codecs - with a comments, and that makes sense to me
2. in the simplerestjson client codecs implementation - that also makes sense because we want the client to understand the information sent by the server

what I don't understand is why it is not specified for the aws json protocols implementation? anyway this is not blocking, more of a misunderstanding on my side rather than anything else

Baccata:
@daddykotex very good remark ! I think I just didn't think it through, but you're absolutely correct in that there was a bit of a disconnect. 

It should be addressed here : https://github.com/disneystreaming/smithy4s/pull/1040/commits/2f41f5a563665b4ceec4e8da26ad2fe004e27956

PR #1042, opened 2023-06-16
Backport middleware exception handling changes

daddykotex:


kubukoz (approved):
I think I got this in #1041, probably going to be an easy resolve though

PR #1045, opened 2023-06-19
AWS-Rest-Json feature branch tweaks 

Baccata:
* Rename `smithy4s.capability.Encoder` to `smithy4s.Writer` (it's not a capability per se), and also "write" is closer to what it does than "encode", as "writing" is the action of modifying existing messages  
* Centralises the logic that allows for combining writers for the usecase of REST (ie http-binding aware protocols) to smithy4s-core, as it's now third-party agnostic and could be reused in a reasonably large range of http interpreters. 
* Temporarily disables the MD5-related test so that we can progress independently


modules/core/src/smithy4s/Encoder.scala
```diff
@@ -0,0 +1,98 @@
+/*
```
daddykotex: should we call this file Writer instead?

modules/core/src/smithy4s/Encoder.scala
```diff
@@ -0,0 +1,98 @@
...
+package smithy4s
+
+import smithy4s.kinds.PolyFunction
+import smithy4s.capability.EncoderK
+
+/**
+  * An abstraction that codifies the notion of writing a piece of data into a message.
+  *
+  * This is particularly useful for http requests/responses, where different subsets of data
+  * have different impacts different locations of the http message : some fields may
```
daddykotex:
```suggestion
  * have a different impact in different locations of the http message : some fields may
```

modules/core/src/smithy4s/Encoder.scala
Baccata: doh !

PR #1047, opened 2023-06-19
add md5 checksum middleware

yisraelU:
I created a Client => Client for handling md5
The core logic is very simple and passes the test necessary

A couple of thoughts .
Obviously, this is not in line with anything in the kernel package, as  they operate at the Message codec level
I didn't think of a way to do this without modifying existing encoder structures.
Perhaps this should be an AwsClient feature only
Currently left it in the kernel , however, if we don't care if simpleRestJson implements this part of the protocol, I would say to move it into the aws-http4s unary endpoint


modules/http4s-kernel/src/smithy4s/http4s/kernel/Md5CheckSumClient.scala
```diff
@@ -0,0 +1,56 @@
...
+      }
+  }
+
+  private def reqWithChecksum[F[_]: Concurrent](
+      client: Client[F]
+  ): Client[F] = {
+    val md5HeaderPipe: Pipe[F, Byte, String] =
+      fs2.hash.md5[F] andThen fs2.text.base64.encode[F]
+    Client { request =>
+      val bodyF = request.body.through(md5HeaderPipe).compile.to(Chunk)
```
kubukoz: **issue:** I think you might need a strict request for this. Otherwise, future attempts to consume the body stream may fail (this might not be a concern for bodies created from plain byte arrays, but it's good practice nevertheless to keep that stream consumable)

yisraelU: Should the same apply here ?https://github.com/disneystreaming/smithy4s/blob/5902d2fd81d1eee6703daaab6e2ad95eb69684ab/modules/aws-http4s/src/smithy4s/aws/internals/AwsSignatureClient.scala#LL94C71-L94C71

Baccata: @yisraelU you are correct, it should indeed. Right now it's worked because we've only wired it to strict requests. I'll fix :)

yisraelU: any idea why `Request(..).toStrict` ,requires an instance of Concurrent ?
compiling a stream can accept multiple targets including Sync .

kubukoz: It's probably a relic of the past, before this was a thing: https://github.com/typelevel/fs2/pull/2898

kubukoz: I think we could copy-paste the implementation to avoid bringing in `Concurrent`. Up to you.

Baccata: since it's client-side, doing ` val bodyF = request.body.through(md5HeaderPipe).compile.to(Chunk)` and then `withEntityBody` at the end re-using the produced chunk might be more straightforward actually. 

`toStrict` is more useful to protect servers against ill-intended clients.

yisraelU: Isn't it possible that this client would be called to proxy up a request on the server side?

Baccata: In the grand scheme of things that's true, but the purpose of this middleware is not to be used with "any" http4s client, it's more about clients that are used under the hood for smithy4s clients, the requests of which is gonna be derived from high-level data types that live fully in memory

yisraelU: > since it's client-side, doing ` val bodyF = request.body.through(md5HeaderPipe).compile.to(Chunk)` and then `withEntityBody` at the end re-using the produced chunk might be more straightforward actually.

Did you mean `withEntity` ?
Also, we can't "reuse" the chunk as it's hashed , 
Does it make sense to compile to a F[Chunk[Byte]] and then Stream.emits to pipe it through md5 to create header?

kubukoz: That checks out.

Issue #1051, opened 2023-06-22
Generated code doesn't compile when there's a smithy shape using a reserved name


miguel-vila:
for example this smithy code:

```smithy
$version: "2.0"

namespace smithy4s.example.reservedNameOverride

use smithy4s.api#simpleRestJson

@simpleRestJson
service ReservedNameOverrideService {
    version: "1.0.0",
    operations: [SetOp]
}

@http(method: "POST", uri: "/api/set/", code: 204)
operation SetOp {
    input := {
       @required
       set: Set
    }
}

structure Set {
    @required
    someField: String,
    @required
    otherField: Integer
}
```

generates this code which doesn't compile:

```scala
package smithy4s.example.reservedNameOverride

import smithy4s.Schema
import smithy4s.Hints
import smithy4s.ShapeId
import smithy4s.schema.Schema.struct
import smithy4s.ShapeTag

case class SetOpInput(set: _Set)
object SetOpInput extends ShapeTag.Companion[SetOpInput] {
  val id: ShapeId = ShapeId("smithy4s.example.reservedNameOverride", "SetOpInput")

  val hints : Hints = Hints(
    smithy.api.Input(),
  )

  implicit val schema: Schema[SetOpInput] = struct(
    _Set.schema.required[SetOpInput]("set", _.set).addHints(smithy.api.Required()),
  ){
    SetOpInput.apply
  }.withId(id).addHints(hints)
}
```

because it's referring to `_Set` instead of `Set` in the same package

reproduced at #1050

kubukoz:
related: https://github.com/disneystreaming/smithy4s/issues/389

Baccata:
Wanna have a go at fixing ?

miguel-vila:
@Baccata yeah, will give it a go, any hint on the related code parts?

Baccata:
not off the top of my head

kubukoz:
`CollisionAvoidance.scala` is probably your man.

miguel-vila:
btw, what is the preferred solution here? to override the name (e.g. make it `case class SetOpInput(set: Set)` in the code above and `Set` refers to the type in the same package)? or to use the new name (make it `case class SetOpInput(set: _Set)`)? My thinking is that it's better to avoid name overriding

Baccata:
@miguel-vila, actually, I'll fix it. 

[...]

PR #1052, opened 2023-06-26
Remove reserved types

Baccata:
Remove the out-dated type-reservation list. The only things that are reserved are language keywords now. 

Closes https://github.com/disneystreaming/smithy4s/issues/1051


Baccata:
@miguel-vila, @kubukoz  although my fix works and is actually much better in the grand scheme of things (no more reserved types), it has the annoying side effect of drastically impacting the UX for AWS clients, has AWS has the annoying tendency of redefining primitive types in all its specs : 

https://github.com/aws/aws-sdk-js-v3/blob/main/codegen/sdk-codegen/aws-models/dynamodb.json#L10134-L10136 

So I think we need to take action within this same unit of work to avoid this problem by means of a ProjectionTransformer that'd turn all simple shapes that are coming from aws-namespaces into references to the "standard" shapes (as long as the shapes we're transforming don't have any traits, and have the same names/shape-type as their targets). This ProjectionTransformer would be applied by default. 

@miguel-vila, if you want to pick this, it'd be greatly appreciated

daddykotex:
> @miguel-vila, @kubukoz although my fix works and is actually much better in the grand scheme of things (no more reserved types), it has the annoying side effect of drastically impacting the UX for AWS clients, has AWS has the annoying tendency of redefining primitive types in all its specs :
> 
> https://github.com/aws/aws-sdk-js-v3/blob/main/codegen/sdk-codegen/aws-models/dynamodb.json#L10134-L10136
> 
> So I think we need to take action within this same unit of work to avoid this problem by means of a ProjectionTransformer that'd turn all simple shapes that are coming from aws-namespaces into references to the "standard" shapes (as long as the shapes we're transforming don't have any traits, and have the same names/shape-type as their targets). This ProjectionTransformer would be applied by default.
> 
> @miguel-vila, if you want to pick this, it'd be greatly appreciated

Do you think they have something similar in-house? If they use those specs to generate their SDK, that'd mean, for example, that the Dynamo client works with dynamo.String rather than java.lang.String?

Baccata:
@daddykotex they don't have newtypes in house, so they always render the underlying type, which solves the problem (which obviously we don't want to do, because we have newtypes)

kubukoz:
@Baccata so the transformation would:

- replace occurrences of said types (no traits, unqualified name matches the underlying type) with the underlying type
- remove said types from the model

and it would be included as a default `smithy4sTransformations` entry? or would you make it a default in another way?

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -658,13 +658,13 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
+      line"override def getMessage(): $string_ = ${field.name}"
     case field if field.tpe.isResolved =>
-      line"override def getMessage(): String = ${field.name}.orNull"
+      line"override def getMessage(): $string_  = ${field.name}.orNull"
     case field if field.required =>
-      line"override def getMessage(): String = ${field.name}.value"
+      line"override def getMessage(): $string_  = ${field.name}.value"
     case field =>
-      line"override def getMessage(): String = ${field.name}.map(_.value).orNull"
+      line"override def getMessage(): $string_  = ${field.name}.map(_.value).orNull"
```
kubukoz: **issue:** too many spaces (causing unnecessary changes)

Baccata:
> replace occurrences of said types (no traits, unqualified name matches the underlying type) with the underlying type
remove said types from the model

absolutely 

> and it would be included as a default smithy4sTransformations entry? or would you make it a default in another way?

yes

kubukoz:
I believe @denisrosca from my team will look into this next week. @baccata given this is 0.17, I suppose you want to target **this** branch with the transformation PR, so that #1052 doesn't get released without that?

Baccata:
> I suppose you want to target this branch with the transformation PR, so that https://github.com/disneystreaming/smithy4s/pull/1052 doesn't get released without that? 

Indeed I do

denisrosca:
@kubukoz @Baccata 

I opened https://github.com/disneystreaming/smithy4s/pull/1110 to remove AWS newtypes during code generation.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
Baccata: [a67eb61](https://github.com/disneystreaming/smithy4s/pull/1052/commits/a67eb61cc5c3567bc3646a1390fd1d17734aa8f1)

PR #1053, opened 2023-06-26
Document service product feature

miguel-vila:

PR #1054, opened 2023-06-26
Add documentation / examples for ServiceProduct usage

kubukoz:
Includes changes from #1053 by @miguel-vila.

image


modules/bootstrapped/src/generated/smithy4s/example/product/ExampleOperationInput.scala
```diff
@@ -0,0 +1,23 @@
+package smithy4s.example.product
```
kubukoz, line 1: **thought:** side note, I think the recent codegen changes around the `bootstrapped` module could've broken something wrt. detecting stale files - e.g. if you delete a shape from the smithy specs, you need to run a `clean` in order for that to be reflected on the filesystem

modules/bootstrapped/src/generated/smithy4s/example/product/ObjectService.scala
kubukoz, line 1: **thought:** removed this one in favor of the `ExampleService` that @miguel-vila added in the docs in #1053

modules/docs/markdown/04-codegen/01-customisation/10-service-product.md
```diff
@@ -0,0 +1,217 @@
...
+      override def apply[I, E, O, SI, SO](
+        fa: sp.service.Endpoint[I, E, O, SI, SO]
+      ): List[I] => List[O] = _.map(in => interp(fa.wrap(in)))
+
+    },
+  )
+
+// Usage
+
+listClient(impl)( /* implicit scope problem here - TODO */ ExampleService.serviceProduct)
```
kubukoz, line 134: **issue:** we're lacking something that would make this visible implicitly.

I tried the following (all combinations):

- making `serviceProduct` an implicit val in the interface
- making `serviceProduct` an implicit val in the generated code

I'm not sure why it's not being detected.

modules/docs/markdown/04-codegen/01-customisation/10-service-product.md
```diff
@@ -0,0 +1,217 @@
...
+              }
+
+          }
+
+      },
+    )
+
+  def build: Alg[ToList] = sp
+    .service
+    .algebra(new sp.service.EndpointCompiler[ToList] {
```
kubukoz, line 191: **praise:** for @Baccata - the whole `algebra` / `EndpointCompiler` thing is very cool.

modules/bootstrapped/src/generated/smithy4s/example/product/ExampleOperationInput.scala
Baccata: >  e.g. if you delete a shape from the smithy specs, you need to run a clean in order for that to be reflected on the filesystem 

That is correct. But at least now it's wired in `clean`, previously you had to manually remove stuff.

modules/docs/markdown/04-codegen/01-customisation/10-service-product.md
Baccata: > making serviceProduct an implicit val in the generated code 

where was the implicit located when you tried ?

kubukoz: exactly where the non-implicit val is (companion object of `ExampleServiceGen`)

Baccata:
I'm gonna merge it as is, it's good and we can amend later if we find a way to make the implicit work.

Baccata (approved):

PR #1056, opened 2023-06-29
endpoint trait implementation along with hostLabel trait

yisraelU:
- This is a first run at implementing client-side endpoint and hostLabel traits
   -  since hostLabel has no effect on the protocol serialization, I see no reason for there to be a server-side implementation , since the interpreter only matches on Path, not the hostname
- this passes restJson1 and AwsJson0 and AwsJson1 endpoint related tests
- Naming and file location are not great

#### currently failing due to OpenApi conversion as the hostLabel trait is unsupported 

~~TODO~~
- ~~add the ability for the client to disable host prefix rendering per https://smithy.io/2.0/spec/endpoint-traits.html#smithy-api-endpoint-trait~~
> Clients SHOULD provide a way for users to disable the hostPrefix injection behavior. If a user sets this flag, the client MUST NOT perform any hostPrefix expansion and MUST NOT prepend the prefix to the client derived host. The client MUST serialize members to any modeled target location regardless of this flag.

UPDATE
- added ability to disable hostPrefix injection
- included it in rest json builder


Baccata:
Closing in favour of https://github.com/disneystreaming/smithy4s/pull/1078

PR #1062, opened 2023-06-30
0.18 refactors

daddykotex:
Following the upgrade of our internal libraries on a 0.18 snapshot versions, we found some rough edges. Everything in this PR is up for discussion but it made things easier to upgrade and implement.

1. new Blob type
2. update to the codec api that uses the Blob type
3. theres a change in behaviour for the nativeBlobAndString codec (see below)
4. expose getResponseMetadata and getRequestMetadata from the http4s-kernel

the rest are really just refactors

-------

Before, with a shape like:

```
structure VersionOutput {
    @httpPayload
    @required
    version: String
}
```

You could feed the schema of that shape into the nativeBlobAndString codec and expect a `text/plain`, this is not the case anymore. Users are expected to prepare the schema ahead of time: in our case that meant going through `HttpRestSchema` that gives you a specialized schema without all the metadata noise. In that example, it basically gives you a `Schema.string`, which when given to the nativeBlobAndString does what we want.


modules/core/src/smithy4s/Blob.scala
```diff
@@ -0,0 +1,88 @@
...
+  final def toUTF8String: String = new String(toArray, StandardCharsets.UTF_8)
+}
+
+object Blob {
+
+  val empty: Blob = new Blob.ByteArrayBlob(Array.emptyByteArray)
+
+  def apply(bytes: Array[Byte]): Blob = new ByteArrayBlob(bytes)
+  def apply(buffer: ByteBuffer): Blob = new ByteBufferBlob(buffer)
+  def apply(string: String): Blob = new ByteArrayBlob(string.getBytes())
```
kubukoz: thought: UTF-8 is only a standard encoding since Java 18, previously it was platform-dependent. Shall we hardcode UTF-8 while we support older JVMs?

modules/core/src/smithy4s/schema/Schema.scala
```diff
@@ -180,6 +180,7 @@ object Schema {
   val boolean: Schema[Boolean] = Primitive.PBoolean.schema(prelude, "Boolean")
   val byte: Schema[Byte] = Primitive.PByte.schema(prelude, "Byte")
   val bytes: Schema[ByteArray] = Primitive.PBlob.schema(prelude, "Blob")
+  val blob: Schema[ByteArray] = Primitive.PBlob.schema(prelude, "Blob")
```
kubukoz, line 188: question: aren't these the same? I would've thought you'd replace `ByteArray` with the new `Blob`.

modules/core/src/smithy4s/Blob.scala
kubukoz: another thought, shall we make another variant of `Blob` that's optimized for strings?

modules/core/src/smithy4s/http/CodecAPI.scala
```diff
@@ -121,35 +83,19 @@ object CodecAPI {
...
     def imap[B](to: A => B, from: B => A): Codec[B] = new Codec[B] {
       def mediaType: HttpMediaType = self.mediaType
 
-      def decodeFromByteArrayPartial(
-          bytes: Array[Byte]
-      ): Either[PayloadError, BodyPartial[B]] =
-        self.decodeFromByteArrayPartial(bytes).map(_.map(to))
-
-      def decodeFromByteBufferPartial(
-          bytes: ByteBuffer
```
kubukoz, line 143: good riddance

modules/core/src/smithy4s/schema/Schema.scala
Baccata: @daddykotex has jumped on a WIP branch I started. I should have made it clearer that I had further changes planned for this and didn't want to open right away.

I'm gonna turn this PR into a draft that I'll pick up later. Nevertheless, I appreciate the initiative

modules/core/src/smithy4s/Blob.scala
Baccata: I thought about it making a string specialised case and decided against it as it'd make the integrations more annoying.

I'm happy to hardcore utf-8 in the smart constructors though.

modules/core/src/smithy4s/schema/Schema.scala
Baccata: Well for some reason turning into draft doesn't work from a phone

daddykotex:
By the way, I'm force pushing to this branch, I will close the PR, and maybe reopen in another PR if need be to align with other PRs that will be merge in the meantime

Baccata:
Closing this as I think a lot of it has already been handled in other PRs. Happy for you to open a new one though.

PR #1071, opened 2023-07-04
Revamp of the notion of codec 

Baccata:
**YES, IT'S YET ANOTHER HUMONGOUS REFACTORING, COME AND FIGHT ME**

* Introduces `Blob` as a construct that can capture Array[Byte] and ByteBuffers. A later PR will get rid of `smithy4s.ByteArray` in favour of `smithy4s.Blob` 
* Introduces `smithy4s.codecs.Writer` and `smithy4s.codecs.Reader` as ultra-generic, compositional abstractions. 
* Introduces `smithy4s.codecs.ReaderWriter` that is just a higher-kinded tuple with specialised accessors. 
* Introduces `smithy4s.capability.Zipper` as an abstraction that can be used in smithy4s-core in the absence of 
`cats.Applicative`. It's less typesafe, but is specialised for zips. 
* Moves most of the logic for `Rest`-decoding to core. 
* Moves `smithy4s.http.PayloadError` to `smithy4s.codec`,  separating it from the `HttpContractError` type (which now has a new ADT member wrapping a PayloadError) 
* Reworks the `json` module to make it easier to work with for users. In particular, `smithy4s.json.Json` contains static methods to parse/write documents. The configuration of writers and parsers, as well as the configuration of codecs is also made easier, via fluent-builder patterns. 
* Removes CodecAPI altogether. CodecAPI was basically a glorified `CachedSchemaCompiler`
* Gets rid of the `smithy4s.http.json` package in favour of `smithy4s.json`
* Moves the bespoke json-related logic that was hosted by the AWS module as opt-in behaviour in the Json module (flexible maps, NaN/Infinity, etc). 

#### Seriously though, why do you keep rewriting sh*t up, @Baccata ? 

Two reasons : 

* I'm definitely insane and enjoying killing 10 birds with one stone. 
* I'm trying to optimise for two axis of extensibility : 
   1. addition of protocols (that tend to share semantics) 
   2. implementation of the same protocols on top of different libraries-backends. Even though smithy4s is http4s-friendly/pure-FP friendly, I want to facilitate the implementation of protocols using other http libraries. For instance : JS clients directly built on top of fetch. Or SN-servers directly built on top of SN-unit. This kind of stuff. 

#### TODO 

- [x] Scaladoc all the things
- [x] Some more unit tests wouldn't hurt


modules/core/src/smithy4s/Blob.scala
```diff
@@ -0,0 +1,118 @@
...
+        buffer.compareTo(ByteBuffer.wrap(otherBlob.toArray)) == 0
+    }
+
+    override def equals(other: Any): Boolean = {
+      other.isInstanceOf[ByteArrayBlob] &&
+      buffer.compareTo(other.asInstanceOf[ByteBufferBlob].buffer) == 0
+    }
+
+    private val arr: Array[Byte] = null
+    def toArray: Array[Byte] = {
```
daddykotex, line 105: fixed a bug in this: https://github.com/disneystreaming/smithy4s/pull/1062/commits/7093509a06f9cb60d083333e822120d74b54370e

Baccata: Good catch, addressed in [5752764](https://github.com/disneystreaming/smithy4s/pull/1071/commits/57527646997196bae5f7aacd7aff855497b8f8bc)

modules/json/src/smithy4s/json/internals/JsoniterCodecCompilerImpl.scala
```diff
@@ -0,0 +1,97 @@
...
+        .map(mask => schema.transformHintsTransitively(mask.apply))
+        .getOrElse(schema)
+    amendedSchema.compile(visitor)
+  }
+
+}
+
+private[smithy4s] object JsoniterCodecCompilerImpl {
+
+  val defaultHintMask: HintMask =
```
daddykotex: I think you meant right by having this public, because with `withHintMask` we can only override and not combine the hint mask, but the object is private so it's hard to use when defining a new protocol

Baccata: Done

modules/core/src/smithy4s/http/package.scala
```diff
@@ -16,11 +16,21 @@
...
 package object http {
 
   val errorTypeHeader = "X-Error-Type"
   val amazonErrorTypeHeader = "X-Amzn-Errortype"
 
+  type HttpPayloadReader[A] = Reader[Either[HttpContractError, *], Blob, A]
+  type HttpPayloadCodec[A] = ReaderWriter[HttpPayloadReader, PayloadWriter, A]
+  type HttpMediaReader[A] = HttpMediaTyped[HttpPayloadReader, A]
+  type HttpMediaWriter[A] = HttpMediaTyped[PayloadWriter, A]
+  type HttpMediaCodec[A] = HttpMediaTyped[HttpPayloadCodec, A]
```
daddykotex, line 32: Is it possible that this should be: 
```scala
type HttpMediaCodec[A] =
    HttpMediaTyped[ReaderWriter[HttpMediaReader, HttpMediaWriter, *], A]
```

kubukoz:
Bro went and merged the largest refactor of the year with no reviews 💀

modules/core/src/smithy4s/http/package.scala
daddykotex: ah nvm, just picked it up lol

daddykotex:
my opinion is that it's fine because 0.18 was in a sort of influx state: partially completed refactor and that showed when I tried to integrate in our internal tooling

Baccata:
Happy to address comments later on, but many pieces are moving and I cannot afford to wait.

modules/core/src/smithy4s/capability/Covariant.scala
```diff
@@ -16,11 +16,11 @@
...
 package smithy4s.capability
 
-import smithy4s.ConstraintError
-
+/**
+  * Abstraction that encodes Functors
+  */
 trait Covariant[F[_]] {
   def map[A, B](fa: F[A])(f: A => B): F[B]
-  def emap[A, B](fa: F[A])(f: A => Either[ConstraintError, B]): F[B]
```
daddykotex, line 23: yes!!! I had not spotted that!

PR #1079, opened 2023-07-06
RestXml Protocol Support

lewisjkl:
Adds support for the RestXML protocol. 

* Adds json parsing option to respect ordering of map keys (backing the resulting maps with ListMap) 
* Comparison is relaxed in the case of `Option[List[A]]` so that `Some(List.empty)` and `None` are considered equivalent
* Adds utilities to facilitate the usecase of `SchemaVisitor[Option[F[...]]` 
* Enables RestXml compliance tests 
* Fixes everything that needs to be fixed in the Xml codecs for protocol compliance tests to pass 
* Wires a new `UnaryClientCodec` instance with all the details that are needed for RestXml


lewisjkl:
Spent a while trying to figure out why xmlName hints aren't coming through to collections.. no luck so far. Here are my other notes about some of the failing tests:

- Think this test is flawed, nested structure is not using xmlName? https://github.com/awslabs/smithy/blob/d6e01f9a62d4b31f6f3216f68bd0faa2ebbf827c/smithy-aws-protocol-tests/model/restXml/http-payload.smithy#L320

- Seemingly not capturing/passing in hints for xmlName on collection members: https://github.com/awslabs/smithy/blob/e6ee644a8d15e650c197b08b011707e87ea19c85/smithy-aws-protocol-tests/model/restXml/document-lists.smithy#L381

- SimpleScalarPropertiesWithEscapedCharacter - on a quick first pass it looks like fs2 xml lib is swallowing the escaped characters

- Need to handle string/blob content types types like in json protocols

- Content-Length header is missing which is causing 5-10 tests to fail out

yisraelU:
> Think this test is flawed, nested structure is not using xmlName? https://github.com/awslabs/smithy/blob/d6e01f9a62d4b31f6f3216f68bd0faa2ebbf827c/smithy-aws-protocol-tests/model/restXml/http-payload.smithy#L320

So this one is confusing to me, however, I believe the test is correct.
When there is a nested structure, the xmlName on the nested structure has no impact as the actual member name is whats used for the enclosing tag
see https://smithy.io/2.0/spec/protocol-traits.html#xmlname-on-structures-and-unions
> An xmlName trait applied to a structure or union changes the element name of the serialized shape; **however, it does not influence the serialization of members that target it.**

lewisjkl:
> > Think this test is flawed, nested structure is not using xmlName? https://github.com/awslabs/smithy/blob/d6e01f9a62d4b31f6f3216f68bd0faa2ebbf827c/smithy-aws-protocol-tests/model/restXml/http-payload.smithy#L320
> 
> So this one is confusing to me, however, I believe the test is correct. When there is a nested structure, the xmlName on the nested structure has no impact as the actual member name is whats used for the enclosing tag see https://smithy.io/2.0/spec/protocol-traits.html#xmlname-on-structures-and-unions
> 
> > An xmlName trait applied to a structure or union changes the element name of the serialized shape; **however, it does not influence the serialization of members that target it.**

Hey Yisrael 🙂 , yeah those are the docs I was missing. Appreciate you finding that!

lewisjkl:
Okay so stuck on two things that I will keep working through tomorrow...

```
org.http4s.Entity.apply(
        body,
        contentLength = None // How to set this?
      )
```

Not sure how to get the content length because the content is in a stream and we don't have any effects we can return inside of `EntityEncoder.encodeBy`.. I think the content length might get set automatically at a higher layer before the response is returned? Not really sure though, but the tests for this are still failing

The next thing is that, for example, the `XmlEmptyMaps` request test is failing because we are unable to decode the example for the tests. It seems like 0 bytes are even being passed in to the `media.body` for us to decode. I think in general there is some mismatch between what requests and responses need on the decoding side of things..

yisraelU:
> Okay so stuck on two things that I will keep working through tomorrow...

Thats how http4s handles creating requests with streams , they dont set the content length
> Encodes an entity body. Chunking of the stream is preserved. A Transfer-Encoding: chunked header is set, as we cannot know the content length without running the stream.

one solution is to call `toStrict` on the request which removes Transfer-Encoding and sets content-length

Though , Isn't it inherently the problem that we are using an effectful stream for something that is already in memory?    
for http4s we don't use stream

Baccata:
Disclaimer : my currently ongoing PR will fix (at least some of) the xmlName-related issues : https://github.com/disneystreaming/smithy4s/pull/1081

modules/xml/test/src/smithy4s/xml/XmlCodecSpec.scala
```diff
@@ -340,22 +339,6 @@ object XmlCodecSpec extends SimpleIOSuite {
       checkContent[Foo](xmlRight, Right("hello"))
   }
 
-  test("union") {
```
Baccata, line 343: this was a duplicate

PR #1081, opened 2023-07-10
Overhaul of the `smithy4s.schema.Field` datatype. 

Baccata:
Turn the Field GADT into a case-class. The concept of optionality is now handled by `Schema.NullableSchema` (which may be renamed to `OptionSchema`).

The requirement for adding `NullableSchema` came from having to handle cases like `sparse` collections, where the nullability is not contextualised to a field. Unfortunately, this addition means that there are two ways of defining optionality : one directly on the schema, one on the field.

Fortunately, it appears we can eliminate this duplication by making the `NullableSchema` the only way of modelling optionality. This seems to have the very nice quality of simplifying a lot of the code, at the cost of changing the `Field`'s interface in order to make the concept of `default` easy to access. A `NullableSchema` simply yields default values equal to `None` when the underlying schema has no default value.

**Additionally** : this recovers the distinction between field-level hints and schema-hints, by amending the Hints data-type to be able to carry two different layers of hints. This is much less invasive than the alternative, which would require changing a lot more method signatures (think that maps/collections have member hints too) 

- [x] Align Alt's model with the new Field's model
- [x] Rename `nullable` to `optional` ? 
- [x] Add section of Schema docs with `Nullable|OptionSchema` 
- [x] Add test spec around `Hints` (in particular testing the new distinction) 
- [x] Add smoke test in the bootstrapped code, to ensure that `addMemberHints` is called when relevant
- [x] Add tests around the dynamic module, to ensure that `addMemberHints` is called when relevant.


modules/core/src/smithy4s/Hints.scala
```diff
@@ -25,11 +25,23 @@ package smithy4s
...
+  *
+  * Under the hood, the hints are composed of two maps : one for member-level hints,
+  * one for target-level hints.
   */
 trait Hints {
   def isEmpty: Boolean
   def all: Iterable[Hints.Binding]
+
+  def memberHintsMap: Map[ShapeId, Hints.Binding]
+  def targetHintsMap: Map[ShapeId, Hints.Binding]
```
Baccata, line 37: the distinction between member hints and target hints is here. What we've learnt is that most of the time, the distinction won't matter in interpreters. However, it may matter (in cases such as Xml, where `XmlName` MUST be handled differently based on whether it's on a shape or a member). 

Allowing for the `Hints` construct to own this distinction minimises the impact to existing SchemaVisitors, as members can not only be field members, but also list members, map key/value, etc. Handling all these cases just to make Xml happy would induce a lot of frustrating changes. 

NB : the Hints construct existing methods behave exactly the same as before. We now just store the hints differently based on whether they come from members or non-members.

modules/core/src/smithy4s/schema/Field.scala
```diff
@@ -17,166 +17,73 @@
 package smithy4s
 package schema
 
-import smithy4s.kinds.PolyFunction
-
 /**
   * Represents a member of product type (case class)
   */
-sealed abstract class Field[F[_], S, A] {
```
Baccata, line 25: This is one the main change of this PR : this simplification is allowed by the recent introduction of `NullableSchema` (renamed here to `OptionSchema`). Not having to handle fields as GADTs now ripples through the codebase in a very nice way, allowing for the removal of a lot of code. 

Additionally, no more "FolderK" and co. The logic should look a lot easier to digest to the external eye.

modules/core/src/smithy4s/schema/Field.scala
```diff
@@ -17,166 +17,73 @@
...
-    override def foldK[G[_]](folder: Field.FolderK[F, S, G]): G[Option[A]] =
-      folder.onOptional(label, instance, get)
-    override def leftFolder[B](folder: Field.LeftFolder[F, B]): (B, S) => B = {
-      val partiallyApplied = folder.compile(label, instance)
-      (b, s) =>
-        get(s) match {
-          case Some(a) => partiallyApplied(b, a)
-          case None    => b
-        }
+  def getIfNonDefault(s: S): Option[A] = {
```
Baccata: This is a **very important change**, in that we now have a method that essentially say "dear structure, please give me the value corresponding to this field ONLY IF that value is not the default value for the field". 

Starting from the rule that the default value for option values is `None`, we can process optional values and default values in the same exact fashion, and decide to omit to encode certain fields based on whether the value they contain is the default one.

modules/core/src/smithy4s/schema/Field.scala
```diff
@@ -17,166 +17,73 @@
...
 
-  trait Folder[F[_], S, B] {
-    def onRequired[A](label: String, instance: F[A], get: S => A): B
-    def onOptional[A](label: String, instance: F[A], get: S => Option[A]): B
-  }
-
-  trait LeftFolder[F[_], B] {
-    def compile[T](label: String, instance: F[T]): (B, T) => B
-  }
+  def hasDefaultValue: Boolean = getDefaultValue.isDefined
```
Baccata, line 57: this can be used to statically decide whether to invoke `get` or `getIfNonDefault` when implementing encoders. `get` is obviously more performant, as no comparison is performed, nor boxing.

modules/core/src/smithy4s/schema/Alt.scala
```diff
@@ -23,36 +23,54 @@ import kinds._
 /**
   * Represents a member of coproduct type (sealed trait)
   */
-final case class Alt[F[_], U, A](
+final case class Alt[U, A](
```
Baccata, line 26: parts of the changes in this PR is to eliminate the higher-kinded-ness of `Field` and `Alt`, as it's not been used at all ever since the initial-encoding of `Schema` became central to the codebase.

modules/core/src/smithy4s/schema/Field.scala
```diff
@@ -17,166 +17,73 @@
...
+  def isStrictlyRequired: Boolean = !hasDefaultValue
```
Baccata, line 58: this can be used to statically decide whether to invoke `get` or `getIfNonDefault` when implementing encoders. `get` is obviously more performant, as no comparison is performed, nor boxing.

modules/core/src/smithy4s/Hints.scala
kubukoz: I think this is great and a really nice way to keep a strong degree of compatibility with existing users.

modules/core/src/smithy4s/Hints.scala
```diff
@@ -38,26 +50,68 @@ trait Hints {
...
 
-  val empty: Hints = new Impl(Map.empty)
+  val empty: Hints = new Impl(Map.empty, Map.empty)
+
+  def apply(bindings: Hint*): Hints = fromSeq(bindings)
+  def member(bindings: Hint*): Hints = Impl(mapFromSeq(bindings), Map.empty)
 
-  def apply[S](bindings: Hint*): Hints = fromSeq(bindings)
+  def fromSeq(bindings: Seq[Hint]): Hints =
+    Impl(Map.empty, mapFromSeq(bindings))
```
kubukoz: suggestion: all these `Impl` constructor calls should use named params

modules/core/src/smithy4s/schema/Alt.scala
kubukoz: used at all... by you ;P 

like I said privately, I do think it was useful because of `mapK(compile)`, but it's not essential.

modules/core/src/smithy4s/schema/Field.scala
kubukoz: nitpick: `getUnlessDefault`?

lewisjkl (approved):

kubukoz (approved):
forgot to commit comments earlier

LGTM with some nitpicks

Baccata:
> like I said privately, I do think it was useful because of mapK(compile), but it's not essential. 

I hear you, but I think getting rid of the HKT helps readability for "normal" developers who'd like to understand Schemas and SchemaVisitors, which is desirable.

PR #1083, opened 2023-07-11
Implement server endpoint handlers as middlewares

### LINKED ISSUE #1077: Make flatMapErrors aware about service and endpoint context
Currently `flatMapErrors` is a `PartialFunction[Throwable, F[Throwable]]` that is the same for all the endpoints in the service.
In some cases, it would make sense to make this error handler aware about the service/endpoint context (similar to endpoint middlewares). To keep backwards compatibility, we could add an more generic method that allows for such fine grained handling, and leave `flatMapErrors` as is (with implementation delegating the the specialized version). 
Interested what are your thoughts about this

msosnicki:
closes #1077 

If accepted, I will backport it to `0.17`


Baccata (approved):
Very nice, I'm glad you suggested it 👍

PR #1088, opened 2023-07-14
Union member constructors

yisraelU:
adds smart constructors for unions - unit cases and type cases which are cast as the union type
- Still need to add in product case
- please let me know about location in file, naming or anything else 
- i left empty parentheses for unit case


modules/bootstrapped/src/generated/com/amazonaws/dynamodb/DynamoDB.scala
```diff
@@ -165,7 +165,9 @@ object DynamoDBOperation {
     val hints: Hints = Hints.empty
 
     final case class InternalServerErrorCase(internalServerError: InternalServerError) extends ListTablesError
+    def internalServerErrorCase(internalServerErrorCase:InternalServerError): ListTablesError = InternalServerErrorCase(internalServerErrorCase)
```
kubukoz: thought: this is probably not as useful as it could be. According to [Oli's message from April](https://discord.com/channels/1045676621761347615/1045676622491164724/1093807442644701236), this method should take the same arguments as `InternalServerError` and wrap them into that case class itself. Otherwise, it's just a different casing of `InternalServerErrorCase`

Baccata: @kubukoz the benefit of should be two-folds :

1. The return type is the union, whereas calling the ADT member constructor follows the type of the member, which is annoying in many aspects 
2. The names shouldn't suck as much.

I'd be willing to add overloads in the case where the member targets a structure.

Edit: @yisraelU, the methods shouldn't have the "Case" suffix

kubukoz: I was referring to this part:

> with flatten constructor parameters when the union member targets a structure

so what you said you'd be willing to do ;) I thought that was a given

Baccata:
@yisraelU now that you're working from a fork, it's likely that your way of syncing is not taking into consideration the upstream. 

You'll need to do something like : 

``` 
git remote add upstream git@github.com:disneystreaming/smithy4s.git
git fetch upstream 
git merge upstream/series/0.18
```

yisraelU:
> git remote add upstream git@github.com:disneystreaming/smithy4s.git
> git fetch upstream 
> git merge upstream/series/0.18

i was syncing my fork with smithy4s .
Ill try this

Baccata:
Don't know what's happening but the diff is way off still

Baccata:
@yisraelU, looks like you've made a mistake during a merge : https://github.com/disneystreaming/smithy4s/pull/1088/commits/05e2afb4782ca3546e731b2e6958bc07ffe99208

Would you mind cherry-picking the relevant commits onto a new branch and recreating the PR ?

PR #1091, opened 2023-07-17
Ordinal-based coproducts

Baccata:
* Changes the internal models of `Service` and `UnionSchema` to be based on coproduct's `ordinals`, in order to allow for indexed-based dispatch (which should be more performant than map-based dispatch) 
* In particular, split the `endpoint` method in the service interface, so that the `input` can be accessed on its own. 
* Changes the model of `Alt` to carry a projector partial function. `PartialFunction[U, A]` is preferred to `U => Option[A]` because it doesn't necessarily induce an instantiation. It's also trivial to go from `PartialFunction[U, A]` to `U => Option[A]` via the `.lift` method 
* Amends the rendering logic to render ordinal values where they're needed (all operations, all union members)
* Adds `Schema.either` and `Schema.tuple` smart-constructors 

- [x] Amend documentation where needed
- [x] ~Render unions as sealed abstract classes instead of sealed traits~ not doing it because of how it adds a field to the bytecode representation of ADT member classes which uses more memory
- [x] Generalise `Schema.tuple` via `Boilerplate` to handle all tuples from arity 2 to 22


modules/core/src/smithy4s/schema/Alt.scala
```diff
@@ -26,7 +27,8 @@ import kinds._
 final case class Alt[U, A](
     label: String,
     schema: Schema[A],
-    inject: A => U
+    inject: A => U,
+    project: PartialFunction[U, A]
```
Baccata, line 30: New projection function, which removes the need to have `projector` in `Dispatcher`

modules/core/src/smithy4s/schema/Schema.scala
```diff
@@ -221,11 +222,23 @@ object Schema {
...
   def recursive[A](s: => Schema[A]): Schema[A] = Schema.LazySchema(Lazy(s))
 
-  def union[U](alts: Alt[U, _]*)(dispatch: U => Alt.WithValue[U, _]): Schema.UnionSchema[U] =
-    Schema.UnionSchema(placeholder, Hints.empty, alts.toVector, dispatch)
+  def union[U](alts: Vector[Alt[U, _]]): PartiallyAppliedUnion[U] = new PartiallyAppliedUnion(alts)
+  def union[U](alts: Alt[U, _]*) : PartiallyAppliedUnion[U] = new PartiallyAppliedUnion(alts.toVector)
 
-  def union[U](alts: Vector[Alt[U, _]])(dispatch: U => Alt.WithValue[U, _]): Schema.UnionSchema[U] =
-    Schema.UnionSchema(placeholder, Hints.empty, alts, dispatch)
+  def either[A, B](left: Schema[A], right: Schema[B]) : Schema[Either[A, B]] = {
```
Baccata, line 228: new smart constructors for `Schema[Either[A, B]]` and `Schema[Tuple[A, B]]`

modules/dynamic/src/smithy4s/dynamic/internals/DynamicService.scala
```diff
@@ -23,25 +23,19 @@ import smithy4s.kinds.PolyFunction5
...
     hints: Hints
 ) extends Service.Reflective[DynamicOp]
     with DynamicSchemaIndex.ServiceWrapper {
 
   type Alg[P[_, _, _, _, _]] = PolyFunction5.From[DynamicOp]#Algebra[P]
   override val service: Service[Alg] = this
 
-  private lazy val endpointMap: Map[ShapeId, Endpoint[_, _, _, _, _]] =
-    endpoints.map(ep => ep.id -> ep).toMap
+  private lazy val ordinalMap: Map[ShapeId, Int] =
```
Baccata, line 34: this could be changed by finding a way for the `DynamicEndpoint` to carry the ordinal

project/Boilerplate.scala
```diff
@@ -222,21 +222,6 @@ object Boilerplate {
       -   final def andThen[H[${`_.._`}]](other: PolyFunction$suffix[G, H]): PolyFunction$suffix[F, H] = new PolyFunction$suffix[F, H]{
       -      def apply[${`A..N`}](fa: F[${`A..N`}]): H[${`A..N`}] = other(self(fa))
       -   }
-      -
-      -   import Kind$arity._
```
Baccata, line 226: removing this because the unsafe caching needs to be backed by maps or arrays depending on the situation

modules/aws-http4s/test/src/smithy4s/aws/query/AwsQueryCodecSpec.scala
```diff
@@ -232,14 +232,7 @@ object AwsQueryCodecSpec extends SimpleIOSuite {
...
     type Foo = Either[Int, String]
-    implicit val schema: Schema[Foo] = {
-      val left = int.oneOf[Foo]("left", Left(_))
-      val right = string.oneOf[Foo]("right", Right(_))
-      union(left, right) {
-        case Left(int)     => left(int)
-        case Right(string) => right(string)
-      }
-    }
+    implicit val schema: Schema[Foo] = Schema.either(int, string)
```
daddykotex, line 235: that looks good, wow

daddykotex (commented):
that `def either` change makes it so much cleaner

good job once more

modules/core/src/generated/schema/PartiallyAppliedTuple.scala
```diff
@@ -0,0 +1,46 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/////// THIS FILE WAS GENERATED AT BUILD TIME, AND CHECKED-IN FOR DISCOVERABILITY ///////
+
+package smithy4s
+package schema
+
+class PartiallyAppliedTuple protected[schema](placeholder: ShapeId) {
```
Baccata, line 22: Allows for `Schema.tuple(schemaA,  schemaB, ...)`

Baccata:
@kubukoz if you can give it a review, I'd appreciate it :)

kubukoz:
I will

modules/core/src/smithy4s/schema/Alt.scala
kubukoz: inject + project form something similar to a prism, wondering if it'd be useful to define it that way...

edit: at the time I was originally writing this, we didn't have #1103 - that other PR adds the primitives, so we could actually use them in alts

modules/core/src/smithy4s/schema/Alt.scala
```diff
@@ -59,10 +61,23 @@ final case class Alt[U, A](
 object Alt {
 
   final case class WithValue[U, A](
```
kubukoz: **question:** can we remove this now? I don't see anything failing to compile when I do this

```diff
diff --git a/modules/core/src/smithy4s/schema/Alt.scala b/modules/core/src/smithy4s/schema/Alt.scala
index 45eb6db6..20e41ad0 100644
--- a/modules/core/src/smithy4s/schema/Alt.scala
+++ b/modules/core/src/smithy4s/schema/Alt.scala
@@ -34,8 +34,8 @@ final case class Alt[U, A](
   @deprecated("use .schema instead", since = "0.18.0")
   def instance: Schema[A] = schema
 
-  def apply(value: A): Alt.WithValue[U, A] =
-    Alt.WithValue(this, value)
+  // def apply(value: A): Alt.WithValue[U, A] =
+  //   Alt.WithValue(this, value)
 
   def hints: Hints = schema.hints
   def memberHints: Hints = schema.hints.memberHints
@@ -60,10 +60,10 @@ final case class Alt[U, A](
 }
 object Alt {
 
-  final case class WithValue[U, A](
-      private[schema] val alt: Alt[U, A],
-      value: A
-  )
+  // final case class WithValue[U, A](
+  //     private[schema] val alt: Alt[U, A],
+  //     value: A
+  // )
 
   type ClassTagged[U] = WithClassTag[U, _]
   final case class WithClassTag[U, A <: U](

```

modules/core/src/smithy4s/schema/Alt.scala
```diff
@@ -59,10 +61,23 @@ final case class Alt[U, A](
...
-      private[Alt] val alt: Alt[U, A],
+      private[schema] val alt: Alt[U, A],
       value: A
   )
 
+  type ClassTagged[U] = WithClassTag[U, _]
```
kubukoz: **nitpick:** I don't see this type being used anywhere

modules/core/src/smithy4s/schema/Alt.scala
```diff
@@ -59,10 +61,23 @@ final case class Alt[U, A](
...
+      alt: Alt[U, A],
+      classTag: scala.reflect.ClassTag[A]
+  )
+
+  object WithClassTag {
+    implicit def fromAlt[U, A <: U: ClassTag](
+        alt: Alt[U, A]
+    ): WithClassTag[U, A] =
+      WithClassTag(alt, implicitly[ClassTag[A]])
+  }
```
kubukoz: **question:** are these used anywhere?

modules/core/src/smithy4s/schema/PartiallyAppliedUnion.scala
```diff
@@ -0,0 +1,49 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.schema
+
+import smithy4s.Hints
+
+class PartiallyAppliedUnion[U](val alts: Vector[Alt[U, _]]) extends AnyVal {
```
kubukoz: **nitpick:** final class?

modules/core/src/smithy4s/http/HttpContractError.scala
```diff
@@ -154,12 +154,19 @@ object MetadataError {
...
+      arityError,
+      failedConstraint,
+      impossibleDecoding
+    ) {
+      case _: NotFound           => 0
+      case _: WrongType          => 1
+      case _: ArityError         => 2
+      case _: FailedConstraint   => 3
+      case _: ImpossibleDecoding => 4
     }
```
kubukoz, line 170: **thought:** rant time: I have a tiny gripe with the fact that you can't pass Scala types as values without coming up with custom constructs like `Hint.Binding` or `ShapeTag`.

ClassTag doesn't count.

modules/core/src/smithy4s/http/HttpErrorSelector.scala
```diff
@@ -80,20 +79,24 @@ private[http] final class HttpErrorSelector[F[_]: Covariant, E](
...
-      alts.map(Kind1.existential(_)),
-      identity(_)
-    )
+      val builder = Map.newBuilder[Any, Any]
+      alts.foreach { alt =>
+        builder += alt -> compileAlt(alt)
+      }
+      val resultCache = builder.result()
+      def apply[A](alt: Alt[E, A]): F[E] = {
+        resultCache(alt).asInstanceOf[F[E]]
```
kubukoz, line 97: **question:** is this something we could encapsulate in a public pattern? I suppose it'd be a sort of dual to `Dispatcher`

modules/core/src/smithy4s/schema/Alt.scala
```diff
@@ -86,51 +101,48 @@ object Alt {
         encoderK: EncoderK[G, Result]
     ): G[U]
 
-    def projector[A](alt: Alt[U, A]): U => Option[A]
+    def ordinal(u: U): Int
```
kubukoz, line 82: **thought:** could we make it more constrained than just `Int`, in dispatchers? maybe a type member?

then, we'd also need methods like `def get[A](i: Ordinal)(items: Seq[A]): A`.

modules/dynamic/src/smithy4s/dynamic/internals/DynamicService.scala
kubukoz: is there any issue with making it do so?

kubukoz: although tbh I don't mind it either way, the map seems just as fine

modules/core/src/generated/schema/PartiallyAppliedTuple.scala
kubukoz: gah dayum

modules/core/src/smithy4s/schema/Alt.scala
Baccata: I thought about it and I don't really want to because it'd impact things too much. Moreover we're making the generation of optics an opt-in.

Baccata: And we can't make the `Alt` a bonafide `Prism` and expose it to users because of recursion concerns. Alts need to be hidden in the computation of the schema.

modules/core/src/smithy4s/schema/Alt.scala
Baccata: We could have `get[A](u: U)(items: Seq[A]): A` ?

modules/core/src/smithy4s/http/HttpErrorSelector.scala
Baccata: We might be able to but I've not thought about it enough to know if it's possible

modules/core/src/smithy4s/http/HttpContractError.scala
Baccata: for sure

Baccata: really there should be a dual to constructor that verifies that you have one "thing" per case, without having to actually handcraft a pattern-match.

modules/core/src/smithy4s/http/HttpErrorSelector.scala
Baccata: Right, the bare minimum for us to be able to generalise would be to have a flatmap constraint on the F

modules/core/src/smithy4s/schema/Alt.scala
kubukoz: hmm I don't know about the recursion concerns, can you elaborate?

other than that I get it

modules/core/src/smithy4s/http/HttpErrorSelector.scala
kubukoz: I'd be fine with such a constraint. `flatten` + `map` (from Covariant) would be fine, no?

modules/core/src/smithy4s/schema/Alt.scala
kubukoz: that... works. but now I'm afraid of people passing a seq of the wrong length. It's equally bad :( 

the whole idea is reminiscent of `Representable` from cats

modules/dynamic/src/smithy4s/dynamic/internals/DynamicService.scala
Baccata: > is there any issue with making it do so? 

It's not that simple, because the DynamicOps are not currently tied to a service.

modules/core/src/smithy4s/http/HttpErrorSelector.scala
Baccata: I gave it a go and it's a lot more complicated to generalise than the encoding side, partially because the action of decoding may "consume" data. 

So it ends up looking like this : 

```scala
    def compileDecoder[Decoder[_], F[_], Message, Discriminator](
        precompile: Precompiler[Decoder],
        discriminator: Alt[_, _] => Discriminator,
        discriminate: Message => F[(Discriminator, Message)]
    )(implicit
        decoder: DecoderK[Decoder, F, Message]
    ): Decoder[U]
```

It may be worth exploring further but certainly not in this PR

modules/core/src/smithy4s/schema/Alt.scala
Baccata: > hmm I don't know about the recursion concerns, can you elaborate?

Actually I was partially wrong. What I meant was that : 

```scala 
val schema: Schema[A] = Schema.recursive {
    val aField = schema.optional[A]("a", _.a) 
    Schema.struct(aField)(A(_))
} 
```

Rendering it as follows to make `aField` visible to the external world doesn't work because `aField` refers to the initialised schema. 

```scala 
val aField = schema.optional[A]("a", _.a) 
val schema: Schema[A] = Schema.recursive {
    Schema.struct(aField)(A(_))
} 
```

However we could potentially render it as follows, which works : 

```scala 
val schema: Schema[A] = Schema.recursive {
    Schema.struct(aField)(A(_))
} 

val aField = schema.optional[A]("a", _.a) 
```

Baccata: I'll sync with Jeff to explore an alternative to his PR after this one is merged.

kubukoz (approved):

PR #1092, opened 2023-07-17
empty response content-type

### LINKED ISSUE #1072: 0.18: optional bodies and content-type
Forgive me if this has already been discussed, but anyway:

I noticed that an operation without an output body, e.g.

```smithy
@http(method: "GET", uri: "/no-output")
@readonly
operation NoOutput {
}
```

ends up being:

```http
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 0
Date: Tue, 04 Jul 2023 
```

which makes sense and I think it's a change since 0.17.

However, when the optional has an optional output body:

```smithy
@http(method: "GET", uri: "/optional-output")
@readonly
operation OptionalOutput {
    output := {
        @httpPayload
        bod: String
    }
}
```

and that body is a `None`, the HTTP response has a content type:

```http
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 0
Content-Type: application/json
Date: Tue, 04 Jul 2023 
```

which doesn't seem right - there's no content and the response isn't valid JSON. Is this a bug?

Tested on `0.18.0-595-e8a6dc6`, made from [this commit](https://github.com/disneystreaming/smithy4s/tree/e8a6dc6d1d1d0cfd50de52d2b99be64a9c694a70), which is pretty fresh (here's [a diff against 0.18](https://github.com/disneystreaming/smithy4s/compare/e8a6dc6d1d1d0cfd50de52d2b99be64a9c694a70...series/0.18))

lewisjkl:
Before any changes, the response was coming in as `{}` with `content-length: 2` and `content-type: application/json`. After changes, it is now an empty body with `content-length: 0`, but still has `content-type: application/json`. Currently it looks like the content type is set globally on all operations in simpleRestJson. I'm not sure what the best way would be to remove it when the body is empty.

Closes #1072.

### DIFF
--- a/modules/json/src/smithy4s/json/internals/JsonPayloadCodecCompilerImpl.scala
+++ b/modules/json/src/smithy4s/json/internals/JsonPayloadCodecCompilerImpl.scala
@@ -53,13 +53,25 @@ private[json] case class JsonPayloadCodecCompilerImpl(
   def fromSchema[A](schema: Schema[A], cache: Cache): PayloadCodec[A] = {
     val jcodec = jsoniterCodecCompiler.fromSchema(schema, cache)
     val reader: PayloadReader[A] = new JsonPayloadReader(jcodec)
-    val writer: PayloadWriter[A] = Writer.encodeBy { (value: A) =>
-      val intermediate = Blob(
-        writeToArray(value, jsoniterWriterConfig)(jcodec)
-      )
-      if (intermediate.sameBytesAs(Blob("null"))) Blob("{}")
-      else intermediate
-    }
+    val writer: PayloadWriter[A] =
+      schema.hints.get(smithy.api.HttpPayload) match {
+        case Some(_) =>
+          Writer.encodeBy { (value: A) =>
+            val intermediate = Blob(
+              writeToArray(value, jsoniterWriterConfig)(jcodec)
+            )
+            if (intermediate.sameBytesAs(Blob("null"))) Blob("")
+            else intermediate
+          }
+        case None =>
+          Writer.encodeBy { (value: A) =>
+            val intermediate = Blob(
+              writeToArray(value, jsoniterWriterConfig)(jcodec)
+            )
+            if (intermediate.sameBytesAs(Blob("null"))) Blob("{}")
+            else intermediate
+          }
+      }
     ReaderWriter(reader, writer)
   }
 
--- a/modules/json/src/smithy4s/json/internals/JsonPayloadCodecCompilerImpl.scala
+++ b/modules/json/src/smithy4s/json/internals/JsonPayloadCodecCompilerImpl.scala
@@ -53,25 +53,11 @@ private[json] case class JsonPayloadCodecCompilerImpl(
   def fromSchema[A](schema: Schema[A], cache: Cache): PayloadCodec[A] = {
     val jcodec = jsoniterCodecCompiler.fromSchema(schema, cache)
     val reader: PayloadReader[A] = new JsonPayloadReader(jcodec)
-    val writer: PayloadWriter[A] =
-      schema.hints.get(smithy.api.HttpPayload) match {
-        case Some(_) =>
-          Writer.encodeBy { (value: A) =>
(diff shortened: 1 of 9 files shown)


modules/json/src/smithy4s/json/internals/JsonPayloadCodecCompilerImpl.scala
```diff
@@ -53,13 +53,25 @@ private[json] case class JsonPayloadCodecCompilerImpl(
...
     val reader: PayloadReader[A] = new JsonPayloadReader(jcodec)
-    val writer: PayloadWriter[A] = Writer.encodeBy { (value: A) =>
-      val intermediate = Blob(
-        writeToArray(value, jsoniterWriterConfig)(jcodec)
-      )
-      if (intermediate.sameBytesAs(Blob("null"))) Blob("{}")
-      else intermediate
-    }
+    val writer: PayloadWriter[A] =
+      schema.hints.get(smithy.api.HttpPayload) match {
```
yisraelU: perhaps move this match statement inside the Writer to just indicate the empty body as seems like the codec and writer are the same

Baccata: This is not the right fix for this : the json module MUST NO LONGER have any reference to any of the http-binding traits. That's the whole point of moving the `smithy.json` package to `smithy4s.json`

The problem is more that the original line changing nulls to empty-object is ill-located, and should be the responsibility of another component, probably [here](https://github.com/disneystreaming/smithy4s/blob/series/0.18/modules/http4s/src/smithy4s/http4s/internals/SimpleRestJsonCodecs.scala#L40-L46). That same component should also deal with the querying of `httpPayload` to decide whether to turn `null` into `""` or `{}`

lewisjkl: Makes sense, thanks 👍

yisraelU:
> I'm not sure what the best way would be to remove it when the body is empty.

I would add it in the RequestEncoder and call the Entity Encoder myself [here](https://github.com/disneystreaming/smithy4s/blob/c1f08838a9c3aeffc97540dddb1e3b77fe8e847e/modules/http4s-kernel/src/smithy4s/http4s/kernel/RequestEncoder.scala#L58) 
```
val entity = entityEncoder.toEntity(a)
val headers = entityEncoder.headers
// if entity.length >0 ....

modules/json/src/smithy4s/json/internals/JsonPayloadCodecCompilerImpl.scala
Baccata: Another problem is that the reliance of `org.http4s.EntityEncoder` creates the very problem of static content-types: 

https://github.com/disneystreaming/smithy4s/blob/c1f08838a9c3aeffc97540dddb1e3b77fe8e847e/modules/http4s-kernel/src/smithy4s/http4s/kernel/EntityEncoders.scala#L30-L38

Right now the path of transformation goes : 

`HttpMediaWriter => EntityEncoder => ResponseEncoder => REST-respecting ResponseEncoder`

So it may be that we want to disappear the `EntityEncoder` step in favour of a more direct `HttpMediaWriter => ResponseEncoder` that'd take care of dynamically setting the content-type based on whether the blob is empty or not. 

That change would ripple all the way to the `restSchemaCompiler` methods that currently take `CachedSchemaCompiler[EntityEncoder[F, *]]`

lewisjkl: Getting closer with this approach.. only thing now is that some of the compliance tests are failing because I removed the fallback logic for setting empty bodies to have a content-type of application-json because now we don't always want it set. Will keep looking at it tomorrow

Baccata:
Good effort, but I'm sorry to say I'm gonna be closing this one in favour of that one : https://github.com/disneystreaming/smithy4s/pull/1094/files

PR #1094, opened 2023-07-18
SimpleRestJson empty optional payloads == null

### LINKED ISSUE #1072: 0.18: optional bodies and content-type

Baccata:
Closes https://github.com/disneystreaming/smithy4s/issues/1072.

Changes the logic so that the original json PayloadCodec compilers do not take the responsibility of turning `null` to `{}` as this is a protocol specific concern.

#### The case of SimpleRestJson

Moreover, in the context of SimpleRestJson (which literally has all payloads be json), it makes sense that an optional payload field set to empty should be serialised as json null. For instance, given this operation : 

```smithy
operation Foo {
  output: {
      @httpPayload 
      string: String
  } 
}
```

the only sensical translation for the case where `string == None` in scala is to serialise the value as json `null`. This is because the other side should rightfully expect for a payload to be present and for that payload to abide by `application/json`, since `simpleRestJson` aims at being a straightforward Json-in/Json-out RESTful protocol. 

Note that this case is different from an empty operation 

``` 
operation Foo {
} 
```

In this case, the `SimpleRestJson` protocol ought to consider that no-body should be serialised 

#### In the case of AWS Rest Json 

However, this PR ensures that the behaviour of turning `null` into `{}` in requests is kept for the AWS Rest Json protocol. Indeed, AWS seemingly doesn't care about having any coherence whatsoever, probably because of a desire to keep behaviour of their existing tools. 

I believe that this very situation is one of the few cases where a divergence from AWS is entirely justified, in particular because the behaviour is [unclear when the payload is a optional structure](https://github.com/smithy-lang/smithy/blob/main/smithy-aws-protocol-tests/model/restJson1/http-payload.smithy#L210-L270). 

If we ever get around to supporting AWSRestJson server-side, then we'll add a opt-in behaviour in the json codecs to allow for decoding empty options from `{}`.

### DIFF
--- a/modules/aws-http4s/src/smithy4s/aws/internals/AwsRestJsonCodecs.scala
+++ b/modules/aws-http4s/src/smithy4s/aws/internals/AwsRestJsonCodecs.scala
@@ -17,6 +17,7 @@
 package smithy4s.aws
 package internals
 
+import smithy4s.Blob
 import smithy4s.codecs._
 import smithy4s.http._
 import smithy4s.json.Json
@@ -52,8 +53,13 @@ private[aws] object AwsRestJsonCodecs {
           .andThen[HttpMediaReader](HttpMediaTyped.mediaTypeK(mediaType))
       }
 
+    def nullToEmptyObject(blob: Blob): Blob =
+      if (blob.sameBytesAs(Blob("null"))) Blob("{}") else blob
+
     val jsonMediaWriters = jsonPayloadCodecs.mapK {
-      PayloadCodec.writerK.andThen[HttpMediaWriter](HttpMediaTyped.mediaTypeK(mediaType))
+      PayloadCodec.writerK
+        .andThen[PayloadWriter](Writer.andThenK(nullToEmptyObject))
+        .andThen[HttpMediaWriter](HttpMediaTyped.mediaTypeK(mediaType))
     }
 
     val mediaReaders =
--- a/modules/aws-http4s/src/smithy4s/aws/internals/AwsRestJsonCodecs.scala
+++ b/modules/aws-http4s/src/smithy4s/aws/internals/AwsRestJsonCodecs.scala
@@ -54,7 +54,7 @@ private[aws] object AwsRestJsonCodecs {
       }
 
     def nullToEmptyObject(blob: Blob): Blob =
-      if (blob.sameBytesAs(Blob("null"))) Blob("{}") else blob
+      if (blob.sameBytesAs(Json.NullBlob)) Json.EmptyObjectBlob else blob
 
     val jsonMediaWriters = jsonPayloadCodecs.mapK {
       PayloadCodec.writerK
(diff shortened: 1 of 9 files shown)


modules/aws-http4s/src/smithy4s/aws/internals/AwsRestJsonCodecs.scala
```diff
@@ -52,8 +53,13 @@ private[aws] object AwsRestJsonCodecs {
           .andThen[HttpMediaReader](HttpMediaTyped.mediaTypeK(mediaType))
       }
 
+    def nullToEmptyObject(blob: Blob): Blob =
+      if (blob.sameBytesAs(Blob("null"))) Blob("{}") else blob
```
daddykotex: We could have a `Blob.NullString` in place of `Blob("null")`

daddykotex (approved):

modules/aws-http4s/src/smithy4s/aws/internals/AwsRestJsonCodecs.scala
Baccata: Well `Blob` ain't supposed to be biased towards Json but we can certainly have a `JsonNullBlob` constant in the Json module :+1:

Baccata: [1506755](https://github.com/disneystreaming/smithy4s/pull/1094/commits/15067558e2834867f1f18c50b0594ff1e9f83c5b)

Baccata:
@kubukoz this is gonna be the fix for your issue. It obviously deviates from previous behaviour, but in a way we can hopefully agree is sensical

kubukoz:
yep, it's sensical!

PR #1097, opened 2023-07-18
Builder classes for service and endpoint

### LINKED ISSUE #836: Create `Builder` data classes for `Service` and `Endpoint`
Context : https://github.com/disneystreaming/smithy4s/pull/832#issuecomment-1446002920

Let's create `Endpoint.Builder` and `Service.Builder` data classes that can be easily access from `Endpoint` and `Service`, that would facilitate the manipulation and transformation of `Endpoint` and `Service`

They should follow the [Scala best practice around binary compatibility](https://docs.scala-lang.org/overviews/core/binary-compatibility-for-library-authors.html#changing-a-case-class-definition-in-a-backwards-compatible-manner).  

We could enhance the UX of using builders by having `mapX` methods for every element.

GaryAghedo:
Closes #836.


modules/core/src/smithy4s/Service.scala
```diff
@@ -187,4 +187,66 @@ object Service {
...
+
+        (input, endpointMapper(baseEndpoint))
+      }
+
+      // todo: these should be easier ones
+      override val id: ShapeId = ???
+
+      override val version: String = ???
+
+      override val hints: Hints = ???
```
kubukoz: @GaryAghedo can you add methods for these? It'd be the usual, just like everything in the endpoint builder.

kubukoz (commented):
Thanks for kicking this off! I took your code as a baseline and implemented the endpoint transformation function in the service mapper. I'm not sure it's fully correct, we'll have to get @Baccata's take on this, but I think we're onto something.

We'll need some tests before this can be merged, there are probably also other things...

modules/core/src/smithy4s/Service.scala
GaryAghedo: nice!
Yes sure.. I will add the code now

modules/core/src/smithy4s/Endpoint.scala
```diff
@@ -66,4 +66,99 @@ object Endpoint {
...
+    def mapHints(f: Hints => Hints): Builder[I, E, O, SI, SO] =
+      copy(baseHints = f(baseHints))
+
+    def withInput(input: Schema[I]): Builder[I, E, O, SI, SO] =
+      copy(baseInput = input)
+
+    def mapInput(f: Schema[I] => Schema[I]): Builder[I, E, O, SI, SO] =
+      copy(baseInput = f(baseInput))
+
+    def withOutput(output: Schema[O]): Builder[I, E, O, SI, SO] =
```
kubukoz: **question:** should we allow using a schema of a different type (another `O`, in this case)? I'm thinking yes, because it's certainly possible.

However, when that's used in conjunction with a service builder, I think we'd be pretty limited here - not only because `PolyFunction5` forbids changing type parameters, but also I don't think we can change the operation type (which would be necessary to keep things like `def endpoint: (I, Endpoint)` working).

Baccata: I think there could eventually be two flavours of builders : 

1. One for the static usecase, that respects all the types, which probably means that the only possible transformations will be hints and labels 
2. One that would be more useful in the context of dynamic services, where we could amend both the types involved in operations, and the list of operations 

I believe this PR is only dealing with point 1 (which is totally fine). In this context, you shouldn't allow for different types.

modules/bootstrapped/test/src/smithy4s/builder/EndpointBuilderSpec.scala
```diff
@@ -0,0 +1,70 @@
+package smithy4s
```
kubukoz, line 1: **issue:** this should go to `core/test` - it has a dependency on `bootstrapped` so you should be able to move it pretty much directly

Baccata: actually it doesn't have a dependency and the test is right where it should be. Attempting to add a dependency from `core/Test` to `bootstrapped` created a cycle because of SBT's model which led the build to crash. 

`bootstrapped/Test` now holds all the tests that were previously in core/Test. Blame SBT.

kubukoz: oh. welp, I should've double-checked then :P

modules/bootstrapped/test/src/smithy4s/builder/EndpointBuilderSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+  ) {
+
+
+    val newEndpoint = builder
+      .mapId(shapeId => ShapeId.apply(shapeId.namespace, "getCityEndpoint"))
+      .mapHints { hints =>
+        hints.++(
+          Hints(
+            Hints.Binding.DynamicBinding.apply(ShapeId.apply("smithy.api", "documentation"), Document.fromString("new endpoint")))
+        )
```
kubukoz: **suggestion:** 
```suggestion
        hints ++ Hints(Documentation("new endpoint"))
```

with `Documentation` coming from the `smithy.api` package

modules/bootstrapped/test/src/smithy4s/builder/EndpointBuilderSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+      }
+      .mapInput(s => s.withId(ShapeId.apply("smithy4s.example", "inputSchema")))
+      .mapOutput(s => s.withId(ShapeId.apply("smithy4s.example", "outputSchema")))
```
kubukoz: **nitpick:** `ShapeId.apply` -> `ShapeId`

modules/bootstrapped/test/src/smithy4s/builder/EndpointBuilderSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+      .mapErrorable(_ => None)
+      .build
+
+    assertEquals(newEndpoint.id, ShapeId.apply("smithy4s.example", "getCityEndpoint"))
+    assertEquals( newEndpoint.hints,
```
kubukoz: **nitpick:** this will probably fail formatting

modules/bootstrapped/test/src/smithy4s/builder/EndpointBuilderSpec.scala
```diff
@@ -0,0 +1,58 @@
+package smithy4s
+
+import munit._
+
+class EndpointBuilderSpec extends FunSuite {
+
+  val endpoint = smithy4s.example.WeatherGen.endpoints.apply(2)
```
kubukoz: **suggestion:** you can access this particular endpoint with `smithy4s.example.WeatherOperation.GetForecast` rather tahn indexing

modules/bootstrapped/test/src/smithy4s/builder/EndpointBuilderSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+      .mapId(shapeId => ShapeId.apply(shapeId.namespace, "getCityEndpoint"))
+      .mapHints { hints =>
+        hints.++(
+          Hints(
+            Hints.Binding.DynamicBinding.apply(ShapeId.apply("smithy.api", "documentation"), Document.fromString("new endpoint")))
+        )
+      }
+      .mapInput(s => s.withId(ShapeId.apply("smithy4s.example", "inputSchema")))
+      .mapOutput(s => s.withId(ShapeId.apply("smithy4s.example", "outputSchema")))
+      .mapErrorable(_ => None)
```
kubukoz, line 48: **thought:** this could have some interesting implicitations if someone does that on an endpoint that has errors: I wonder what happens if they use the "surfaced errors" way of writing services and raise an error that way...

modules/bootstrapped/test/src/smithy4s/builder/ServiceBuilderSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  val builder =  smithy4s.Service.Builder.fromService(service)
+
+  test(
+    "can replace the following values (Id, Version and Hints) using withId, withVersion, withHints"
+  ) {
+
+    val newService = builder
+      .withId(ShapeId.apply("smithy4s.example", "newService"))
+      .withHints(Hints(Hints.Binding.DynamicBinding.apply(ShapeId.apply("smithy.api", "documentation"), Document.fromString("this is new Service"))))
```
kubukoz: **issue:** same suggestion as in the endpoint test, you can pass hints with less boilerplate

modules/bootstrapped/test/src/smithy4s/builder/ServiceBuilderSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+  test(
+    "can modify the following values (Id, Version and Hints) using mapId, mapVersion, mapHints"
+  ) {
+
+
+    val newService = builder
+      .mapId(shapeId => ShapeId.apply(shapeId.namespace, "myService"))
+      .mapHints{ hints  =>
+        hints.++(
+          Hints(Hints.Binding.DynamicBinding.apply(ShapeId.apply("smithy.api", "documentation2"), Document.fromString("new Service"))))
```
kubukoz: **suggestion:** 

```suggestion
          Hints(ShapeId.apply("smithy.api", "documentation2") -> Document.fromString("new Service")))
```

(there's an implicit conversion)

modules/bootstrapped/test/src/smithy4s/builder/ServiceBuilderSpec.scala
```diff
@@ -0,0 +1,72 @@
+package smithy4s
+
+
+import munit._
+import smithy4s.kinds.PolyFunction5
+import smithy4s.example.FooServiceGen
+
+class ServiceBuilderSpec extends FunSuite {
+
+  val service = smithy4s.example.FooService
```
kubukoz, line 10: **nitpick:** I would've liked to see a service with multiple endpoints used in this test - allows for covering more ground in the tests (e.g. you see that endpoints aren't all replaced with the same thing)

modules/core/src/smithy4s/Endpoint.scala
```diff
@@ -66,4 +66,103 @@ object Endpoint {
...
+
+  final case class Builder[Op[_, _, _, _, _], I, E, O, SI, SO] private (
+      private val baseId: ShapeId,
+      private val baseHints: Hints,
+      private val baseInput: Schema[I],
+      private val baseOutput: Schema[O],
+      private val baseErrorable: Option[Errorable[E]],
+      private val baseStreamedInput: StreamingSchema[SI],
+      private val baseStreamedOutput: StreamingSchema[SO],
+      private val baseWrap: I => Op[I, E, O, SI, SO]
```
kubukoz, line 81: **thought:** I didn't suggest adding the ability to customize `wrap` because, frankly, what can you do? other than side effects like printing the input, I don't see how else you might extend this. Just leaving this for other maintainers if they're wondering why `wrap` doesn't have builder functions.

modules/core/src/smithy4s/Service.scala
```diff
@@ -187,4 +187,84 @@ object Service {
...
+      private val endpointMapper: PolyFunction5[Endpoint.ForOperation[Op]#e, Endpoint.ForOperation[Op]#e],
+      private val baseId: ShapeId,
+      private val baseVersion: String,
+      private val baseHints: Hints,
+  ) {
+
+    def mapEndpointEach(
+        mapper: PolyFunction5[Endpoint.ForOperation[Op]#e, Endpoint.ForOperation[Op]#e]
+    ): Builder[Alg, Op] = copy(
+      endpointMapper = (endpointMapper.andThen(mapper)).unsafeCacheBy(
```
kubukoz: **thought:** this transformation is kind of lazy (only happens when you call `.build`), perhaps we should document that in its Scaladoc.

also, perhaps we should only cache it inside `build`, otherwise if you call `mapEndpointEach` multiple times you get multiple layers of caching.

modules/bootstrapped/test/src/smithy4s/builder/EndpointBuilderSpec.scala
kubukoz: I think it's worth a test in here, at least.

kubukoz: just checked, the behavior is:

- if you remove an errorable and the service raises an error, it behaves just like on any other Throwable
- adding errors to an existing errorable, or creating custom Alts in the errorable otherwise, seems impossible (legally) because `E` isn't known: so you can't biject an arbitrary value into it

modules/core/src/smithy4s/Service.scala
Baccata: NB : syncing with series `0.18`  will remove `unsafeCacheBy`, but it's okay because the exposure of ordinals should remove the need to cache.

modules/bootstrapped/test/src/smithy4s/builder/EndpointBuilderSpec.scala
GaryAghedo: https://github.com/disneystreaming/smithy4s/pull/1097/commits/748de5300df1a328d5c8eb09dcf1c613783e7b70

modules/bootstrapped/test/src/smithy4s/builder/EndpointBuilderSpec.scala
GaryAghedo: https://github.com/disneystreaming/smithy4s/pull/1097/commits/6346866132e5c7ce91ea50417345b6d6488ab705

modules/bootstrapped/test/src/smithy4s/builder/EndpointBuilderSpec.scala
GaryAghedo: https://github.com/disneystreaming/smithy4s/pull/1097/commits/6346866132e5c7ce91ea50417345b6d6488ab705

modules/bootstrapped/test/src/smithy4s/builder/ServiceBuilderSpec.scala
GaryAghedo: https://github.com/disneystreaming/smithy4s/pull/1097/commits/8ffcacfa0cc5a7e2446723e3efd0e8e165f8feb1

modules/bootstrapped/test/src/smithy4s/builder/ServiceBuilderSpec.scala
GaryAghedo: https://github.com/disneystreaming/smithy4s/pull/1097/commits/a01028fa70381c2d8b7db7a0bfb4741f1b306421

modules/bootstrapped/test/src/smithy4s/builder/EndpointBuilderSpec.scala
GaryAghedo: https://github.com/disneystreaming/smithy4s/pull/1097/commits/e0ebde221291632b48a0f58372ef5eeb8dbf600c

modules/bootstrapped/test/src/smithy4s/builder/EndpointBuilderSpec.scala
GaryAghedo: https://github.com/disneystreaming/smithy4s/pull/1097/commits/f31a91264de502e1a1e2b721bcd61c746f60b541

modules/core/src/smithy4s/Endpoint.scala
GaryAghedo: if customization of the wrap is not available, we must offer an alternative solution that involves providing an implementation. Its seems unavoidable in this case

kubukoz: I had a thought the other day, now I'm thinking `wrap` could actually be customizable on an individual endpoint level. so, `withWrap` and `mapWrap` would still be doable.

I think we can leave them out of this PR though, and make a separate ticket.

kubukoz: https://github.com/disneystreaming/smithy4s/issues/1152

modules/core/src/smithy4s/Service.scala
```diff
@@ -196,4 +196,82 @@ object Service {
...
+
+  final case class Builder[Alg[_[_, _, _, _, _]], Op[_, _, _, _, _]] private(
+      private val base: Service.Aux[Alg, Op],
+      private val endpointMapper: PolyFunction5[Endpoint.ForOperation[Op]#e, Endpoint.ForOperation[Op]#e],
+      private val baseId: ShapeId,
+      private val baseVersion: String,
+      private val baseHints: Hints,
+  ) {
+
+    def mapEndpointEach(
```
kubukoz, line 218: **thought:** lost this comment somewhere, but I had some doubts about this method: it's lazy (`mapper` doesn't get called until you call `build`), which could be surprising to users. However, mapping the endpoints eagerly proved to be a bit difficult last time I tried to do it, so I don't have a great suggestion either.

Only one: let's document the lazy behavior: sth like

```scala
/**
  Maps each endpoint in the service with the given polymorphic function. Note: unlike other `map*` methods here, this one is lazy and non-memoized - the mapping is only performed when you call `build`, every time you call it.
```

modules/core/src/smithy4s/Service.scala
```diff
@@ -196,4 +196,82 @@ object Service {
...
+      copy(baseHints = hints)
+
+    def mapHints(f: Hints => Hints): Builder[Alg, Op] =
+      copy(baseHints = f(baseHints))
+
+    def build: Service.Aux[Alg, Op] = new Service[Alg] {
+
+      override type Operation[I, E, O, SI, SO] = Op[I, E, O, SI, SO]
+
+      override def endpoints: IndexedSeq[Endpoint[_, _, _, _, _]] =
```
kubukoz: **nitpick:** this could probably be a `val`

build.sbt
```diff
@@ -671,6 +671,9 @@ lazy val http4s = projectMatrix
         Dependencies.Alloy.`protocol-tests` % Test
       )
     },
+    Compile / allowedNamespaces := Seq(
+      "smithy4s.example.guides.auth"
+    ),
```
kubukoz: **nitpick:** I thought `bootstrapped` was the only module that generates code, but apparently not.

however, this should probably be scoped to `Test` instead of `Compile`, just like the other codegen-specific settings here.

kubukoz (approved):

modules/core/src/smithy4s/Service.scala
kubukoz: it did, because we got rid of `def endpoint`. We'll defer to the default impl, unless you prefer otherwise.

modules/core/src/smithy4s/Service.scala
kubukoz: this is outdated now, the transformation is no longer lazy.

kubukoz:
@Baccata can you take another look?

Baccata (approved):
Thanks !

PR #1098, opened 2023-07-19
Allow throttled webjars update PRs

### LINKED ISSUE #1069: Outdated webjars-locator version with security vulnerabilities due to Scalasteward ignore
The current version of org.webjars:webjars-locator used relies on com.fasterxml.jackson.core:jackson-databind version 2.12.5, which [has a security vulnerability](https://nvd.nist.gov/vuln/detail/CVE-2022-42003) that was fixed in later versions. I'm fairly doubtful that this could actually be exploited, but it's being flagged by our security tooling.

The latest version of webjars-locator is 0.47, which uses jackson-databind 2.15.2 . Upgrading would solve the issue, but updates aren't picked up by scala-steward because it's been [explicitly set to ignore webjars](https://github.com/disneystreaming/smithy4s/blob/c179457ba49b5c012d9b3fdf0b0cb40618f07592/.scala-steward.conf#L3).

I'm not sure if you have some specific reason not to want automatic update PRs for webjars, but getting it to a version that has the patch would be very helpful for us!

ybasket:
Let Scala Steward sometimes send an update PR for the webjars so security findings are mitigated.

Closes #1069

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -1,8 +1,11 @@
 updates.ignore = [
   { groupId = "org.scalameta" },
-  { groupId = "org.webjars" },
-  { groupId = "org.webjars.npm"},
   { groupId = "software.amazon.awssdk" }
+]
+
+dependencyOverrides = [
+  { pullRequests = { frequency = "@monthly" }, dependency = { groupId = "org.webjars" }},
+  { pullRequests = { frequency = "@monthly" }, dependency = { groupId = "org.webjars.npm" }}
 ]
 
 updates.limit = 4


ybasket:
`series/0.18` seems to be the default branch, but maybe this should better go to another one? 🤷 
Would be cool to get a 0.17 release with updated webjars if easily possible.

Baccata:
thanks !

daddykotex:
@ybasket I can definitely back manually update the webjar dependency and backport it to 0.17

Baccata:
@daddykotex that'd be great I think :D

PR #1102, opened 2023-07-19
Add mechanism for granular bloop-config generation 

kubukoz:
@Baccata's edit : 

Adds a mechanism for granularly enabling bloop-config generation, defaulting to the Scala 2.13 / JVM axes. 

In order to change  the default, one can create a `user.sbt` in the top directory of the project and tweak the `bloopAllowedCombos` setting, as such : 

```scala 
ThisBuild / bloopAllowedCombos := Seq(
  Seq(
    VirtualAxis.jvm,
    VirtualAxis.scalaABIVersion("3.3.0")
  )
)
```

<del>
Usage:

1. Close metals
1. Remove `.bloop/`
1. Determine which Scala version you want to work on. For example, 2.13 on the JVM
1. Run the appropriate command: `sbt -Dbloop.export-jar-classifiers=sources bloopGenerate_2_13_jvm`
1. Open metals in the project, DON'T import the build.
1. Wait for indexing and use Metals

Got the idea from here: https://scalacenter.github.io/bloop/docs/build-tools/sbt#dont-export-certain-targets
</del>


daddykotex:
I knew something like that was possible. Thanks for finding it out why.

We had discussed that the other day with @lewisjkl  as he has some issue with its setup when working in smithy4s

kubukoz:
my main issue was... it's slow as heck to set up (because of all the cells in the matrix), and you compile everything like 7 times (same reason).

Baccata:
Honestly I think I'd rather disable bloop generation by default for anything that's not JVM/2.13 (or 2.12 for the SBT plugin and its dependencies). That'd be less of a barrier of entry for contributors

Baccata:
> Honestly I think I'd rather disable bloop generation by default for anything that's not JVM/2.13 (or 2.12 for the SBT plugin and its dependencies). That'd be less of a barrier of entry for contributors

To elaborate : let's keep what you've added here, but let's also set the settings we want by default

Baccata:
@kubukoz apologies, I was meaning to branch-out, but mis-happed and pushed to your branch. Anyway, I can revert and branch out, but what do you think of this https://github.com/disneystreaming/smithy4s/pull/1102/commits/df8e0fefca7cbfbf50025e7892e31c1f85e60d02 ?

kubukoz:
no worries, I'm actually glad you pushed here :P 

looks good, but what you we do when you want to work on other versions?

Baccata:
> looks good, but what you we do when you want to work on other versions? 

you just don't 😄 

Joke aside. I guess "bloopEnabled" could be derived from something else that could be more easily tweaked project wise. Let's ask @keynmol and @armanbilge if they have some ideas

armanbilge:
We recently added a similar feature to sbt-typelevel.

- https://github.com/typelevel/sbt-typelevel/pull/465#issuecomment-1598120437

We define a setting with a reasonable global default and expect that a dev can modify the setting in a `user.sbt` in case they need something different.

README.md
```diff
@@ -46,3 +46,19 @@ If you don't have Flakes support:
 ```bash
 nix-shell
 ```
+
+## Note for metals/bloop users
```
kubukoz, line 50: **thought:** this and the above contributing notes should eventually go to a `CONTRIBUTING` file. One Day!

kubukoz (commented):
✅ 

I trust in this not blowing my machine up more than the current state

Baccata (approved):
Approving my own changes :p

PR #1103, opened 2023-07-19
Optics Rendering

lewisjkl:
Adds the ability to render optics (Prisms and Lenses) in the generated code. The idea is that there will be very basic optic capabilities built into the smithy4s.core and then if users want something more advanced they can use implicit conversions to use something like monocle (see docs in the PR for more info).

Currently Lens instances are generated for structures and Prisms are generated for Unions and Enums.

For newtypes, I added a `.value` function on prisms/lenses that uses the bijection to automatically create a new lens/prism that targets the inner type of the newtype. I opted for this because then users can choose to create lenses/prisms that target the newtype or the inner type depending on their preferences. Lmk if you think there is a cleaner approach though.


modules/core/src/smithy4s/optics/Prism.scala
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+object Prism {
+
+  /**
+   * Construct a new [[Prism]] by providing functions for getting
+   * an Option[A] from S and getting an S given an A.
+   */
+  def apply[S, A](_get: S => Option[A])(_project: A => S): Prism[S, A] =
```
Baccata, line 88: `project` needs to be called `inject`

modules/core/src/smithy4s/optics/Prism.scala
```diff
@@ -0,0 +1,100 @@
...
+trait Prism[S, A] extends Optional[S, A] { self =>
+
+  /**
+   * Returns a [[Some]] of A from S if it is able to obtain an A.
+   * Else returns [[None]].
+   */
+  def getOption(s: S): Option[A]
+
+  /** Returns an S given an A */
+  def project(a: A): S
```
Baccata: project needs to be called `inject`

modules/core/src/smithy4s/optics/Prism.scala
```diff
@@ -0,0 +1,100 @@
...
+ * Lens implementation which can be used to abstract over accessing/updating
+ * a member of a coproduct type
+ */
+trait Prism[S, A] extends Optional[S, A] { self =>
+
+  /**
+   * Returns a [[Some]] of A from S if it is able to obtain an A.
+   * Else returns [[None]].
+   */
+  def getOption(s: S): Option[A]
```
Baccata: this can be called `project`

Baccata:
@lewisjkl, in the view of the size of the diff, I think it'd be worth having two different ways of enabling optics : 

1. one trait-driven 
2. one metadata-driven 

The inlined samples would only use the trait-driven one on a select few shapes (a struct, a union, an enum, etc) to showcase the feature

modules/core/src/smithy4s/optics/Lens.scala
```diff
@@ -0,0 +1,95 @@
...
+        that.get(self.get(s))
+      def replace(a: A0): S => S =
+        self.modify(that.replace(a))
+    }
+
+  /** 
+   * Allows abstracting over an optional target by pointing to 
+   * the inside of the optional value (the value inside of the [[Some]]).
+   */
+  final override def some[A0](implicit
```
Baccata, line 60: 👍

modules/core/src/smithy4s/optics/Lens.scala
```diff
@@ -0,0 +1,95 @@
...
+  private[this] final def adapt[A0](implicit
+      evA: A =:= A0
+  ): Lens[S, A0] =
+    evA.substituteCo[Lens[S, *]](this)
+
+  /**
+   * Helper function for targeting the value inside of a [[smithy4s.Newtype]]
+   * or other type with an implicit [[Bijection]] available.
+   */
+  final def value[A0](implicit bijection: Bijection[A0, A]): Lens[S, A0] =
```
Baccata: :+1:

modules/core/src/smithy4s/optics/Lens.scala
```diff
@@ -0,0 +1,95 @@
...
+
+import smithy4s.Bijection
+
+// inspired by and adapted from https://www.optics.dev/Monocle/ under the MIT license
+
+/**
+ * Lens implementation which can be used to abstract over accessing/updating
+ * a member of a product type
+ */
+trait Lens[S, A] extends Optional[S, A] { self =>
```
Baccata, line 27: 👍

modules/bootstrapped/src/generated/smithy4s/example/AddMenuItemRequest.scala
```diff
@@ -15,8 +15,8 @@ object AddMenuItemRequest extends ShapeTag.Companion[AddMenuItemRequest] {
   val hints: Hints = Hints.empty
 
   object Optics {
-    val restaurant = Lens[AddMenuItemRequest, String](_.restaurant)(n => a => a.copy(restaurant = n))
-    val menuItem = Lens[AddMenuItemRequest, MenuItem](_.menuItem)(n => a => a.copy(menuItem = n))
+    val restaurantLens = Lens[AddMenuItemRequest, String](_.restaurant)(n => a => a.copy(restaurant = n))
```
Baccata: is it necessary to have the `Lens` suffix ?

lewisjkl: Scala 2.12 was complaining that it couldn't tell if the name inside of the .copy(myName = ???) was the name of the parameter of the case class or the name of the val.. not sure if there is another way around this or not, but that's why I added the suffix to disambiguate

kubukoz: strange, this seems to compile for me:

```scala
//> using scala "2.12.18"
trait Lens[A, B]

object Lens {
  def apply[A, B](f: A => B)(g: (A, B) => A): Lens[A, B] = ???
}

case class Hello(name: String)

object Hello {

  val name: Lens[Hello, String] = Lens[Hello, String](_.name)((c, n) => c.copy(name = n))
}
```

lewisjkl: Ahh the issue is that 2.12 needs the type annotation. The following fails:

```scala
//> using scala 2.12

trait Lens[A, B]

object Lens {
  def apply[A, B](f: A => B)(g: B => A => A): Lens[A, B] = ???
}

case class Hello(name: String)

object Hello {
  val name  = Lens[Hello, String](_.name)(n => c => c.copy(name = n))
}
```

Baccata:
@kubukoz, Jeff and I tried to unify the rendering of Lenses/Prisms with Fields/Alts. Our conclusion is that trying for this unification requires way to many changes for little benefit, so we're gonna roll with this current PR, after Jeff syncs it with the latest 0.18 

If you're curious about our attempt to unify, here's a diff (it introduces a bug that makes 4 aws-related tests fail, not gonna bother attempting to fix)  :https://github.com/disneystreaming/smithy4s/compare/series/0.18...optics-in-fields-and-alts

kubukoz:
I appreciate the info!

Baccata (approved):

Issue #1070, opened 2023-07-04
Errors without stack trace


msinton:
Is it possible to create errors that extend `NoStackTrace` ?

This will be an efficiency and cost saving measure if we can remove the stack trace from business errors

daddykotex:
Hi @msinton , thanks for the issue

To be clear, do you mean that you'd like to have the errors you define in your Smithy spec to be code generated as Throwable without a stack trace? For example [this NotFoundError generated](https://github.com/disneystreaming/smithy4s/blob/series/0.18/modules/bootstrapped/src/generated/smithy4s/example/ServerError.scala#L10) for the [error defined in this spec](https://github.com/disneystreaming/smithy4s/blob/series/0.18/sampleSpecs/example.smithy#L134).

I'm a fan of stack traces, so if we were to do such a thing, it would probably be opt-in via a `meta` annotations, like [adtMember](https://github.com/disneystreaming/smithy4s/blob/series/0.18/modules/protocol/resources/META-INF/smithy/smithy4s.meta.smithy#L26) that you'd add on the error structures

msinton:
Hi @daddykotex 

Yes, there are many cases where the stack trace adds no value for debugging because it is generated by validation for example. If we can remove the stack trace at source then it will also not be logged and will not percolate through the system.

I think opt-in is a good idea

yisraelU:
@msinton Can this be closed ?

Baccata:
Closing : this feature has been merged and will be released in version 0.18 which should happen sometime in the next couple months.

PR #1104, opened 2023-07-20
Generate NoStackTrace error types

yisraelU:
Not sure if this is a great idea or too specific of a Scala idiom . Inspired by #1070 
- Adds a smithy trait `@noStackTrace` to annotate structures that are annotated with the  `@error` trait.
- This influences the  codegen that error data types extend `scala.util.control.NoStackTrace` instead of directly `Throwable`


Baccata:
@yisraelU I'm cool with the idea. Can you sync with the latest 0.18 and we'll merge it

Baccata (approved):

PR #1105, opened 2023-07-20
Minor refactors for 0.18

daddykotex:
Two changes, one of which is not strictly necessary:

1. visibility of metadata functions in the http4s-kernel
2. make the client codecs one level above to avoid dragging endpoint further down in the implementation

I have another one [that add a way to merge two readerwriter](https://github.com/disneystreaming/smithy4s/commit/2e6dbc5dc6e3298afcb6cd178e65e96b4b7a7470) but I'm not sure if you would want that or something more generic

### DIFF
--- a/modules/http4s-kernel/src/smithy4s/http4s/kernel/package.scala
+++ b/modules/http4s-kernel/src/smithy4s/http4s/kernel/package.scala
@@ -108,7 +108,7 @@ package object kernel {
       .groupBy(_._1)
       .map { case (k, v) => k -> v.map(_._2).toList }
 
-  private[smithy4s] def getRequestMetadata[F[_]](
+  def getRequestMetadata[F[_]](
       pathParams: PathParams,
       request: Request[F]
   ): Metadata =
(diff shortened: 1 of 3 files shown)


Baccata (approved):

modules/http4s-kernel/src/smithy4s/http4s/kernel/package.scala
```diff
@@ -108,7 +108,7 @@ package object kernel {
       .groupBy(_._1)
       .map { case (k, v) => k -> v.map(_._2).toList }
 
-  private[smithy4s] def getRequestMetadata[F[_]](
+  def getRequestMetadata[F[_]](
```
kubukoz, line 111: **praise:** thanks, I'll be able to remove my copies of these ;)

kubukoz (approved):

modules/http4s-kernel/src/smithy4s/http4s/kernel/package.scala
daddykotex: hehe

PR #1110, opened 2023-07-20
Flatten AWS newtypes

denisrosca:
AWS in their smithy specs have a tendency to create newtypes for the smithy standard shapes. As part of this change, introduce a projection transformer that will run as a preprocessor before code generation and will replace the newtypes (if they match some criteria) with references to the standard smithy shapes.

For a newtype to be considered as being "replaceable" it has to match the following criteria:
- doesn't have any traits (other than @default)
- is a simple shape
- the name matches the smithy shape (one exception is Date, as it will be replaced with Timestamp even though it doesn't actually match).
- comes from an AWS namespace (i.e. starts with com.amazonaws)

### DIFF
--- a/modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/build.sbt
+++ b/modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/build.sbt
@@ -2,16 +2,8 @@ lazy val root = (project in file("."))
   .enablePlugins(Smithy4sCodegenPlugin)
   .settings(
     scalaVersion := "2.13.10",
-    Compile / smithy4sAllowedNamespaces := List(
-      "aws.api",
-      "aws.auth",
-      "aws.customizations",
-      "aws.protocols",
-      "com.amazonaws.dynamodb",
-    ),
     libraryDependencies ++= Seq(
-      "com.disneystreaming.smithy4s" %% "smithy4s-core" % smithy4sVersion.value,
-      "com.disneystreaming.smithy"   % "aws-dynamodb-spec" % "2023.02.10"
+      "com.disneystreaming.smithy4s" %% "smithy4s-aws-kernel" % smithy4sVersion.value
     ),
     Compile / smithy4sOutputDir := baseDirectory.value / "smithy_output"
   )
--- a/modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/build.sbt
+++ b/modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/build.sbt
@@ -1,0 +1,17 @@
+lazy val root = (project in file("."))
+  .enablePlugins(Smithy4sCodegenPlugin)
+  .settings(
+    scalaVersion := "2.13.10",
+    Compile / smithy4sAllowedNamespaces := List(
+      "aws.api",
+      "aws.auth",
+      "aws.customizations",
+      "aws.protocols",
+      "com.amazonaws.dynamodb",
+    ),
+    libraryDependencies ++= Seq(
+      "com.disneystreaming.smithy4s" %% "smithy4s-core" % smithy4sVersion.value,
+      "com.disneystreaming.smithy"   % "aws-dynamodb-spec" % "2023.02.10"
+    ),
+    Compile / smithy4sOutputDir := baseDirectory.value / "smithy_output"
+  )
--- a/modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/src/main/scala/Main.scala
+++ b/modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/src/main/scala/Main.scala
@@ -1,0 +1,20 @@
+
+import com.amazonaws.dynamodb.ArchivalSummary
+import com.amazonaws.dynamodb.BackupArn
+import smithy4s.Timestamp
+
+object Main extends App {
+
+  val archivalSummary = new ArchivalSummary(
+    // archivalDateTime needs to be rewritten from a com.amazonaws.dynamodb.Date to a simple timestamp
+    archivalDateTime = Some(Timestamp.fromEpochSecond(java.time.Instant.now().getEpochSecond)),
+
+    // archivalReason needs to be rewritten from com.amazonaws.dynamodb.ArchivalReason to a String
+    archivalReason = Some("This is just a string"),
+
+    // archivalBackupArn remains as a newtype because com.amazonaws.dynamodb.BackupArn
+    // has the @length trait applied
+    archivalBackupArn = Some(BackupArn("this-is-a-back-up-arn-with-a-minimum-length"))
+  )
+
+}
--- a/modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/src/main/scala/Main.scala
+++ b/modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/src/main/scala/Main.scala
@@ -1,20 +1,32 @@
-
-import com.amazonaws.dynamodb.ArchivalSummary
-import com.amazonaws.dynamodb.BackupArn
-import smithy4s.Timestamp
+import com.amazonaws.dynamodb.DescribeEndpointsResponse
+import com.amazonaws.dynamodb.Endpoint
+import com.amazonaws.dynamodb.ListTablesInput
+import com.amazonaws.dynamodb.ListTablesInputLimit
+import com.amazonaws.dynamodb.TableName
 
 object Main extends App {
 
-  val archivalSummary = new ArchivalSummary(
-    // archivalDateTime needs to be rewritten from a com.amazonaws.dynamodb.Date to a simple timestamp
-    archivalDateTime = Some(Timestamp.fromEpochSecond(java.time.Instant.now().getEpochSecond)),
+  val listTablesInput = new ListTablesInput(
+    // TableName is not flattened because it has
+    // @length and @pattern traits applied
+    exclusiveStartTableName = Some(TableName("example-table-name")),
 
-    // archivalReason needs to be rewritten from com.amazonaws.dynamodb.ArchivalReason to a String
-    archivalReason = Some("This is just a string"),
-
-    // archivalBackupArn remains as a newtype because com.amazonaws.dynamodb.BackupArn
-    // has the @length trait applied
-    archivalBackupArn = Some(BackupArn("this-is-a-back-up-arn-with-a-minimum-length"))
+    // ListTablesInputLimit is not flattened because it has
+    // a @range constraint
+    limit = Some(ListTablesInputLimit(5))
   )
+
+  val describeEndpointsResponse = List(
+    Endpoint(
+      // Endpoint address was flattened from a com.amazonaws.dynamodb#String
+      // to a standard String shape
+      address = "example-endpoint",
+
+      // CachePeriodInMinutes was flattened from a com.amazonaws.dynamodb#Long
+      // to a standard long
+      cachePeriodInMinutes = 5L
+    )
+  )
+
 
 }
--- a/modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/src/main/scala/Main.scala
+++ b/modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/src/main/scala/Main.scala
@@ -1,3 +1,19 @@
+/*
+ *  Copyright 2021-2022 Disney Streaming
+ *
+ *  Licensed under the Tomorrow Open Source Technology License, Version 1.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     https://disneystreaming.github.io/TOST-1.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
 import com.amazonaws.dynamodb.DescribeEndpointsResponse
 import com.amazonaws.dynamodb.Endpoint
 import com.amazonaws.dynamodb.ListTablesInput
--- a/modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/test
+++ b/modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/test
@@ -1,8 +1,8 @@
 # check if smithy4sCodegen works
 > compile
 
-$ exists  smithy_output/com/amazonaws/dynamodb/ArchivalSummary.scala
+# This new type was not flattened because it has range constraints
+$ exists  smithy_output/com/amazonaws/dynamodb/ListTablesInputLimit.scala
 
-# This was a type alias to string `string ArchivalReason` and should be removed
-# by the AWS newtype flattener
--$ exists smithy_output/com/amazonaws/dynamodb/ArchivalReason.scala
+# Flattened and removed by the projection transformer
+-$ exists smithy_output/com/amazonaws/dynamodb/Long.scala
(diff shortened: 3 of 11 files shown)


modules/codegen/src/smithy4s/codegen/internals/CodegenImpl.scala
```diff
@@ -34,7 +35,7 @@ private[codegen] object CodegenImpl { self =>
       args.specs.map(_.toIO).toSet,
       args.dependencies,
       args.repositories,
-      args.transformers,
+      withAwsTypeTransformer(args.transformers),
```
denisrosca, line 38: Not sure if this is the best place to put this "default" transformer, but thought that it's a good place to start.

Open to suggestions on how to improve this.

modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/src/main/scala/Main.scala
```diff
@@ -0,0 +1,20 @@
+
```
denisrosca, line 16: If we decide that this test is good enough, I'll add a similar one to the mill plugin

modules/codegen/src/smithy4s/codegen/internals/CodegenImpl.scala
```diff
@@ -148,12 +149,15 @@ private[codegen] object CodegenImpl { self =>
...
```
denisrosca, line 152: same here

modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/build.sbt
```diff
@@ -0,0 +1,17 @@
+lazy val root = (project in file("."))
+  .enablePlugins(Smithy4sCodegenPlugin)
+  .settings(
+    scalaVersion := "2.13.10",
+    Compile / smithy4sAllowedNamespaces := List(
```
kubukoz: **question:** can we just omit this and generate everything in scope?

modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/src/main/scala/Main.scala
kubukoz: looks good enough to me

modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/test
```diff
@@ -0,0 +1,8 @@
+# check if smithy4sCodegen works
+> compile
+
+$ exists  smithy_output/com/amazonaws/dynamodb/ArchivalSummary.scala
+
+# This was a type alias to string `string ArchivalReason` and should be removed
+# by the AWS newtype flattener
+-$ exists smithy_output/com/amazonaws/dynamodb/ArchivalReason.scala
```
kubukoz: **praise:** good stuff

modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/src/main/scala/Main.scala
```diff
@@ -0,0 +1,20 @@
...
+  val archivalSummary = new ArchivalSummary(
+    // archivalDateTime needs to be rewritten from a com.amazonaws.dynamodb.Date to a simple timestamp
+    archivalDateTime = Some(Timestamp.fromEpochSecond(java.time.Instant.now().getEpochSecond)),
+
+    // archivalReason needs to be rewritten from com.amazonaws.dynamodb.ArchivalReason to a String
+    archivalReason = Some("This is just a string"),
+
+    // archivalBackupArn remains as a newtype because com.amazonaws.dynamodb.BackupArn
+    // has the @length trait applied
```
kubukoz: **thought:** perhaps we should flatten these too but add the constraints to the members? @baccata

modules/codegen/src/smithy4s/codegen/internals/CodegenImpl.scala
kubukoz: hmmmm so this makes sense to me because it'd also work for the CLI with no sbt/mill plugin...

but there's no way to remove this transformation, unlike if it was part of the default value:

```diff
diff --git a/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala b/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
index 3e375184..4d496b13 100644
--- a/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
+++ b/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
@@ -16,6 +16,7 @@
 
 package smithy4s.codegen
 
+import smithy4s.codegen.transformers.AwsStandardTypesTransformer
 import sbt.Keys._
 import sbt.util.CacheImplicits._
 import sbt.{fileJsonFormatter => _, _}
@@ -255,7 +256,9 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
     ),
     config / cleanFiles += (config / smithy4sOutputDir).value,
     config / cleanFiles += (config / smithy4sResourceDir).value,
-    config / smithy4sModelTransformers := List.empty,
+    config / smithy4sModelTransformers := List(
+      AwsStandardTypesTransformer.name
+    ),
     config / packageBin / packageOptions += {
       // This piece of logic aims at tracking the dependencies that Smithy4s used to generate
       // code at build time, in the manifest of the jar. This helps automatically pulling
diff --git a/modules/codegen/src/smithy4s/codegen/internals/CodegenImpl.scala b/modules/codegen/src/smithy4s/codegen/internals/CodegenImpl.scala
index 9358e07c..4e77ae67 100644
--- a/modules/codegen/src/smithy4s/codegen/internals/CodegenImpl.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/CodegenImpl.scala
@@ -18,7 +18,6 @@ package smithy4s.codegen
 package internals
 
 import alloy.openapi._
-import smithy4s.codegen.transformers.AwsStandardTypesTransformer
 import software.amazon.smithy.model.Model
 
 import scala.jdk.CollectionConverters._
@@ -35,7 +34,7 @@ private[codegen] object CodegenImpl { self =>
       args.specs.map(_.toIO).toSet,
       args.dependencies,
       args.repositories,
-      withAwsTypeTransformer(args.transformers),
+      args.transformers,
       args.discoverModels,
       args.localJars
     )
@@ -149,7 +148,7 @@ private[codegen] object CodegenImpl { self =>
       args.specs.map(_.toIO).toSet,
       args.dependencies,
       args.repositories,
-      withAwsTypeTransformer(args.transformers),
+      args.transformers,
       discoverModels = false,
       args.localJars
     )
@@ -157,7 +156,4 @@ private[codegen] object CodegenImpl { self =>
     Node.prettyPrintJson(ModelSerializer.builder().build.serialize(model))
   }
 
-  private def withAwsTypeTransformer(transformers: List[String]): List[String] =
-    transformers :+ AwsStandardTypesTransformer.name
-
 }

```

so originally I wanted to do that second thing, but now I don't know anymore 😬 

perhaps we can add another set of flags for removing default transformations.

modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/test
```diff
@@ -0,0 +1,8 @@
+# check if smithy4sCodegen works
+> compile
```
Baccata, line 2: ~~let's change it to only run the source generation : the dynamodb specs are huge and compiling just for the sake of checking whether the file exists is detrimental to build times~~ 

revised my judgement. HOWEVER, rather than pulling the full dynamodb spec, let's inline a minimalistic mock spec so that we don't compile the full dynamodb thing. We do this in the `bootstrapped` module : https://github.com/disneystreaming/smithy4s/blob/series/0.18/sampleSpecs/lightweight-dynamo.json

modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/src/main/scala/Main.scala
Baccata: I think the newtype is justified when a constraint is present.

modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/build.sbt
```diff
@@ -0,0 +1,17 @@
...
+      "aws.api",
+      "aws.auth",
+      "aws.customizations",
+      "aws.protocols",
+      "com.amazonaws.dynamodb",
+    ),
+    libraryDependencies ++= Seq(
+      "com.disneystreaming.smithy4s" %% "smithy4s-core" % smithy4sVersion.value,
+      "com.disneystreaming.smithy"   % "aws-dynamodb-spec" % "2023.02.10"
```
Baccata: see my comment [here](https://github.com/disneystreaming/smithy4s/pull/1110/files#r1270308761). Compiling the full dynamodb generated code is detrimental to build times.

modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/build.sbt
Baccata: if you add the `smithy4s-aws-kernel` library dependency (instead of core), then yes. We treat `aws*` as a [reserved namespace](https://github.com/disneystreaming/smithy4s/blob/series/0.18/modules/codegen/src/smithy4s/codegen/internals/CodegenImpl.scala#L126) and disable its generation by default.

modules/codegen/src/smithy4s/codegen/internals/CodegenImpl.scala
Baccata: Honestly I think it's fine to hardcode it, considering what the transformation does. Let's not overthink it until somebody actually has a problem with it.

modules/codegen/src/smithy4s/codegen/transformers/AwsStandardTypesTransformer.scala
```diff
@@ -0,0 +1,99 @@
...
+      .flatMap[Shape](shape => {
+        val replacementShape = replaceWith(shape.getTarget)
+        val target = model.expectShape(shape.getTarget)
+
+        val builder = MemberShape
+          .builder()
+          .id(shape.getId)
+          .source(shape.getSourceLocation)
+          .target(replacementShape)
+          .copyDefaultTrait(target, replacementShape)
```
Baccata: this is flawed as it doesn't keep the traits of the original member. Also there's a simpler way to achieve what you're trying to do. Here's a patch you can apply : 

```patch
diff --git a/modules/codegen/src/smithy4s/codegen/transformers/AwsStandardTypesTransformer.scala b/modules/codegen/src/smithy4s/codegen/transformers/AwsStandardTypesTransformer.scala
index 3536db21..7b41dda0 100644
--- a/modules/codegen/src/smithy4s/codegen/transformers/AwsStandardTypesTransformer.scala
+++ b/modules/codegen/src/smithy4s/codegen/transformers/AwsStandardTypesTransformer.scala
@@ -22,11 +22,15 @@ class AwsStandardTypesTransformer extends ProjectionTransformer {
     )
   }
 
-  private def mapFunction(model: Model): java.util.function.Function[Shape, Shape] = shape => {
+  private def mapFunction(
+      model: Model
+  ): java.util.function.Function[Shape, Shape] = shape => {
     def replaceWith(shapeId: ShapeId): ShapeId = {
       val shapeType =
         model
-          .expectShape(shapeId) // this should be safe as we already did some filtering before calling the method
+          .expectShape(
+            shapeId
+          ) // this should be safe as we already did some filtering before calling the method
           .getType
 
       // Use the shape type instead of relying on the name
@@ -37,25 +41,22 @@ class AwsStandardTypesTransformer extends ProjectionTransformer {
       ShapeId.fromParts(smithyStandardNamespace, shapeName)
     }
 
-    shape.asMemberShape()
-      .filter{ memberShape =>
-        model.getShape(memberShape.getTarget)
+    shape
+      .asMemberShape()
+      .filter { memberShape =>
+        model
+          .getShape(memberShape.getTarget)
           .map[Boolean](canReplace)
           .orElse(false)
       }
-      .flatMap[Shape](shape => {
-        val replacementShape = replaceWith(shape.getTarget)
+      .map[Shape] { shape =>
         val target = model.expectShape(shape.getTarget)
-
-        val builder = MemberShape
-          .builder()
-          .id(shape.getId)
-          .source(shape.getSourceLocation)
+        val replacementShape = replaceWith(shape.getTarget)
+        shape.toBuilder
           .target(replacementShape)
-          .copyDefaultTrait(target, replacementShape)
-
-        java.util.Optional.of(builder.build().asInstanceOf[Shape])
-      })
+          .copyDefaultTrait(target)
+          .build()
+      }
       .orElse(shape)
   }
 
@@ -64,14 +65,18 @@ class AwsStandardTypesTransformer extends ProjectionTransformer {
 
   private def canReplace(shape: Shape): Boolean = {
     shape.isInstanceOf[SimpleShape] &&
-      isAwsShape(shape) &&
-      onlySupportedTraits(shape.getAllTraits)
+    isAwsShape(shape) &&
+    onlySupportedTraits(shape.getAllTraits)
   }
 
   @annotation.nowarn
-  private def onlySupportedTraits(traits: java.util.Map[ShapeId, Trait]) = traits.values().stream().allMatch(t => {
-      t.isInstanceOf[DefaultTrait] || t.isInstanceOf[BoxTrait]
-  })
+  private def onlySupportedTraits(traits: java.util.Map[ShapeId, Trait]) =
+    traits
+      .values()
+      .stream()
+      .allMatch(t => {
+        t.isInstanceOf[DefaultTrait] || t.isInstanceOf[BoxTrait]
+      })
 
 }
 
@@ -79,13 +84,16 @@ object AwsStandardTypesTransformer {
 
   val name: String = "AwsStandardTypesTransformer"
 
-  private[transformers] final implicit class MemberShapeBuilderOps(val builder: MemberShape.Builder) extends AnyVal {
-    def copyDefaultTrait(shape: Shape, replacementShape: ShapeId): MemberShape.Builder = {
+  private[transformers] final implicit class MemberShapeBuilderOps(
+      val builder: MemberShape.Builder
+  ) extends AnyVal {
+    def copyDefaultTrait(
+        shape: Shape
+    ): MemberShape.Builder = {
       val defaultTraitOpt = toOption(shape.getTrait(classOf[DefaultTrait]))
 
-      defaultTraitOpt.map { df =>
-        new DefaultTrait.Provider().createTrait(replacementShape, df.toNode)
-      }.fold(builder){defaultTrait => builder.addTrait(defaultTrait)}
+      defaultTraitOpt
+        .fold(builder)(builder.addTrait(_))
     }
   }
 
@@ -96,4 +104,4 @@ object AwsStandardTypesTransformer {
       None
     }
   }
-}
\ No newline at end of file
+}
```

modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/src/main/scala/Main.scala
Baccata: @denisrosca because we can share the logic between SBT/mill, I'm happy with just having this test, no need for a mill-specific test.

Baccata:
Pretty good work overall :)

modules/codegen/src/smithy4s/codegen/transformers/AwsStandardTypesTransformer.scala
denisrosca: Fixed + added a test to validate that member traits are kept.

modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/test
denisrosca: Switched to using a lightweight dynamo spec in 578d114

denisrosca:
Need some help figuring out the reason for the build failure. Unfortunately, I can't see the error message 🤔

Baccata:
@denisrosca, just sync your branch with the latest `series/0.17` and it should fix it

modules/codegen/src/smithy4s/codegen/internals/CodegenImpl.scala
kubukoz: fine

kubukoz (approved):

kubukoz:
@baccata we cool here?

Baccata (approved):

Baccata:
Thanks a lot for this !

PR #1112, opened 2023-07-21
UrlForm encoding and AWS query client support

dhpiggott:
This is good for review now. Some notes:
1. I still want to run Scalafix against just the 26 paths touched by this PR to ensure imports are organised, but I haven't figured out a good way to do that yet. Not sure if that's essential but definitely nice to have.
2. I haven't waited for CI to be all green since my last change so there could be a surprise failure I've not seen yet. Hopefully not though.
3. Given the number of commits and the presence of decoding stuff in some of the earlier ones, I'd like this to be squash merged.


modules/aws-http4s/src/smithy4s/aws/internals/AwsQueryCodecs.scala
```diff
@@ -0,0 +1,226 @@
...
+                            endpoint.id.name
+                          )
+                          .toPathedValues ++
+                          UrlForm.FormData
+                            .PathedValue(
+                              PayloadPath(PayloadPath.Segment("Version")),
+                              version
+                            )
+                            .toPathedValues ++
+                          fa.encode(a).values.values
```
Baccata: aesthetically speaking, I'm not a fan of [hadouken code](https://i.pinimg.com/originals/98/bd/38/98bd38321fad0e8b14ebf3a5a9ec70c9.jpg) 

I'd rather the `Action`/`Version` related was extracted in a `addOperationHeader` method

dhpiggott: Yeah this mess isn't meant to be final, it's just another step along the way.

modules/core/src/smithy4s/http/UrlForm.scala
```diff
@@ -0,0 +1,146 @@
...
+
+      override def prepend(segment: PayloadPath.Segment): FormData = this
+
+      override def render: String = ""
+
+      override def toPathedValues: Vector[FormData.PathedValue] = Vector.empty
+
+    }
+
+    // TODO: Rename as Value, replace uses by PathedValue?
```
dhpiggott, line 61: Will handle this as part of the decoding PR.

modules/core/src/smithy4s/http/UrlForm.scala
```diff
@@ -0,0 +1,146 @@
...
+            URLEncoder.encode(value, StandardCharsets.UTF_8.name())
+        }
+        renderedKey + "=" + renderedValue
+      }
+
+      override def toPathedValues: Vector[FormData.PathedValue] = Vector(this)
+
+    }
+
+    // TODO: Rename as Values?
```
dhpiggott, line 113: Ditto, because it won't be clear if the model is good until decoding support is finalised.

modules/aws-http4s/src/smithy4s/aws/AwsCredentialsProvider.scala
```diff
@@ -62,6 +62,8 @@ class AwsCredentialsProvider[F[_]](implicit F: Temporal[F]) {
     Resource
       .eval(fromEnv)
       .map(F.pure)
+      // TODO: Ensure minimal delay when these endpoints don't exist, e.g.
+      // when running locally.
```
dhpiggott, line 66: Didn't feel it was appropriate to make any more changes to the provider as part of this PR, so just leaving this here.

modules/aws-http4s/src/smithy4s/aws/internals/AwsQueryCodecs.scala
dhpiggott: Improved it a bit in https://github.com/disneystreaming/smithy4s/pull/1112/commits/74cf808b3c3a76e7282bde7280bb6e78b5bca5f8. My plan is to add a function like `addValue` in the decoding PR which is the one where I plan to finalise the `UrlForm` model.

modules/aws-http4s/src/smithy4s/aws/AwsCredentialsProvider.scala
Baccata: Fair nuff

modules/sandbox/src/Sandbox.scala
```diff
@@ -0,0 +1,96 @@
...
+        ),
+        IO.realTime.map(_.toSeconds).map(Timestamp(_, 0))
+      )
+      cloudWatch <- AwsClient(CloudWatch, awsEnv)
+    } yield cloudWatch
+
+  // This is probably something that's gonna get reimplemented a lot in
+  // user-land. Perhaps we could use pagination hints from the specs to avoid
+  // having to manually wire up the accessors, and to generate synthetic service
+  // functions that handle pagination?
```
Baccata, line 72: yup I have a plan for this

modules/core/src/smithy4s/http/UrlForm.scala
```diff
@@ -0,0 +1,146 @@
...
+
+            case (builder, (Segment.Label(label), _)) =>
+              builder.append(
+                URLEncoder.encode(label, StandardCharsets.UTF_8.name())
+              )
+
+            case (builder, (Segment.Index(index), _)) =>
+              builder.append(index)
+          }
+          .toString()
```
Baccata: I'd rather this was an imperative for-loop, and avoided `zipWithIndex` to save on allocation. Moreover, mixing folds with mutable-data (StringBuilder) is a code-smell.

modules/core/src/smithy4s/http/UrlForm.scala
```diff
@@ -0,0 +1,146 @@
...
+        val renderedValue = maybeValue match {
+          case None => ""
+          case Some(value) =>
+            URLEncoder.encode(value, StandardCharsets.UTF_8.name())
+        }
+        renderedKey + "=" + renderedValue
```
Baccata: this could keep appending to the `StringBuilder`, rather than allocating new strings via concatenation

modules/core/src/smithy4s/http/UrlForm.scala
```diff
@@ -0,0 +1,146 @@
...
+
+private[smithy4s] final case class UrlForm(
+    formData: UrlForm.FormData.MultipleValues
+)
+
+private[smithy4s] object UrlForm {
+
+  sealed trait FormData extends Product with Serializable {
+    def prepend(segment: PayloadPath.Segment): FormData
+    def render: String
```
Baccata: I think this should delegate to a `def writeTo(sb: StringBuilder): Unit` method

modules/core/src/smithy4s/http/UrlForm.scala
dhpiggott: I'd left this mostly unchanged relative to the original in https://github.com/disneystreaming/smithy4s/blob/8b8cba7819cef902b3bc81c714f00d431b00125f/modules/aws-http4s/src/smithy4s/aws/query/FormData.scala#L56

Makes sense to improve it like that though! Done in https://github.com/disneystreaming/smithy4s/pull/1112/commits/ff5a7c14dd60fc08bf699551c3a8e77e280a8b02.

modules/core/src/smithy4s/http/UrlForm.scala
dhpiggott: Ah yeah. Done in https://github.com/disneystreaming/smithy4s/pull/1112/commits/ff5a7c14dd60fc08bf699551c3a8e77e280a8b02.

modules/core/src/smithy4s/http/UrlForm.scala
dhpiggott: Nice. Done in https://github.com/disneystreaming/smithy4s/pull/1112/commits/ff5a7c14dd60fc08bf699551c3a8e77e280a8b02.

Baccata (approved):
Great work !

Issue #701, opened 2023-01-03
Improve document decoding failure error messages


yisraelU:
Currently, when a Document Decoder fails, one receives a PayloadError with the field and type it was looking for.
The message is simply `Wrong Json shape` 
It doesnt tell you what type of document it did receive.
It would be more informative and greatly assist with debugging if it would say expected string , received Integer etc..
Thats all I could think of for now , perhaps additional infromation would be useful too

kubukoz:
@yisraelU is this fixed in #1114?

PR #1114, opened 2023-07-21
add method to display json shape received when failing document decoder

yisraelU:
RE: #701
adds method to display shape of document received when failing due to unexpected json shape

### DIFF
--- a/modules/core/src/smithy4s/Document.scala
+++ b/modules/core/src/smithy4s/Document.scala
@@ -35,6 +35,17 @@ sealed trait Document extends Product with Serializable {
 
   override def toString(): String = this.show
 
+  def name: String = this match {
+    case DNumber(_)  => "number"
+    case DString(_)  => "string"
+    case DBoolean(_) => "boolean"
+    case DNull       => "null"
+    case DArray(value) =>
+      s"array[${value.headOption.map(_.name).getOrElse("null")}]"
+    case DObject(map) =>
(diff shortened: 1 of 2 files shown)


modules/core/src/smithy4s/Document.scala
```diff
@@ -35,6 +35,17 @@ sealed trait Document extends Product with Serializable {
 
   override def toString(): String = this.show
 
+  def name: String = this match {
+    case DNumber(_)  => "number"
```
kubukoz: **suggestion:** these are lowercase, but AFAICS the usage of `DocumentDecoder.instance` uses uppercase names for the expected JSON shape. Would you like to update the call sites so that it's consistenly lowercase? (or the opposite)

image

modules/core/src/smithy4s/Document.scala
```diff
@@ -35,6 +35,17 @@ sealed trait Document extends Product with Serializable {
 
   override def toString(): String = this.show
 
+  def name: String = this match {
+    case DNumber(_)  => "Number"
+    case DString(_)  => "String"
+    case DBoolean(_) => "Boolean"
+    case DNull       => "Null"
+    case DArray(value) =>
+      s"Array[${value.headOption.map(_.name).getOrElse("Null")}]"
```
Baccata: there is absolutely no guarantee that an json array may be homogenous in shape. I'd rather it was just a static `"Array"`, and for object to be `"Object"`.

Baccata (approved):

PR #1115, opened 2023-07-21
use blob in place of bytearray

lewisjkl:
Deprecate smithy4s.ByteArray and use smithy4s.Blob all over instead.


modules/core/src/smithy4s/ByteArray.scala
```diff
@@ -18,6 +18,8 @@ package smithy4s
 
 import java.util.Base64
 
+@deprecated("Use smithy4s.Blob instead", since = "0.18.0")
```
kubukoz: Thought: if it's deprecated and we're replacing all its usages, chances are people will get breakage because of all the type mismatches. At this point I'd say remove the type altogether

Baccata: agreed

modules/bootstrapped/test/src/smithy4s/http/StringAndBlobSpec.scala
```diff
@@ -93,8 +93,8 @@ class StringAndBlobSpec() extends munit.FunSuite {
 
   test("Blobs") {
     check(
-      Schema.bytes,
-      ByteArray("hello".getBytes()),
+      Schema.blob,
+      Blob("hello".getBytes()),
```
Baccata: we have smart constructors that take strings directly now

modules/bootstrapped/test/src/smithy4s/http/StringAndBlobSpec.scala
```diff
@@ -103,7 +103,7 @@ class StringAndBlobSpec() extends munit.FunSuite {
   test("Blobs (custom media-type)") {
     check(
       PNG.schema,
-      PNG(ByteArray("hello".getBytes())),
+      PNG(Blob("hello".getBytes())),
```
Baccata: ditto

modules/bootstrapped/test/src/smithy4s/schema/DefaultValueSpec.scala
```diff
@@ -59,7 +59,7 @@ final class DefaultValueSpec extends FunSuite {
   }
 
   test("blob") {
-    testCase(Schema.bytes, ByteArray(Array.empty))
+    testCase(Schema.bytes, Blob(Array.empty[Byte]))
```
Baccata: `Blob.empty`

modules/cats/src/smithy4s/interopcats/instances/HashInstances.scala
```diff
@@ -17,18 +17,18 @@
...
 import smithy4s.schema.Primitive
 
 private[interopcats] trait HashInstances {
 
-  implicit val byteArrayHash: Hash[ByteArray] =
-    new Hash[ByteArray] {
-      override def hash(x: ByteArray): Int = x.array.hashCode()
+  implicit val blobHash: Hash[Blob] =
+    new Hash[Blob] {
+      override def hash(x: Blob): Int = x.toArray.hashCode()
```
Baccata: x.hashCode

modules/cats/src/smithy4s/interopcats/instances/ShowInstances.scala
```diff
@@ -17,14 +17,16 @@
...
-import smithy4s.{ByteArray, Document, ShapeId, Timestamp}
+import smithy4s.{Blob, Document, ShapeId, Timestamp}
 import smithy4s.kinds.PolyFunction
 
 private[interopcats] trait ShowInstances {
 
   implicit val sId: Show[ShapeId] = Show.fromToString
-  implicit val byteArray: Show[ByteArray] = Show.fromToString
+  implicit val blob: Show[Blob] = (b: Blob) =>
+    Base64.getEncoder().encodeToString(b.toArray)
```
Baccata: b.toBase64String

modules/cats/test/src/HashVisitorSpec.scala
```diff
@@ -105,11 +105,11 @@ object HashVisitorSpec extends FunSuite with CompatProvider {
 
   }
 
-  test("smithy4s ByteArray") {
-    val schema: Schema[ByteArray] = bytes
-    val fooBar = ByteArray("fooBar".getBytes)
+  test("smithy4s Blob") {
+    val schema: Schema[Blob] = blob
+    val fooBar = Blob("fooBar".getBytes)
```
Baccata:
```suggestion
    val fooBar = Blob("fooBar")
```

Baccata (approved):

PR #1120, opened 2023-07-25
Wire up EC2 Query codec based on Query codec, plus compliance tests

dhpiggott:
Approach:

The compliance tests caught a few issues, fixes for which are included. I then used it in the sandbox module to call some EC2 APIs, and that found a few more issues, fixes for which are also included. The remaining issue not fixed in this PR is the need for optional open enum support. We're agree that's out of scope for this PR. The consequence is that there are a decent number of EC2 API responses which fail to decode due to AWS including enum values that are not present in the published spec. To get a good sandbox run for now I just used an EC2 endpoint that happens to not include any unknown enum values.

Summary of changes:

1. Wire up AWS EC2 Query codec based on AWS Query codec, plus compliance tests.
2. Add flags to `UrlFormDataEncoderSchemaVisitor` so it can accomodate the quirks of the AWS EC2 protocol without having to be aware of all the details. Instead, the relevant hints on the schema are transformed so that `UrlFormDataEncoderSchemaVisitor` can work only in terms of `XmlName`.
3. Add comments explaining quirks of both query protocols.
4. Sort imports in all files touched by #1112 and this branch.
5. Patch `AwsStandardTypesTransformer` to fix codgen error (thanks Oli for the patch).
6. Also fix sandbox compile errors.
7. Patch renderer to escape $ when rendering literal strings, necessary because the EC2 spec includes a documentation string with `<p>Specify the fields using the <code>${field-id}</code> format, separated by spaces.` in it.
8. Make `XmlDecoder` aware of default hint on structure members. Without this, sandbox runs failed with `HttpPayloadError(.DescribeInstancesResponse.reservationSet.item.0.instancesSet.item.0.placement.partitionNumber, expected = , message=Expected a single node with text content)` due to the use of `@default` in
https://raw.githubusercontent.com/aws/aws-sdk-js-v3/main/codegen/sdk-codegen/aws-models/ec2.json.
9. Make XML schema visitors use cache.

Merging:

Like #1112 and for the same reason, let's squash merge this, using the above summary of changes as the commit message.

### DIFF
--- a/modules/aws-http4s/src/smithy4s/aws/internals/AwsEc2QueryCodecs.scala
+++ b/modules/aws-http4s/src/smithy4s/aws/internals/AwsEc2QueryCodecs.scala
@@ -109,21 +109,28 @@ private[aws] object AwsEcsQueryCodecs {
       action: String,
       version: String
   ): CachedSchemaCompiler[RequestEncoder[F, *]] = {
-    val urlFormEntityEncoderCompilers = UrlForm.Encoder.mapK(
-      new PolyFunction[UrlForm.Encoder, EntityEncoder[F, *]] {
-        def apply[A](fa: UrlForm.Encoder[A]): EntityEncoder[F, A] =
-          urlFormEntityEncoder[F].contramap((a: A) =>
-            UrlForm(
-              formData = UrlForm.FormData.MultipleValues(
-                values = Vector(
-                  UrlForm.FormData.PathedValue(PayloadPath("Action"), action),
-                  UrlForm.FormData.PathedValue(PayloadPath("Version"), version)
-                ) ++ fa.encode(a).formData.values
+    val urlFormEntityEncoderCompilers = UrlForm
+      .Encoder(
+        // Per
+        // https://github.com/smithy-lang/smithy/blob/0d7a26d4880cdb7b3c21ba414c2642d93245b19e/smithy-aws-protocol-tests/model/ec2Query/main.smithy#L5.
+        ignoreXmlFlattened = true
+      )
+      .mapK(
+        new PolyFunction[UrlForm.Encoder, EntityEncoder[F, *]] {
+          def apply[A](fa: UrlForm.Encoder[A]): EntityEncoder[F, A] =
+            urlFormEntityEncoder[F].contramap((a: A) =>
+              UrlForm(
+                formData = UrlForm.FormData.MultipleValues(
+                  values = Vector(
+                    UrlForm.FormData.PathedValue(PayloadPath("Action"), action),
+                    UrlForm.FormData
+                      .PathedValue(PayloadPath("Version"), version)
+                  ) ++ fa.encode(a).formData.values
+                )
               )
             )
-          )
-      }
-    )
+        }
+      )
     RequestEncoder.restSchemaCompiler[F](
       metadataEncoderCompiler = Metadata.AwsEncoder,
       entityEncoderCompiler = urlFormEntityEncoderCompilers,
--- a/modules/aws-http4s/src/smithy4s/aws/internals/AwsEc2QueryCodecs.scala
+++ b/modules/aws-http4s/src/smithy4s/aws/internals/AwsEc2QueryCodecs.scala
@@ -18,15 +18,15 @@ package smithy4s
 package aws
 package internals
 
-import smithy.api.XmlName
+import _root_.aws.protocols.AwsQueryError
+import _root_.aws.protocols.Ec2QueryName
 import cats.effect.Concurrent
 import cats.syntax.all._
 import fs2.compression.Compression
+import smithy.api.XmlName
 import smithy4s.Endpoint
 import smithy4s.http._
 import smithy4s.http4s.kernel._
-import _root_.aws.protocols.Ec2QueryName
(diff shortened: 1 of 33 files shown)


modules/sandbox/src/Sandbox.scala
```diff
@@ -14,18 +14,23 @@
...
 
 object Main extends IOApp.Simple {
 
   type NextToken = String
 
-  override def run: IO[Unit] = cloudWatchClientResource.use(cloudWatchClient =>
+  override def run: IO[Unit] = awsEnvironmentResource.use { awsEnvironment =>
+    // Per
+    // https://disneystreaming.github.io/smithy4s/docs/protocols/aws/aws/#awsquery,
+    // CloudWatch is one of a few services that use the ec2query protocol.
```
Baccata: you mean `awsQuery` protocol

dhpiggott: Oops! Nice catch, thanks.
```suggestion
    // CloudWatch is one of a few services that use the awsQuery protocol.
```

modules/aws-http4s/src/smithy4s/aws/internals/AwsEc2QueryCodecs.scala
```diff
@@ -0,0 +1,146 @@
...
+                case None =>
+                  hints.memberHints.get(XmlName) match {
+                    case Some(xmlName) =>
+                      hints.addMemberHints(
+                        XmlName(xmlName.value.capitalize)
+                      )
+
+                    case _ =>
+                      hints
+                  }
```
Baccata: good thinking, Batman

modules/aws-http4s/src/smithy4s/aws/internals/AwsEc2QueryCodecs.scala
```diff
@@ -0,0 +1,146 @@
...
+  ): UnaryClientCodecs.Make[F] =
+    new UnaryClientCodecs.Make[F] {
+      def apply[I, E, O, SI, SO](
+          endpoint: Endpoint.Base[I, E, O, SI, SO]
+      ): UnaryClientCodecs[F, I, E, O] = {
+        val requestEncoderCompilers = AwsQueryCodecs
+          .requestEncoderCompilers[F](
+            // These are set to fulfil the requirements of
+            // https://smithy.io/2.0/aws/protocols/aws-ec2-query-protocol.html?highlight=ec2%20query%20protocol#query-key-resolution.
+            // without UrlFormDataEncoderSchemaVisitor having to have more aware
```
daddykotex:
```suggestion
            // without UrlFormDataEncoderSchemaVisitor having to be more aware
```

modules/aws-kernel/src/smithy4s/aws/AwsProtocol.scala
```diff
@@ -52,6 +57,7 @@ private[aws] object AwsProtocol {
       )
 
   // See https://awslabs.github.io/smithy/1.0/spec/aws/aws-json-1_0-protocol.html#differences-between-awsjson1-0-and-awsjson1-1
```
daddykotex: the comment feels a bit out of place, I assume it was meant to be above the json  protocols

daddykotex: maybe we could add a link to https://smithy.io/2.0/aws/protocols/aws-ec2-query-protocol.html

modules/xml/src/smithy4s/xml/internals/XmlDecoder.scala
```diff
@@ -80,55 +86,82 @@ private[smithy4s] object XmlDecoder {
...
-              other.history,
+              history,
               s"Expected a single node with text content"
             )
           )
       }
-    }
 
+    case AttrNode(history, values) =>
+      if (values.tail.nonEmpty) {
```
daddykotex, line 150: can you add a check for empty list? because calling tail on a empty `values` will blow up

```suggestion
      if (values.nonEmpty && values.tail.nonEmpty) {
```

daddykotex (commented):
great job

mostly minor comments, one important bit related to checking for empty list,

modules/aws-kernel/src/smithy4s/aws/AwsProtocol.scala
dhpiggott: Oops! Done.

modules/xml/src/smithy4s/xml/internals/XmlDecoder.scala
dhpiggott: `values` is a `NonEmptyList`: `case class AttrNode(history: XPath, values: NonEmptyList[String]) extends XmlCursor { ... }`, and `tail` is a `List`: `final case class NonEmptyList[+A](head: A, tail: List[A]) extends NonEmptyCollection[A, List, NonEmptyList] { ... }` so this should be OK.

daddykotex: ahhh, good!

daddykotex (approved):

lewisjkl (approved):

PR #1123, opened 2023-07-26
Fix rendering of deprecated annotation in mixins

### LINKED ISSUE #1121: @deprecated on mixin member generates invalid code
Deprecating mixin member causes the code generation to output invalid code. 
Tested with 0.17.12

Given:
```smithy
@mixin
structure Extra {
    @deprecated
    @required
    name: String
}

structure Greeting with [Extra] {
    @required
    message: String
}
```

actual generated code:
```scala
trait Extra {
  def @deprecated(message = "N/A", since = "N/A") name: String
}
```

expected generated code:
```scala
trait Extra {
  @deprecated(message = "N/A", since = "N/A") 
  def name: String
}
```

Full code to reproduce the issue can be found here https://github.com/ghostbuster91/demos/tree/deprecated-issue-mixin

ghostbuster91:
This fixes #1121

### DIFF
--- a/modules/example/src/smithy4s/example/DeprecatedMixin.scala
+++ b/modules/example/src/smithy4s/example/DeprecatedMixin.scala
@@ -3,5 +3,7 @@ package smithy4s.example
 
 @deprecated(message = "A compelling reason", since = "0.0.1")
 trait DeprecatedMixin {
-  @deprecated(message = "N/A", since = "N/A") def strings: Option[List[String]]
+  @deprecated(message = "N/A", since = "N/A")
+  def strings: Option[List[String]]
+  def other: Option[List[String]]
 }
--- a/modules/example/src/smithy4s/example/DeprecatedMixin.scala
+++ b/modules/example/src/smithy4s/example/DeprecatedMixin.scala
@@ -1,0 +1,7 @@
+package smithy4s.example
+
+
+@deprecated(message = "A compelling reason", since = "0.0.1")
+trait DeprecatedMixin {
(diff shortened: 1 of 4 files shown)


kubukoz (commented):
can you make this target 0.17? It'd be great to add in a patch rather than wait for 0.18.

kubukoz (approved):

ghostbuster91:
> can you make this target 0.17? It'd be great to add in a patch rather than wait for 0.18.

Good idea, changed.

daddykotex:
Thank you so much

daddykotex:
I'll happily port it to 0.18

modules/example/src/smithy4s/example/DeprecatedMixin.scala
```diff
@@ -0,0 +1,7 @@
+package smithy4s.example
+
+
+@deprecated(message = "A compelling reason", since = "0.0.1")
+trait DeprecatedMixin {
+  @deprecated(message = "N/A", since = "N/A") def strings: Option[List[String]]
```
daddykotex: should we add a new line between the annotation and the def?

daddykotex: added it, hope you don't mind

daddykotex (approved):

kubukoz (approved):

kubukoz:
@daddykotex I'll open a backport, there's another PR to sync as well so I'll deal with them both

daddykotex:
ahh thanks :)

PR #1127, opened 2023-07-27
Only transform AWS shapes named after standard shapes

Baccata:
I messed up when reviewing https://github.com/disneystreaming/smithy4s/pull/1110 : the requirements were to only transform shapes that are named after standard simple shapes, such as [this one](https://github.com/aws/aws-sdk-js-v3/blob/main/codegen/sdk-codegen/aws-models/dynamodb.json#L7433-L7438). 

This PR fixes that

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/transformers/AwsStandardTypesTransformer.scala
+++ b/modules/codegen/src/smithy4s/codegen/transformers/AwsStandardTypesTransformer.scala
@@ -84,7 +84,8 @@ class AwsStandardTypesTransformer extends ProjectionTransformer {
   private def canReplace(shape: Shape): Boolean = {
     shape.isInstanceOf[SimpleShape] &&
     isAwsShape(shape) &&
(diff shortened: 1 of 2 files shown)


daddykotex (approved):

modules/codegen/src/smithy4s/codegen/transformers/AwsStandardTypesTransformer.scala
```diff
@@ -96,12 +97,33 @@ class AwsStandardTypesTransformer extends ProjectionTransformer {
         t.isInstanceOf[DefaultTrait] || t.isInstanceOf[BoxTrait]
       })
 
+  private def hasStandardName(shape: Shape): Boolean =
+    AwsStandardTypesTransformer.standardSimpleShapes.contains(
```
daddykotex, line 101:
```suggestion
    AwsStandardTypesTransformer.standardSimpleShapes(
```

I think you could get away by calling apply on the Set directly, but this is minor

PR #1128, opened 2023-07-27
add nicer api in new Xml object

lewisjkl:
This is meant to be similar to the `smithy4s.json.Json` object. Let me know what improvements you think I can make to the methods here.


modules/xml/src/smithy4s/xml/Xml.scala
```diff
@@ -0,0 +1,126 @@
...
+      .take(1)
+      .compile
+      .last
+      .flatMap(
+        SyncIO.fromOption(_)(
+          new XmlDecodeError(XPath(List.empty), "Unable to decode XML input")
+        )
+      )
+      .attempt
+      .unsafeRunSync()
```
daddykotex: is this expected?

lewisjkl: Yeah, since it is SyncIO it is basically just so that users don't need to deal with an effect unless they want to (in which case they can use the other methods). Wdyt?

kubukoz: I think if you only need `RaiseThrowable`, you can use `F=EitherThrow`. IIRC fs2 has a compiler for that, so you could compile without SyncIO.

kubukoz: sorry, it's `Fallible`:

```scala
//> using scala "2.13.10"
//> using lib "co.fs2::fs2-core:3.7.0"

val s = fs2.Stream(1, 2, 3).covary[fs2.Fallible]
// s: fs2.Stream[fs2.Fallible, Int] = Stream(..)

s.compile.toList
// res0: Either[Throwable, List[Int]] = Right(value = List(1, 2, 3))
```

daddykotex: never took the time to read the docs: https://typelevel.org/cats-effect/api/3.x/cats/effect/SyncIO.html

TIL

modules/xml/src/smithy4s/xml/Xml.scala
```diff
@@ -0,0 +1,126 @@
...
+      }
+  }
+
+  /**
+    * Writes the XML representation for an instance of `A` into an [[fs2.Stream]] of String.
+    *
+    * Beware : using this method with a non-static schema (for instance, dynamically generated) may
+    * result in memory leaks.
+    */
+  def writeToString[A: Schema](a: A): Stream[fs2.Pure, String] = {
```
Baccata: I wouldn't return a `Stream` here, really

modules/xml/src/smithy4s/xml/Xml.scala
```diff
@@ -0,0 +1,126 @@
...
+object Xml {
+
+  private val decoderCacheGlobal = XmlDocument.Decoder.createCache()
+  private val encoderCacheGlobal = XmlDocument.Encoder.createCache()
+
+  private def deriveXmlDecoder[A: Schema]: XmlDocument.Decoder[A] =
+    XmlDocument.Decoder.fromSchema(Schema[A], decoderCacheGlobal)
+
+  private def deriveXmlEncoder[A: Schema]: XmlDocument.Encoder[A] =
+    XmlDocument.Encoder.fromSchema(Schema[A], encoderCacheGlobal)
```
Baccata: when possible, it's better to have public members first, private members afterwards for people who like to read sources

modules/xml/src/smithy4s/xml/Xml.scala
```diff
@@ -0,0 +1,126 @@
...
+    * result in memory leaks.
+    */
+  def readToStream[F[_]: RaiseThrowable, A: Schema](
+      blob: Blob
+  ): Stream[F, A] = {
+    val decoder = deriveXmlDecoder[A]
+    readToDocumentStream[F](blob).flatMap(document =>
+      Stream.fromEither(decoder.decode(document))
+    )
+  }
```
Baccata: I don't think this one is necessary (or at least it shouldn't be public). I'd rather we exposed an API that works on the assumption that 1 Blob == 1 XmlDocument == 1 piece of data

modules/xml/src/smithy4s/xml/Xml.scala
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+    * Reads an instance of `A` from a [[smithy4s.Blob]] holding an XML payload.
+    *
+    * Beware : using this method with a non-static schema (for instance, dynamically generated) may
+    * result in memory leaks.
+    */
+  def read[A: Schema](blob: Blob): Either[XmlDecodeError, A] = {
```
Baccata: 👍

modules/xml/src/smithy4s/xml/Xml.scala
```diff
@@ -0,0 +1,126 @@
...
+    * Parses an [[XmlDocument]] from a [[Blob]] inside of an
+    * fs2.Stream.
+    */
+  def readToDocumentStream[F[_]: RaiseThrowable](
+      blob: Blob
+  ): Stream[F, XmlDocument] =
+    Stream
+      .emit(blob.toUTF8String)
+      .through(events[F, String]())
+      .through(documents[F, XmlDocument])
```
Baccata: remove (or make private), see my comment there : https://github.com/disneystreaming/smithy4s/pull/1128/files#r1277576714

modules/xml/src/smithy4s/xml/Xml.scala
```diff
@@ -0,0 +1,126 @@
...
+    * Reads an instance of `A` from a [[smithy4s.Blob]] holding an XML payload.
+    *
+    * Beware : using this method with a non-static schema (for instance, dynamically generated) may
+    * result in memory leaks.
+    */
+  def read[A: Schema](blob: Blob): Either[XmlDecodeError, A] = {
+    readToStream[SyncIO, A](blob)
+      .take(1)
+      .compile
+      .last
```
Baccata: now there's `.compile.onlyOrError` that exists, that's what we want really

modules/xml/src/smithy4s/xml/Xml.scala
```diff
@@ -0,0 +1,126 @@
...
+      .flatMap(
+        SyncIO.fromOption(_)(
+          new XmlDecodeError(XPath(List.empty), "Unable to decode XML input")
+        )
+      )
+      .attempt
+      .unsafeRunSync()
+      .leftMap {
+        case x: XmlDecodeError => x
+        case other => new XmlDecodeError(XPath(List.empty), other.getMessage)
```
Baccata, line 42: 👍

modules/xml/src/smithy4s/xml/Xml.scala
```diff
@@ -0,0 +1,126 @@
...
+  def writeToBytes[A: Schema](a: A): Stream[fs2.Pure, Byte] =
+    writeToString[A](a).through(fs2.text.utf8.encode[fs2.Pure])
+
+  /**
+    * Writes the XML representation for an instance of `A` into a [[smithy4s.Blob]].
+    *
+    * Beware : using this method with a non-static schema (for instance, dynamically generated) may
+    * result in memory leaks.
+    */
+  def write[A: Schema](a: A): Blob = {
```
Baccata: 👍

modules/xml/src/smithy4s/xml/Xml.scala
Baccata: Let's make it return `String`. What we want is an API that's as unceremonious as possible.

Baccata:
Alright. So the goal of this object is to offer easy/un-ceremonious methods, for quickly encoding/decoding to/from in-memory data. So let's remove/hide any method that exposes fs2.Stream whilst taking an implicit `Schema`

modules/xml/src/smithy4s/xml/Xml.scala
```diff
@@ -0,0 +1,123 @@
...
+    * Beware : using this method with a non-static schema (for instance, dynamically generated) may
+    * result in memory leaks.
+    */
+  def writeToString[A: Schema](a: A): Option[String] =
+    writeToStringStream[A](a).compile.last
+
+  type XmlByteStreamEncoder[F[_], A] =
+    smithy4s.codecs.Writer[Any, Stream[F, Byte], A]
+  type XmlByteStreamDecoder[F[_], A] =
+    smithy4s.codecs.Reader[F, Stream[F, Byte], A]
```
Baccata: cool, so I'm thinking actually that we need an `smithy4s-fs2` module, that could carry this : 

* a lot of what we do could actually be fs2-specific as opposed to http4s-specific, and that'd give us a chance to be more precise and avoid some coupling 
* the pagination work could go there 
* these types could move over-there, as there is nothing Xml specific to them

modules/xml/src/smithy4s/xml/internals/XmlByteStreamDecoderCompilerImpl.scala
```diff
@@ -0,0 +1,68 @@
...
+    with XmlByteStreamDecoderCompiler[F] {
+
+  type Aux[A] = XmlByteStreamDecoder[F, A]
+
+  def fromSchema[A](
+      schema: Schema[A],
+      cache: Cache
+  ): XmlByteStreamDecoder[F, A] =
+    new Reader[F, Stream[F, Byte], A] {
+      val fa = XmlDocument.Decoder.fromSchema(schema)
```
lewisjkl, line 42: @Baccata wanted to double check with you.. do I somehow need to pass a cache down here? I'm not exactly sure.

modules/xml/src/smithy4s/xml/internals/XmlByteStreamEncoderCompilerImpl.scala
```diff
@@ -0,0 +1,46 @@
...
+    extends CachedSchemaCompiler.Impl[XmlByteStreamEncoder[F, *]]
+    with XmlByteStreamEncoderCompiler[F] {
+
+  type Aux[A] = XmlByteStreamEncoder[F, A]
+
+  def fromSchema[A](
+      schema: Schema[A],
+      cache: Cache
+  ): XmlByteStreamEncoder[F, A] = new Writer[Any, Stream[F, Byte], A] {
+    val fa = XmlDocument.Encoder.fromSchema(schema)
```
lewisjkl, line 37: Here as well.

lewisjkl:
It might be worth going with something different for the package name than `smithy4s.fs2` since we often import `smithy4s._` and then it conflicts. Not sure what a good name would be though, maybe `smithy4s.fs2lib`? Wdyt?

---

Edit I went ahead and made the change to `fs2lib` if people don't like that we can easily revert that commit or change it to something different.

modules/aws-http4s/src/smithy4s/aws/internals/AwsXmlCodecs.scala
```diff
@@ -33,7 +33,7 @@ import smithy4s.http4s.kernel._
 import smithy4s.kinds.PolyFunction
 import smithy4s.schema.CachedSchemaCompiler
 import smithy4s.xml.Xml
-import smithy4s.fs2._
+import smithy4s.fs2lib._
```
Baccata: 👍 good idea, though we've gone for `interopcats` for cats, so let's keep consistent and call it `interopfs2`

modules/fs2/src/smithy4s/fs2lib/package.scala
```diff
@@ -0,0 +1,12 @@
+package smithy4s
+
+import _root_.fs2.Stream
+
+package object fs2lib {
+
+  type XmlByteStreamEncoder[F[_], A] =
+    smithy4s.codecs.Writer[Any, Stream[F, Byte], A]
+  type XmlByteStreamDecoder[F[_], A] =
```
Baccata: let's call those `ByteStreamEncoder` and `ByteStreamDecoder`, since they don't need to be XML specific

Baccata (approved):

PR #1129, opened 2023-07-28
Fixes AWS AwsStandardTypesTransformer bug

Baccata:
Only structure members can carry the `@default` trait.

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/transformers/AwsStandardTypesTransformer.scala
+++ b/modules/codegen/src/smithy4s/codegen/transformers/AwsStandardTypesTransformer.scala
@@ -65,16 +65,20 @@ class AwsStandardTypesTransformer extends ProjectionTransformer {
           .map[Boolean](canReplace)
           .orElse(false)
       }
-      .map[Shape](shape => {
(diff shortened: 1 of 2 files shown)


kubukoz (approved):

daddykotex:
So the spec was not invalid, we were just transforming the model incorrectly, interesting.

daddykotex:
Do you think it could be good user experience to:

1. load the model before any transformation (keep the result aside, do not throw if it fail)
2. load the model after transformation

Then if an error occur during the second step, we can print a better error message like:

```scala
val msg: 
if (failedAt1) "The model was not loading before transformation and still does not load after transformations."
else "The model was loading before transformation but does not load after transformations."
val errors = ??? // errors from 1 if failed, else errors from 2, and JUST errors, we can print warnings when the thing is successful
s"""
Code generation failed because we were not able to load the model.
$msg
Here are the errors:
$errors
"""
```

Baccata:
Possibly. Honestly I'm not too sure, because we're not attempting transformations of models that do not pass validation. 

However, one thing that'd improve user experience would be to remove warnings from the reported validation events when there's an error or danger.

PR #1133, opened 2023-08-02
Render idRefs as smithy4s.ShapeId

### LINKED ISSUE #889: Add refinement provider for IdRef
`@idRef` is a Smithy standard trait that you can put on a `String` type to ensure it's a valid Smithy shape.

Although schemas don't have access to all the possible shapes, we could at least provide a refinement that a value tagged with the trait uses valid shape syntax, in a typed way.

It would be something like this:

```scala
implicit val shapeIdRefinemend =
    Refinement
      .drivenBy[IdRef]
      .apply(ShapeId.parse(_: String).toRight("Not a valid ShapeId"), (_: ShapeId).show)
```

This would also be nice to special-case at codegen time, so that the `String` type would automatically become a `ShapeId` in the generated code - just like a custom refinement would change the type of any other primitive. This would be source breaking in the generated code, and possibly binary breaking for generated `smithy.api` code in smithy4s.

lewisjkl:
There are a few reasons why we need to have bespoke rendering logic for this PR rather than using the refinements feature out of the box:

1) There is no way to apply refinements to prelude shapes. Using the apply syntax to attempt adding a trait to a prelude shape results in an error. This is why we treat IdRef traits as though they are refinements even though they aren't technically.
2) There is no way to generically render refinements in the `hints` val of the generated code. The reason for this is that the constructors for the external refinement types can look like anything. While we could use the `apply` function from the `Refinement` for this, this function is not total so it would open the door to all kinds of issues in constructing the hints. As such, I chose to instead render the ShapeId for IdRefs with direct logic treating this case as a one-off that won't scale to other types.

I think it is justified to treat IdRef specially like this since IdRefs are such an integral part of the smithy model. I think it also makes the generated code nicer to work with now that things like protocolIds are ShapeIds rather than simple Strings.

---

Note, as part of this PR I also got rid of double rendering of hints in refinements when the refinement is directly in a structure field rather than in a newtype (see modules/bootstrapped/src/generated/smithy4s/example/StructureWithRefinedMember.scala in this PR for an example of what I mean).

---

Closes #889.


modules/bootstrapped/src/generated/smithy4s/example/StructureWithRefinedMember.scala
```diff
@@ -15,7 +15,7 @@ object StructureWithRefinedMember extends ShapeTag.Companion[StructureWithRefine
   val hints: Hints = Hints.empty
 
   implicit val schema: Schema[StructureWithRefinedMember] = struct(
-    int.refined[smithy4s.refined.Age](smithy4s.example.AgeFormat()).optional[StructureWithRefinedMember]("otherAge", _.otherAge).addHints(smithy4s.example.AgeFormat()),
```
daddykotex, line 18: is it expected that his addHints is gone?

modules/bootstrapped/src/generated/smithy4s/example/TestIdRef.scala
```diff
@@ -0,0 +1,22 @@
+package smithy4s.example
+
+import smithy4s.Hints
+import smithy4s.Schema
+import smithy4s.ShapeId
+import smithy4s.ShapeTag
+import smithy4s.schema.Schema.string
+import smithy4s.schema.Schema.struct
+
+final case class TestIdRef(test: Option[ShapeId] = None, test2: Option[TestIdRefTwo] = None)
```
daddykotex, line 10: nice: `Option[ShapeId]`

daddykotex (approved):
LGTM, initial request came from @kubukoz so I'd wait for a review from them before moving on

modules/bootstrapped/src/generated/smithy4s/example/StructureWithRefinedMember.scala
lewisjkl: Yeah that reason for this is that the hint is rendered inside of the refined schema already. We already did the same thing for NewTypes (see how they are rendered) so this is just bringing it into parity

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -569,11 +569,22 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+
+            refinement.orElse {
+              // Handle IdRefs as a refinement
+              if (trt.toShapeId == IdRefTrait.ID) {
+                val rt = RefinementTrait
+                  .builder()
+                  .targetType("smithy4s.ShapeId")
+                  .build()
+                Some(trt -> ExternalTypeInfo.RefinementInfo(rt))
+              } else None
```
kubukoz: suggestion: extract this to a parameterless `def` (for laziness) so that the `orElse` call is a simple one-liner

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -1027,8 +1037,12 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
             hints(member) ++ default ++ noDefault
           )
         }
-        .collect { case (name, Some(tpe), required, hints) =>
-          Field(name, tpe, required, hints)
+        .collect {
+          case (name, Some(tpe: Type.ExternalType), required, hints) =>
+            val newHints = hints.filterNot(_ == tpe.refinementHint)
+            Field(name, tpe, required, newHints)
```
kubukoz, line 1063: question: is it enough to do it on fields? I may not understand `Field` correctly, but I'd guess that we should do this everywhere we do `hints(shape)`

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -1288,9 +1302,38 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+  private object IdRefCase {
+    def unapply(tpe: Type): Boolean = tpe match {
+      case Type.ExternalType(
+            _,
+            fqn,
+            _,
+            _,
+            Type.PrimitiveType(Primitive.String),
+            _
+          ) if fqn == "smithy4s.ShapeId" =>
```
kubukoz: suggestion: `==` => `===`, at least it forces you to use the same type on both sides

sampleSpecs/idref.smithy
```diff
@@ -0,0 +1,12 @@
+$version: "2"
+
+namespace smithy4s.example
+
+structure TestIdRef {
+  @idRef
```
kubukoz, line 6: suggestion: this relates to another comment, I'd like to see tests for non-struct cases with idrefs. E.g. a custom `string` type, a member of a union / collection etc.

kubukoz (commented):
very nice, just had a concern about non-struct-member usecases

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
lewisjkl: Yeah that's a good call, I will check it out

lewisjkl:
@kubukoz appreciate your feedback. I've added support for using IdRefs in Maps, Lists, and Union alternatives (and previously structure members and String shapes/newtypes). So now I think I've handled them all since the selector is `:test(string, member > string)`. Lmk if you think of any others though.

kubukoz:
Should we support set shapes, just for the sake of compatibility with Smithy 1.x?

kubukoz (approved):

lewisjkl:
> Should we support set shapes, just for the sake of compatibility with Smithy 1.x?

I was debating.. I guess it wouldn't hurt to do so. I'll add that now

kubukoz:
cool, consider it approved from my side when you do

PR #1137, opened 2023-08-07
Open enumerations

lewisjkl:
This PR introduces open enumerations that are codegenerated with an extra `Unknown` case for handling values that are not defined in the specification. The presence of the `alloy#openEnum` trait on an `enum` or `intEnum` shape will trigger this to take place in the codegen.

The main need for this is that AWS treats enums as open in several places in their SDKs/APIs where as we treat them as closed always (in the past). This PR will allow for maintaining the default of closed enumerations, while also enabling interop with AWS and the like.


lewisjkl:
Still going to add docs and changelog and such, just working on getting CI happy in the meantime

lewisjkl:
One thing I still need to handle are collisions on the name `Unknown` since that could be a name that someone uses for an enum. Not sure the best thing to do in that case, but leaning toward making our `Unknown` into `$Unknown` or something like that. Will create an example and solve it tomorrow.

kubukoz:
That's interesting and well-timed as well, I was just thinking of how I'd support open enums in smithy-playground :)

will try to review after the weekend as I'm OOO

Baccata:
>  but leaning toward making our Unknown into $Unknown or something like that. Will create an example and solve it tomorrow.

I'm happy for you to roll with this :+1:

lewisjkl:
Ultimately decided to render all unknown enum members with a leading `$`. This simplifies things quite a bit and will make it very clear that the member is coming from an open enum and not from the spec. However, I'm happy to change it to only use a `$` when there is a collision if others think that is better.

lewisjkl:
So apparently using the `-Xlint:missing-interpolator` option with scala 2.12 means that it will warn on the following:

```scala
final case class $Unknown()
```

So I changed it to be:

```scala
final case class $$Unknown()
```

which works fine, but it is obviously kind of ugly.. so open to other suggestions.

kubukoz:
Frankly if it's just 2.12 I would keep the single $ and tell users who are stuck on 2.12 to ignore the warning in the build.

Baccata:
I agree with Jakub

Baccata:
Great work on this ! The only thing that bothers me is the `$$` to make 2.12 happy. We can remove this linter from 2.12 and roll with a single `$`

Baccata (approved):

PR #1140, opened 2023-08-09
update doc to include smithy4s.meta.noStackTrace usage

yisraelU:


modules/docs/markdown/04-codegen/01-customisation/11-no-stacktrace
```diff
@@ -0,0 +1,38 @@
+---
+sidebar_label: Error Stack Trace
+title: Throwable Usage
+---
+
+
+By default, smithy4s generates error data types that extend `java.lang.Throwable` , For example the following spec:
+
+```kotlin
```
Baccata, line 9: you can use `smithy` now, it's now supported by github. We may still have some `kotlin` annotations but it's just because we've not got around to changing them.

Baccata (approved):

PR #1143, opened 2023-08-09
switch kotlin marker to smithy

yisraelU:
update all references of kotlin to smithy now that MD supports smithy annotation


lewisjkl:
Won't this get rid of all syntax highlighting? Or does prism/highlight/whatever support smithy now?

Baccata:
> Or does prism/highlight/whatever support smithy now?

~It does :) it's called "linguist"~

eeeeeergh yeah you're right, we're using prism on the microsite -_____-'

lewisjkl:
Yeah, which I think we could provide some custom regexes and Prism can use them. Never tried it myself, but maybe we could reuse from the VSCode highlighting... https://prismjs.com/extending.html

Baccata:
Apparently docusaurus can use the shiki highlighter, which uses textmate grammars, which is what we use to define highlighting in vscode.

https://www.npmjs.com/package/docusaurus-preset-shiki-twoslash

So I believe if we found a way to load our textmate grammar in shiki, we could do the full switch without angering docusaurus. Docs for shiki are there : https://github.com/shikijs/shiki#configuration-and-options

Baccata:
And here are the docs on how to do just that : https://github.com/shikijs/shiki/blob/main/docs/languages.md#supporting-your-own-languages-with-shiki 

Now it'd just be a matter of understanding how to wire this altogether

yisraelU:
ok , so IIUC , we can keep the smithy annotation and benefit from linguist support for Smithy, we would just need to provide prism with a custom language grammar.

Baccata:
> ok , so IIUC , we can keep the smithy annotation and benefit from linguist support for Smithy, we would just need to provide prism with a custom language grammar.

Exactly, that, or we switch highlighters to shiki which actually supports the same grammar language that we already use for vscode, which means we wouldn't have to re-implement the grammar

PR #1144, opened 2023-08-10
Add Projections and Visitors to Unions

lewisjkl:
- Adds projection methods for each union alternative (basically `U => Option[A]`)

Note that I placed these methods inside of an object called `project`. The reason for this is to avoid collisions (there were already quite a few when I had them at the top level).

- Adds a Visitor trait in each union companion object and an `accept` method inside of each union trait

---

Also note that some changes in this PR are from correcting some places where strings were used in the renderer when `NameDef` or `NameRef` should have been used. Using strings rather than these was leading to compile errors from collisions in several places.

See the docs in the PR for more information.

### DIFF
--- a/modules/bootstrapped/src/generated/smithy4s/example/CheckedOrUnchecked.scala
+++ b/modules/bootstrapped/src/generated/smithy4s/example/CheckedOrUnchecked.scala
@@ -50,6 +50,14 @@ object CheckedOrUnchecked extends ShapeTag.Companion[CheckedOrUnchecked] {
     def raw(value: String): A
   }
 
+  object Visitor {
+    trait Default[A] extends Visitor[A] {
+      def default: A
+      def checked(value: String): A = default
+      def raw(value: String): A = default
+    }
+  }
+
   implicit val schema: Schema[CheckedOrUnchecked] = union(
     CheckedOrUnchecked.CheckedCase.alt,
     CheckedOrUnchecked.RawCase.alt,
--- a/modules/bootstrapped/src/generated/smithy4s/example/CheckedOrUnchecked.scala
+++ b/modules/bootstrapped/src/generated/smithy4s/example/CheckedOrUnchecked.scala
@@ -8,14 +8,24 @@ import smithy4s.schema.Schema.bijection
 import smithy4s.schema.Schema.string
 import smithy4s.schema.Schema.union
 
-sealed trait CheckedOrUnchecked extends scala.Product with scala.Serializable {
+sealed trait CheckedOrUnchecked extends scala.Product with scala.Serializable { self =>
   @inline final def widen: CheckedOrUnchecked = this
   def $ordinal: Int
+
+  object project {
+    def checked: Option[String] = CheckedOrUnchecked.CheckedCase.alt.project.lift(self).map(_.checked)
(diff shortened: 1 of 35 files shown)


modules/bootstrapped/src/generated/smithy4s/example/CheckedOrUnchecked.scala
```diff
@@ -26,18 +36,23 @@ object CheckedOrUnchecked extends ShapeTag.Companion[CheckedOrUnchecked] {
...
     val alt = schema.oneOf[CheckedOrUnchecked]("checked")
   }
   object RawCase {
     val hints: Hints = Hints.empty
-    val schema: Schema[RawCase] = bijection(string.addHints(hints), RawCase(_), _.raw)
+    val schema: Schema[CheckedOrUnchecked.RawCase] = bijection(string.addHints(hints), CheckedOrUnchecked.RawCase(_), _.raw)
     val alt = schema.oneOf[CheckedOrUnchecked]("raw")
   }
 
+  trait Visitor[A] {
```
Baccata, line 48: this needs to be accompanied by a a companion object that goes : 

``` 
object Visitor {
      def default: A
      // implement all methods in terms of `default`
   }
}
```

Baccata:
> Note that I placed these methods inside of an object called project. The reason for this is to avoid collisions (there were already quite a few when I had them at the top level). 

That's fine I think. I wouldn't have assumed it was the case, but maybe when the union is an ADT and all members share a mixin it creates a risk ...

kubukoz:
@Baccata I was also thinking it'd conflict with the fields of the union's members if you use adtMember etc.

Baccata:
> @Baccata I was also thinking it'd conflict with the fields of the union's members if you use adtMember etc.

yeah makes sense

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -892,11 +892,47 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     )
   }
 
-  private def caseName(alt: Alt): NameRef = alt.member match {
-    case UnionMember.ProductCase(product) => NameRef(product.name)
-    case UnionMember.TypeCase(_) | UnionMember.UnitCase =>
-      NameRef(alt.name.dropWhile(_ == '_').capitalize + "Case")
-  }
+  private def caseName(unionName: NameRef, alt: Alt): NameRef =
```
yisraelU, line 895: wondering if it might be beneficial to add a fold method on Union , as this pattern is repeated 4 times

yisraelU, line 895: wondering if it might be beneficial to add a fold method on Union , as this pattern is repeated 4 times

lewisjkl: Idk I think the hit to overall readability is substantial with something like the following vs a pattern match:

```scala
private def caseNameForMatch(unionName: NameRef, alt: Alt): NameRef =
    alt.member.fold(_ =>
      unionName.down(alt.name.dropWhile(_ == '_').capitalize + "Case")
    )(product => unionName.down(product.name))(
      unionName.down(alt.name.dropWhile(_ == '_').capitalize + "Case.type")
    )
```

Baccata (approved):

Issue #1145, opened 2023-08-11
Services should not return bodies in responses of `HEAD` requests 


Baccata:
We've recently changed the logic in both the 0.17 and 0.18 lineages to autofill response bodies with `{}` under certain circumstances, in order to align with AWS semantics. 

This new behaviour is arguably incorrect in the context of `HEAD` requests, as the corresponding responses are not supposed to contain a body. 

On 0.17, the fix is tricky as the addition of an empty json object body is coupled to the entity decoder which doesn't have the information. We may have to resort some http4s-level middleware under the hood. 
On 0.18, the fix should be easier

kubukoz:
Q: should this affect reponses with the NoContent status?

Baccata:
Well, somebody's got an issue for it : https://github.com/disneystreaming/smithy4s/issues/1169

kubukoz:
Not an accident :p

PR #1148, opened 2023-08-11
Head request empty body

lewisjkl:
Proposed solution for #1145

### DIFF
--- a/modules/http4s/src/smithy4s/http4s/internals/SimpleRestJsonCodecs.scala
+++ b/modules/http4s/src/smithy4s/http4s/internals/SimpleRestJsonCodecs.scala
@@ -117,7 +117,7 @@ private[http4s] class SimpleRestJsonCodecs(
       val input = messageDecoderCompiler
       val output = responseEncoderCompiler
 
-      val error = responseEncoderCompiler.mapK {
+      val error = responseEncoderCompiler.mapK[EncoderFromHints[F, *], ResponseEncoder[F, *]] {
         new PolyFunction[EncoderFromHints[F, *], ResponseEncoder[F, *]] {
           def apply[A](fromHints: EncoderFromHints[F, A]): ResponseEncoder[F, A] = fromHints(Hints.empty)
         }
--- a/modules/http4s/src/smithy4s/http4s/internals/SimpleRestJsonCodecs.scala
+++ b/modules/http4s/src/smithy4s/http4s/internals/SimpleRestJsonCodecs.scala
@@ -31,7 +31,6 @@ import smithy4s.json.Json
 import smithy4s.schema.CachedSchemaCompiler
 import cats.syntax.all._
 import cats.Eq
-import smithy4s.kinds._
 
 private[http4s] class SimpleRestJsonCodecs(
     val maxArity: Int,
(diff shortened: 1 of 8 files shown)


modules/http4s/src/smithy4s/http4s/internals/SmithyHttp4sServerEndpoint.scala
```diff
@@ -94,14 +95,22 @@ private[http4s] class SmithyHttp4sServerEndpointImpl[F[_], Op[_, _, _, _, _], I,
...
+
   override val httpApp: HttpApp[F] = {
     val baseApp = HttpApp[F] { req =>
       val run: F[O] = for {
         input <- inputDecoder.read(req)
         output <- (impl(endpoint.wrap(input)): F[O])
       } yield output
 
-      run.map(outputEncoder.write(successResponseBase, _))
+      run.map(outputEncoder.write(successResponseBase, _)).map(headRemoveBody)
```
Baccata: unfortunately, that's not a suitable solution (for 0.18 anyway), as the code you've modified is meant to be protocol agnostic and the removal of bodies is a protocol specific detail. 

What needs to change is this, we essentially need to NOT set a body if the method is HEAD https://github.com/disneystreaming/smithy4s/blob/series/0.18/modules/http4s/src/smithy4s/http4s/internals/SimpleRestJsonCodecs.scala#L97

modules/http4s/src/smithy4s/http4s/internals/SimpleRestJsonCodecs.scala
```diff
@@ -73,38 +79,73 @@ private[http4s] class SimpleRestJsonCodecs(
...
         def fromSchema[A](schema: Schema[A]) = fromSchema(schema, createCache())
 
-        def fromSchema[A](schema: Schema[A], cache: Cache) = if (schema.isUnit) {
-          restSchemaCompiler.fromSchema(schema, cache)
-        } else {
-          restSchemaCompiler
-            .fromSchema(schema, cache)
-            .andThen(addEmptyJsonToResponse(_))
-        }
+        def fromSchema[A](schema: Schema[A], cache: Cache) = endpointHints =>
```
Baccata: That's better, but an arguably better approach would be to modify Schema before it's passed to this EndpointCompiler by forwarding the hint to the output schema, instead of changing the type the compiler works against : 

``` 
output.fromSchema(endpoint.output.addHints(endpoint.hints), responseEncoderCache)
```

daddykotex: is there a possibility that the hints we would be adding from the endpoint affect the compiler behaviour?

I don't think the `smithy.api.Http` has a chance of doing that but maybe other hints could?

can't think of such hints at the moment but eh

Baccata: @daddykotex I don't think it's much of concern here, but if you're actually worried about it, it's totally possible to create private hints directly in Scala and use them it ways that would have pretty much zero chance of interfering with existing interpreter behaviour.

Baccata:
Closing as this is taken care of here : https://github.com/disneystreaming/smithy4s/blob/d3339368ed44e4209f81d1c146164e4b36d91d39/modules/core/src/smithy4s/http/HttpUnaryServerCodecs.scala#L180-L182

PR #1151, opened 2023-08-14
Add support for ternary `Removable` type, to facilitate "merge-patch"-like usecases 

Baccata:
Adds a `smithy4s.Removable` construct, to help model patchable data where a field can be in three different states : 

1. present
2. absent 
3. removed

At the schema level, it is modelled as `Option[Option[A]]` which prevents yet another impactful change. The `alloy#nullable` trait is also expected to be applied on the field. Without it the deserialisation cannot distinguish between `Option[None]` and `None`. 

At the codegen level, the signal to render `Removable` is the presence of the `alloy#nullable` trait on a field that's not also annotated with `smithy.api#required`. 

#### TODOs 

- [ ] In alloy, add the  `alloy#nullable` to the list of traits known by the `simpleRestJson` protocol
- [ ] See whether the addition of compliance tests is doable (it may not be the case, considering how smithy works) 
- [ ] add documentation in both alloy and smithy4s 
- [ ] add support in the dynamic module


modules/core/src/smithy4s/internals/DocumentDecoderSchemaVisitor.scala
```diff
@@ -209,12 +210,13 @@ class DocumentDecoderSchemaVisitor(
   def option[A](schema: Schema[A]): DocumentDecoder[Option[A]] =
     new DocumentDecoder[Option[A]] {
       val decoder = schema.compile(self)
+      val aIsNullable = schema.hints.get(Nullable).isDefined && schema.isOption
```
daddykotex, line 213: minor we have `schema.hints.has` which might make this easier to read

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -1255,7 +1258,10 @@ private[smithy4s] class SchemaVisitorJCodec(
       }
 
       def decodeValue(cursor: Cursor, in: JsonReader): Option[A] =
-        if (in.isNextToken('n'))
+        // if `A` is an option and has nullable, we delegate the handling of `null` to it.
+        // This allows for supporting Json-merge patches, where the absence of value
+        // and the presence of "null" have different meanings.
+        if (in.isNextToken('n') && !aIsNullable)
```
daddykotex, line 1270: shouldn't it be `if (in.isNextToken('n') && aIsNullable)` ? my understanding is that if the next token is a `n` AND we are looking for `null`, then we can read a `null`

daddykotex: but the test right below tells me everything is good

Baccata: It's testing if the underlying schema is nullable, which means the option can delegate to the underlying decoder for null-handling. If the `A` is not nullable however, the option needs to take care of it.

lewisjkl:
> In alloy, add the alloy#nullable to the list of traits known by the simpleRestJson protocol

If I'm not mistaken, this means we'll need to do a minor version bump (rather than patch) as to not mess with any existing functionality. I think it probably wouldn't mess anything up, but I'm not positive since it will change the HintMasks we derive from the protocolTraits

Baccata:
Superseded by https://github.com/disneystreaming/smithy4s/pull/1408

PR #1155, opened 2023-08-15
HEAD response fix (again)

lewisjkl:
I realized that my excluding of the Content-Type and Content-Length headers from HEAD requests was incorrect. From the HTTP Spec, it is supposed to be the case that the headers on a HEAD request are exactly as they would be on the same request with a GET method, the only difference is the body is not actually returned. I also updated the test to be more permissive of having extra headers in place as some HTTP libraries add other headers by default which is fine.

(I will update #1148 to match this behavior)

### DIFF
--- a/modules/tests/src/smithy4s/tests/PizzaSpec.scala
+++ b/modules/tests/src/smithy4s/tests/PizzaSpec.scala
@@ -355,7 +355,7 @@ abstract class PizzaSpec
           "Content-Type" -> List("application/json")
         )
         val containsProperHeaders =
-          expectedHeaders.forall(h => headers.get(h._1).isDefined)
+          expectedHeaders.forall(h => headers.get(h._1).contains(h._2))
         expect.same(code, 200) &&
         expect.same(body, "") &&
         expect(
--- a/modules/tests/src/smithy4s/tests/PizzaSpec.scala
+++ b/modules/tests/src/smithy4s/tests/PizzaSpec.scala
@@ -329,7 +329,7 @@ abstract class PizzaSpec
         "Content-Type" -> List("application/json")
       )
       val containsProperHeaders =
-        expectedHeaders.forall(h => headers.get(h._1).isDefined)
+        expectedHeaders.forall(h => headers.get(h._1).contains(h._2))
       expect.same(code, 200) &&
       expect.same(body, "") &&
       expect(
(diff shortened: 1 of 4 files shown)


modules/tests/src/smithy4s/tests/PizzaSpec.scala
```diff
@@ -312,27 +312,59 @@ abstract class PizzaSpec
...
     } yield {
       val (code, headers, body) = res
+      val expectedHeaders = Map(
+        "Test" -> List("test"),
+        // response would have been `{}` if this was a GET
+        "Content-Length" -> List("2"),
+        "Content-Type" -> List("application/json")
+      )
+      val containsProperHeaders =
+        expectedHeaders.forall(h => headers.get(h._1).isDefined)
```
daddykotex: should you also use `expectedHeaders.forall(h => headers.get(h._1).contains(h._2))` here?

lewisjkl: Ah yeah, thanks will update that now

daddykotex (approved):

lewisjkl:
Okay, so reading the [full HTTP spec for HEAD requests](https://www.rfc-editor.org/rfc/rfc9110.html#name-head), I think we should actually leave it as is with no headers for content length or content type. You can see justification for this in the spec:

>The server SHOULD send the same header fields in response to a HEAD request as it would have sent if the request method had been GET. However, a server MAY omit header fields for which a value is determined only while generating the content. For example, some servers buffer a dynamic response to GET until a minimum amount of data is generated so that they can more efficiently delimit small responses or make late decisions with regard to content selection. Such a response to GET might contain [Content-Length](https://www.rfc-editor.org/rfc/rfc9110.html#field.content-length) and [Vary](https://www.rfc-editor.org/rfc/rfc9110.html#field.vary) fields, for example, that are not generated within a HEAD response. These minor inconsistencies are considered preferable to generating and discarding the content for a HEAD request, since HEAD is usually requested for the sake of efficiency.

daddykotex:
god that spec is complicated
so many implementation details spills out

lewisjkl:
> god that spec is complicated so many implementation details spills out

For real.. wish it was just explicit in this case because every http framework seems to handle it differently _sigh_.

PR #1159, opened 2023-08-18
[0.18] Fix ClassCastException in document encoding

### LINKED ISSUE #1158: Scala 3: ClassCastException when encoding empty document arrays
To reproduce:

```scala
//> using dep "com.disneystreaming.smithy4s::smithy4s-json:0.17.17"
//> using scala "3.3.0"
import smithy4s.Document
import smithy4s.schema.Schema

val capi = smithy4s.http.json.codecs()
val codec = capi.compileCodec(Schema.document)

println(
  new String(
    capi.writeToArray(
      codec,
      Document.array()
    )
  )
)

```

This throws:

```
java.lang.ClassCastException: class [Ljava.lang.Object; cannot be cast to class [Lsmithy4s.Document; ([Ljava.lang.Object; is in module java.base of loader 'bootstrap'; [Lsmithy4s.Document; is in unnamed module of loader java.net.URLClassLoader @7558c8ef)
	at smithy4s.http.json.SchemaVisitorJCodec$$anon$17.encodeValue(SchemaVisitorJCodec.scala:328)
	at smithy4s.http.json.SchemaVisitorJCodec$$anon$17.encodeValue(SchemaVisitorJCodec.scala:320)
	at com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter.write(JsonWriter.scala:934)
	at com.github.plokhotnyuk.jsoniter_scala.core.package$.writeToArray(package.scala:405)
	at smithy4s.http.json.JsonCodecAPI.writeToArray(JsonCodecAPI.scala:91)
	at repl.MdocSession$MdocApp.<init>(Main.worksheet.sc:26)
	at repl.MdocSession$.app(Main.worksheet.sc:3)

```

Affects Scala 3 only (2.13 is fine, at least). 

I did a quick check of a possible regression and:

- Works fine in 0.13.0 (the first release with the current Schema encoding) and 0.13.7 (the last 0.13 release)
- Breaks in [0.14.0](https://github.com/disneystreaming/smithy4s/releases/tag/v0.14.0) (didn't narrow it down to a specific PR but my guess is https://github.com/disneystreaming/smithy4s/pull/297 as it touched the relevant lines) and in the latest 0.17 release, 0,17.17.

kubukoz:
Closes #1158.

As far as I understand, `case x: ArraySeq[Document]` isn't actually doing a class check because of type erasure.

The unsafe array inside turns out to actually be an `Array[Object]`, so casting it to `Array[Document]` throws because arrays aren't erased (unlike `ArraySeq` which is a normal generic class).

I'm sure this solution is a bit inefficient, another one I had in mind would confirm the identity of the wrapped array:

```scala
a.value match {
  case x: ArraySeq[_]
      if x.unsafeArray.getClass == classOf[Array[Document]] =>
    val xs = x.unsafeArray.asInstanceOf[Array[Document]]
    var i = 0
    while (i < xs.length) {
      encodeValue(xs(i), out)
      i += 1
    }
  case xs =>
    xs.foreach(encodeValue(_, out))
}
```

but I'm not sure it's better as I didn't run any benchmarks. @plokhotnyuk any thoughts on this PR?

Note: the fix for the issue should be backported to 0.17.

EDIT: updated the implementation to be closer to the original, just checking for empty arrays instead. Perhaps it's worth a property-based test of roundtripping arbitrary documents just in case.

### DIFF
--- a/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
+++ b/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
@@ -396,11 +396,15 @@ private[smithy4s] class SchemaVisitorJCodec(
         case a: DArray =>
           out.writeArrayStart()
           a.value match {
+            // short-circuiting on empty arrays to avoid the downcast to array of documents
+            // which has proven to be dangerous in Scala 3:
+            // https://github.com/disneystreaming/smithy4s/issues/1158
+            case x: ArraySeq[_] if x.isEmpty =>
             case x: ArraySeq[_] =>
-              val xs = x.unsafeArray
+              val xs = x.unsafeArray.asInstanceOf[Array[Document]]
               var i = 0
               while (i < xs.length) {
-                encodeValue(xs(i).asInstanceOf[Document], out)
+                encodeValue(xs(i), out)
                 i += 1
               }
             case xs =>
--- a/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
+++ b/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
@@ -399,13 +399,15 @@ private[smithy4s] class SchemaVisitorJCodec(
             // short-circuiting on empty arrays to avoid the downcast to array of documents
             // which has proven to be dangerous in Scala 3:
             // https://github.com/disneystreaming/smithy4s/issues/1158
-            case x: ArraySeq[_] if x.isEmpty =>
             case x: ArraySeq[_] =>
-              val xs = x.unsafeArray.asInstanceOf[Array[Document]]
-              var i = 0
-              while (i < xs.length) {
-                encodeValue(xs(i), out)
-                i += 1
+              if (x.isEmpty) ()
+              else {
+                val xs = x.unsafeArray.asInstanceOf[Array[Document]]
+                var i = 0
+                while (i < xs.length) {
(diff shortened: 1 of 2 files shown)


modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -396,7 +396,11 @@ private[smithy4s] class SchemaVisitorJCodec(
         case a: DArray =>
           out.writeArrayStart()
           a.value match {
-            case x: ArraySeq[Document] =>
+            // short-circuiting on empty arrays to avoid the downcast to array of documents
+            // which has proven to be dangerous in Scala 3:
+            // https://github.com/disneystreaming/smithy4s/issues/1158
+            case x: ArraySeq[_] if x.isEmpty =>
```
Baccata: I have no idea whether the compiler optimises this to avoid doing two class-checks : 

```scala
case a: A if (...) => 
case a: A
```

so I'd just nest the conditional : 

```scala
case x: ArraySeq[_] => 
  if (x.nonEmpty){
     ...
  } 
```

kubukoz: 4de846993612bf94944ff29b46e2f7532a6c81db

Baccata (approved):

PR #1163, opened 2023-08-21
remove rendering of required hint as it is redundant

### LINKED ISSUE #890: Proposal: bake `Required` trait into the `.required` method on schemas
In generated struct schemas, a field needs 2 things to be considered required:

- `.addHints(Required())`
- `.required`

```scala
case class SubscriptionCreated(subscriptionId: String)
object SubscriptionCreated extends ShapeTag.Companion[SubscriptionCreated] {
//...
  implicit val schema: Schema[SubscriptionCreated] = struct(
    string.required[SubscriptionCreated]("subscriptionId", _.subscriptionId).addHints(smithy.api.Required()),
  ){
    SubscriptionCreated.apply
  }.withId(id).addHints(hints)
}
```

I find this a bit strange - could we instead build `.addHints(Required)` into the `required` method and filter the trait out? Alternatively, we could avoid the special case but still have the hint added within `.required` to make it easier to handcraft these schemas, without complicating the codegen code.

lewisjkl:
Removes required trait from being rendered in hints since fields are already capturing whether or not they are required.

Closes #890.


kubukoz (commented):
This should probably be reflected in Dynamic too...

besides, I have a few concerns:

- this could be very much breaking for anyone who passes around Hints to determine if something is required or not. For example, I did this in a generalized [completion function in smithy-playground](https://github.com/kubukoz/smithy-playground/blob/78706948d4fc14230fa62a113b8df597d06ce656/modules/language-support/src/main/scala/playground/language/CompletionVisitor.scala#L142-L188), which isn't a biggie because I have tests, but still.
- if what we seek is less boilerplate in the generated code, how about we add the hint to all required fields (using an implementation in `required` itself)? I originally suggested something similar.
- it feels strange to generate definitions (`case class Required()`) for traits from the stdlib but hide some them from the smithy4s representation of the model.

Baccata:
I don't totally think we should do this anymore, now that the Field is not carrying the information about its optionality. 

> if what we seek is less boilerplate in the generated code, how about we add the hint to all required fields (using an implementation in required itself)? I originally suggested something similar. 

I now agree with you on that. 

> it feels strange to generate definitions (case class Required()) for traits from the stdlib but hide some them from the smithy4s representation of the model. 

I'm not sure I follow. Is that to reinforce your point about retaining the hint whilst not rendering it ?

kubukoz:
> Is that to reinforce your point about retaining the hint whilst not rendering it

It may've had the result of reinforcing that point, but I only meant to highlight my gut feeling (about it feeling strange) :)

kubukoz (approved):

Baccata (approved):

PR #1164, opened 2023-08-22
[0.17] Update smithyVersion to 1.36.0

ryanjackson27:


kubukoz (approved):

project/Dependencies.scala
```diff
@@ -20,7 +20,7 @@ object Dependencies {
 
   val Smithy = new {
     val org = "software.amazon.smithy"
-    val smithyVersion = "1.31.0"
```
daddykotex, line 23: if we rename that to `smithyModelVersion` we'll probably get scala-steward to work

kubukoz: how Steward works is a constant mystery to me

daddykotex: yeah the algorithm is kinda funky :P

PR #1165, opened 2023-08-22
Add parse member function

miguel-vila:
Adds a function to parse shape ids with members


daddykotex (approved):
That makes sense too me. Plus TIL that you can deconstruct an `Array` directly like that!

daddykotex:
Your branch is outdated though

modules/core/src/smithy4s/ShapeId.scala
```diff
@@ -40,12 +40,29 @@ object ShapeId extends ShapeTag.Has[ShapeId] { self =>
...
-  final case class Member(shapeId: ShapeId, member: String)
-
   val id: ShapeId = ShapeId("smithy4s", "ShapeId")
   lazy val schema =
     Schema.string.refined[ShapeId](IdRef()).withId(id)
 
+  final case class Member(shapeId: ShapeId, member: String)
+
+  object Member {
+    def parse(string: String): Option[Member] = {
```
miguel-vila, line 50: think if you call `ShapeId.parse` with something like `smithy4s#ShapeId$member` you will get `ShapeId("smithy4s", "ShapeId$member")` which is not entirely right

modules/core/src/smithy4s/ShapeId.scala
```diff
@@ -40,12 +40,29 @@ object ShapeId extends ShapeTag.Has[ShapeId] { self =>
     }
   }
 
-  final case class Member(shapeId: ShapeId, member: String)
```
kubukoz, line 43: Interesting, didn't know we had that. Was this used for anything?

miguel-vila: seems like not, there's a `withMember` method and that one is not used either

Baccata:
Sorry for only reviewing this now. Is there a specific usecase for it ? I was tempted to remover the `Member` case class, as we don't use it anywhere

Baccata:
Closing this as it doesn't help with any of the library features, and is simple enough to implement in userland. 

Happy to revisit if there's ever a justification that would warrant it.

PR #1166, opened 2023-08-22
Add constraints remover transformer for AWS specs

### LINKED ISSUE #1142: Add a pre-processor for removing constraints from shapes in the `com.amazon` namespaces
As Smithy4s evolves to take on the role of an AWS SDK, we need to be wary of potential warts of AWS. In particular, although some constraints seem to be present in the AWS specs found in the [javascript SDK repo](https://github.com/aws/aws-sdk-js-v3), they seem to not be completely respected by the AWS services. 

Considering we cannot easily ignore constraints in our interpreters, it feels like a safer bet to just remove the constraints at build time (at least length/range constraints) in order to ensure that the AWS clients we derive in smithy4s do not raise errors when AWS doesn't respect its own constraints. AWS even states that [constraints should not be enforced when deserialising an output](https://smithy.io/2.0/spec/constraint-traits.html?highlight=constraints#constraint-trait-enforcement) 

NB : one of the reasons we cannot easily disable validation of response deserialisation is that the mechanism that constraints are implemented with is the same as [type refinements](https://disneystreaming.github.io/smithy4s/docs/codegen/customisation/refinements). Refinements require a partial type transformation `A => Either[String, B]`, and disabling constraints would mean being able to dynamically test that the input type `A` and the output type `B` are one and the same, which we are not able to do with the current model. Changing the model is very impactful, and we want to reserve it for situations where we don't see any other possible solution. 

For a point of reference to this unit of work, head over to this PR : https://github.com/disneystreaming/smithy4s/pull/1110

zetashift:
Fixed https://github.com/disneystreaming/smithy4s/issues/1142


modules/codegen/src/smithy4s/codegen/transformers/AwsConstraintsRemover.scala
```diff
@@ -0,0 +1,48 @@
...
+    val transformer = context.getTransformer()
+    transformer.removeTraitsIf(
+      context.getModel(),
+      (shape, t) => {
+        val shapeName = t.toShapeId().getName()
+        shape.getId.getNamespace.startsWith(
+          awsNamespacePrefix
+        ) && constraintsToRemove.contains(shapeName)
+      }
+    )
```
zetashift: Is this how it should be done? The PR https://github.com/disneystreaming/smithy4s/pull/1110 was quite a lot more involved, so I'm not sure if I'm missing something crucial here.

modules/codegen/src/smithy4s/codegen/transformers/AwsConstraintsRemover.scala
```diff
@@ -0,0 +1,48 @@
...
+import software.amazon.smithy.build.ProjectionTransformer
+import software.amazon.smithy.build.TransformContext
+import software.amazon.smithy.model.Model
+
+/** A preprocessor that removes constraints from specifications of AWS. */
+class AwsConstraintsRemover extends ProjectionTransformer {
+
+  override def getName: String = AwsConstraintsRemover.name
+
+  private val constraintsToRemove = List("length", "range", "pattern")
```
zetashift: I just picked these 3 out initially from the list of constraint traits, but if any more need to be added, let me know. So far I've only had problems with `length` and `range`

yisraelU: Any reason to avoid including all listed here https://smithy.io/2.0/spec/constraint-traits.html

kubukoz: side note, no clue why `private` is considered a constraint trait because it doesn't conform to the definition:

> Constraint traits are used to constrain the values that can be provided for a shape.

it's a smithy model thing only.

zetashift: > Any reason to avoid including all listed here https://smithy.io/2.0/spec/constraint-traits.html

I had some doubts if the other ones really needed to be included, so I put it up like this and if needed I'll just add it more! :P

Baccata: Others don't need to be included : the only reason we're including those is because of how they are treated in smithy4s.

modules/codegen/src/smithy4s/codegen/transformers/AwsConstraintsRemover.scala
Baccata: Well, in theory you should only remove them if they are used in output types, which does make it quite more involved. 

There's a construct provided by the "smithy-model" library called "NeighbourVisitor" or something, which allows to get a transitive closure from a shape in the model. 

So the algorithm that I'm looking for is : 

1. lookup for all operations in the aws namespace
2. map them to get their input shapes 
3. flatMap the input shapes with the neighbour visitor to transitively gather all shapes their are referencing 
4. remove the constraint traits on the set of shapes you get in 3

zetashift: I've gotten this:
```scala
        val shapeName = t.toShapeId().getName()
        val isOperation = shape.isOperationShape()
        val isInAwsNamespace =
          shape.getId.getNamespace.startsWith(awsNamespacePrefix)

        if (isOperation && isInAwsNamespace ) {
          val np = NeighborProvider.of(context.getModel())
          val neighbors = np.getNeighbors(shape).asScala.toList
        }
```

But I don't know Smithy so well that I can work the algorithm out... I've pushed the code I have so far, could use some more guidance on how to get it finished :P

modules/codegen/src/smithy4s/codegen/transformers/AwsConstraintsRemover.scala
```diff
@@ -0,0 +1,56 @@
...
+  private val awsNamespacePrefix = "com.amazonaws"
+
+  override def transform(context: TransformContext): Model = {
+    val transformer = context.getTransformer()
+    transformer.removeTraitsIf(
```
Baccata, line 57: Okay, so you need to build a collection of all shapes connected to AWS outputs, before calling the `removeTraitsIf`. I've not tested it, but it should look like this (I should have pointed to the `Walker` construct in addition to the NeighborVisitor one, my bad)

```scala
    val model = context.getModel()
    val walker: Walker = new Walker(
      NeighborProvider.cached(NeighborProvider.of(context.getModel()))
    )
    val awsOperations = model
      .getOperationShapes()
      .asScala
      .filter(_.getId().getNamespace().startsWith(awsNamespacePrefix))

    val awsOutputShapes =
      awsOperations.flatMap(_.getInput().toScala.toSet).map(model.expectShape)
    val awsOutputConnectedShapes: Set[ShapeId] =
      awsOutputShapes.flatMap(walker.walkShapeIds(_).asScala)

```

The condition for removing the trait is basically a check to verify whether the shape's ID is in the `awsOutputConnectedShape` set.

daddykotex:
Hey @zetashift , I took the initiative to try and help on this PR because we'll need it for the 0.18 release and it's coming shortly. Hopefully, it does not bother you, if it does, feel free to remove my commit and I'll step back and let you wrap it.

With that said, I tried to implement what @Baccata suggested in their comments. I still have one question: the spec says constraint should not be enforce when serializing input or deserializing output.

So far the code I've written, based off @Baccata 's suggestion, but this bit left me wondering:

```
 val awsOutputShapes =
      awsOperations.flatMap(_.getInput().toScala.toSet).map(model.expectShape)
```

Do we want to remove constraints from input only, output only, or both? The suggested code uses `awsOutputShapes` name, but also `getInput` so I was a little bit confused.

zetashift:
@daddykotex oh sweet thank you! No complaints here from me, wish I had the time to finish it before a 0.18 release(I've been a bit busy with new job hunting).

>Do we want to remove constraints from input only, output only, or both? The suggested code uses awsOutputShapes name, but also getInput so I was a little bit confused.

I believe the goal is to only remove constraints from the output.

daddykotex:
No worries, and good luck on the job hunt :)
Thanks for your work

Baccata:
> Do we want to remove constraints from input only, output only, or both? The suggested code uses awsOutputShapes name, but also getInput so I was a little bit confused. 

The `getInput` is a mistake 😅, but yeah, remove constraints from anything transitively accessible from output types 👍

Baccata (approved):

PR #1179, opened 2023-08-31
Large rewrite of the internals

Baccata:
This really large restructuring has the following goals : 

1. streamline the construction of server/client http codecs, in particular to make the AWS module more maintainable 
2. decouples http routing/reverse-routing logic from http4s as much as possible, so that it's easier to integrate smithy4s with other http-backends or protocols 

In order to achieve these goals, a few things are happening here : 

1. introducing of `smithy4s.capability.MonadThrowLike . It's not great, but we're working with the constraint that the core module should not have any dependency. This is not meant to be something that end-users are exposed to, just something that people who want to integrate smithy4s with other libraries than what we provide OOTB can use to facilitate this integration
2. introducing `smithy4s.http.HttpRequest/HttpResponse`. This drastically facilitates the generalisation of some constructs in a backend-agnostic. The integration is then reduced to some functions that take care of converting from those models to the models of their http libraries of choice. 

The changes in the `http4s` and `aws-http4s` module show the benefits : 
* the integration for each AWS protocol is much more concise and doesn't require too much mental contorsion 
* the `http4s-kernel` and `http4s` module has now each have much less code in it, which implies that integrating with other http libraries (which is a $work requirement) will be much less painful, as many bugs in the protocol implementation will now require a single fix instead of N.


modules/core/src/smithy4s/Endpoint.scala
```diff
@@ -63,75 +63,114 @@ object Endpoint {
...
+      }
 
   }
 
   type ForOperation[Op[_, _, _, _, _]] = {
     type e[I, E, O, SI, SO] = Endpoint[Op, I, E, O, SI, SO]
   }
 
-  final case class Builder[Op[_, _, _, _, _], I, E, O, SI, SO] private (
+  final case class Builder[I, E, O, SI, SO] private (
```
Baccata: @kubukoz, so one thing that's appearing more and more obvious is that the coupling of `Endpoint.Base` is a very valuable construct on its own, and it's annoying that `Endpoint` uses inheritance instead of composition. So I think it's gonna be important to decouple `Endpoint` with `Endpoint.Base`, which also implies decoupling the builders. 

I think this is one of the things that will be extracted in a separate PR

modules/core/src/smithy4s/Endpoint.scala
```diff
@@ -63,75 +63,114 @@ object Endpoint {
     def errorable: Option[Errorable[E]]
     def streamedInput: StreamingSchema[SI]
     def streamedOutput: StreamingSchema[SO]
+    def builder: Builder[I, E, O, SI, SO] = Builder.fromEndpoint(this)
+  }
+
+  trait Middleware[A] { self =>
```
Baccata, line 76: `Endpoint.Middleware` can be generalised and hosted by the core module. It's actually very valuable to have access to it, as it allows to generalise the routing logic.

daddykotex:
Pushed a commit on here to resolve compilation issues that I noticed

daddykotex:
Published under: `0.18.0-942-9120462`

PR #1187, opened 2023-09-07
Fix and optimize AWS kernel JS/Native-platform specific code

armanbilge:
- JS stuff was essentially just tidying / refactoring
- Native stuff was leaking memory and had no error checking in its previous state. This is all fixed up now.

### DIFF
--- a/modules/aws-kernel/src-js/smithy4s/aws/CryptoPlatformCompat.scala
+++ b/modules/aws-kernel/src-js/smithy4s/aws/CryptoPlatformCompat.scala
@@ -24,16 +24,10 @@ trait CryptoPlatformCompat {
   def binaryFromString(str: String): Binary = Buffer.from(str, utf8)
 
   def toHexString(binary: Binary): String =
-    binary
-      .entries()
-      .toIterator
-      .flatMap(_.lastOption)
-      .map(n => f"$n%02x")
-      .mkString
+    binary.toString("hex")
 
-  def sha256HexDigest(message: Array[Byte]): String = {
+  def sha256HexDigest(message: Array[Byte]): String =
     this.sha256HexDigest(new String(message))
-  }
 
   def sha256HexDigest(message: String): String = {
     val hash = Crypto.createHash(sha256)
@@ -50,9 +44,9 @@ trait CryptoPlatformCompat {
 }
 
 object CryptoPlatformCompat {
-  val sha256 = "sha256"
-  val utf8 = "UTF-8"
-  val hex = "hex"
-  val binary = "binary"
-  val ascii = "ascii"
+  final val sha256 = "sha256"
+  final val utf8 = "UTF-8"
+  final val hex = "hex"
+  final val binary = "binary"
+  final val ascii = "ascii"
 }
--- a/modules/aws-kernel/src-native/smithy4s/aws/CryptoPlatformCompat.scala
+++ b/modules/aws-kernel/src-native/smithy4s/aws/CryptoPlatformCompat.scala
@@ -16,12 +16,15 @@
 
 package smithy4s.aws.kernel
 
-import scala.scalanative.unsafe
 import scala.scalanative.unsafe._
 import scala.scalanative.unsigned._
 
+import crypto._
+
 trait CryptoPlatformCompat {
   type Binary = Array[Byte]
+
+  private[this] val EvpMdSha256 = EVP_sha256()
 
   def binaryFromString(str: String): Binary =
     str.getBytes("UTF-8")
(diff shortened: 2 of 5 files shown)


modules/aws-kernel/src-js/smithy4s/aws/CryptoPlatformCompat.scala
```diff
@@ -50,9 +44,9 @@ trait CryptoPlatformCompat {
 }
 
 object CryptoPlatformCompat {
-  val sha256 = "sha256"
-  val utf8 = "UTF-8"
-  val hex = "hex"
-  val binary = "binary"
-  val ascii = "ascii"
+  final val sha256 = "sha256"
```
Baccata, line 47: weird, I'd have thought that was implied in static objects. What's the effect of adding final here ? 

Also, do you happen to have a `list of do's/don'ts for Scala library maintainers` somewhere ?

armanbilge: I learned this one from Seb :) per spec the Scala compiler treats these specially.

> A constant value definition is of the form
> ```scala
> final val x = e
> ```
> where `e` is a [constant expression](https://www.scala-lang.org/files/archive/spec/2.13/06-expressions.html#constant-expressions). The final modifier must be present and no type annotation may be given. References to the constant value `x` are themselves treated as constant expressions; in the generated code they are replaced by the definition's right-hand side `e`.

https://www.scala-lang.org/files/archive/spec/2.13/04-basic-declarations-and-definitions.html#value-declarations-and-definitions

modules/aws-kernel/src-native/smithy4s/aws/CryptoPlatformCompat.scala
```diff
@@ -16,90 +16,63 @@
...
+  def hmacSha256(data: String, key: Binary): Binary = {
+    val md = new Array[Byte](EVP_MAX_MD_SIZE)
+    val mdLen = stackalloc[CUnsignedInt]()
+    if (
+      HMAC(
+        EvpMdSha256,
+        key.atUnsafe(0),
+        key.size.toInt,
+        data.getBytes("UTF-8").atUnsafe(0),
+        data.length.toULong,
```
armanbilge: Actually I made a bug here. The length of the string may not be equivalent to its length in UTF-8 bytes. Let me push a fix.

modules/aws-kernel/src-native/smithy4s/aws/CryptoPlatformCompat.scala
```diff
@@ -16,90 +16,64 @@
...
-        i += 1
-      }
-
-      out.map("%02x".format(_)).mkString
-    }
+  def sha256HexDigest(message: Array[Byte]): String = {
+    val md = new Array[Byte](EVP_MAX_MD_SIZE)
+    val size = stackalloc[CUnsignedInt]()
+    if (
+      EVP_Digest(
```
Baccata, line 39: well I feel very dumb 😄 

Although, this reminds me of something I wanted to ask your opinion about : in the large majority of usecases, the AWS signature algorithm implies that the whole body of the outgoing http request lives in memory. 
Because that's the case, and because at this stage correctness matters much more than performance, I kinda went "well, might as well use byte arrays, duh". Without thinking twice about it, I did this : 

https://github.com/disneystreaming/smithy4s/blob/series/0.18/modules/aws-http4s/src/smithy4s/aws/internals/AwsSignatureClient.scala#L126

I'm wondering whether it'd be beneficial for the digest to be computed with knowledge of how the Chunk is structured, instead of just forcing a `.toArray` call. At the moment, the chunk is likely to be backed by a single `Array[Byte]`, but it may evolve in the future.

armanbilge: > I'm wondering whether it'd be beneficial for the digest to be computed with knowledge of how the Chunk is structured

Hmm, since that's the http4s integration, couldn't we just use `fs2.hash.sha256` there?

Baccata: lol, good point

Baccata (approved):

Baccata:
Thanks !

modules/aws-kernel/src-js/smithy4s/aws/CryptoPlatformCompat.scala
kubukoz: I think this has an impact on the inferred type too, as `final` fields are singleton types. BuildInfo also has an option to turn this on btw

kubukoz: not sure but maybe a type annotation would work _if_ it's the singleton type 🤔

PR #1199, opened 2023-09-18
Revamped Endpoint to use composition instead of inheritance

Baccata:
This PR makes changes that are required https://github.com/disneystreaming/smithy4s/pull/1179

* Added smithy4s.schema.OperationSchema
* Changed the structure of `smithy4s.Endpoint`
* Change a number of names
* Changed the rendering of operations 
* Removed `Endpoint.Builder` as the composition of `OperationSchema` solves the same problem.


modules/core/src/smithy4s/schema/OperationSchema.scala
```diff
@@ -0,0 +1,98 @@
...
+ *  limitations under the License.
+ */
+
+package smithy4s
+
+package schema
+
+import smithy4s.Errorable
+
+final case class OperationSchema[I, E, O, SI, SO](
```
Baccata, line 23: Start review here. This has the responsibility of carrying schemas associated to an operation.

modules/core/src/smithy4s/Errorable.scala
```diff
@@ -16,8 +16,14 @@
 
 package smithy4s
 
-trait Errorable[E] {
-  def error: schema.Schema.UnionSchema[E]
+trait Errorable[E] extends ShapeTag[E] {
```
Baccata, line 19: I think I'm gonna move this in the `schema` package in a subsequent PR, and rename to `ErrorSchema`

modules/core/src/smithy4s/Endpoint.scala
```diff
@@ -39,131 +41,48 @@ package smithy4s
...
-trait Endpoint[Op[_, _, _, _, _], I, E, O, SI, SO]
-    extends Endpoint.Base[I, E, O, SI, SO] {
+// scalafmt: {maxColumn = 120}
+trait Endpoint[Op[_, _, _, _, _], I, E, O, SI, SO] {
+
+  final def mapSchema(
+      f: OperationSchema[I, E, O, SI, SO] => OperationSchema[I, E, O, SI, SO]
+  ): Endpoint[Op, I, E, O, SI, SO] = Endpoint(f(schema), wrap)
 
+  def schema: OperationSchema[I, E, O, SI, SO]
```
Baccata, line 51: this is now using composition over inheritance. Means we can have a `mapSchema` easily, rendering the builder pattern obsolete.

lewisjkl (approved):

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -563,33 +558,33 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
       )(
-        renderId(op.shapeId),
-        line"val input: $Schema_[${op.input}] = ${op.input.schemaRef}.addHints(smithy4s.internals.InputOutput.Input.widen)",
-        line"val output: $Schema_[${op.output}] = ${op.output.schemaRef}.addHints(smithy4s.internals.InputOutput.Output.widen)",
-        renderStreamingSchemaVal("streamedInput", op.streamedInput),
-        renderStreamingSchemaVal("streamedOutput", op.streamedOutput),
-        renderHintsVal(op.hints),
-        line"def wrap(input: ${op.input}) = ${opNameRef}($input)",
-        renderErrorable(op)
+        line"""def schema: $OperationSchema_[${op.renderAlgParams(
```
kubukoz, line 563: **question:** TLDR why is this a `def` and not a `val`?

modules/core/src/smithy4s/schema/OperationSchema.scala
kubukoz: it's very cool

kubukoz: should it have a private constructor for bincompat? you'd need something like `Schema.operation` in the companion, or make it package-private, I guess.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
Baccata: could be a val, np

modules/core/src/smithy4s/schema/OperationSchema.scala
Baccata: You mean to future-proof it ? I suppose it should.

kubukoz: that's what I meant yep

PR #1201, opened 2023-09-19
Ergnomics improvements

Baccata:


modules/bootstrapped/test/src/smithy4s/HintsSpec.scala
```diff
@@ -61,4 +62,23 @@ class HintsSpec() extends FunSuite {
     expect.same(concat.targetHints, Hints(HttpLabel()))
   }
 
+  test("Hints#add adds to the member layer") {
+    val concat = Hints.empty.add(HttpHeader("X-Member"))
+    expect.same(concat, Hints.member(HttpHeader("X-Member")))
```
daddykotex, line 67: I'm assuming member was already there, just _untested_

TIL

daddykotex: forget that, now I understand what you did

PR #1205, opened 2023-09-19
Revamped `Errorable`

Baccata:
* Moved to the `schema` package and renamed to `ErrorSchema`
* Turned into a case class
* Removed enforcement of `UnionSchema`
* Exposing alternatives and ordinal methods instead

The idea is to facilitate the transformation of the underlying schema, which is really helpful for interpreters to pre-process the schemas with additional hints. That is because the `transformHints` methods returns `Schema`. I do think that losing the `UnionSchema` constraint is gonna open some interesting things later on.


lewisjkl (approved):

modules/core/src/smithy4s/schema/ErrorSchema.scala
```diff
@@ -0,0 +1,66 @@
...
+package smithy4s.schema
+
+import smithy4s.ShapeTag
+
+/**
+  * An ErrorSchema is similar to a UnionSchema in that it exposes `alternatives` and `ordinal` values,
+  * and therefore can be manipulated similarly to UnionSchemas.
+  *
+  * Additionally, it carries functions to go from E to Throwable and vice-versa. This is used by
+  * interpreters to
```
daddykotex: unfinished comment

PR #1206, opened 2023-09-20
Rename Readers to Decoders

Baccata:
For consistency across the codebase.


daddykotex:
Love this effort

modules/aws-http4s/src/smithy4s/aws/internals/AwsEc2QueryCodecs.scala
```diff
@@ -49,7 +49,7 @@ private[aws] object AwsEcsQueryCodecs {
 
   private val addErrorStartingPath = (_: Hints).add(XmlStartingPath(List("Response", "Errors", "Error")))
   private val discriminatorReaders =
```
daddykotex: `discriminatorDecoders`

I'm happy to make the changes as I do my review if you want, it will avoid comment spam

modules/core/src/smithy4s/capability/MonadThrowLike.scala
```diff
@@ -18,7 +18,8 @@ package smithy4s.capability
 
 import smithy4s.kinds.PolyFunction
 
-trait MonadThrowLike[F[_]] extends Zipper[F] {
```
Baccata: @daddykotex, this is a bit annoying for downstream, because it means that there are few more constructs to implement to use the generic routers, but it feels more "right", as otherwise there are some annoying bits wrt orphan instances, which make the decoders less usable.

modules/decline/src/core/OptsVisitor.scala
```diff
@@ -129,7 +129,7 @@ object OptsVisitor extends SchemaVisitor[Opts] { self =>
   }
 
   private def parseJson[A](schema: Schema[A]): String => Either[String, A] = {
-    val reader = smithy4s.json.Json.payloadCodecs.readers.fromSchema(schema)
+    val reader = smithy4s.json.Json.payloadCodecs.decoders.fromSchema(schema)
```
daddykotex: val decoder

modules/core/src/smithy4s/capability/MonadThrowLike.scala
Baccata: Decided to move instances to orphan locations instead. It's more pragmatic.

PR #1208, opened 2023-09-20
Generate smithy-build.json from sbt config

### LINKED ISSUE #497: Generate smithy-build.json
It's been discussed in the past, and it seems like the consensus is that we should have the ability to generate `smithy-build.json` based on the sbt/mill build settings.

It should be customizable to some extent, e.g. to allow users to specify a plugin like [Smithy Playground](https://github.com/kubukoz/smithy-playground). Maybe the whole JSON structure should just be passed verbatim at the end.

To be discussed:

- should `imports` be generated? I'm thinking yes, because otherwise files from `target` are seen by the language server. Either that or users would be greatly encouraged to specify the directories they wanna put their Smithy files in.
- how do we filter dependencies that should be added to `mavenDependencies`? Should it be opt-in or can we somehow filter out `libraryDependencies` to see which ones contain a smithy manifest?

In general I'm thinking [sbt-github-actions](https://github.com/djspiewak/sbt-github-actions/)-style configuration would be nice, but users would be encouraged to gitignore the file (unlike in the case of GHA). That makes it easier as we don't need a `check*` task.

related: #151

daddykotex:
This is not perfect, but I guess it's a good first step and it might help our user setup their original configuration.

This current version has no support for customization though. Should I do that in this PR or we're okay revisiting this when required.

for sbt: `sbt generateSmithyBuildJson`

for mill: `mill  smithy4s.codegen.mill.Smithy4sExternalModule/generateSmithyBuildJson`

Fixes #497


modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -255,6 +263,24 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
...
+        .flatMap(moduleIdEncode(_, scalaBin))
+      val imports = (config / smithy4sInputDirs).value
+        .collect(prepareInputDirs(baseDir, sourceManagedDir))
+      val repositories = externalResolvers.value.collect(prepareResolvers)
+      SmithyBuildJson.toJson(imports, deps, repositories)
+    },
+    config / smithy4sGeneratedSmithyBuildFile := {
+      val out = baseDirectory.value / "smithy-build.json"
+      IO.write(out, (config / smithy4sGeneratedSmithyBuild).value)
+      out
```
Baccata: not sure this cuts it : ideally we want to aggregate all dependencies from all modules and create the file from this aggregation.

daddykotex:
For now, I'll investigate on the following alternatives to improve the current implementation:

* the task is global to all the modules and will merge the results from all modules where the smithy4s codegen plugins is enabled
* the task in each module that has the plugin enabled will read the current smithy-build file and append to it if necessary (aware of concurrency)

modules/codegen-plugin/src/sbt-test/codegen-plugin/defaults/build.sbt
```diff
@@ -5,5 +5,41 @@ lazy val root = (project in file("."))
...
+          .mkString("\n")
+          .trim()
+      val expected = """|{
+                        |    "version": "1.0",
+                        |    "imports": [
+                        |        "src/main/smithy"
+                        |    ],
+                        |    "maven": {
+                        |        "dependencies": [
+                        |            "com.disneystreaming.alloy:alloy-core:0.2.6"
```
daddykotex: just realized that this test will be sensitive to dependency upgrade (right now alloy is bumped in 0.18 and after merged series/0.18 here, it fails :/ )

kubukoz: TBH I wouldn't check the file against a text version of JSON in the first place. I wouldn't want this to break because e.g. jsoniter changes ordering of serialized fields, or if we change config.

Just checking whether the file contains relevant strings should do (e.g. the `com.disneystreaming.alloy:alloy-core` prefix, `src/main/smithy`, `1.0`), and maybe also verifying that it all parses as JSON.

If you want to be more strict I'd still compare JSON ASTs rather than strings, and then the version numbers could maybe be ignored

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
kubukoz: wish the LSP supported multiple config files in a workspace :/

contributions welcome I guess

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -290,6 +316,22 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
       )
     )
 
+  private def prepareResolvers: PartialFunction[Resolver, String] = {
```
kubukoz: nitpick: `val`

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -255,6 +263,24 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
...
+    config / smithy4sGeneratedSmithyBuild := {
+      val baseDir = baseDirectory.value
+      val sourceManagedDir = (config / sourceManaged).value
+
+      val scalaBin = (config / scalaBinaryVersion).?.value
+      val deps = (config / libraryDependencies).value
+        .filter(_.configurations.exists(_.contains(Smithy4s.name)))
+        .flatMap(moduleIdEncode(_, scalaBin))
+      val imports = (config / smithy4sInputDirs).value
+        .collect(prepareInputDirs(baseDir, sourceManagedDir))
```
kubukoz: suggestion: named parameters for `baseDir` / `sourceManagedDir`

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -290,6 +316,22 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
...
 
+  private def prepareResolvers: PartialFunction[Resolver, String] = {
+    case mr: MavenRepository if !mr.root.contains("repo1.maven.org") => mr.root
+  }
+
+  private def prepareInputDirs(
+      base: File,
+      sourceManaged: File
+  ): PartialFunction[File, String] = {
+    // exclude files that are under sourceManaged
```
kubukoz: question: can you explain why we wanna do this? I was thinking that if these files are in `smithy4sInputDirs` and affect codegen, we'd want to import them

daddykotex: it's a good question. when I saw it in the list, the first time I generated it
my brain automatically marked it as unwanted because it was under `target`

maybe the fact that they are generated and user modifications are discarded because the build tool handles those files

I can keep it no problem though

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -290,6 +314,59 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
...
     )
 
+  private def makeSmithyBuildMerge(
+      config: Configuration
+  ): Def.Initialize[Task[File]] = Def.task {
+    val sbFile = (config / smithy4sSmithyBuildFile).value
+
+    val smithyBuildJson = makeSmithyBuild(config).value
+
+    synchronized {
```
daddykotex: it seems like `task / aggregate := false` did not prevent parallel execution as I had hoped and sometimes I was left with a broken json file. so I slapped a synchronized

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
daddykotex: kept it

kubukoz: what files are those? the ones that set metadata for the Scala import syntax, perchance?

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
kubukoz: tbh that sounds like a bit of a band aid rather than a long-term solution. I'd rather have one top-level task that you run on your `root` project, which eliminates any concurrency in side-effecting IO

daddykotex: we discussed this internally, and I don't know how to implement such a task

especially not in a sbt plugin

maybe @keynmol would know?

context: I'd like a task that I can run on the root module (or any aggregate module) that runs a task defined in a plugin. ideally, the task would aggregate the result of all the task executions, and then write to a file. the idea is that every module that has `Smithy4sCodegenPlugin` has a task to generate a `smithy-build.json` definition, but we'd like to have one task to reduce them all together and write them to disk

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
daddykotex:
```
> tree bar/target/scala-2.13/src_managed/main/smithy/
bar/target/scala-2.13/src_managed/main/smithy/
└── generated-metadata.smithy
> cat bar/target/scala-2.13/src_managed/main/smithy/generated-metadata.smithy 
$version: "2"
metadata smithy4sWildcardArgument = "_"
metadata smithy4sRenderOptics = false
```

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
daddykotex: got rid of that 👍

modules/mill-codegen-plugin/src/smithy4s/codegen/mill/Smithy4sExternalModule.scala
```diff
@@ -0,0 +1,101 @@
...
+import mill.eval.Evaluator
+import mill.define.ExternalModule
+import mill.define.Target
+import coursier.maven.MavenRepository
+import smithy4s.codegen.SmithyBuildJson
+import smithy4s.codegen.SmithyBuildData
+import mill.define.Command
+import mill.api.PathRef
+
+object Smithy4sExternalModule extends ExternalModule {
```
Baccata: when creating mill external module, it's important to think about the UX imposed onto the users, as they don't get autocompletion. You want the command to be as easy to remember as possible. I think it should be something like : 

`smithy4s.LSP/config`

modules/mill-codegen-plugin/src/smithy4s/codegen/mill/Smithy4sModule.scala
```diff
@@ -38,23 +38,23 @@ import mill.scalalib.CrossVersion.Full
 trait Smithy4sModule extends ScalaModule {
 
   /** Input directory for .smithy files */
-  protected def smithy4sInputDirs: Target[Seq[PathRef]] = T.sources {
```
Baccata, line 41: just make them package private

modules/codegen/src/smithy4s/codegen/internals/SmithyBuild.scala
```diff
@@ -0,0 +1,81 @@
...
+ *     https://disneystreaming.github.io/TOST-1.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.codegen.internals
```
Baccata: all the constructs in here should probably be package-private (`private [smithy4s]`)

Baccata:
I've taken the liberty of rewriting the mill integration more idiomatically, and added some tests : [7d60f5c](https://github.com/disneystreaming/smithy4s/pull/1208/commits/7d60f5ccf9d72d4d85f9add2a965e1164628a1bf)

modules/mill-codegen-plugin/src/smithy4s/codegen/mill/Smithy4sExternalModule.scala
```diff
@@ -21,81 +21,48 @@ import mill.define.ExternalModule
...
+      .collect { case s: Smithy4sModule => s }
 
-      def extractImports(m: Smithy4sModule): Seq[String] = {
-        ev
-          .evalOrThrow()(m.smithy4sInputDirs)
-          .map(p => p.path.relativeTo(rootPath))
-          .map(_.toString)
-          .distinct
-      }
+    val depsTask = Target
```
daddykotex: I had a similar approach at first and I rewrote to avoid traversing the s4smodule seq multiple time (like I did in sbt)

modules/mill-codegen-plugin/test/src/smithy4s/codegen/mill/SmithyLSPConfigSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+import mill._
+import mill.scalalib._
+import mill.testkit.MillTestKit
+import sourcecode.FullName
+
+class SmithyLSPConfigSpec extends munit.FunSuite {
+
+  private object testKit extends MillTestKit
+
+  test("config gets generated") {
```
daddykotex, line 12: I did not write the test because I was not sure how to, that `root` trick does it perfectly

thanks for stepping up

modules/mill-codegen-plugin/test/src/smithy4s/codegen/mill/SmithyLSPConfigSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+      ev.evaluator.rootModule.millModuleBasePath.value / "smithy-build.json"
+    )
+    val writtenJson = ujson.read(fileContent)
+    val expectedJson = ujson.read {
+      s"""|{
+          |  "version": "1.0",
+          |  "imports": ["./foo/smithy"],
+          |  "maven": {
+          |    "dependencies": [
+          |       "com.disneystreaming.alloy:alloy-core:${smithy4s.codegen.BuildInfo.alloyVersion}",
```
daddykotex, line 45: clever, the version is in the `BuildInfo`

modules/mill-codegen-plugin/src/smithy4s/codegen/mill/Smithy4sExternalModule.scala
daddykotex: That would mean moving the object under the `smithy4s` package, right?

modules/mill-codegen-plugin/src/smithy4s/codegen/mill/Smithy4sExternalModule.scala
Baccata: Yeah, the approach you had was similar to chaining IO using `unsafeRunSync`, which ain't great. 

I can't comment much on whether the SBT approach is idiomatic because I don't know SBT as well 😅

modules/codegen/src/smithy4s/codegen/internals/SmithyBuild.scala
daddykotex: yeah, all of that should be private, because I hide all of that functionality behind the `SmithyBuildJson` object

modules/mill-codegen-plugin/src/smithy4s/codegen/mill/Smithy4sExternalModule.scala
daddykotex: would `smithy4s.codegen.LSP/smithyBuild` be fine with you?

I feel like we're gonna add other commands in the future in this module and `config` sounds too generic

I added `smithy4s.codegen` so package visiblity of the `SmithyBuildJson` still make some sense

modules/mill-codegen-plugin/test/src/smithy4s/codegen/mill/SmithyLSPConfigSpec.scala
```diff
@@ -0,0 +1,58 @@
+package smithy4s.codegen.mill
+
+import mill._
+import mill.scalalib._
+import mill.testkit.MillTestKit
+import sourcecode.FullName
+
+class SmithyLSPConfigSpec extends munit.FunSuite {
```
kubukoz, line 8: thought: to be precise it's the smithy-cli config, the LSP happens to use it

kubukoz (commented):
I like the idea to merge with the existing files, that should allow custom options, right?

modules/mill-codegen-plugin/src/smithy4s/codegen/mill/Smithy4sExternalModule.scala
Baccata: package is fine. For reference, similar commands in the mill ecosystem are : 

``` 
mill.idea.GenIdea/idea 
mill.contrib.bloop.Bloop/install
```

I think the first one is silly, there's 3 repetitions of `idea`. I however like the `GenIdea` object
I like `install` but it's a bit generic. 

Combining the two, we could have something like : 

``` 
smithy4s.codegen.LSPConfig/update
smithy4s.codegen.LSP/updateConfig
```

NB : after a decision on this, the SBT command should be updated to be similar. `smithy4sUpdateLSPConfig` would be alright.

modules/mill-codegen-plugin/test/src/smithy4s/codegen/mill/SmithyLSPConfigSpec.scala
Baccata: <a href="https://imgflip.com/i/80to1e"></a><div></div>

:trollface:

daddykotex:
> I like the idea to merge with the existing files, that should allow custom options, right?

absolutely, it does

there is a sbt-test that tests that

modules/mill-codegen-plugin/src/smithy4s/codegen/mill/Smithy4sExternalModule.scala
daddykotex: done, I've added some docs in the editor support page as well

Baccata (approved):

modules/mill-codegen-plugin/test/src/smithy4s/codegen/mill/SmithyLSPConfigSpec.scala
kubukoz: I missed this 😂 

it was a good comment. If you use smithy-cli (which has some nice things in it), you won't think to use `smithyLspConfig`

PR #1209, opened 2023-09-20
Fix for Json Cursor

lewisjkl:
Fixes an issue where the cursor would reflect the wrong item being popped off when converting to a PayloadPath. This happened since we the labelStack could have an item still in place at the current `top` index since we were never actually clearing it. I also added a check to make sure we never decrement the top index under zero. This will guard against any possible index out of bounds issues from uneven push/pop calls.

### DIFF
--- a/modules/json/src/smithy4s/json/internals/Cursor.scala
+++ b/modules/json/src/smithy4s/json/internals/Cursor.scala
@@ -22,7 +22,7 @@ import com.github.plokhotnyuk.jsoniter_scala.core.JsonReader
 import com.github.plokhotnyuk.jsoniter_scala.core.JsonReaderException
 import smithy4s.codecs._
 
-private[internals] class Cursor private () {
+private[internals] class Cursor private[internals] () {
   private[this] var indexStack: Array[Int] = new Array[Int](8)
   private[this] var labelStack: Array[String] = new Array[String](8)
   private[this] var top: Int = _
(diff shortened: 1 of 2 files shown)


daddykotex (approved):
How did you stumble on this?!

lewisjkl:
> How did you stumble on this?!

Lol long story.. basically an internal use case

modules/json/src/smithy4s/json/internals/Cursor.scala
```diff
@@ -65,7 +65,12 @@ private[internals] class Cursor private () {
     top += 1
   }
 
-  def pop(): Unit = top -= 1
+  def pop(): Unit = {
+    if (top > 0) {
+      top -= 1
+    }
+    labelStack(top) = null
```
Baccata, line 72: it's suspicious to me that you wouldn't apply the same to the indexStack.

lewisjkl: Well you can, but don't strictly need to because the function for getting the items out of the stack looks at labelStack(top) to check if it has a value before continuing on to pull the value from the index stack if it doesn't. So the issue came from the labelStack having a value, where it should have been deferring to the index stack

Baccata: gotcha 👍

Baccata (approved):

PR #1211, opened 2023-09-21
Rewrite Blob to facilitate interop with other libs

Baccata:
Rewrite the structure of Blob to help interop with other libraries and patterns, such as fs2 or scodec.

Inspired from : 
* https://github.com/typelevel/fs2/blob/356b5c79c985f867cee0fbe9e45803040aecdaa7/core/shared/src/main/scala/fs2/Chunk.scala 
* https://github.com/armanbilge/scodec-bits/blob/ad81e86d47dc1859e48e453df5da402bc8f13abb/core/shared/src/main/scala/scodec/bits/ByteVector.scala

### DIFF
--- a/modules/bootstrapped/test/src/smithy4s/BlobSpec.scala
+++ b/modules/bootstrapped/test/src/smithy4s/BlobSpec.scala
@@ -20,6 +20,7 @@ import munit._
 
 import java.nio.ByteBuffer
 import java.io.ByteArrayOutputStream
+import scala.util.Using
 class BlobSpec() extends FunSuite {
 
   test("sameBytesAs works across data structures") {
@@ -89,12 +90,7 @@ class BlobSpec() extends FunSuite {
     }
 
     test(s"$name: out of bounds access") {
-      try {
-        val _ = data(6)
-        fail("expected exception")
-      } catch {
-        case _: IndexOutOfBoundsException => ()
-      }
+      intercept[IndexOutOfBoundsException] { data(6) }
     }
 
     test(s"$name: utf8String") {
@@ -121,12 +117,9 @@ class BlobSpec() extends FunSuite {
     }
 
     test(s"$name: copyToStream") {
-      val stream = new ByteArrayOutputStream()
-      try {
+      Using.resource(new ByteArrayOutputStream()) { stream =>
         data.copyToStream(stream, 0, data.size)
         assert(stream.toByteArray().sameElements("foobar".getBytes()))
-      } finally {
-        stream.close()
       }
     }
   }
--- a/modules/bootstrapped/test/src/smithy4s/BlobSpec.scala
+++ b/modules/bootstrapped/test/src/smithy4s/BlobSpec.scala
@@ -19,7 +19,6 @@ package smithy4s
 import munit._
 
 import java.nio.ByteBuffer
-
 class BlobSpec() extends FunSuite {
 
   test("sameBytesAs works across data structures") {
@@ -35,13 +34,6 @@ class BlobSpec() extends FunSuite {
         ByteBuffer.wrap("foo".getBytes)
       )
     )
-  }
-
-  test("ByteBufferBlob.toArray is idempotent, instantiation-wise") {
-    val blob = Blob(ByteBuffer.wrap("foo".getBytes))
-    assert(blob.toArray != null)
-    assert(blob.toArray.eq(blob.toArray))
-    assertEquals(Blob(blob.toArray), Blob("foo"))
   }
 
   test("ByteArrayBlob.hashcode is consistent") {
@@ -60,6 +52,15 @@ class BlobSpec() extends FunSuite {
     val blob3 = makeBlob("bar")
     assertEquals(blob1.hashCode, blob2.hashCode)
     assertNotEquals(blob1.hashCode, blob3.hashCode)
+  }
+
+  test("Concat works as expected") {
+    val blob = Blob("foo") ++ Blob("bar")
+    assertEquals(blob.size, 6)
+    assertEquals(blob(2), 'o'.toByte)
+    assertEquals(blob(4), 'a'.toByte)
+    java.util.Arrays
+      .equals(blob.toArray, "foo".getBytes ++ "bar".getBytes())
   }
 
 }
--- a/modules/bootstrapped/test/src/smithy4s/BlobSpec.scala
+++ b/modules/bootstrapped/test/src/smithy4s/BlobSpec.scala
@@ -19,6 +19,7 @@ package smithy4s
 import munit._
 
 import java.nio.ByteBuffer
+import java.io.ByteArrayOutputStream
 class BlobSpec() extends FunSuite {
 
   test("sameBytesAs works across data structures") {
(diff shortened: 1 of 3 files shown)


Baccata:
@armanbilge, if you could spare the time to give me you your opinion on this, it'd be well appreciated

modules/core/src/smithy4s/Blob.scala
```diff
@@ -41,78 +117,171 @@ object Blob {
     string.getBytes(StandardCharsets.UTF_8)
   )
 
-  trait Encoder[A] {
-    def encode(a: A): Blob
-  }
+  final class ByteArrayBlob(val arr: Array[Byte]) extends Blob {
```
armanbilge: This one is a special case of `ArraySliceBlob` right? Is it worth to make it a separate case in the ADT?

modules/core/src/smithy4s/Blob.scala
```diff
@@ -18,17 +18,93 @@ package smithy4s
...
+    while (i < size) { f(apply(i), i); i += 1 }
+  }
+
+  def toArray: Array[Byte] = {
+    val result = Array.ofDim[Byte](size)
+    foreachWithIndex((b, i) => result(i) = b)
+    result
+  }
+
+  def toArrayUnsafe: Array[Byte] = toArray
```
armanbilge, line 56: Another useful one to have is `toArraySliceBlob: ArraySliceBlob`. This would make it extremely easy to convert to/from `fs2.Chunk.ArraySlice`. It also provides direct (unsafe) access to the array without copying if its a view and without having to grapple with the confusing `ByteBuffer` API.

modules/core/src/smithy4s/Blob.scala
```diff
@@ -41,78 +117,171 @@ object Blob {
...
+      n.get(xs, start, size)
+      ()
+    }
+
+    override def toArray: Array[Byte] = {
+      val arr = Array.ofDim[Byte](buf.remaining())
+      copyToArray(arr, 0, 0, size)
+      arr
+    }
+
```
armanbilge, line 149: Many `ByteBuffer`s are actually backed by an `Array[Byte]` so we can override `toArrayUnsafe` (and `toArraySliceBlob`) with something more efficient.

https://github.com/typelevel/fs2/blob/196995d58d5b82da38252ec81d003f91a740503a/core/shared/src/main/scala/fs2/Chunk.scala#L1026-L1031

armanbilge (commented):
Seems fine. FWIW I think you could get away with just `ArraySliceBlob` and `QueueBlob` (also I think JVM special-cases classes with no more than 2 implementations, but don't quote me).

Many `ByteBuffer`s are backed by `Array[Byte]`s so you can convert them to `ArraySliceBlob` without copying. In the case where they're not (e.g. off-heap buffer), you'd have to copy to bring it into the ADT, but chances are you'd have to copy at some point _anyway_. The only case you lose is when you are literally just shoveling bytes between sockets/files without any parsing/transformation, then keeping things in off-heap `ByteBuffer`s could win. Edit: note that FS2 doesn't use off-heap buffers anyway, actually they are kind of hostile to immutable datatypes since they don't GC well.

modules/core/src/smithy4s/Blob.scala
Baccata: Good point

modules/core/src/smithy4s/Blob.scala
```diff
@@ -18,101 +18,238 @@ package smithy4s
...
+  }
+
+  def asByteBuffer(offset: Int, size: Int): ByteBuffer = {
+    val arr = new Array[Byte](size)
+    copyToArray(arr, 0, offset, size)
+    ByteBuffer.wrap(arr)
+  }
+  def asByteBuffer: ByteBuffer = asByteBuffer(0, size)
+  def asByteBufferUnsafe(offset: Int, size: Int): ByteBuffer = asByteBuffer(offset, size)
+  def asByteBufferUnsafe: ByteBuffer = asByteBuffer(0, size)
```
daddykotex, line 77: what are the `unsafe` alternative for? they seem to have that same underlying implementation

armanbilge: `asByteBuffer` may return a read-only `ByteBuffer` to protect the backing memory. `asByteBufferUnsafe` unsafely exposes the backing memory. This makes a meaningful difference if you want to obtain the `Array[Byte]` backing a `ByteBuffer`.

modules/core/src/smithy4s/Blob.scala
```diff
@@ -18,101 +18,238 @@ package smithy4s
...
-        hashCode += bytes(i).hashCode()
-        i += 1
+    override def asByteBufferUnsafe(offset: Int, size: Int): ByteBuffer = {
+      val b = buf
+      if (offset == 0 && b.position() == 0 && size == b.remaining()) b
+      else {
+        b.position(offset.toInt)
+        b.limit(offset.toInt + size)
+        b.slice()
       }
```
armanbilge: @daddykotex see here.

modules/core/src/smithy4s/Blob.scala
```diff
@@ -18,101 +18,238 @@ package smithy4s
...
+      asByteBufferUnsafe(offset, size).asReadOnlyBuffer()
 
-    override def hashCode(): Int = {
-      var hashCode = 0
-      var i = 0
-      while (i < bytes.length) {
-        hashCode += bytes(i).hashCode()
-        i += 1
+    override def asByteBufferUnsafe(offset: Int, size: Int): ByteBuffer = {
+      val b = buf
```
armanbilge: `ByteBuffer`s are really annoying and hard to work correctly with. This is why I'm skeptical if its worth to give them a designated position in the ADT.

```suggestion
      val b = buf.duplicate()
```

armanbilge: Hmmm 😅 I now realize this is possibly my own mistake in scodec-bits. I need to think about this.

https://github.com/scodec/scodec-bits/blob/ad81e86d47dc1859e48e453df5da402bc8f13abb/core/shared/src/main/scala/scodec/bits/ByteVector.scala#L1466-L1474

modules/core/src/smithy4s/Blob.scala
daddykotex: ahhh I was looking at the default implementation, and I did not realize there was an override for it

modules/core/src/smithy4s/Blob.scala
Baccata: Yup, this was pretty much copied verbatim from scodecs-bits.

armanbilge: - https://github.com/scodec/scodec-bits/pull/467

modules/core/src/smithy4s/Blob.scala
```diff
@@ -18,101 +18,238 @@ package smithy4s
...
+        length == o.length &&
+        java.util.Arrays.equals(arr, o.arr)
+      }
+    }
+
+  }
+
+  final class QueueBlob private[smithy4s] (val blobs: Queue[Blob], val size: Int) extends Blob {
+    def apply(i: Int): Byte = {
+      if (i > size) throw new IndexOutOfBoundsException()
```
lewisjkl:
```suggestion
      if (i >= size) throw new IndexOutOfBoundsException()
```

daddykotex: cursor trauma in action

lewisjkl: Hahah I am scarred for life. This one is a much smaller issue because the `IndexOutOfBoundsException` would still get thrown, just after doing some unnecessary work.. so really a very minor optimization in this case lol

modules/core/src/smithy4s/Blob.scala
Baccata: Addressed in [aabf826](https://github.com/disneystreaming/smithy4s/pull/1211/commits/aabf8268f0b2893a451d744710eee631cd0fb68c)

modules/core/src/smithy4s/Blob.scala
Baccata: Addressed in [aabf826](https://github.com/disneystreaming/smithy4s/pull/1211/commits/aabf8268f0b2893a451d744710eee631cd0fb68c)

lewisjkl (approved):

modules/bootstrapped/test/src/smithy4s/BlobSpec.scala
```diff
@@ -62,4 +55,105 @@ class BlobSpec() extends FunSuite {
...
+      assertEquals(data(2), 'o'.toByte)
+    }
+
+    test(s"$name: out of bounds access") {
+      try {
+        val _ = data(6)
+        fail("expected exception")
+      } catch {
+        case _: IndexOutOfBoundsException => ()
+      }
```
daddykotex:
```suggestion
      intercept[IndexOutOfBoundsException]{ data(6) }
```

modules/bootstrapped/test/src/smithy4s/BlobSpec.scala
```diff
@@ -62,4 +55,105 @@ class BlobSpec() extends FunSuite {
...
+
+    test(s"$name: copyToBuffer") {
+      val target = ByteBuffer.wrap(Array.fill[Byte](6)(0))
+      data.copyToBuffer(target, 0, data.size)
+      assert(target.array.sameElements("foobar".getBytes()))
+    }
+
+    test(s"$name: copyToStream") {
+      val stream = new ByteArrayOutputStream()
+      try {
```
daddykotex: the other day I discovered `Using` and [`Using.resource`](https://www.scala-lang.org/api/2.13.12/scala/util/Using$.html#resource[R,A](resource:R)(body:R=%3EA)(implicitreleasable:scala.util.Using.Releasable[R]):A) to do this kind of resource management

modules/core/src/smithy4s/Blob.scala
```diff
@@ -18,101 +18,241 @@ package smithy4s
...
 
-    def toArray: Array[Byte] = {
-      if (arr == null) {
-        this.synchronized {
-          if (arr == null) {
-            arr = Array.ofDim[Byte](buffer.remaining())
-            val _ = buffer.get(arr)
-          }
+    override def hashCode(): Int = {
+      import util.hashing.MurmurHash3
```
daddykotex, line 210: well that's nice

daddykotex (approved):
I'm no expert, so take this review with a grain of salt. I have a few suggestion to improve test readability but that's about it, the rest seems fine to me.

modules/bootstrapped/test/src/smithy4s/BlobSpec.scala
Baccata: Thanks ! [b50d8d4](https://github.com/disneystreaming/smithy4s/pull/1211/commits/b50d8d4b22cf0095e0b4e4f0e92c505ffb3d6540)

modules/bootstrapped/test/src/smithy4s/BlobSpec.scala
Baccata: Thanks. I was under the impression it was a 2.13+ only thing 

[b50d8d4](https://github.com/disneystreaming/smithy4s/pull/1211/commits/b50d8d4b22cf0095e0b4e4f0e92c505ffb3d6540)

PR #1078, opened 2023-07-06
Endpoint/HostLabel implementation 

Baccata:
Supersedes https://github.com/disneystreaming/smithy4s/pull/1056

@yisraelU, once again, apologies for branching out. The tools I use to pull your branch (namely, the hub CLI) are favouring using ssh remotes, and the disney SSO is unhappy about it. 

Anyway : 
* implements the [endpoint/hostLabel traits](https://smithy.io/2.0/spec/endpoint-traits.html?highlight=endpoint) on client-side.
* restructure some things to help inference, in particular with the fact that some request encoders ought to work on Pure requests

PR #1213, opened 2023-09-25
Endpoint Host Prefix

lewisjkl:
Supersedes #1078 (same code, but updated to work with new http refactor)

Implements the [endpoint/hostLabel traits](https://smithy.io/2.0/spec/endpoint-traits.html?highlight=endpoint) on client-side.


modules/core/src/smithy4s/codecs/Writer.scala
```diff
@@ -98,6 +98,11 @@ object Writer {
     */
   def lift[Message, A](f: (Message, A) => Message): Writer[Message, A] = f(_, _)
 
+  /**
+    * Creates an writer which returns a constant value
```
daddykotex: minor: a writer ?

modules/core/src/smithy4s/http/HttpRequest.scala
```diff
@@ -96,6 +96,28 @@ object HttpRequest {
...
+        case Some(prefixEncoder) =>
+          new Writer[Body, I] {
+            def write(
+                request: HttpRequest[Body],
+                input: I
+            ): HttpRequest[Body] = {
+              val hostPrefix = prefixEncoder.write(List.empty, input)
+              val oldUri = request.uri
+              val newUri =
+                oldUri.copy(host = s"${hostPrefix.mkString("")}.${oldUri.host}")
```
daddykotex: there is no separator between the prefix parts? no `.`? anyhow, I think you could discard the empty string and just do `.mkString` if it is no separator

modules/core/src/smithy4s/http/internals/package.scala
```diff
@@ -24,6 +24,10 @@ package object internals {
   private[http] type HttpCode[A] = A => Option[Int]
   private[http] val httpHints = HintMask(HttpBinding)
 
+  private[internals] type HostPrefixEncode[A] =
```
daddykotex:
```suggestion
  private[internals] type HostPrefixEncoder[A] =
```

I'd probably add that the type alias `MaybeHostPrefixEncode` is not particularly useful

modules/core/src/smithy4s/http/internals/HostPrefixSchemaVisitor.scala
```diff
@@ -0,0 +1,111 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.http.internals
+
+import smithy4s.{Schema, _}
```
daddykotex, line 19:
```suggestion
import smithy4s._
```

modules/core/src/smithy4s/http/HttpUnaryClientCodecs.scala
```diff
@@ -192,7 +197,10 @@ object HttpUnaryClientCodecs {
               case None => inputEncoders.fromSchema(endpoint.input, inputEncoderCache)
             }
 
-          val inputEncoder = (i: I) => F.map(baseRequest(endpoint))(inputWriter.write(_, i))
+          val prefixedInputWriter: HttpRequest.Writer[Blob, I] =
+            if (hostPrefixInjection) inputWriter.combine(HttpRequest.Writer.hostPrefix(endpoint)) else inputWriter
```
daddykotex, line 201: a `combineIf` method could make this easier to digest:
```
val prefixedInputWriter: HttpRequest.Writer[Blob, I] = inputWriter.combineIf(hostPrefixInjection, HttpRequest.Writer.hostPrefix(endpoint))`

```

modules/core/src/smithy4s/http/HttpRequest.scala
daddykotex: a test just to see the effect on the uri would be valuable I think

modules/core/src/smithy4s/http/internals/HostPrefixSchemaVisitor.scala
Baccata: It's needed because there's still a `smithy4s.Schema` type alias, which conflicts with the import of `smithy4s.schema.Schema`.

modules/core/src/smithy4s/http/HttpUnaryClientCodecs.scala
Baccata: it's not a bad idea but I don't think a single pain point warrants it. Let's avoid increasing the API surface for now

modules/http4s/src/smithy4s/http4s/SimpleRestJsonBuilder.scala
```diff
@@ -17,28 +17,47 @@
...
       explicitDefaultsEncoding: Boolean
   ): SimpleRestJsonBuilder =
     new SimpleRestJsonBuilder(
       simpleRestJsonCodecs.maxArity,
-      explicitDefaultsEncoding
+      explicitDefaultsEncoding,
+      simpleRestJsonCodecs.hostPrefixInjection
+    )
+
+  def disableHostPrefixInjection(): SimpleRestJsonBuilder =
```
Baccata, line 57: I don't think this customisation is needed at this level, I think we can assume it's always true. What's your rationale for exposing it ?

modules/core/src/smithy4s/http/HostPrefix.scala
```diff
@@ -0,0 +1,35 @@
...
+ *  limitations under the License.
+ */
+
+package smithy4s.http
+
+import smithy4s.http.internals.HostPrefixSchemaVisitor
+import smithy4s.codecs.Writer
+import smithy4s.schema.OperationSchema
+
+object HostPrefix {
```
Baccata: let's rename this to `HttpHostPrefix `

modules/http4s/src/smithy4s/http4s/SimpleRestJsonBuilder.scala
daddykotex: I also thought that, but when looking up something in the spec, I saw: 

> Clients SHOULD provide a way for users to disable the hostPrefix injection behavior.

lewisjkl: Yeah I just copied it over from the other PR/branch, but I kept it as I saw the language David quoted there

modules/core/src/smithy4s/http/HttpRequest.scala
lewisjkl: Separators are already put into those segments (as they can be anything as defined in the endpoint trait), but yeah I will definitely look at adding a test there (currently is covered by compliance tests also)

Baccata (approved):

PR #1219, opened 2023-09-27
Facilitate the loading of AWS specifications

Baccata:
* Adds logic to code-generate a list of artifacts containing AWS specifications in a Scala object 
* Add build tasks (for SBT and Mill) to lower the barrier of entry for referencing those artifacts in their build. Since the artifacts are listed as Scala values, the users get auto-completion to help set-up the code generation of AWS services. 
* Added corresponding tests for both SBT and Mill, avoiding the compilation of the generated code to save on CI time.

### DIFF
--- a/project/AwsBoilerplate.scala
+++ b/project/AwsBoilerplate.scala
@@ -1,0 +1,103 @@
+import sbt._
+import scala.annotation.tailrec
+import java.net.URLClassLoader
+import scala.jdk.CollectionConverters._
+import sjsonnew._
+import sjsonnew.:*:
+import sjsonnew.BasicJsonProtocol._
+import scala.io.Source
+import sjsonnew.shaded.scalajson.ast.JValue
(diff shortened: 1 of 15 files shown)


kubukoz (approved):

project/AwsBoilerplate.scala
```diff
@@ -0,0 +1,103 @@
...
+      .getLines()
+      .mkString(System.lineSeparator())
+
+    com.github.plokhotnyuk.jsoniter_scala.core
+      .readFromString[Summary](jsonString)
+  }
+
+  private case class Module(service: String, name: String)
+  private type GenModule = String :*: String :*: LNil
+  private object Module {
```
daddykotex: do you still need that `sjsonnew` stuff? since you're parsing with jsoniter?

Baccata: good point, I can remove :+1:

PR #1220, opened 2023-09-27
Attempt to compute the original url, otherwise keep behaviour

### LINKED ISSUE #439: Adding a base path prefix to API routes causes UI not to load
Hi everyone 👋

There is probably an easy solution to this but I can't seem to figure it out. 

When the server is behind a proxy which routes requests with a specific prefix to the root of the server (i.e., the server will not see the prefix in the request path), the Swagger UI no longer loads and fails with a 404 error. Note that the API routes still work as expected.

I was wondering, is there a setting where I can define a prefix to the API routes or to make the UI work based on the relative path instead of the absolute path?

Please let me know if my explanation is unclear and I'll put together an example

daddykotex:
An attempt at fixing https://github.com/disneystreaming/smithy4s/issues/439

If we have the info available, we use it to generate the redirect path. For example, a request that goes like:

```
browser -> nginx (http://proxy.com/admin/docs) -> backend (http://localhost:8080/docs) => "http://localhost:8080/docs/index.html"
```

The proxy can then rewrite it appropriately and send the right response to the client.

```
browser -> backend (http://localhost:8080/docs) => "http://localhost:8080/docs/index.html"
```

If we force no host header:

```
curl -H 'Host:' ... -> backend (http://localhost:8080/docs) => "/docs/index.html"
```

IMO, this is a low hanging fruit type of fix that will help working from behind a proxy w/o affecting other uses cases.

Fixes #439

### DIFF
--- a/modules/http4s-swagger/src-jvm/smithy4s/http4s/swagger/Docs.scala
+++ b/modules/http4s-swagger/src-jvm/smithy4s/http4s/swagger/Docs.scala
@@ -19,13 +19,14 @@ package http4s
 package swagger
 
 import cats.data.NonEmptyList
+import cats.data.OptionT
 import cats.effect.Sync
+import org.http4s.HttpRoutes
 import org.http4s._
 import org.http4s.dsl.Http4sDsl
+import org.http4s.headers.Host
 import org.http4s.headers.Location
-import org.http4s.HttpRoutes
 import org.webjars.WebJarAssetLocator
-import cats.data.OptionT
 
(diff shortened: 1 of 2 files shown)


Baccata:
Can we have unit tests for this ?

daddykotex:
> Can we have unit tests for this ?

yup, dropped the ball here

daddykotex:
There was already a bunch of test covering the redirect behaviour with a relative uri. I've added one to trigger the case where an absolute uri is generated.

Baccata (approved):

PR #1222, opened 2023-09-29
Change the DynamicSchemaIndex API

Baccata:
* Return `Iterable` instead of `List`/`Vector`
* Have `loadModel` not return an Either. This is mostly to future-proof it as we should be able to change the implementation to circumvent the Node and Document jumps, thus turning it into a bonafide total function.


daddykotex (approved):
> circumvent the Node and Document jumps

Are you saying that we should be able to load it w/o error if we don't use `NodeToDocument`

Baccata:
> Are you saying that we should be able to load it w/o error if we don't use NodeToDocument 

Well I'm saying that with some effort we could write the function directly, avoiding both the `ModelSerializer` call and the `NodeToDocument` call.

PR #1225, opened 2023-09-29
Tweak codegen api

daddykotex:
This changes comes after a discussion regarding a way to run codegen, from the outside of this project without writing to the file system. This PR changes the api in a breaking way:

From `processSpecs(args: CodegenArgs): Set[os.Path]` to the dual:

- `def generate(args: CodegenArgs): CodegenResult`
- `def generateToDisk(args: CodegenArgs): Set[os.Path]`

I could have kept `processSpecs` or slap a deprecate on it but I don't think it will be particularly useful

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/CodegenResult.scala
+++ b/modules/codegen/src/smithy4s/codegen/CodegenResult.scala
@@ -16,16 +16,19 @@
 
 package smithy4s.codegen
 
-object Codegen {
-
-  def generate(args: CodegenArgs): CodegenResult =
-    internals.CodegenImpl.generate(args)
-
-  def generateToDisk(args: CodegenArgs): Set[os.Path] = {
-    internals.CodegenImpl.write(generate(args))
+sealed trait CodegenEntry {
+  def toPath: os.Path = this match {
+    case CodegenEntry.FromMemory(path, _) => path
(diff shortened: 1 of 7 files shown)


modules/codegen/src/smithy4s/codegen/CodegenResult.scala
```diff
@@ -0,0 +1,34 @@
...
+
+sealed trait CodegenEntry {
+  def toPath: os.Path = this match {
+    case CodegenEntry.FromMemory(path, _) => path
+    case CodegenEntry.FromDisk(path, _)   => path
+  }
+}
+
+object CodegenEntry {
+  case class FromMemory(path: os.Path, content: String) extends CodegenEntry
```
Baccata: replace `path` with `targetPath`

modules/codegen/src/smithy4s/codegen/CodegenResult.scala
```diff
@@ -0,0 +1,34 @@
...
+  case class FromDisk(path: os.Path, sourceFile: os.Path) extends CodegenEntry
```
Baccata: replace `path` with `targetPath`, and `sourceFile` with `sourcePath`

Baccata (approved):

Issue #1028, opened 2023-06-14
codegen not working with watch mode


ptrdom:
Changing `smithy` sources in watch mode (`~compile`) does not trigger codegen. I suggest adding the directory to `unmanagedSourceDirectories` or `unmanagedResourceDirectories`, for example:
```
Compile / unmanagedSourceDirectories+= (Compile / sourceDirectory).value / "smithy"
```
`smithy` sources could just be added to `watchSources`, but they do seem like a good fit for what additional source directory settings are used. For example, `scalapb` seems to have `src/main/protobuf` in `unmanagedSourceDirectories`.

PR #1231, opened 2023-10-02
Misc tweaks 

### LINKED ISSUE #1238: cognito-identity-provider `notify()` method clash
Example project

```sbt
lazy val lambda = project
  .in(file("lambda"))
  .settings(
    libraryDependencies ++= Seq(
      "com.disneystreaming.smithy4s" %%% "smithy4s-aws-http4s" % smithy4sVersion.value,
      "com.disneystreaming.smithy" % "aws-cognito-identity-provider-spec" % "2023.09.22" 
    )
  )
```

With no source files of my own, this is the compiler output:

```
/lambda/target/scala-2.13/src_managed/main/scala/com/amazonaws/cognitoidentityprovider/AccountTakeoverActionType.scala:36:83: cannot override final member:
[error] final def notify(): Unit (defined in class Object)
[error] case class AccountTakeoverActionType(eventAction: AccountTakeoverEventActionType, notify: AccountTakeoverActionNotifyType = com.amazonaws.cognitoidentityprovider.AccountTakeoverActionNotifyType(false))
[error]                                                                                   ^
[error] one error found
[error] (lambda / Compile / compileIncremental) Compilation failed
```

This is a ScalaJS project, intended to be used with feral. There are no differences with the LambdaJSPlugin applied.

Also validated with v2023.02.10 of the specs

Baccata:
* Add a more Play-friendly routing function.   
* Change Router requirement to extract the whole HttpUri instead of just the path, as the `@http` trait allows for capturing query parameters which should have an incidence in the routing logic eventually. Although this does not change behaviour, it helps future-proofing the API 
* Remove unused Xml constructs
* Change visibility of some http4s-kernel methods
* Change Unary{Client|Server}Codecs.Make interface in order to allow for transforming the request/response types. This should eventually facilitate the sharing of protocol codecs across several integrations.
* Resolve #1238 by avoiding collisions with Java symbols like `notify()`.

For github: Closes #1238.


modules/core/src/smithy4s/http/HttpUnaryServerRouter.scala
```diff
@@ -22,91 +22,181 @@ import smithy4s.server.UnaryServerCodecs
...
+      impl: FunctorInterpreter[Op, F],
+      makeServerCodecs: UnaryServerCodecs.Make[F, Request, Response],
+      endpointMiddleware: Endpoint.Middleware[Request => F[Response]],
+      getMethod: Request => HttpMethod,
+      getUri: Request => HttpUri,
+      addDecodedPathParams: (Request, PathParams) => Request
+  )(implicit F: MonadThrowLike[F])
+      extends (Request => Option[F[Response]]) {
+
+    @nowarn
```
kubukoz, line 89: **question:** what's the warning? Is it the `final` outer reference thing?

Perhaps worth a comment _or_ a more specific `@nowarn` rule (filter)

kubukoz (approved):

modules/fs2/src/smithy4s/interopfs2/package.scala
```diff
@@ -16,14 +16,5 @@
...
 // scalafmt: {maxColumn = 120}
-package object interopfs2 {
-
-  type ByteStreamEncoder[F[_], A] =
-    smithy4s.codecs.Encoder[Stream[F, Byte], A]
-  type ByteStreamDecoder[F[_], A] =
-    smithy4s.codecs.Decoder[F, Stream[F, Byte], A]
-
-}
+package object interopfs2 {}
```
lewisjkl, line 20: Is this just being left as a placeholder for future use?

lewisjkl (approved):

modules/http4s-kernel/src/smithy4s/http4s/kernel/package.scala
```diff
@@ -37,7 +37,7 @@ import fs2.Chunk
 // scalafmt: { maxColumn = 120}
 package object kernel {
 
-  private[smithy4s] def toSmithy4sHttpRequest[F[_]: Concurrent](req: Request[F]): F[Smithy4sHttpRequest[Blob]] = {
+  def toSmithy4sHttpRequest[F[_]: Concurrent](req: Request[F]): F[Smithy4sHttpRequest[Blob]] = {
```
kubukoz, line 40: **suggestion (if-minor, non-blocking):** perhaps all these functions should go to some kind of `conversions` object, keeping them in the package object seems a bit strange.

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -183,6 +183,7 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
       (config / sourceDirectory).value / "smithy",
       (config / sourceManaged).value / "smithy"
     ),
+    config / unmanagedSourceDirectories ++= (config / smithy4sInputDirs).value,
```
daddykotex, line 186: this closes an opened issue if I recall correctly
or at least relates to it

daddykotex: https://github.com/disneystreaming/smithy4s/issues/1028

modules/core/src/smithy4s/http/HttpUnaryServerRouter.scala
```diff
@@ -22,91 +22,181 @@ import smithy4s.server.UnaryServerCodecs
...
+      addDecodedPathParams
+    )
   }
 
-}
+  /**
+    * Creates a router in the form of a partial function, which is useful for integration with frameworks
+    * such as Play.
+    */
+  def partialFunction[Alg[_[_, _, _, _, _]], F[_], RequestHead, Request, Response](
```
daddykotex, line 57: sometimes the simplest solution turns out to be the one to use

good thinking, it sucks that we sort of have duplicated router logic in those two classes, but it will make usage easier during integration, that's for sure

kubukoz: that's mostly due to `RequestHead` being a distinct type from `Request`, right? I suppose this kind of duplication is just inevitable in cases like this.

daddykotex: yes :/

and the nature of the partial function means that we have to run the logic twice, once to see if it should be applied, and the second time around to actually perform the routing

modules/fs2/src/smithy4s/interopfs2/package.scala
Baccata: yes

kubukoz (approved):

modules/http4s-kernel/src/smithy4s/http4s/kernel/package.scala
kubukoz: FYI this is a really minor thing that could be resolved in 0.19 ;P or never.

PR #1233, opened 2023-10-02
Keep protocol in smithy4s->http4s URI conversion

### LINKED ISSUE #1232: Strange behavior in http4s URIs
In this code:

```scala
//> using lib "com.disneystreaming.smithy4s::smithy4s-http4s:0.17.19"
import cats.effect.unsafe.IORuntime
import cats.effect.IO

import demo.DemoGen
import demo.Demo
import org.http4s.Response
import cats.effect.kernel.Resource
import cats.implicits._
import cats.effect.implicits._

val c: Client[IO] = Client { req =>

  IO.println(req.uri).toResource *>
    Resource.pure(Response())
}

@main def run =
  SimpleRestJsonBuilder(Demo)
    .client(c)
    .make
    .toTry
    .get
    .getDemo()
    .unsafeRunSync()(IORuntime.global)
```

In 0.17.19, the captured URI is:

```
http://localhost:8080/demo
```

However, on the latest series/0.18 snapshot:

```scala
//localhost:8080/demo
```

That doesn't seem like a valid URL. I think we should dig into it before 0.18 hits shelves.

kubukoz:
Closes #1232

### DIFF
--- a/modules/http4s-kernel/src/smithy4s/http4s/kernel/package.scala
+++ b/modules/http4s-kernel/src/smithy4s/http4s/kernel/package.scala
@@ -97,9 +97,16 @@ package object kernel {
 
   def fromSmithy4sHttpUri(uri: Smithy4sHttpUri): Uri = {
     val path = Uri.Path.Root.addSegments(uri.path.map(Uri.Path.Segment(_)).toVector)
+
     Uri(
       path = path,
-      authority = Some(Uri.Authority(host = Uri.RegName(uri.host), port = uri.port))
+      authority = Some(Uri.Authority(host = Uri.RegName(uri.host), port = uri.port)),
+      scheme = Some {
+        uri.scheme match {
+          case Smithy4sHttpUriScheme.Http  => Uri.Scheme.http
+          case Smithy4sHttpUriScheme.Https => Uri.Scheme.https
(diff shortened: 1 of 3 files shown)


kubukoz:
Something else I'm wondering about:

```scala
    val uriScheme = uri.scheme match {
      case Some(Uri.Scheme.http) => Smithy4sHttpUriScheme.Http
      case _                     => Smithy4sHttpUriScheme.Https
    }
```

this would make `https` the default scheme. I'm not sure we're ready for such a revolution... but it certainly would be worth at least documenting somewhere.

daddykotex:
> Something else I'm wondering about:
> 
> ```scala
>     val uriScheme = uri.scheme match {
>       case Some(Uri.Scheme.http) => Smithy4sHttpUriScheme.Http
>       case _                     => Smithy4sHttpUriScheme.Https
>     }
> ```
> 
> this would make `https` the default scheme. I'm not sure we're ready for such a revolution... but it certainly would be worth at least documenting somewhere.

I'd also do the opposite, match if we have `Some(https)` and use `http` otherwise

daddykotex (approved):
Thanks

Baccata (approved):
Thanks !

PR #1241, opened 2023-10-05
Extra dollar sign to avoid env var substitution in docs

daddykotex:
Previous implementation left the dollar sign as is, this triggered a warning:

```
[warn] [bootstrapped] /Users/David.Francoeur/workspace/dev/smithy4s/modules/bootstrapped/src/generated/smithy4s/example/EnvVarString.scala:12:38: Variable ENV undefined in comment for class EnvVarString in class EnvVarString
[warn] [bootstrapped]   *   This is meant to be used with $ENV_VAR
[warn] [bootstrapped]                                      ^
[warn] [bootstrapped] one warning found
```

Current implementation uses a double `$` sign to avoid this. I tried using `{@literal $}` but they must be using some sort of basic regex that fails, because when I try, it fails with an `Illegal group reference` similar to [this SO question](https://stackoverflow.com/questions/11913709/why-does-replaceall-fail-with-illegal-group-reference)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,6 @@ version
 .bsp
 
 .direnv
+
+#smithy cli metadata
+build/smithy/classpath.json
--- a/modules/bootstrapped/src/generated/smithy4s/example/EnvVarString.scala
+++ b/modules/bootstrapped/src/generated/smithy4s/example/EnvVarString.scala
@@ -1,0 +1,28 @@
+package smithy4s.example
+
+import smithy4s.Hints
+import smithy4s.Schema
+import smithy4s.ShapeId
+import smithy4s.ShapeTag
+import smithy4s.schema.Schema.string
+import smithy4s.schema.Schema.struct
+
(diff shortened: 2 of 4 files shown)


Baccata (approved):

Baccata:
Good thinking !

daddykotex:
I got curious and wondered why `literal` failed, and looking into the compile I found: https://github.com/scala/scala/blob/40beefbb750d326a5ae71ed852c58dc9a31d7e17/src/scaladoc/scala/tools/nsc/doc/base/CommentFactoryBase.scala#L143

which just wraps the character in backtick. So I gave it a shot and it works:

Screenshot 2023-10-05 at 12 10 23
Screenshot 2023-10-05 at 12 09 02

Do you mind if I open another PR for that?

daddykotex:
Did it anyway: https://github.com/disneystreaming/smithy4s/pull/1242

We can close if not needed

PR #1244, opened 2023-10-05
Dynamic: support open enums

### LINKED ISSUE #1243: Open enums: Dynamic support
In #1137, open enum support was added. Unfortunately, the `DynamicModelCompiler` treats all enums as closed. I don't think this can be worked around in userland, because ultimately you end up having to go from `String` / `Int` to `A`, which is abstract to you.

I propose that we add a condition for `@openEnum` just like in the codegen, and produce open enum tags in these cases.

kubukoz:
Closes #1243

### DIFF
--- a/modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
+++ b/modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
@@ -168,18 +168,11 @@ private[dynamic] object Compiler {
         shapeId: ShapeId,
         traits: Map[IdRef, Document],
         lSchema: Eval[Schema[A]]
-    ): Unit =
-      updateWithHints(shapeId, allHints(traits), lSchema)
-
-    private def updateWithHints[A](
-        shapeId: ShapeId,
-        hints: Hints,
-        lSchema: Eval[Schema[A]]
     ): Unit = {
       schemaMap += (shapeId -> lSchema.map { sch =>
         sch
           .withId(shapeId)
-          .addHints(hints)
+          .addHints(allHints(traits))
           .asInstanceOf[Schema[DynData]]
       })
     }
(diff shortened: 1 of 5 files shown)


modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
```diff
@@ -256,9 +256,33 @@ private[dynamic] object Compiler {
             )
           }
 
-      val theEnum = stringEnumeration(values, values)
+      if (shape.traits.contains(IdRef("alloy#openEnum"))) {
+        // the runtime representation of normal enums is Int, but for open enums it's String to support arbitrary unknown values.
```
kubukoz, line 255: **question:** should we go for a `String` representation for all the cases?

modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
```diff
@@ -289,14 +313,34 @@ private[dynamic] object Compiler {
...
-        theEnum
-      )
+        updateWithHints(
+          id, {
+            allHints(shape.traits)
+          },
+          theEnum
+        )
+      }
     }
```
kubukoz, line 332: **issue:** pending - support for 1.x enums. I'll add those later on  in this PR.

kubukoz: added now

modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
Baccata: I suppose that'd make sense.

Baccata: Although, int is fine for the closed-enum case

Baccata (approved):

Baccata (approved):

PR #1254, opened 2023-10-09
Expose UrlForm.parse and UrlFormDecodeError

BalmungSan:
At work, we built our own `SimpleUrlFormProtocol` and the implementation mostly reuses what **smithy4s** already has.
However, we had to use runtime reflection to access these bits. Which, IMHO, could _(should?)_ be public.

Let me know what you think and if I should do something else.

-----

## PR Checklist

- [x] Added unit-tests _(already there)_
- [x] Added documentation
- [x] Updated changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ In 0.18.0, support was added for [open enums](https://disneystreaming.github.io/
 This model-preprocessor aims at removing constraints from output types in AWS specs (as AWS doesn't seem to respect said constraints)
 https://github.com/disneystreaming/smithy4s/pull/1251
 
+## Expose UrlForm.parse and UrlFormDecodeError
+
+In 0.18.0, support was added for `application/x-www-form-urlencoded` data. But, many of its related constructs were private, they are now public for users to access them directly.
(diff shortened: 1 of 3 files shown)


kubukoz (approved):

kubukoz:
oops, didn't check the description earlier, thought it was such an obvious change 🤪

I'd say we should update the changelog. Don't know if there's any documentation about forms so far.

mazurio (approved):

BalmungSan:
@kubukoz  added some docs and the `CHANGELOG` entry.
Let me know if there is any feedback about those or if I missed something else :)

kubukoz:
Looks like it's failing formatting, can you run `sbt formatAll` and commit?

dhpiggott:
> At work, we built our own SimpleUrlFormProtocol

Interesting! What does it do/what is it used for? Is it URL form data in and JSON out by any chance, or something else?

mazurio:
@dhpiggott We use this for OKTA (Token Exchange) - that endpoint uses `x-www-form-urlencoded` that accepts `grant_type=client_credentials` and returns an `access_token` with `scope` etc.

PR #1263, opened 2023-10-16
Throwable toString override

daddykotex:
Default `toString` on `Throwable` prints the classname. This is suboptimal when we could have the case class `toString` implementation available.

- [x] Update bootstrapped code + smoke tests (when the rendering logic is modified)
- [ ] Added documentation
- [x] Updated changelog

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
@@ -663,17 +663,16 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
           val ext = mixins.map(m => line"$m").intercalate(line" with ")
           line" with $ext"
         } else Line.empty
-        val messageField = fields
-          .find { f =>
-            f.hints.contains_(Hint.ErrorMessage) ||
-            f.name === "message"
-          }
-          .filter {
-            _.tpe.dealiased == Type.PrimitiveType(Primitive.String)
-          }
-
         block(line"$decl extends $exception$mixinExtensions") {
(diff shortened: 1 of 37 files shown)


lewisjkl (approved):

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -801,6 +802,10 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     )
   }
 
+  // Default toString on Throwable prints the classname, instead we use the case class toString.
+  private def defaultThrowableGetMessage =
+    line"override def toString(): $string_ = $scalaRuntime._toString(this)"
```
Baccata: Sorry for reviewing this only now. Anyway, the documentation of the ScalaRunTime object states the following : 

```scala
/** The object ScalaRunTime provides support methods required by
 *  the scala runtime.  All these methods should be considered
 *  outside the API and subject to change or removal without notice.
 */
```

therefore, I'm not comfortable with this implementation. Can you elaborate on the rationale for wanting to use this ? I guess it's just for the convenience of it. 

In any case, we can render an actual implementation that'd be using a StringBuilder quite easily.

daddykotex: I missed that completely, that's fair.

> In any case, we can render an actual implementation that'd be using a StringBuilder quite easily.

Any format we prefer? For `Foo(bar: String)`, default implementation renders `Foo(barValue)`, we can choose how to render if we implement it ourselves

Baccata (approved):

PR #1264, opened 2023-10-16
Throwable toString override

daddykotex:

- [x] Update bootstrapped code + smoke tests (when the rendering logic is modified)
- [ ] Added documentation
- [x] Updated changelog

0.17 backport of https://github.com/disneystreaming/smithy4s/pull/1263

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/internals/LineSegment.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/LineSegment.scala
@@ -52,7 +52,7 @@ private[codegen] object LineSegment {
 
     def isAutoImported: Boolean = {
       val value = pkg.mkString(".")
-      NameRef.autoImportedNames.exists(_.equalsIgnoreCase(value))
+      value.startsWith("scala") || value.equalsIgnoreCase("java.lang")
     }
     def getNamePrefix: String = name.split("\\.").head
     def +(piece: String): NameRef = {
@@ -73,13 +73,6 @@ private[codegen] object LineSegment {
   }
 
   object NameRef {
-    val autoImportedNames: List[String] = List(
-      "scala",
-      "java.lang",
-      "scala.Predef",
-      "scala.collection.immutable"
-    )
-
     implicit val nameRefShow: Show[NameRef] = Show.show[NameRef](_.asImport)
     def apply(pkg: String, name: String): NameRef =
       NameRef(pkg.split("\\.").toList, name, List.empty)
--- a/modules/codegen/src/smithy4s/codegen/internals/LineSegment.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/LineSegment.scala
@@ -52,7 +52,7 @@ private[codegen] object LineSegment {
 
     def isAutoImported: Boolean = {
       val value = pkg.mkString(".")
-      value.startsWith("scala") || value.equalsIgnoreCase("java.lang")
+      NameRef.autoImportedNames.exists(_.equalsIgnoreCase(value))
     }
     def getNamePrefix: String = name.split("\\.").head
     def +(piece: String): NameRef = {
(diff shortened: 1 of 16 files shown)


lewisjkl (approved):

modules/codegen/src/smithy4s/codegen/internals/LineSegment.scala
```diff
@@ -73,6 +73,13 @@ private[codegen] object LineSegment {
   }
 
   object NameRef {
+    val autoImportedNames: List[String] = List(
+      "scala",
+      "java.lang",
+      "scala.Predef",
+      "scala.collection.immutable"
```
kubukoz: **question:** should this be here? I don't think scala imports this by default, correct me if I'm wrong but some things like `List`/`Set`/`Map` are available because of aliases in `scala` / `scala.Predef`.

daddykotex: You're probably right, I grabbed that code from the 0.18 branch.

If I remove it, some files in `bootstrapped` will get an additional import for things you mentionned: like `import scala.collection.immutable.Map`. So although it might be incorrect, the incorrect behaviour does not show because we don't use shapes that would be in `scala.collection.immutable` but not imported by the predef.

kubukoz: Can we change occurrences of `NameRef("scala.collection.immutable", "Map")` to `NameRef("Predef", "Map")`? wouldn't that give us the right behavior too?

kubukoz: btw. this sent me on a spiral of madness because I remembered some 2.12-specific issue with immutable/general collections. Thought it was about Set/Map, but it was about `Seq`:

from [2.13 release notes](https://github.com/scala/scala/releases/tag/v2.13.0):

```
Immutable scala.Seq
    Seq is now an alias for collection.immutable.Seq
        Before, it was an alias for the possibly-mutable collection.Seq.
```

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -663,6 +665,9 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     )
   }
 
+  private def defaultThrowableGetMessage =
+    line"override def toString(): $string_ = $scalaRuntime._toString(this)"
```
kubukoz: **issue:** needs a changelog entry (on both versions) ;)

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -663,6 +665,9 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     )
   }
 
+  private def defaultThrowableGetMessage =
```
kubukoz: **suggestion:** I would add a code comment with your explanation from the 0.18 PR:

> Default `toString` on `Throwable` prints the classname. This is suboptimal when we could have the case class `toString` implementation available.

daddykotex: done, good suggestion

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -663,6 +665,10 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     )
   }
 
+  // Default toString on Throwable prints the classname, instead we use the case class toString.
+  private def defaultThrowableGetMessage =
+    line"override def toString(): $string_ = $scalaRuntime._toString(this)"
```
Baccata: See my comment [there](https://github.com/disneystreaming/smithy4s/pull/1263/files#r1377405904)

Baccata:
Closing this one : we're not promising LTS in any way.

PR #1266, opened 2023-10-17
Fix AwsInstanceMetadata decoding

msosnicki:

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/modules/aws-http4s/test/src/smithy4s/aws/AwsInstanceMetadataTest.scala
+++ b/modules/aws-http4s/test/src/smithy4s/aws/AwsInstanceMetadataTest.scala
@@ -1,0 +1,22 @@
+package smithy4s.aws
+
+import smithy4s.aws.kernel.AwsInstanceMetadata
+import smithy4s.Blob
+
+object AwsInstanceMetadataTest extends weaver.FunSuite {
+
+  val codec =
+    internals.AwsJsonCodecs.jsonDecoders.fromSchema(AwsInstanceMetadata.schema)
+
+  test("decodes AwsInstanceMetadata correctly") {
+    val result =
+      codec.decode(Blob("""|{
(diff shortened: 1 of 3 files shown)


modules/aws-http4s/test/src/smithy4s/aws/AwsInstanceMetadataTest.scala
```diff
@@ -0,0 +1,22 @@
...
+    internals.AwsJsonCodecs.jsonDecoders.fromSchema(AwsInstanceMetadata.schema)
+
+  test("decodes AwsInstanceMetadata correctly") {
+    val result =
+      codec.decode(Blob("""|{
+                           |  "RoleArn":"arn:aws:iam::123:user:JohnDoe",
+                           |  "AccessKeyId":"access-key",
+                           |  "SecretAccessKey":"secret-key",
+                           |  "Token":"token",
+                           |  "Expiration":"2023-03-03T17:56:46Z"
```
daddykotex, line 18: To be clear, the issue was that the schema did not define the timestamp format and the AwsInstanceMetadata uses `date-time`

correct?

msosnicki: Yes, it's not well documented (at least I cannot find anything), but that's the behavior I'm observing.

Please see https://repost.aws/questions/QUgcf1EIOPS7GZNboeAiyO9Q/renewing-aws-credentials

Baccata: @msosnicki that's good. Could you inline a comment in the code with that link, and put a line with the changelog for the next release ?

msosnicki: Sure, done

kubukoz (approved):

PR #1267, opened 2023-10-17
Make some keys config-scoped where they weren't

kubukoz:

- ~Added unit-tests (for runtime code)~
- ~Added bootstrapped code + smoke tests (when the rendering logic is modified)~
- ~Added build-plugins integration tests (when reflection loading is required at codegen-time)~
- ~Added alloy compliance tests (when simpleRestJson protocol behaviour is expanded/updated)~
- ~Updated dynamic module to match generated-code behaviour~
- ~Added documentation~
- [x] Updated changelog

### DIFF
--- a/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
+++ b/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
@@ -200,7 +200,7 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
       (config / internalDependencyAsJars).value.map(_.data)
     },
     config / smithy4sExplicitCodegenOnlyDependencies := {
-      transitiveLibraryDependencies(config).value
+      transitiveLibraryDependencies.value
         .filter(
           _.configurations.exists(_.contains(Smithy4s.name))
         )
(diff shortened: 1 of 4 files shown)


daddykotex:
We'll need to update https://disneystreaming.github.io/smithy4s/docs/protocols/aws/aws#sbt as well

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
kubukoz, line 1: **thought:** note that I left `transitiveLibraryDependencies` unchanged because it depends on `libraryDependencies`, which I think nobody actually scopes to the config.

kubukoz: if anyone understands why `libraryDependencies` uses `% Test` rather than being `Test/libraryDependencies`-scoped, and you see an issue with leaving `transitiveLibraryDependencies` unscoped, please let me know!

kubukoz:
Discussed on discord, we'll leave these keys be. There's some complexity with `libraryDependencies` and related keys, plus the whole config-specific feature in smithy4s may not be widely used. https://discord.com/channels/1045676621761347615/1045676622491164724/1163910447725023282

PR #1283, opened 2023-10-31
Support range/length refinement providers for enums

### LINKED ISSUE #1282: Invalid code generated when constraints on enums are used
```smithy
$version: "2"

namespace hello

structure MyStruct {
    @length(max: 3)
    s: Str
}

enum Str {
    FOO
    BARS
}
```

This fails to compile on 0.18.2:

```
[error] /Users/kubukoz/projects/demos/target/scala-2.13/src_managed/main/scala/hello/MyStruct.scala:17:25: could not find implicit value for parameter constraint: smithy4s.RefinementProvider.Simple[smithy.api.Length,hello.Str]
[error]     Str.schema.validated(smithy.api.Length(min = None, max = Some(3L))).optional[MyStruct]("s", _.s),
[error]                         ^
[error] /Users/kubukoz/projects/demos/target/scala-2.13/src_managed/main/scala/hello/MyStruct.scala:19:14: missing argument list for method apply in object MyStruct
[error] Unapplied methods are only converted to functions when a function type is expected.
[error] You can make this conversion explicit by writing `apply _` or `apply(_)` instead of `apply`.
[error]     MyStruct.apply
[error]              ^
[error] two errors found
[error] (Compile / compileIncremental) Compilation failed
```

because there's no `RefinementProvider` available for enums, or a bijection that would allow deriving one.

BTW. `@length` / `@range` on the enum/intEnum shape itself are ignored by the codegen, including `val hints`.

Technically, the Smithy validators allow `@length` on enums, and `@range` on int enums. Should we support this somehow? I'm not sure there's a usecase in limiting the amount of allowed enum values further by imposing constraints, but I also can't say that nobody is doing so.

Baccata:
Fixes https://github.com/disneystreaming/smithy4s/issues/1282

- [x] Added bootstrapped code 
- [x] Updated changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # 0.18.3
 
+* Support length/range constraints on structure fields targeting enums
+
+Although it's weird to allow it, it is actually supported in Smithy.
+
 * Fixing AwsInstanceMetadata codec in [#1266](https://github.com/disneystreaming/smithy4s/pull/1266)
 
 Resolves an issue in which AWS credentials would be decoded using the wrong timestamp format, affecting AWS clients on EC2/ECS.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 0.18.3
 
-* Support length/range constraints on structure fields targeting enums
+* Support constraint traits on members targeting enums
 
 Although it's weird to allow it, it is actually supported in Smithy.
 
--- a/modules/core/src/smithy4s/RefinementProvider.scala
+++ b/modules/core/src/smithy4s/RefinementProvider.scala
@@ -78,23 +78,8 @@ object RefinementProvider extends LowPriorityImplicits {
       },
       (_: ShapeId).show
     )
-}
 
-private[smithy4s] trait LowPriorityImplicits {
-
-  implicit def enumLengthConstraint[E <: Enumeration.Value]
-      : RefinementProvider[Length, E, E] =
-    new LengthConstraint[E](e => e.value.size)
-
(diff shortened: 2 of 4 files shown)


modules/core/src/smithy4s/RefinementProvider.scala
```diff
@@ -82,6 +83,12 @@ object RefinementProvider {
   implicit def mapLengthConstraint[K, V]: Simple[Length, Map[K, V]] =
     new LengthConstraint[Map[K, V]](_.size)
 
+  implicit def enumLengthConstraint[E <: Enumeration.Value]: Simple[Length, E] =
```
kubukoz: **question:** should we also have `@pattern`?

kubukoz: `@idRef` would be nasty. If someone ever creates an enum that matches the ShapeId format, I'll personally go to their house and remove their github account.

Baccata: > question: should we also have @pattern? 

Is it allowed ? you mentioned range and length, but ...

kubukoz: I didn't check the others! 😅

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # 0.18.3
 
+* Support length/range constraints on structure fields targeting enums
```
kubukoz: nitpick: 

```suggestion
* Support constraint traits on members targeting enums
```

kubukoz (approved):

Issue #1258, opened 2023-10-11
`DynamicSchemaIndex.loadModel` is slow


kubukoz:
On my machine (M1 Max MBP), loading a relatively large model (15k+ shapes) takes less than a second on the Smithy (aws) side but over 11 seconds on the `DynamicSchemaIndex`side.

You can try yourself (with granular steps) with the following scala-cli script:

```scala
//> using lib "com.disneystreaming.smithy4s::smithy4s-dynamic:0.18.2"
//> using lib "io.get-coursier:coursier_2.13:2.1.7"
//> using scala "3.3.1"
//> using toolkit "latest"
//> using option "-Wunused:imports"
import coursier._
import coursier.parse.DependencyParser
import smithy4s.Document
import smithy4s.dynamic.DynamicSchemaIndex
import smithy4s.dynamic.NodeToDocument
import software.amazon.smithy.model.Model
import software.amazon.smithy.model.shapes.ModelSerializer
import software.amazon.smithy.model.transform.ModelTransformer
import scala.concurrent.ExecutionContext

def measure[A](tag: String)(f: => A): A = {
  val start = System.nanoTime()
  val result = f
  println(s"$tag: ${(System.nanoTime() - start) / 1000000}ms")
  result
}

@main def main = {

  val jars = measure("fetch jars") {
    Fetch()
      .withDependencies(
        List(
          "aws-quicksight-spec",
          "aws-sagemaker-spec"
        ).takeRight(3)
          .map { sdk =>
            DependencyParser
              .dependency(
                s"com.disneystreaming.smithy:$sdk:2023.09.22",
                "2.13"
              )
              .toOption
              .get
          }
      )
      .run()(ExecutionContext.global)
      // poor man's way of filtering out jars without a smithy manifest. don't judge pls
      .filterNot(_.toString().contains("smithy-model"))
      .filterNot(_.toString().contains("smithy-jmespath"))
      .filterNot(_.toString().contains("smithy-utils"))
  }

  val model = measure("load model") {
    val assembler = Model
      .assembler()

    jars.foreach(jar => assembler.addImport(jar.toURI().toURL()))

    assembler
      .assemble()
      .unwrap()
  }

  println(model.shapes().toList().size() + " shapes")

  val codec = measure("derive codec") {
    Document.decoderFromSchema[smithy4s.dynamic.model.Model]
  }

  // the equivalent of loadModel's first part
  val converted = measure("convert model") {
    val flattenedModel =
[...]

PR #1295, opened 2023-11-14
WIP: dynamic recursive hint bindings

### LINKED ISSUE #537: Schema instantiation hangs
Couldn't reproduce it simpler, so here's a ZIO test:

```scala
//> using scala "2.13.10"
//> using lib "com.disneystreaming.smithy4s::smithy4s-json:0.16.6"
//> using lib "dev.zio::zio-test:1.0.17"
//> using lib "dev.zio::zio-test-sbt:1.0.17"

import smithy4s.schema.Schema
import smithy.api.JsonName
import smithy4s.api.Discriminated
import smithy4s.http.CodecAPI
import smithy4s.HintMask
import smithy4s.IntEnum
import smithy4s.http.json.JCodec

import zio.test._
import zio.test.Assertion._
import smithy4s.schema.Schema
import smithy.api.JsonName
import smithy4s.api.Discriminated
import zio.test._
import zio.test.Assertion._
import zio.duration._

object JsonSpec extends DefaultRunnableSpec {

  private val codecs = smithy4s.http.json.codecs(HintMask(IntEnum))

  private def encodeString[A](a: A)(schema: Schema[A]): String = {
    val c = codecs.compileCodec(schema)

    new String(codecs.writeToArray(c, a))
  }

  def spec: ZSpec[Environment, Failure] = suite("Command JSON codec")(
    test("struct schema serialization ignores @jsonName") {
      val schema = Schema
        .struct[String]
        .apply(
          Schema.string
            .addHints(JsonName("b"))
            .required[String]("a", identity)
        )(identity)

      assert(encodeString("test")(schema))(equalTo("""{"a":"test"}"""))
    },
    test("unions are always encoded as tagged") {

      val schema = {
        val i = Schema.int.oneOf[Either[Int, String]]("intKey", Left(_))
        val s = Schema.string.oneOf[Either[Int, String]]("stringKey", Right(_))

        Schema
          .union[Either[Int, String]](i, s)(_.fold(i(_), s(_)))
          .addHints(Discriminated("type"))
      }

      assert(encodeString[Either[Int, String]](Left(42))(schema))(
        equalTo("""{"intKey":42}""")
      )
    }
  ) @@ TestAspect.timeout(2.seconds)

}
```

Feel free to bump the timeout to 10 seconds or more, chances are you'll hit the timeout more often than not - the spec usually doesn't finish.

I bisected this down to 80a5d912ca101266795f5a425d855c3b1ea04f6a.

kubukoz:
Will close #537.

- [] Added unit-tests (for runtime code)
- [] Added bootstrapped code + smoke tests (when the rendering logic is modified)
- [] Added build-plugins integration tests (when reflection loading is required at codegen-time)
- [] Added alloy compliance tests (when simpleRestJson protocol behaviour is expanded/updated)
- [] Updated dynamic module to match generated-code behaviour
- [] Added documentation
- [] Updated changelog


modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -1172,11 +1173,38 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+    Hint.Native(
+      typedNode = recursion.ana(unfoldNodeAndType)(nodeAndType),
+      rawNode = nodeAndType.node,
+      tr = Type.Ref(tr.namespace, tr.name),
+      isRecursive = isRecursiveTraitOf(shape.getId, tr.toShapeId(), Set.empty)
+    )
   }
 
+  // returns true if there's a reference to `shape` in the trait closure of `id`.
+  private def isRecursiveTraitOf(
```
kubukoz: This looks a bit like the recursive check in `DynamicModelCompiler`, which has known performance issues: #1258.

I'm using this implementation for now, as I don't expect trait trees to be as deep (or wide) as shapes used for data, and so far it seems like a sane assumption... that said, I'm happy to use a more optimized implementation if we can make one.

Notably, this isn't a simple `isRecursive(id)` - I'm checking `shape` in particular appears in the closure of the given trait.

Some examples:

- `documentation -> documentation` is true
- `documentation -> trait` is true (the opposite is also true, which makes me think this is commutative)
- `foo -> documentation` is false for any user-defined `foo`.

It doesn't affect any existing shapes in `bootstrapped`. I added new ones in the PR.

Here are some stdlib examples:

```scala
// Documentation
val hints: Hints = Hints(
  ShapeId("smithy.api", "documentation") -> Document.fromString("Adds documentation to a shape or member using CommonMark syntax."),
  ShapeId("smithy.api", "trait") -> Document.obj(),
)

// Trait
val hints: Hints = Hints(
  ShapeId("smithy.api", "documentation") -> Document.fromString("Makes a shape a trait."),
  ShapeId("smithy.api", "trait") -> Document.obj("selector" -> Document.fromString(":is(simpleType, list, map, structure, union)"), "breakingChanges" -> Document.array(Document.obj("change" -> Document.fromString("presence")), Document.obj("path" -> Document.fromString("/structurallyExclusive"), "change" -> Document.fromString("any")), Document.obj("path" -> Document.fromString("/conflicts"), "change" -> Document.fromString("update"), "severity" -> Document.fromString("NOTE"), "message" -> Document.fromString("Adding more conflicts to a trait could cause previously written models to fail validation.")))),
)

// HttpApiKeyAuth (references the above two and something else, for comparison)
val hints: Hints = Hints(
  smithy.api.Documentation("An HTTP-specific authentication scheme that sends an arbitrary API key in\na header or query string parameter."),
  smithy.api.Trait(selector = Some("service"), structurallyExclusive = None, conflicts = None, breakingChanges = Some(List(smithy.api.TraitDiffRule(change = smithy.api.TraitChangeType.REMOVE.widen, severity = smithy.api.TraitChangeSeverity.ERROR.widen, path = None, message = None)))),
  smithy.api.AuthDefinition(traits = None),
)
```

modules/bootstrapped/src/generated/smithy4s/example/traits/IndirectRecursiveViaMembersTrait.scala
```diff
@@ -0,0 +1,26 @@
...
+import smithy4s.Schema
+import smithy4s.ShapeId
+import smithy4s.ShapeTag
+import smithy4s.schema.Schema.recursive
+import smithy4s.schema.Schema.struct
+
+/** A trait that has indirect recursive references to it via members. */
+final case class IndirectRecursiveViaMembersTrait(member: Option[RecursiveMember] = None)
+
+object IndirectRecursiveViaMembersTrait extends ShapeTag.Companion[IndirectRecursiveViaMembersTrait] {
```
kubukoz, line 13: I'm 99% sure about this one.

The recursive reference to this one only happens via members, and the `schema` of this type is already lazy.

So, if you initialize `IndirectRecursiveViaMembersTrait` first, everything should initialize by the time the recursive reference is used.

What happens if you try to instantiate `RecursiveMember` first?

1. run `RecursiveMember.<clinit>`
2. instantiate `hints` of `RecursiveMember`, which includes a static binding to `IndirectRecursiveViaMembersTrait`
3. run `IndirectRecursiveViaMembersTrait.<clinit>`
4. run hints here, not a big deal
5. run `schema` here. This is lazy so we're good

at least that's what I'm thinking at the moment.

sampleSpecs/traits.smithy
```diff
@@ -0,0 +1,47 @@
...
+structure nonRecursiveTraitReferencingOthers {}
+
+// TODO: this can't be generated because of https://github.com/disneystreaming/smithy4s/issues/1296
+// A trait that has direct recursive references to it via members.
+// @trait
+// structure directRecursiveViaMembersTrait {
+// @directRecursiveViaMembersTrait
+// member: String
+// }
+//
```
kubukoz: This is currently impossible to generate but I think the current state should render something like this:

```scala
implicit val schema: Schema[DirectRecursiveViaMembersTrait] = recursive(
  struct(
    string
      .optional[DirectRecursiveViaMembersTrait]("member", _.member)
      .addHints(smithy4s.example.traits.DirectRecursiveViaMembersTrait())
  ) {
    DirectRecursiveViaMembersTrait.apply
  }.withId(id).addHints(hints)
)
```

which is probably OK because of `recursive` laziness.

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -1172,11 +1173,38 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
       s"Unhandled trait binding:\ntype: $tpe\nvalue: ${Node.printJson(node)}"
   }
 
-  private def unfoldTrait(tr: Trait): Hint.Native = {
+  private def unfoldTrait(shape: Shape, tr: Trait): Hint.Native = {
```
Baccata: I'd rather not modify this method. Instead, I'd create a new `Hint` type holding just the ShapeId and the Node (say `Hint.NativeDynamic` for the sake of the convo), and have a new method do `if isRecursive then createNativeDynamic(trait) else unfoldTrait(trait)` and call that where `unfoldTrait` is currently called.

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
Baccata: > that said, I'm happy to use a more optimized implementation if we can make one.

I'd much rather we did that to be honest, and constructed eagerly a set of all shapes that match the recursion criteria.

sampleSpecs/traits.smithy
kubukoz: aaand it's possible now because of #1305. The generated result looks like this:

```scala
implicit val schema: Schema[DirectRecursiveViaMembersTrait] = recursive(
  struct(
    string
      .optional[DirectRecursiveViaMembersTrait]("member", _.member)
      .addHints(
        smithy4s.example.traits.DirectRecursiveViaMembersTrait(member = None)
      )
  ) {
    DirectRecursiveViaMembersTrait.apply
  }.withId(id).addHints(hints)
)
```

which is pretty much as expected.

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
kubukoz: this is so much nicer, thanks!

kubukoz:
Just came up with this:

```smithy
@trait
@traitWithMember(m: "foo")
structure recursiveViaTraitMember {}

@trait
structure traitWithMember {
    m: M
}

@recursiveViaTraitMember
string M
```

I think `traitWithMember` in the hints of `recursiveViaTraitMember` also has to be a document. Otherwise:

```scala
object RecursiveViaTraitMember extends ShapeTag.Companion[RecursiveViaTraitMember] {
  val id: ShapeId = ShapeId("smithy4s.example.traits", "recursiveViaTraitMember")

  val hints: Hints = Hints(
    smithy4s.example.traits.TraitWithMember(m = Some(smithy4s.example.traits.M("foo"))),
    smithy.api.Trait(selector = None, structurallyExclusive = None, conflicts = None, breakingChanges = None),
  )
  ...
}

object M extends Newtype[String] {
  val id: ShapeId = ShapeId("smithy4s.example.traits", "M")
  val hints: Hints = Hints(
    smithy4s.example.traits.RecursiveViaTraitMember(),
  )
  ...
}
```

The initialization of `RecursiveViaTraitMember.hints` requires initializing `M$`. This may not be a problem with e.g. structs (rather than newtypes), but that's probably a difference we can ignore.

kubukoz:
If I got this right:

image

_Edges that should result in a dynamic hint are marked red, also ignoring the edge from anything to `trait` for brevity_

this is going to be strongly connected components again, but unlike in [DynamicSchemaIndex](https://github.com/disneystreaming/smithy4s/issues/1258#issuecomment-1759100870) we're gonna want to find **edges**, not vertices.

kubukoz:
Closing in favor of #1326 which is much less complex. The recursion tracking via Tarjan should come useful in #1258.

PR #1301, opened 2023-11-17
Update meaning of `required` in both codegen and runtime.

### LINKED ISSUE #1289: Required defaults aren't written
According to [Smithy documentation](https://smithy.io/2.0/spec/type-refinement-traits.html?highlight=default#default-value-serialization):

> 2. All effective default values SHOULD be serialized. This ensures that messages are unambiguous and do not change during deserialization if the default value for a member changes after the message was serialized.

> 4. A member marked `@required` MUST be serialized, including members that have a default.

As far as I understand, this means that the following code (edited):

```scala
//> using lib "com.disneystreaming.smithy4s::smithy4s-json:0.18.3"
import smithy4s.json.JsoniterCodecCompiler

import smithy4s.json.Json
import smithy4s.Schema
import smithy4s.Document
import smithy.api.Default

case class C(s: String, i: Option[Int])

object C {

  implicit val schema: Schema[C] =
    Schema.struct[C](
      Schema
        .string
        .required[C]("s", _.s)
        .addHints(Default(Document.fromString("hello"))),
      Schema.int.optional[C]("i", _.i).addHints(Default(Document.fromInt(42))),
    )(C.apply)

}

def write[A: Schema](a: A) = Json.writeBlob(a).toUTF8String

write(C("hello", Some(42)))
// res0: String = {}
write(C("hello2", None))
// res1: String = {"s":"hello2","i":null}

```

should/must write this:

`C("hello", None)` -> `{"s": "hello", "i": 42}`

whereas it serializes as:

`{"i": null}`.

- `s` is a required member so MUST be serialized according to rule 4 from the docs
- `i`'s effective default SHOULD be serialized so that it's unambiguous on deserialization.

Looking forward to thoughts on how we should act: I think the encoders for JSON, HTTP metadata, XML and documents (and any others we may have) should be updated to account for these rules: basically write the default by default, only using another value if it's non-null.

Baccata:
This PR aims at moving towards a more principled way of handling required and defaults. The notion of "required" is now more strictly coupled to the presence of the `smithy.api#required` trait. 

* Update codegen to not use the "required" terminology, as in smithy it is explicitly tied to the presence of the `required` trait.
* Makes sure that the `smithy.api.Required` hint is only generated when the `smithy.api#required` is applied at the Smithy level
* Updates `getUnlessDefault` or `foreachUnlessDefault` to return value/ apply lambda when the `smithy.api#required` is applied.
* Update default Json hint mask to have `smithy.api.Required` explicitly
* Adds relevant unit-tests for Document and Json encodings

Closes https://github.com/disneystreaming/smithy4s/issues/1289

- [x] Updated changelog


sampleSpecs/defaults.smithy
```diff
@@ -1,66 +1,74 @@
...
+    three: String
 }
 
 @mixin()
 structure DefaultInMixinTest {
-  one: String = "test"
+    one: String = "test"
+}
+
+structure DefaultVariants {
```
daddykotex, line 65: ahh, until I saw this structure I was unsure what use case we were not covering properly but now it's clearer

a field *not marked with required* with a default would use `required` before, now it use `field`

modules/codegen/src/smithy4s/codegen/internals/IR.scala
```diff
@@ -141,13 +141,37 @@ private[internals] case class StreamingField(
 
 private[internals] object Field {
 
+  sealed trait Modifier {
```
daddykotex, line 144: great improvement

daddykotex (approved):
great work

modules/bootstrapped/test/src/smithy4s/DefaultSmokeSpec.scala
```diff
@@ -0,0 +1,44 @@
...
+import smithy4s.example._
+
+class DefaultSmokeSpec() extends FunSuite {
+
+  test(
+    "Fields with defaults do not get implicitly `smithy.api.Required` hints"
+  ) {
+
+    val document = Document.encode(
+      DefaultVariants("default", "default", "default", Some("default"))
```
kubukoz, line 29: suggestion: named parameters would be very helpful

modules/json/test/src/smithy4s/json/SchemaVisitorJCodecTests.scala
```diff
@@ -584,4 +584,36 @@ class SchemaVisitorJCodecTests() extends FunSuite {
     }
   }
 
+  test(
+    "fields marked with @required and @default should always be encoded"
```
kubukoz, line 588: nitpick: good candidate for a property test?

kubukoz (approved):

modules/bootstrapped/src/generated/smithy4s/example/DefaultVariants.scala
```diff
@@ -0,0 +1,25 @@
...
+object DefaultVariants extends ShapeTag.Companion[DefaultVariants] {
+  val id: ShapeId = ShapeId("smithy4s.example", "DefaultVariants")
+
+  val hints: Hints = Hints.empty
+
+  implicit val schema: Schema[DefaultVariants] = struct(
+    string.required[DefaultVariants]("req", _.req),
+    string.required[DefaultVariants]("reqDef", _.reqDef).addHints(smithy.api.Default(smithy4s.Document.fromString("default"))),
+    string.field[DefaultVariants]("optDef", _.optDef).addHints(smithy.api.Default(smithy4s.Document.fromString("default"))),
+    string.optional[DefaultVariants]("opt", _.opt),
```
lewisjkl, line 21: I'm not sure I fully understand why `field` is used here rather than `optional` and it is also used in other places where `required` was used prior. Is `field` only used when a default value is present and the field is not required?

Baccata: >  Is field only used when a default value is present and the field is not required 

Pretty much. 

* `optional` forces the type of the getter to be `S => Option[A]`
* `required` lets you do `S => A`, but automatically enriches the hints with `smithy.api.Required`
* `field` lets you do `S => A`, and doesn't enrich the hints with anything

kubukoz:
This isn't going to affect Dynamic, right?

edit: I suppose `getUnlessDefault` and friends _will_ be affected but that's about it

modules/bootstrapped/src/generated/smithy4s/example/DefaultVariants.scala
lewisjkl: Got it yeah that makes sense

lewisjkl (approved):

PR #1304, opened 2023-11-17
Escape astrisks following slashes in documentation

### LINKED ISSUE #1300: Generated Code may contain Scaladoc with Unclosed Comment Errors
I encountered this with the Cloudformation spec, but I suppose it's possible with any spec that has documentation that contains the `/*` string.

Using `v0.18.3`, build an app using `smithy4sAwsSpecs += AWS.cloudformation`, and try to compile it. The generated code contains comments like

```
*               <p>{ "Version": "2012-10-17", "Statement": [{ "Effect": "Deny", "Action": "cloudformation:DescribeStacks",
*       "NotResource": "arn:aws:cloudformation:*:*:stack/\*\/*" }] }</p>
```

and the `/*` causes the Scala compiler to complain of unclosed comments.

We briefly discussed this in Discord and it sounds like this is the function that needs to be tweaked: 

https://github.com/disneystreaming/smithy4s/blob/8893d93ffa1a38a32d49be83ca54f8fe21b4935d/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala#L221-L224

I'll try to address this by replacing any `/*` in the contents of `rawLines` with `/&#42;`, which is the HTML escape code for `*`, as suggested by [this StackOverflow answer](https://stackoverflow.com/a/31741493). There may be a better solution out there but that should at least be a start.

bpholt:
When `/*` is present in Smithy documentation, it causes Unclosed Comment errors in the generated Scala code. Replacing the `*` with `&ast;` avoids the issue.

Fixes #1300.

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/modules/bootstrapped/src/generated/smithy4s/example/DocumentationComment.scala
+++ b/modules/bootstrapped/src/generated/smithy4s/example/DocumentationComment.scala
@@ -1,0 +1,28 @@
+package smithy4s.example
+
+import smithy4s.Hints
+import smithy4s.Schema
+import smithy4s.ShapeId
+import smithy4s.ShapeTag
+import smithy4s.schema.Schema.string
+import smithy4s.schema.Schema.struct
+
+/** We should be able to use comments in documentation /&ast; \*\/
+  * @param member
+  *   /&ast;
+  */
+final case class DocumentationComment(member: Option[String] = None)
+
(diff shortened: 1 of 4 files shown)


Baccata:
Disregarding CLA because it's an arguably simple fix without IP value that would make the signature of a Corporate CLA relevant.

Baccata (approved):

PR #1305, opened 2023-11-21
fix: Do not evaluate traits of traits members recursively

### LINKED ISSUE #1296: Rendering: Traits used recursively in their direct struct members cause a StackOverflowError
Version: 0.18.3

Minimal example:

```smithy
$version: "2"

namespace input

@trait
structure Person {
    @Person
    name: String
}
```

<details><summary>Stack trace</summary>
<p>

```
java.lang.StackOverflowError
	at java.base/java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1694)
	at software.amazon.smithy.model.Model.toSet(Model.java:773)
	at software.amazon.smithy.model.Model.getMemberShapes(Model.java:470)
	at smithy4s.codegen.internals.SmithyToIR.smithy4s$codegen$internals$SmithyToIR$$getAdtParent(SmithyToIR.scala:1124)
	at smithy4s.codegen.internals.SmithyToIR$UnRef$.$anonfun$unapply$3(SmithyToIR.scala:1148)
	at scala.Option.orElse(Option.scala:477)
	at smithy4s.codegen.internals.SmithyToIR$UnRef$.$anonfun$unapply$1(SmithyToIR.scala:1148)
	at scala.Option.map(Option.scala:242)
	at smithy4s.codegen.internals.SmithyToIR$UnRef$.unapply(SmithyToIR.scala:1142)
	at smithy4s.codegen.internals.SmithyToIR.unfoldNodeAndType(SmithyToIR.scala:1183)
	at smithy4s.codegen.internals.SmithyToIR.$anonfun$unfoldTrait$1(SmithyToIR.scala:1177)
	at smithy4s.codegen.internals.recursion$.hylo(recursion.scala:41)
	at smithy4s.codegen.internals.recursion$.ana(recursion.scala:63)
	at smithy4s.codegen.internals.SmithyToIR.smithy4s$codegen$internals$SmithyToIR$$unfoldTrait(SmithyToIR.scala:1177)
	at smithy4s.codegen.internals.SmithyToIR.$anonfun$hints$5(SmithyToIR.scala:968)
	at scala.collection.immutable.List.map(List.scala:246)
	at smithy4s.codegen.internals.SmithyToIR.smithy4s$codegen$internals$SmithyToIR$$hints(SmithyToIR.scala:968)
	at smithy4s.codegen.internals.SmithyToIR$ShapeExt.$anonfun$fields$2(SmithyToIR.scala:1001)
	at scala.collection.StrictOptimizedIterableOps.map(StrictOptimizedIterableOps.scala:100)
	at scala.collection.StrictOptimizedIterableOps.map$(StrictOptimizedIterableOps.scala:87)
	at scala.collection.mutable.ArrayBuffer.map(ArrayBuffer.scala:43)
	at smithy4s.codegen.internals.SmithyToIR$ShapeExt.fields(SmithyToIR.scala:990)
	at smithy4s.codegen.internals.SmithyToIR.unfoldNodeAndType(SmithyToIR.scala:1186)
	at smithy4s.codegen.internals.SmithyToIR.$anonfun$unfoldTrait$1(SmithyToIR.scala:1177)
	at smithy4s.codegen.internals.recursion$.hylo(recursion.scala:41)
	at smithy4s.codegen.internals.recursion$.ana(recursion.scala:63)
	at smithy4s.codegen.internals.SmithyToIR.smithy4s$codegen$internals$SmithyToIR$$unfoldTrait(SmithyToIR.scala:1177)
	at smithy4s.codegen.internals.SmithyToIR.$anonfun$hints$5(SmithyToIR.scala:968)
	at scala.collection.immutable.List.map(List.scala:246)
	at smithy4s.codegen.internals.SmithyToIR.smithy4s$codegen$internals$SmithyToIR$$hints(SmithyToIR.scala:968)
	at smithy4s.codegen.internals.SmithyToIR$ShapeExt.$anonfun$fields$2(SmithyToIR.scala:1001)
	at scala.collection.StrictOptimizedIterableOps.map(StrictOptimizedIterableOps.scala:100)
	at scala.collection.StrictOptimizedIterableOps.map$(StrictOptimizedIterableOps.scala:87)
	at scala.collection.mutable.ArrayBuffer.map(ArrayBuffer.scala:43)
	at smithy4s.codegen.internals.SmithyToIR$ShapeExt.fields(SmithyToIR.scala:990)
	at smithy4s.codegen.internals.SmithyToIR.unfoldNodeAndType(SmithyToIR.scala:1186)
	at smithy4s.codegen.internals.SmithyToIR.$anonfun$unfoldTrait$1(SmithyToIR.scala:1177)
	at smithy4s.codegen.internals.recursion$.hylo(recursion.scala:41)
	at smithy4s.codegen.internals.recursion$.ana(recursion.scala:63)
	at smithy4s.codegen.internals.SmithyToIR.smithy4s$codegen$internals$SmithyToIR$$unfoldTrait(SmithyToIR.scala:1177)
	at smithy4s.codegen.internals.SmithyToIR.$anonfun$hints$5(SmithyToIR.scala:968)
	at scala.collection.immutable.List.map(List.scala:246)
	at smithy4s.codegen.internals.SmithyToIR.smithy4s$codegen$internals$SmithyToIR$$hints(SmithyToIR.scala:968)
	at smithy4s.codegen.internals.SmithyToIR$ShapeExt.$anonfun$fields$2(SmithyToIR.scala:1001)
	at scala.collection.StrictOptimizedIterableOps.map(StrictOptimizedIterableOps.scala:100)
	at scala.collection.StrictOptimizedIterableOps.map$(StrictOptimizedIterableOps.scala:87)
...
``` 

</p>
</details> 

This is probably not a recent regression.

Note: this works fine:

```smithy
$version: "2"

namespace input

@trait
structure Person {
    b: B
}

structure B {
    @Person
    s: String
}
```

Note 2: unions aren't affected:

```smithy
$version: "2"

namespace input

@trait
union Person {
    @Person(u: "demo")
    name: String
    u: String
}
```

Note 3: collections (list member, map key/value) are affected.

Note 4: EDIT: collections are a different issue, actually.

ghostbuster91:
Closes #1296 

- [ ] Added documentation
- [ ] Updated changelog


modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -1025,6 +1025,53 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
       }
     }
 
+    def getFieldsPlain = {
```
ghostbuster91: **thought:** this is just copied `fields` method with a single difference that it does not call `hints(member)` inside. Once we agree that this is the correct solution we can deduplicate code between these two methods. I haven't done it deliberately to make it clearer where the problem was and also not too spend to much time on PoC solution.

modules/codegen/test/src/smithy4s/codegen/internals/DefaultRenderModeSpec.scala
```diff
@@ -151,4 +151,133 @@ final class DefaultRenderModeSpec extends munit.FunSuite {
     TestUtils.runTest(smithy, scalaCode)
   }
 
+  test("allow for self referencing trait - structure") {
```
ghostbuster91: **issue:** I am sure that this is not the right place for these kind of tests, but I wasn't sure where to put them. Also, I don't need to check the generated code, all I need is to assert on the IR.

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -1025,6 +1025,53 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+            if (defaultRenderMode == DefaultRenderMode.Full)
+              maybeDefault(member)
+            else List.empty
+          val defaultable = member.hasTrait(classOf[DefaultTrait]) &&
+            !member.tpe.exists(_.isExternal)
+          (
+            member.getMemberName(),
+            member.tpe,
+            member.hasTrait(classOf[RequiredTrait]) || defaultable,
+            default ++ noDefault // no call to hints(member)
```
ghostbuster91: **thought:** The root cause of the issue is that we end up calling recursively https://github.com/disneystreaming/smithy4s/blob/95c295d3d217ef5e020d8790b8d29e4b40f88276/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala#L1177  

This is why we are getting stack overflow even though recursion schemes are in play.

In general, I see two approaches to fix that:
1. keep the model strict and have a separate function for accessing nested structures fields - current one
2. change the model to be lazy; notice that this might also require change in the renderer (or whatever else is using that structure, to prevent infinite recursion there)

Why it is safe to omit the call?

AFAICT This place is called only when we are calculating members of a trait on a call site e.g.: 
```
@Person()
p: String
```
we attempt creation of `internals.Field` for the member `p`. From its point of view we only care about the id of the trait that targets it and its values (including default values). We don't need to know about its structure.

Baccata: I'm happy with going with option 1, but as you hinted [here](https://github.com/disneystreaming/smithy4s/pull/1305/files#r1400785273) it needs a little bit of deduplication. I'd like for the implementation to be unified with the one of `getField` by means of a private method receiving a flag parameter .

ghostbuster91: @Baccata thanks, I will do the deduplication then. Any suggestion where to put the tests?

modules/codegen/test/src/smithy4s/codegen/internals/DefaultRenderModeSpec.scala
Baccata: Instead of doing a rendering test, you can do a smoke test in the bootstrapped module, adding a smithy file to the sample specs and exercising the code in the tests to verify that the schema doesn't blow up during its initialisation

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -1048,6 +1048,20 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+      *
+      * See https://github.com/disneystreaming/smithy4s/issues/1296 for details.
+      */
+    def fields: List[Field] = fieldsInternal(hintsExtractor = hints)
+
+    /**
+      * Should be used only on the call site 
+      * of the trait application where there is no need to call `unfoldTrait` for every hint of the trait. 
+      */
+    def getFieldsPlain: List[Field] =
```
ghostbuster91, line 1062: **thought:** this could be a different type which would make a mistake impossible. Question is, whether it is worth the effort.

Baccata (approved):
Very well done !

PR #1315, opened 2023-11-27
Default handling

### LINKED ISSUE #1290: Support configuration of explicit defaults in more encoders
In smithy4s, JSON writers and the `SimpleRestJson` builder have a configuration option that controls how default values are handled on the write side: https://disneystreaming.github.io/smithy4s/docs/protocols/simple-rest-json/overview#explicit-null-encoding

**Note:** the documentation is a bit stale, as it only mentions optional fields and nulls. In reality, the behavior of the `@default` trait is also controlled by this option: when disabled, a value that matches the default will simply be omitted from the payload. The documentation **should be updated** to reflect this.

Also, the combination of `optional + @default` makes the field essentially required - with the explicit defaults setting on, you won't see a null in this case, but rather the default value.

All this is only limited to JSON - this feature should also be supported in other encoders.

Here are the ones I know of where it'd make sense:
- [ ] `SchemaVisitorMetadataWriter`: this is used to write HTTP headers, query parameters. Ends up being surfaced via `Metadata.Encoder`.
- [x] `SchemaVisitorPathEncoder`: writes HTTP path segments. This already always writes default values, otherwise URIs might be invalid.
- [ ] `DocumentEncoderSchemaVisitor`: encodes values to documents.

We can also take a look at the others: look for usages of `foreachUnlessDefault` and `getUnlessDefault` methods, as not-writing default values seems to be their primary purpose.

For each of the encoders/writers that are missing such a configuration option, add one.

- this needs to be controllable from a public API, e.g. `Metadata.Encoder.withExplicitDefaults(true)`
- the change needs to be backwards compatible, and implemented in a future-aware way, i.e. following https://docs.scala-lang.org/overviews/core/binary-compatibility-for-library-authors.html (e.g. don't expose the `copy` method of the configuration)
- the change needs to be documented
- (possibly contentious point) the existing `SimpleRestJsonBuilder` configuration should affect these new options within the scope of said builder. I don't have a strong stance on this but it seems to be the most consistent approach.

Feel free to pick up any part of this - the ticket can be implemented incrementally. Just let us know which one you're taking on!

msosnicki:

Closes #1290 

- [x] Added documentation
- [x] Updated changelog

### DIFF
--- a/modules/core/src/smithy4s/Document.scala
+++ b/modules/core/src/smithy4s/Document.scala
@@ -1,5 +1,5 @@
 /*
- *  Copyright 2021-2023 Disney Streaming
+ *  Copyright 2021-2024 Disney Streaming
  *
  *  Licensed under the Tomorrow Open Source Technology License, Version 1.0 (the "License");
  *  you may not use this file except in compliance with the License.
--- a/modules/core/src/smithy4s/Document.scala
+++ b/modules/core/src/smithy4s/Document.scala
@@ -105,9 +105,21 @@ object Document {
     def fromSchema[A](
         schema: Schema[A],
         cache: Cache
+    ): Encoder[A] =
+      fromSchema(schema, cache, explicitDefaultsEncoding = false)
+
+    def fromSchema[A](
+        schema: Schema[A],
+        cache: Cache,
+        explicitDefaultsEncoding: Boolean
     ): Encoder[A] = {
       val makeEncoder =
-        schema.compile(new DocumentEncoderSchemaVisitor(cache))
+        schema.compile(
+          new DocumentEncoderSchemaVisitor(
+            cache,
+            explicitDefaultsEncoding = explicitDefaultsEncoding
+          )
+        )
       new Encoder[A] {
         def encode(a: A): Document = {
           makeEncoder.apply(a)
--- a/modules/core/src/smithy4s/Document.scala
+++ b/modules/core/src/smithy4s/Document.scala
@@ -98,27 +98,29 @@ object Document {
     def encode(a: A): Document
   }
 
-  object Encoder extends CachedSchemaCompiler.DerivingImpl[Encoder] {
+  trait EncoderCompiler extends CachedSchemaCompiler[Encoder] {
+    def withExplicitDefaultsEncoding(
+        explicitDefaultsEncoding: Boolean
+    ): EncoderCompiler
+  }
+
+  object Encoder
+      extends CachedEncoderCompilerImpl(explicitDefaultsEncoding = false)
+
+  private[smithy4s] class CachedEncoderCompilerImpl(
+      explicitDefaultsEncoding: Boolean
+  ) extends CachedSchemaCompiler.DerivingImpl[Encoder]
+      with EncoderCompiler {
 
     protected type Aux[A] = internals.DocumentEncoder[A]
 
     def fromSchema[A](
         schema: Schema[A],
         cache: Cache
-    ): Encoder[A] =
-      fromSchema(schema, cache, explicitDefaultsEncoding = false)
-
-    def fromSchema[A](
-        schema: Schema[A],
-        cache: Cache,
-        explicitDefaultsEncoding: Boolean
     ): Encoder[A] = {
       val makeEncoder =
         schema.compile(
-          new DocumentEncoderSchemaVisitor(
-            cache,
-            explicitDefaultsEncoding = explicitDefaultsEncoding
-          )
+          new DocumentEncoderSchemaVisitor(cache, explicitDefaultsEncoding)
         )
       new Encoder[A] {
         def encode(a: A): Document = {
@@ -127,6 +129,11 @@ object Document {
       }
     }
 
+    def withExplicitDefaultsEncoding(
+        explicitDefaultsEncoding: Boolean
+    ): EncoderCompiler = new CachedEncoderCompilerImpl(
+      explicitDefaultsEncoding = explicitDefaultsEncoding
+    )
   }
 
   type Decoder[A] =
--- a/modules/core/src/smithy4s/http/Metadata.scala
+++ b/modules/core/src/smithy4s/http/Metadata.scala
@@ -1,5 +1,5 @@
 /*
- *  Copyright 2021-2023 Disney Streaming
+ *  Copyright 2021-2024 Disney Streaming
  *
  *  Licensed under the Tomorrow Open Source Technology License, Version 1.0 (the "License");
  *  you may not use this file except in compliance with the License.
--- a/modules/core/src/smithy4s/http/Metadata.scala
+++ b/modules/core/src/smithy4s/http/Metadata.scala
@@ -211,14 +211,42 @@ object Metadata {
     */
   type Encoder[A] = smithy4s.codecs.Encoder[Metadata, A]
 
-  object Encoder extends CachedEncoderCompilerImpl(awsHeaderEncoding = false) {
+  private final class MetadataEncoder(
+      awsHeaderEncoding: Boolean,
+      explicitDefaultsEncoding: Boolean
+  ) extends CachedEncoderCompilerImpl(
+        awsHeaderEncoding = awsHeaderEncoding,
+        explicitDefaultsEncoding = explicitDefaultsEncoding
+      ) {
     type Compiler = CachedSchemaCompiler[Encoder]
   }
-  private[smithy4s] object AwsEncoder
-      extends CachedEncoderCompilerImpl(awsHeaderEncoding = true)
 
-  private[http] class CachedEncoderCompilerImpl(awsHeaderEncoding: Boolean)
-      extends CachedSchemaCompiler.DerivingImpl[Encoder] {
+  object Encoder
+      extends CachedEncoderCompilerImpl(
+        awsHeaderEncoding = false,
+        explicitDefaultsEncoding = false
+      ) {
+    type Compiler = CachedSchemaCompiler[Encoder]
+
+    def make(
+        awsHeaderEncoding: Boolean,
+        explicitDefaultsEncoding: Boolean
+    ): Metadata.Encoder.Compiler = new MetadataEncoder(
+      awsHeaderEncoding = awsHeaderEncoding,
+      explicitDefaultsEncoding = explicitDefaultsEncoding
+    )
+  }
+
+  private[smithy4s] object AwsEncoder
+      extends CachedEncoderCompilerImpl(
+        awsHeaderEncoding = true,
+        explicitDefaultsEncoding = false
+      )
+
+  private[http] class CachedEncoderCompilerImpl(
+      awsHeaderEncoding: Boolean,
+      explicitDefaultsEncoding: Boolean
+  ) extends CachedSchemaCompiler.DerivingImpl[Encoder] {
 
     type Aux[A] = internals.MetaEncode[A]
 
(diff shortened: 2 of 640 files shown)


modules/http4s/test/src/smithy4s/http4s/NullsAndDefaultEncodingSuite.scala
```diff
@@ -0,0 +1,235 @@
...
+  }
+
+  test("client - explicit defaults encoding = true") {
+    runClientTest(explicitDefaults = true, OperationInput())
+      .map { request =>
+        assert.same(
+          Map(
+            ci"optional-header-with-default" -> "optional-header-with-default",
+            ci"required-header-with-default" -> "required-header-with-default",
+            ci"optional-header" -> ""
```
msosnicki: **question**: I think `optional-header` should be omitted here?

modules/http4s/test/src/smithy4s/http4s/NullsAndDefaultEncodingSuite.scala
```diff
@@ -0,0 +1,235 @@
...
+          ),
+          request.headers
+        ) && assert.same(
+          Map(
+            "optional-query-with-default" -> "optional-query-with-default",
+            "required-query-with-default" -> "required-query-with-default",
+            "optional-query" -> ""
```
msosnicki: **question**: same as with header above?

modules/http4s/test/src/smithy4s/http4s/NullsAndDefaultEncodingSuite.scala
```diff
@@ -0,0 +1,235 @@
...
+import io.circe.Json
+import org.typelevel.ci.CIString
+import org.typelevel.ci._
+import org.http4s.circe.CirceInstances
+import smithy4s.Document
+import org.http4s.client.Client
+import smithy4s.example.OperationInput
+import cats.effect.kernel.Deferred
+
+object NullsAndDefaultEncodingSuite extends SimpleIOSuite with CirceInstances {
```
msosnicki, line 18: **question**: wasn't sure if this is a proper way, and place (`http4s` module) to test it.

modules/http4s/test/src/smithy4s/http4s/NullsAndDefaultEncodingSuite.scala
```diff
@@ -0,0 +1,235 @@
...
+            "optional" -> Json.Null,
+            "optionalWithDefault" -> Json.fromString("optional-default"),
+            "requiredWithDefault" -> Json.fromString("required-default")
+          ),
+          request.body
+        )
+      }
+  }
+
+  test("document encoder - all default") {
```
msosnicki: **Question**: Do we need a flag to control the `Document.Encoder` behavior?

kubukoz: imo yes

msosnicki: I think this will be a breaking change, it requires changing `DocumentEncoderSchemaVisitor`, which is public.

Baccata: Considering it's in `internals`, we deem it's okay to break. If you're a third party who depends on what's in an `internals` package, it's your funeral. 

Moreover, the reason we've not been ultra strict wrt private/public is because of a drastic lack of tooling that would help us do so. We'd like to build such tooling (probably by means of scalafix)

But I think it'd be totally possible to retain binary compatibility by means of secondary constructors.

kubukoz: ah, I didn't realize it was in `internals`. Alright then

modules/http4s/test/src/smithy4s/http4s/NullsAndDefaultEncodingSuite.scala
```diff
@@ -0,0 +1,262 @@
...
+  }
+
+  private def runClientTest(
+      explicitDefaults: Boolean,
+      input: OperationInput
+  ): IO[TestRequest] = {
+    val resources = for {
+      promise <- Deferred[IO, Request[IO]].toResource
+      httpClient: Client[IO] = Client(req =>
+        promise.complete(req).as(Response[IO]()).toResource
```
kubukoz: **issue:** should mention that this doesn't necessarily copy over the entire request body. While we're not using the body, let's either use `req.toStrict` first, or leave a comment about it.

modules/http4s/test/src/smithy4s/http4s/NullsAndDefaultEncodingSuite.scala
```diff
@@ -0,0 +1,262 @@
...
+        .client(httpClient)
+        .resource
+    } yield (promise, client)
+    resources.use { case (promise, client) =>
+      client.operation(input) >> promise.get.flatMap { req =>
+        val query =
+          req.uri.query.pairs.flatMap(kv => kv._2.map(kv._1 -> _)).toMap
+        val labels = req.uri.path.segments
+          .map(_.toString)
+          .filterNot(_ == "operation")
```
kubukoz: **question:** why don't we keep the `operation` segment?

modules/http4s/test/src/smithy4s/http4s/NullsAndDefaultEncodingSuite.scala
```diff
@@ -0,0 +1,262 @@
...
+      client <- SimpleRestJsonBuilder
+        .withExplicitDefaultsEncoding(explicitDefaults)
+        .apply(ServiceWithNullsAndDefaults)
+        .client(httpClient)
+        .resource
+    } yield (promise, client)
+    resources.use { case (promise, client) =>
+      client.operation(input) >> promise.get.flatMap { req =>
+        val query =
+          req.uri.query.pairs.flatMap(kv => kv._2.map(kv._1 -> _)).toMap
```
kubukoz: **suggestion:** extract a `queryToMap` so that it sits next to `headersToMap`

modules/http4s/test/src/smithy4s/http4s/NullsAndDefaultEncodingSuite.scala
msosnicki: For easier asserts, the `operation` part identifies the endpoint, and I wanted the test to assert just on the labels that are part of the endpoint. I can change though, so please let me know

kubukoz: Up to you, I just found it a bit strange. If it's not adding a lot to the diff, I'd assert on the whole thing.

msosnicki: Sure, I don't mind, updated

kubukoz (approved):

modules/core/src/smithy4s/http/Metadata.scala
```diff
@@ -211,14 +211,42 @@ object Metadata {
...
   }
+
+  object Encoder
+      extends CachedEncoderCompilerImpl(
+        awsHeaderEncoding = false,
+        explicitDefaultsEncoding = false
+      ) {
+    type Compiler = CachedSchemaCompiler[Encoder]
+
+    def make(
```
Baccata: not a fan of this. `awsHeaderEncoding` shouldn't be exposed to the users here. 

Also, the UX I'd like would be : `Encoder.withExplicitDefaultsEncoding(true)`. The `CachedEncoderCompilerImpl` class should carry the method, which implies that we need to create an public interface for this that it will implement (as `CachedEncoderCompilerImpl` is private). 

Also this `make` method should be removed.

modules/core/src/smithy4s/Document.scala
```diff
@@ -105,9 +105,21 @@ object Document {
     def fromSchema[A](
         schema: Schema[A],
         cache: Cache
+    ): Encoder[A] =
+      fromSchema(schema, cache, explicitDefaultsEncoding = false)
+
+    def fromSchema[A](
+        schema: Schema[A],
+        cache: Cache,
+        explicitDefaultsEncoding: Boolean
```
Baccata: Similarly to my comment [here](https://github.com/disneystreaming/smithy4s/pull/1315/files#r1442591876), this method should be removed in favour of a `withExplicitDefaultsEncoding(explicitDefaults: Boolean)` on the `Encoder`. 

This implies the creation of public interface. You can take [this](https://github.com/disneystreaming/smithy4s/blob/cebdd437b02608b9f9e834c9b1b03c0a37036d5a/modules/json/src/smithy4s/json/JsoniterCodecCompiler.scala#L31) as inspiration.

modules/http4s/test/src/smithy4s/http4s/NullsAndDefaultEncodingSuite.scala
```diff
@@ -0,0 +1,277 @@
...
+          "requiredHeaderWithDefault" -> Document.fromString(
+            "required-header-with-default"
+          )
+        ),
+        result
+      )
+    )
+  }
+
+  test("document encoder - all default - explicit defaults encoding = true") {
```
Baccata: this should go in the `DocumentSpec` in the bootstrapped module.

modules/http4s/test/src/smithy4s/http4s/NullsAndDefaultEncodingSuite.scala
```diff
@@ -0,0 +1,277 @@
...
+          "requiredHeaderWithDefault" -> Document.fromString(
+            "required-header-with-default"
+          )
+        ),
+        result
+      )
+    )
+  }
+
+  test("document encoder - default overrides") {
```
Baccata: this should go in the `DocumentSpec` in the bootstrapped module.

Baccata (changes requested):
Good work overall, but the UX needs to be addressed. Also a couple tests need to be moved

modules/core/src/smithy4s/Document.scala
msosnicki: There is one problem with the above approach, currently, the following doesn't compile:

```
Json.jsoniter.withMaxArity(1).fromSchema(Schema.string, CompilationCache.nop)
```

This is because `JsoniterCodecCompiler` inherits abstract type `Cache` type, wheareas in `CachedSchemaCompiler.Impl` it is set to `type Cache = CompilationCache[Aux]`. So the information about cache is lost when calling methods from `JsoniterCodecCompiler`.

Is it a known limitation?

Baccata: It's not a bug, it's a feature 😄 

If you want to compile a codec without cache, you should be able to call 

```
Json.jsoniter.withMaxArity(1).fromSchema(Schema.string)
```

Otherwise, if you wanted to use a shared cache across several codec compilations, you would have to do : 

```
val compiler = Json.jsoniter.withMaxArity(1) 
val cache = compiler.createCache() 
val stringCodec = compiler.fromSchema(Schema.string, cache) 
```

msosnicki:
@Baccata addressed all the comments, not sure if `EncoderCompilers` should stay where they are or should be moved somewhere.

Baccata (approved):

Baccata:
@kubukoz do you want to review again ? In particular, do you want more documentation ? I wouldn't mind having more of it

kubukoz:
> @kubukoz do you want to review again ? In particular, do you want more documentation ? I wouldn't mind having more of it

I'm happy with the current state, but some docs for the Document encoding/decoding would be welcome. I think that doesn't exist at all, though, so let's make it a separate issue which will cover the basics and the config options.

kubukoz (changes requested):
Needs a changelog entry

msosnicki:
> Needs a changelog entry

Done

kubukoz:
> > Needs a changelog entry
> 
> Done

didn't notice this 😭 thanks, LGTM

kubukoz (approved):

PR #1323, opened 2023-12-08
Use field.getUnlessDefault instead of field.get in UrlFormDataEncoder

### LINKED ISSUE #1321: Serializing default values causes failure in EC2 DescribeRequest
Using the generated EC2 client's `describeInstance` method, set a list of Instance IDs and attempt to make a request. It will fail because the default `MaxResults` value is set on the request as well, which is [forbidden by the AWS API](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html#API_DescribeInstances_RequestParameters).

Reproduction:
```scala
import cats.*
import cats.effect.*
import cats.syntax.all.*
import com.amazonaws.ec2.*
import org.http4s.*
import org.http4s.client.{Client, middleware}
import smithy4s.aws.*
import smithy4s.aws.kernel.AwsRegion

object TestApp extends ResourceApp.Simple:
  private val id = InstanceId("i-00000000000000000")
  override def run: Resource[IO, Unit] =
    EmberClientBuilder.default[IO].build
      .map(middleware.Logger(logHeaders = true, logBody = true))
      .flatMap(AwsEnvironment.default(_, AwsRegion.US_WEST_2))
      .flatMap(AwsClient(EC2, _))
      .evalMap(_.describeInstances(instanceIds = Option(List(id))))
      .evalMap(IO.println(_))
      .void
```

Running this results in
```
HTTP/1.1 POST https://ec2.us-west-2.amazonaws.com/ Headers(Content-Length: 102, Authorization: <REDACTED>, Content-Type: application/x-www-form-urlencoded, host: ec2.us-west-2.amazonaws.com, X-Amz-Date: 20231207T004525Z, X-Amz-Target: AmazonEC2.DescribeInstances) body="Action=DescribeInstances&Version=2016-11-15&DryRun=false&MaxResults=0&InstanceId.1=i-00000000000000000"
HTTP/1.1 400 Bad Request Headers(x-amzn-RequestId: redacted, Cache-Control: no-cache, no-store, Strict-Transport-Security: max-age=31536000; includeSubDomains, vary: accept-encoding, Content-Type: text/xml;charset=UTF-8, Transfer-Encoding: chunked, Date: Thu, 07 Dec 2023 00:45:26 GMT, Connection: close, Server: AmazonEC2) body="<?xml version="1.0" encoding="UTF-8"?>
<Response><Errors><Error><Code>InvalidParameterCombination</Code><Message>The parameter instancesSet cannot be used with the parameter maxResults</Message></Error></Errors><RequestID>redacted</RequestID></Response>"
smithy4s.http.UnknownErrorResponse: status 400, headers: HashMap(Cache-Control -> List(no-cache, no-store), x-amzn-RequestId -> List(redacted), Connection -> List(close), Strict-Transport-Security -> List(max-age=31536000; includeSubDomains), vary -> List(accept-encoding), Date -> List(Thu, 07 Dec 2023 00:45:26 GMT), Content-Type -> List(text/xml;charset=UTF-8), Server -> List(AmazonEC2), Transfer-Encoding -> List(chunked)), body:
<?xml version="1.0" encoding="UTF-8"?>
<Response>
  <Errors>
    <Error>
      <Code>InvalidParameterCombination</Code>
      <Message>The parameter instancesSet cannot be used with the parameter maxResults</Message>
    </Error>
  </Errors>
  <RequestID>redacted</RequestID>
</Response>
```

I worked around this by removing `MaxResults` from the model using a preprocessor:

```scala
override def transform(ctx: TransformContext): Model =
  ctx.getTransformer.filterShapes(
    ctx.getModel,
    !_.toShapeId.equals(ShapeId.from("com.amazonaws.ec2#DescribeInstancesRequest$MaxResults"))
  )
```

bpholt:
Fixes #1321.

- [ ] Added documentation
- [x] Updated changelog

### DIFF
--- a/modules/core/src/smithy4s/http/internals/UrlFormDataEncoderSchemaVisitor.scala
+++ b/modules/core/src/smithy4s/http/internals/UrlFormDataEncoderSchemaVisitor.scala
@@ -136,8 +136,14 @@ private[http] class UrlFormDataEncoderSchemaVisitor(
       make: IndexedSeq[Any] => S
   ): UrlFormDataEncoder[S] = {
     def fieldEncoder[A](field: Field[S, A]): UrlFormDataEncoder[S] =
-      compile(field.schema)
-        .contramap(field.get)
+      new UrlFormDataEncoder[S] {
+        private val cachedEncoder = compile(field.schema)
+        override def encode(value: S): List[UrlForm.FormData] =
+          field
+            .getUnlessDefault(value)
+            .toList
+            .flatMap(cachedEncoder.encode)
+      }
         .prepend(getKey(field.hints, field.label))
     val encoders = fields.map(fieldEncoder(_))
     struct => encoders.toList.flatMap(_.encode(struct))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.18.5
+
+* When encoding to `application/x-www-form-urlencoded`, omit optional fields set to the field's default value.
+
 # 0.18.4
 
 * Changes the behaviour of `Field#getUnlessDefault` and `Field#foreachUnlessDefault` to always take the value into consideration when the `smithy.api#required` trait
--- a/modules/bootstrapped/test/src/smithy4s/http/internals/UrlFormDataEncoderDecoderSchemaVisitorSpec.scala
+++ b/modules/bootstrapped/test/src/smithy4s/http/internals/UrlFormDataEncoderDecoderSchemaVisitorSpec.scala
@@ -175,6 +175,89 @@ object UrlFormDataEncoderDecoderSchemaVisitorSpec extends SimpleIOSuite {
     )
   }
 
+  pureTest(
+    "struct: default (but not required) values are omitted from encoded output"
+  ) {
+    val defaultValue = "default"
+    case class Foo(x: Int, y: String)
+    object Foo {
+      implicit val schema: Schema[Foo] = {
+        val x = int.required[Foo]("x", _.x)
+        val y = smithy4s.schema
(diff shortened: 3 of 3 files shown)


bpholt:
@Baccata could you help me figure out where to add unit tests for this change? It's unclear to me where they belong—maybe there aren't any unit tests for `UrlFormDataEncoder` yet?

Baccata:
https://github.com/disneystreaming/smithy4s/blob/series/0.18/modules/bootstrapped/test/src/smithy4s/http/internals/UrlFormDataEncoderDecoderSchemaVisitorSpec.scala

The unit tests of the core module live in the bootstrapped module, which makes it easier to write assertions against the generated code

modules/core/src/smithy4s/http/internals/UrlFormDataEncoderSchemaVisitor.scala
```diff
@@ -136,8 +136,14 @@ private[http] class UrlFormDataEncoderSchemaVisitor(
       make: IndexedSeq[Any] => S
   ): UrlFormDataEncoder[S] = {
     def fieldEncoder[A](field: Field[S, A]): UrlFormDataEncoder[S] =
-      compile(field.schema)
-        .contramap(field.get)
+      new UrlFormDataEncoder[S] {
+        private lazy val cachedEncoder = compile(field.schema)
```
Baccata: doesn't have to be lazy

kubukoz (approved):

kubukoz (commented):
Please add a CHANGELOG entry for 0.18.5. Other than that, looks good to me

Baccata (approved):

PR #1326, opened 2023-12-13
renderHintsVal: wrap in `lazily`

kubukoz:
Closes #537, the dreaded classloading initialization deadlocks.

## PR Checklist (not all items are relevant to all PRs)

- [x] Added unit-tests (for runtime code)
- ~~Added bootstrapped code + smoke tests (when the rendering logic is modified)~~
- ~~Added build-plugins integration tests (when reflection loading is required at codegen-time)~~
- ~~Added alloy compliance tests (when simpleRestJson protocol behaviour is expanded/updated)~~
- ~~Updated dynamic module to match generated-code behaviour~~
- ~~Added documentation~~
- [x] Updated changelog

PR #1327, opened 2023-12-13
Lazy hint bindings!

kubukoz:
ANOTHER stab at #537.

This one makes the static bindings lazy by utilizing the fact that implicit conversions can take by-name parameters.

- [ ] Added documentation
- [ ] Updated changelog


modules/core/src/smithy4s/Lazy.scala
```diff
@@ -25,6 +25,14 @@ final class Lazy[A](make: () => A) {
   }
 
   def map[B](f: A => B): Lazy[B] = new Lazy(() => f(make()))
+
+  override def equals(obj: Any): Boolean = obj match {
+    case that: Lazy[_] => this.value == that.value
+    case _             => false
+  }
+
+  override def hashCode(): Int = value.hashCode()
```
kubukoz: **issue:** this is going to be a problem whenever `Lazy` is used with recursive schemas.

just spitballing, maybe we make a new `Lazy` for hints, which _does_ have equality, and keep the old one for recursion cases (schemas), without equality?

modules/core/src/smithy4s/Hints.scala
```diff
@@ -170,7 +170,7 @@ object Hints {
   }
 
   object Binding {
-    final case class StaticBinding[A](key: ShapeTag[A], value: A)
+    final case class StaticBinding[A](key: ShapeTag[A], value: Lazy[A])
```
kubukoz: **issue:** breaks bincompat because these are public types.

Maayyyyyyyyybe we can work around it by providing a legacy-style `unapply`, getter, `copy` and all that.

modules/core/src/smithy4s/Lazy.scala
kubukoz: oh, it just struck me.

What if we simply don't implement these, and instead implement custom equality/hashcode in `StaticBinding` itself?

modules/core/src/smithy4s/Hints.scala
```diff
@@ -170,19 +176,81 @@ object Hints {
...
 
-    implicit def fromValue[A, AA <: A](value: AA)(implicit
+    implicit def fromValueLazy[A, AA <: A](value: => AA)(implicit
+        key: ShapeTag[A]
+    ): Binding = new StaticBinding[A](key, Lazy[A](value))
+
+    private[smithy4s] def fromValue[A, AA <: A](value: AA)(implicit
         key: ShapeTag[A]
-    ): Binding = StaticBinding(key, value)
+    ): Binding = fromValueLazy[A, AA](value)
```
kubukoz: thought: the actual change starts here.

kubukoz:
OOF. 

> Exception in thread "io-compute-3" java.lang.ClassFormatError: Duplicate method name "copy$default$1" with signature "()Lsmithy4s.ShapeTag;" in class file smithy4s/Hints$Binding$StaticBinding

Isn't it a compiler bug if a method like this ends up in the classfile?

I'll see about minimizing (update: done https://github.com/lampepfl/dotty/issues/19274) but I think we can just remove it - I needed it when I made the class a non-case-class.

modules/core/src/smithy4s/Hints.scala
```diff
@@ -170,19 +176,81 @@ object Hints {
   }
 
   object Binding {
-    final case class StaticBinding[A](key: ShapeTag[A], value: A)
-        extends Binding {
+    final case class StaticBinding[A](
```
Baccata: Do you think we could make it a normal class ?

modules/core/src/smithy4s/Hints.scala
```diff
@@ -170,19 +176,81 @@ object Hints {
...
+
     final case class DynamicBinding(keyId: ShapeId, value: Document)
         extends Binding {
       override def toString = Document.obj(keyId.show -> value).toString()
     }
 
-    implicit def fromValue[A, AA <: A](value: AA)(implicit
+    implicit def fromValueLazy[A, AA <: A](value: => AA)(implicit
+        key: ShapeTag[A]
+    ): Binding = new StaticBinding[A](key, Lazy[A](value))
```
Baccata, line 207: this is absolutely genius and I love it.

modules/core/src/smithy4s/Hints.scala
kubukoz: Probably, but do we want to? We'll need to make a Mirror for 3.x.

Baccata: right, let's not then. I do want to get rid of most publicly-facing case classes at some point but we'll keep that for a breaking release, provided we can get some cross-version tooling for macro-based datatypes.

kubukoz: lesson learned: case classes are evil for bincompat ESPECIALLY if you're cross-building for 2.x and 3.x

kubukoz:
This solution has some NPE problems on Scala 3 because it encodes lambdas in constructor calls differently from Scala 2.

At the moment, I don't see a great way to use this idea. Closing in favor of #1326.

kubukoz:
Reopening, looks like #1326 may not be any better and it'd require similar codegen inconveniences.

kubukoz:
amazing, I think I've found a nondeterministic compilation in 2.13... twice in 4 commits!

https://github.com/disneystreaming/smithy4s/actions/runs/7427122765/job/20212279041?pr=1327

https://github.com/disneystreaming/smithy4s/actions/runs/7427658446/job/20213734394?pr=1327

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -1249,12 +1249,17 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
         renderPrismsEnum(name, values, hints, isOpen),
         values.map { case e @ EnumValue(value, intValue, _, hints) =>
           val valueName = NameRef(e.name)
-          val valueHints = line"$Hints_(${memberHints(e.hints)})"
+
+          val baseLine = line"""case object $valueName extends $name("$value", "${e.name}", $intValue, $Hints_())"""
```
Baccata, line 1253: `$Hints_.empty` doesn't work ?

kubukoz: it does, I just wanted to avoid increasing the diff. Currently, this PR is a no-op for members that don't have hints.

We can do this change anytime without breakage, I suggest that we do it separately.

Baccata:
Closing in favour of https://github.com/disneystreaming/smithy4s/pull/1326

Issue #1328, opened 2023-12-16
OOM during execution of app


FloWi:
Hi!

Here's the bugreport as discussed with @Baccata on discord.

I tried to reproduce the error in a small [example project](https://github.com/FloWi/smithy-oom-bugreport), but there the code behaves nicely. I used an example json and tried encoding/decoding as I do in my real app. The json-conversions are [here](https://github.com/FloWi/smithy-oom-bugreport/blob/main/modules/example/src/main/scala/smithy4sbugreport/AStarWithFuelDebug.scala#L18-L26) and [here](https://github.com/FloWi/smithy-oom-bugreport/blob/main/modules/example/src/main/scala/smithy4sbugreport/CodecsHelper.scala#L8C22-L33)   

The real app uses the generated client to talk to the [game-server](https://spacetraders.stoplight.io/docs/spacetraders/8e768e6831f6c-getting-started). I also have a 2nd model that I'm using to generate a server.

Here's the description again for convenience.

In the memory snapshots I've seen smithy4s coming up quite often as the top entries in both instance count and memory, so I'm hoping someone here might be able to help me out.

The smithy4s.schema.CompilationCache grows constantly and takes up most memory (see screenshot 1). Next is smithy4s.schema.Schema$EnumerationSchema (this also has a lot of instances) and then smithy4s.schema.EnumValue
Unfortunately I changed quite a lot in my codebase and also upgraded to smithy4s 0.18.x, so I don't know what might cause this behavior. I'm using the generated types of smithy4s everywhere in my codebase.

At first I thought I might be storing stuff and not cleaning it up (see number of instances of scala.immutable.List in screenshot 1), but when I'm drilling down, the top-entry is smithy4s.schema.Schema$EnumerationSchema (screenshot 2). 

If you have a spontaneous idea, what I might be the issue, please let me know. Otherwise I'll try to reproduce it in a minimal project and maybe open a bugticket if I was successful.
If someone has experience with debugging memory leaks and knows a good article, please let me know. I'm not experienced with that and mostly explored using both VisualVM and IntelliJ.

![image](https://github.com/disneystreaming/smithy4s/assets/39078/df7ed180-ceeb-4d05-aa97-e2f647631ee9)

![image](https://github.com/disneystreaming/smithy4s/assets/39078/5158df9d-6b87-44a9-b7ad-8fe1f899685e)

PR #1329, opened 2023-12-18
[0.19] Address caching issue in schema compiler that can lead to OOMs

Baccata:
Closes https://github.com/disneystreaming/smithy4s/issues/1328

- [x] Updated changelog

### DIFF
--- a/modules/core/src/smithy4s/schema/CachedSchemaCompiler.scala
+++ b/modules/core/src/smithy4s/schema/CachedSchemaCompiler.scala
@@ -25,7 +25,7 @@ trait CachedSchemaCompiler[+F[_]] { self =>
   def fromSchema[A](schema: Schema[A]): F[A]
   def fromSchema[A](schema: Schema[A], cache: Cache): F[A]
 
-  final def mapK[F0[x] >: F[x], G[_]](
+  def mapK[F0[x] >: F[x], G[_]](
       fk: PolyFunction[F0, G]
   ): CachedSchemaCompiler[G] =
     new CachedSchemaCompiler[G] {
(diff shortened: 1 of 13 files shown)


modules/core/src/smithy4s/schema/CachedSchemaCompiler.scala
```diff
@@ -86,14 +85,50 @@ object CachedSchemaCompiler { outer =>
         self.mapK(fk)
     }
 
-  abstract class Impl[F[_]] extends CachedSchemaCompiler[F] {
+  abstract class Impl[F[_]] extends CachedSchemaCompiler[F] { self =>
     protected type Aux[_]
-    type Cache = CompilationCache[Aux]
+    type AuxCache = CompilationCache[Aux]
+    case class Cache(outer: CompilationCache[F], inner: AuxCache)
```
Baccata, line 91: this is the core of the change : we need an inner cache to be passed to some schema visitor, but we also need an outer cache to prevent schema-preprocessing from being re-applied

kubukoz: question: is it only a problem because the preprocessing doesn't produce deterministic (as per hashCode/equals) outputs?

Do we know which transformation was causing the particular issue?

Baccata: There's 2 problems : most pernicious one is what you describe, and it's the total function held by enumeration schemas which is at fault. Hence the changes in my other PR.

Less pernicious one is that the it's the input of the schema visitor call that gets cached instead of the input of the schema compiler.

This means that we're not protecting against re-running the inefficient pre-processing of schemas that may occur (like hint masks), which is really bad performance wise

kubukoz: That's good info, thanks

kubukoz (approved):

PR #1340, opened 2024-01-04
Rework EnumerationSchema

Baccata:
* Changes the EnumTag modelling by having just two cases : IntEnum, StringEnum. The open/close aspect is determined by an optional function. Additionally, EnumTags now carry a `value` function allowing to extract the String/Int value out of the enumeration.
* Rework EnumerationSchema to avoid carrying a `total` function. The `value` function carried by the EnumerationSchema can now be used instead, which gets the logic closer to smithy semantics.
* Changes the dynamic data so that string enumerations are now represented by the string values, instead of integers.
* Rename the `value` method to `stringValue` in the EnumerationValue datatype.

- [x] Updated documentation
- [x] Updated changelog

### DIFF
--- a/modules/bootstrapped/src/generated/smithy4s/example/AudioEnum.scala
+++ b/modules/bootstrapped/src/generated/smithy4s/example/AudioEnum.scala
@@ -5,13 +5,12 @@ import smithy4s.Hints
 import smithy4s.Schema
 import smithy4s.ShapeId
 import smithy4s.ShapeTag
-import smithy4s.schema.EnumTag
-import smithy4s.schema.Schema.enumeration
+import smithy4s.schema.Schema.stringEnumeration
 
-sealed abstract class AudioEnum(_value: String, _name: String, _intValue: Int, _hints: Hints) extends Enumeration.Value {
+sealed abstract class AudioEnum(_name: String, _stringValue: String, _intValue: Int, _hints: Hints) extends Enumeration.Value {
   override type EnumType = AudioEnum
-  override val value: String = _value
   override val name: String = _name
+  override val stringValue: String = _stringValue
   override val intValue: Int = _intValue
   override val hints: Hints = _hints
   override def enumeration: Enumeration[EnumType] = AudioEnum
@@ -24,13 +23,12 @@ object AudioEnum extends Enumeration[AudioEnum] with ShapeTag.Companion[AudioEnu
     smithy.api.MediaType("audio/mpeg3"),
   )
 
-  case object GUITAR extends AudioEnum("guitar", "GUITAR", 0, Hints())
-  case object BASS extends AudioEnum("bass", "BASS", 1, Hints())
+  case object GUITAR extends AudioEnum("GUITAR", "guitar", 0, Hints())
+  case object BASS extends AudioEnum("BASS", "bass", 1, Hints())
 
   val values: List[AudioEnum] = List(
     GUITAR,
     BASS,
   )
-  val tag: EnumTag[AudioEnum] = EnumTag.ClosedStringEnum
-  implicit val schema: Schema[AudioEnum] = enumeration(tag, values).withId(id).addHints(hints)
+  implicit val schema: Schema[AudioEnum] = stringEnumeration(values).withId(id).addHints(hints)
 }
--- a/modules/core/src/smithy4s/schema/EnumTag.scala
+++ b/modules/core/src/smithy4s/schema/EnumTag.scala
@@ -22,20 +22,4 @@ sealed trait EnumTag[E]
 object EnumTag {
   case class StringEnum[E](value: E => String, unknown: Option[String => E]) extends EnumTag[E]
   case class IntEnum[E](value: E => Int, unknown: Option[Int => E])          extends EnumTag[E]
-
-  object OpenStringEnum {
-    def unapply[E](enumTag: EnumTag[E]): Option[String => E] = enumTag match {
-      case t: StringEnum[E] => t.unknown
-      case _                => None
-    }
-
-  }
-
-  object OpenIntEnum {
-    def unapply[E](enumTag: EnumTag[E]): Option[Int => E] = enumTag match {
-      case t: IntEnum[E] => t.unknown
-      case _             => None
-    }
-  }
-
 }
--- a/modules/core/src/smithy4s/schema/EnumTag.scala
+++ b/modules/core/src/smithy4s/schema/EnumTag.scala
@@ -16,28 +16,26 @@
 
 package smithy4s.schema
 
-sealed trait EnumTag[+E]
+// scalafmt: { maxColumn = 120, align.preset = most}
+sealed trait EnumTag[E]
 
 object EnumTag {
-  case object ClosedStringEnum extends EnumTag[Nothing]
-  case object ClosedIntEnum extends EnumTag[Nothing]
-  case class OpenStringEnum[E](unknown: String => E) extends EnumTag[E]
-  case class OpenIntEnum[E](unknown: Int => E) extends EnumTag[E]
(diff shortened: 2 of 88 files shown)


modules/core/src/smithy4s/schema/EnumTag.scala
```diff
@@ -16,28 +16,26 @@
 
 package smithy4s.schema
 
-sealed trait EnumTag[+E]
```
Baccata, line 19: This is the most meaningful change of this PR : the changes in modelling of the tag allows for the removal of the `total` function in EnumerationSchema. Additionally, the `Closed` vs `Open` distinction is now carried by an optional `unknown` value in the 2 case classes that now compose the ADT. It simplifies some pattern match.

modules/core/src/smithy4s/schema/EnumTag.scala
```diff
@@ -16,28 +16,26 @@
...
-  case class OpenIntEnum[E](unknown: Int => E) extends EnumTag[E]
+  case class StringEnum[E](value: E => String, unknown: Option[String => E]) extends EnumTag[E]
+  case class IntEnum[E](value: E => Int, unknown: Option[Int => E])          extends EnumTag[E]
 
-  object StringEnum {
-    def unapply[E](enumTag: EnumTag[E]): Boolean = enumTag match {
-      case ClosedStringEnum  => true
-      case OpenStringEnum(_) => true
-      case _                 => false
+  object OpenStringEnum {
```
Baccata: I'm probably gonna remove these : these custom extractors are getting in the way of exhaustive pat-mats

modules/core/src/smithy4s/schema/Schema.scala
```diff
@@ -158,7 +158,7 @@ object Schema {
   final case class PrimitiveSchema[P](shapeId: ShapeId, hints: Hints, tag: Primitive[P]) extends Schema[P]
   final case class CollectionSchema[C[_], A](shapeId: ShapeId, hints: Hints, tag: CollectionTag[C], member: Schema[A]) extends Schema[C[A]]
   final case class MapSchema[K, V](shapeId: ShapeId, hints: Hints, key: Schema[K], value: Schema[V]) extends Schema[Map[K, V]]
-  final case class EnumerationSchema[E](shapeId: ShapeId, hints: Hints, tag: EnumTag[E], values: List[EnumValue[E]], total: E => EnumValue[E]) extends Schema[E]
```
Baccata, line 162: This is the second meaningful change in this PR : all the encoding logic that was counting on this function can now use the `value` functions held by the members of the `EnumTag` ADT.

modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
```diff
@@ -216,7 +216,7 @@ private[dynamic] object Compiler {
       update(id, shape.traits, bytes)
 
     override def enumShape(id: ShapeId, shape: EnumShape): Unit = {
-      val values: List[EnumValue[Int]] =
+      val values: List[EnumValue[String]] =
```
Baccata, line 219: Another meaningful change : now the string enumerations are represented as a raw string in the dynamic data. This allows for fulfilling the new `EnumTag` contract

modules/bootstrapped/src/generated/smithy4s/example/AudioEnum.scala
```diff
@@ -5,13 +5,12 @@ import smithy4s.Hints
 import smithy4s.Schema
 import smithy4s.ShapeId
 import smithy4s.ShapeTag
-import smithy4s.schema.EnumTag
-import smithy4s.schema.Schema.enumeration
+import smithy4s.schema.Schema.stringEnumeration
 
-sealed abstract class AudioEnum(_value: String, _name: String, _intValue: Int, _hints: Hints) extends Enumeration.Value {
```
daddykotex, line 11: I'm fine with the rename, but the reordering seems a little tricky, especially because both values have type `String` and compilation would not fail

modules/core/src/smithy4s/schema/EnumTag.scala
daddykotex: using a function `E => String` and `E => Int` just clicks in so nicely with all the use cases in the various visitor, wow

modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
```diff
@@ -349,7 +316,7 @@ private[dynamic] object Compiler {
                 stringValue = value,
                 intValue = intValue,
                 // Using the intValue as a runtime value
```
daddykotex: the comment is irrelevant now

modules/dynamic/test/src-jvm/smithy4s/dynamic/EnumSpec.scala
```diff
@@ -112,14 +112,14 @@ class EnumSpec extends DummyIO.Suite {
         EnumValue(
           stringValue = "Ice",
           intValue = 0,
-          value = 0,
+          value = "Ice",
           name = "ICE",
           hints = Hints.empty
         ),
         EnumValue(
```
daddykotex, line 119: related to comment on reordering, here, we're _lucky_ that we used name params, otherwise this would have gone under the rader

daddykotex (commented):
good job, I also am in favour of removing the custom unapply

modules/bootstrapped/src/generated/smithy4s/example/AudioEnum.scala
Baccata: for the most part, this stuff is meant to be code-generated, we only handcraft some of them for unit tests, so I think it's fine. 

Good point though.

modules/core/src/smithy4s/schema/EnumTag.scala
```diff
@@ -16,28 +16,26 @@
...
+// scalafmt: { maxColumn = 120, align.preset = most}
+sealed trait EnumTag[E]
 
 object EnumTag {
-  case object ClosedStringEnum extends EnumTag[Nothing]
-  case object ClosedIntEnum extends EnumTag[Nothing]
-  case class OpenStringEnum[E](unknown: String => E) extends EnumTag[E]
-  case class OpenIntEnum[E](unknown: Int => E) extends EnumTag[E]
+  case class StringEnum[E](value: E => String, unknown: Option[String => E]) extends EnumTag[E]
+  case class IntEnum[E](value: E => Int, unknown: Option[Int => E])          extends EnumTag[E]
```
lewisjkl, line 24: Would there be value now in also having `String => Option[E]` and `Int => Option[E]` at this level? Mostly just thinking of how we build the same-ish `Map[String, E]` and `Map[Int, E]` in the different interpreters

Baccata: Don't think so. The functions I added are not there for convenience, but rather because the removal of "total" makes it necessary for these to be present. 

It's easy enough to reconstruct maps on demand from the list of enum values.

modules/dynamic/test/src-jvm/smithy4s/dynamic/EnumSpec.scala
kubukoz: Named params are a life-saver.

modules/core/src/smithy4s/schema/EnumTag.scala
kubukoz: it kinda forces you to think about open enums explicitly because, when you do `values.groupBy(_.value).get(value)`, it's optional - meaning you have to consider kind of explicitly what happens if the value isn't there (which would be the case in open enums, as I understand)

kubukoz (approved):

kubukoz:
shall we go ahead and merge?

Baccata:
Yeah let's

PR #1341, opened 2024-01-04
Remove default localhost on our uri

### LINKED ISSUE #1245: 0.18.0 SimpleRestJson for http4s Fetch client issue
I'm having a headscratcher, which I'm beginning to think, might be upgrade to smithy 0.18 related. I have this service - simply defined.
```
@readonly
@http(method: "GET", uri: "/api/events")
operation GetAllIlsEvents {  
  output: AllIlsEvents
}
```
I want to call it in browser... through vite's proxy 

test1 - http://localhost:8080/api/events - http4s server directly, success

test2 - browser console - 
```
 let res = await fetch('http://localhost:8080/api/events', {method: 'GET'}).then(function (response) {
    console.log(response.text());
  }) ;
```
 suceeds. 

test3 - browser console ... but going through vite proxy
```
let res = await fetch('/api/events', {method: 'GET'}).then(function (response) {
    console.log(response.text());
  }) ;
```
succeeds.

test4 - http4s fetch client, direct use, compiled to scalaJS in the frontend.
```
    val vanillaClient = FetchClientBuilder[IO].create
    vanillaClient.get("/api/events")(r => IO.println(r)).unsafeToFuture()
```

Succeeds.

test5 - 
```
SimpleRestJson
  def eventClient(
      http4sClient: Client[IO]
  ): IlsEventService[IO] =
    SimpleRestJsonBuilder
      .withMaxArity(arity)
      .apply(IlsEventService)
      .client(http4sClient)
      .uri(Uri.unsafeFromString("/"))
      .make
      .fold(throw _, identity)


    Clients.eventClient(vanillaClient).getAllIlsEvents().unsafeToFuture().map { r =>
      println(r)
    }  
```

Fails (!). I believe there is a difference in browser between the requests.
In the browser Network tab
The successful "raw" request from the http4s vanilla fetch client looks like;
` http://localhost:5173/api/events `
 

The request coming out of the SimpleRestJsonBuilder, appears to look like. 
` http://localhost/api/events `

I think I claim, that the SimpleRestJson is making the Fetch client "forget" about some part of it's environment... the port perhaps? Is this possible?

daddykotex:
Fixes #1245 

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/modules/core/src/smithy4s/http/HttpUri.scala
+++ b/modules/core/src/smithy4s/http/HttpUri.scala
@@ -18,7 +18,7 @@ package smithy4s.http
 
 final case class HttpUri(
     scheme: HttpUriScheme,
-    host: String,
+    host: Option[String],
     port: Option[Int],
     /**
       * A sequence of URL-decoded URI segment.
--- a/modules/aws-http4s/src/smithy4s/aws/AwsClient.scala
+++ b/modules/aws-http4s/src/smithy4s/aws/AwsClient.scala
@@ -74,7 +74,7 @@ object AwsClient {
           val endpointPrefix = awsService.endpointPrefix.getOrElse(endpoint.id.name)
           val baseUri = HttpUri(
             scheme = HttpUriScheme.Https,
-            host = s"$endpointPrefix.$region.amazonaws.com",
+            host = Some(s"$endpointPrefix.$region.amazonaws.com"),
             port = None,
             path = IndexedSeq.empty,
             queryParams = Map.empty,
--- a/modules/bootstrapped/test/src/smithy4s/http/HttpRequestSpec.scala
+++ b/modules/bootstrapped/test/src/smithy4s/http/HttpRequestSpec.scala
@@ -28,7 +28,7 @@ final class HttpRequestSpec extends FunSuite {
 
     val uri = HttpUri(
       HttpUriScheme.Https,
-      "example.com",
+      Some("example.com"),
       None,
       IndexedSeq.empty,
       Map.empty,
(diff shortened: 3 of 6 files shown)


daddykotex:
tried it and it allows me to revert https://github.com/daddykotex/smithy4s-code-generation/pull/13 successfully
I'll do further testing with regards to the different internal implementation we have to make sure everything still works there as well, then I'll mark as ready to be merged

kubukoz:
> tried it and it allows me to revert [daddykotex/smithy4s-code-generation#13](https://github.com/daddykotex/smithy4s-code-generation/pull/13) successfully

tbh I wouldn't revert it _fully_, removing CORS is a security improvement :) but if you mean the `resetBaseUri` part, absolutely let's kill it.

modules/core/src/smithy4s/http/HttpUri.scala
```diff
@@ -18,7 +18,7 @@ package smithy4s.http
 
 final case class HttpUri(
```
kubukoz, line 19: **suggestion:** we could safeguard against bincompat breakage in the future, now that we're at it:

https://docs.scala-lang.org/overviews/core/binary-compatibility-for-library-authors.html

this would mean e.g. hiding the `copy` method, and opting for `.withXXX(newXXX)` methods instead, and would need to touch more than just this type though.

cc @Baccata

daddykotex:
yeah, no need to reintroduce cors in that case so I just want to remove the resetBaseUri middleware

modules/core/src/smithy4s/http/HttpUri.scala
Baccata: Good point :+1:

daddykotex: Do you mind if I make a dedicated issue for that so we can track it for the whole of smithy4s?
It could be an important part of a 0.19 update

Baccata: Sure. In that issue, you should add the presence of a dedicated scalafix rule aimed at finding such cases.

daddykotex: Ok, I'll create the issue and look into it

daddykotex: https://github.com/disneystreaming/smithy4s/issues/1349

kubukoz (approved):

Baccata (approved):

PR #1377, opened 2024-01-25
Support `@httpResponseCode` over newtypes

miguel-vila:
not sure if there was a reason to miss this case, let me know what u think

### DIFF
--- a/modules/core/src/smithy4s/http/internals/HttpResponseCodeSchemaVisitor.scala
+++ b/modules/core/src/smithy4s/http/internals/HttpResponseCodeSchemaVisitor.scala
@@ -101,8 +101,18 @@ class HttpResponseCodeSchemaVisitor()
           case OptionalResponseCode(f) => f(value)
         }
     }
-
   }
+
+  override def biject[A, B](
+      schema: smithy4s.schema.Schema[A],
+      bijection: smithy4s.Bijection[A, B]
+  ): ResponseCodeExtractor[B] = {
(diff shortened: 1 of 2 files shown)


modules/core/src/smithy4s/http/internals/HttpResponseCodeSchemaVisitor.scala
```diff
@@ -101,8 +101,18 @@ class HttpResponseCodeSchemaVisitor()
           case OptionalResponseCode(f) => f(value)
         }
     }
+  }
 
+  override def biject[A, B](
```
kubukoz, line 107: **suggestion:** should we handle `def refine` as well?

Baccata: Might as well ^^

Baccata:
> not sure if there was a reason to miss this case, let me know what u think

No particular reason, sometimes we just miss things

modules/core/src/smithy4s/http/internals/HttpResponseCodeSchemaVisitor.scala
miguel-vila: done at [2dfd250](https://github.com/disneystreaming/smithy4s/pull/1377/commits/2dfd25048c65b2b4a46a90e7b692a55b69c44f4f)

kubukoz (approved):

kubukoz:
gah, I forgot the CHANGELOG. @miguel-vila please check the PR template next time 😅

PR #1382, opened 2024-01-26
Add 0.18 forward-porting into 0.19

kubukoz:

- [ ] Added documentation
- [ ] Updated changelog


Baccata:
besides the fact that this actually sets up fore-port (not back-port), do you want to revisit this in order to automate the process a little more ?

kubukoz:
yes. PR is called "backport action" because that's how Mergify calls it in the config ;)

I may try revisiting it now that you've sniped me

.mergify.yml
```diff
@@ -2,11 +2,29 @@ pull_request_rules:
   - name: Automatically approve/merge Scala Steward PRs on CI success
     conditions:
       - author=scala-steward
-      - status-success="build-success-checkpoint"
+      - check-success="build-success-checkpoint"
```
kubukoz, line 5: status-success apparently isn't a thing - https://docs.mergify.com/configuration/conditions/

kubukoz:
@Baccata that should do it. Tested in a repo of mine (without the required status check, but I trust that'll work properly) - https://github.com/kubukoz/demos/pull/8

.mergify.yml
```diff
@@ -2,11 +2,29 @@ pull_request_rules:
...
     actions:
       review:
         type: APPROVE
         message: Automatically approving scala-steward
       merge:
         method: merge
+
+  - name: forward-port labeled PRs to the 0.19 branch
+    conditions:
+      - label="forwardport-0.19"
```
kubukoz, line 16: **question:** do we need a label? I'd say we should do it automatically (that way it'll be much less trouble with things like Steward), or with opt-out via a "do not port" label.

kubukoz:
this is outdated and not necessary anymore. Let us please never have parallel series ever again. Or we setup backporting on day 1.

PR #1391, opened 2024-02-05
case class rewrite for core

daddykotex:

- [ ] Updated changelog

I'll reach out to scalafix to try and upstream the linting rule. If we can't get it in there in, I'll open another PR to get it in here.

### DIFF
--- a/modules/core/src/smithy4s/http/HttpRestSchema.scala
+++ b/modules/core/src/smithy4s/http/HttpRestSchema.scala
@@ -45,14 +45,16 @@ sealed trait HttpRestSchema[A]
 
 object HttpRestSchema {
 
-  final case class OnlyMetadata[A](schema: Schema[A]) extends HttpRestSchema[A]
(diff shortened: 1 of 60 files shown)


modules/core/src/smithy4s/http/HttpRestSchema.scala
```diff
@@ -64,22 +126,22 @@ object HttpRestSchema {
       field.memberHints.has[HttpPayload]
 
     fullSchema.findPayload(isPayloadField) match {
-      case TotalMatch(schema) => OnlyBody(schema)
-      case NoMatch() =>
+      case tm: TotalMatch[_] => OnlyBody(tm.schema)
```
Baccata, line 129: this is typically the kind of thing that I think warrants a visitor

kubukoz:
I think we're not going with this idea - #1584 got us halfway there, and this can't happen until 0.20 anyway.

PR #1392, opened 2024-02-05
[codegen] add scala 3 keywords to collision avoidance

jpassaro:

- [ ] (_N/A_) Added build-plugins integration tests (when reflection loading is required at codegen-time)
- [ ] (_N/A_) Added alloy compliance tests (when simpleRestJson protocol behaviour is expanded/updated)
- [ ] Added documentation
- [x] Updated changelog


modules/codegen/src/smithy4s/codegen/internals/CollisionAvoidance.scala
```diff
@@ -279,7 +279,30 @@ private[internals] object CollisionAvoidance {
...
+    "end",
+    "erased",
+    "extension",
+    "infix",
+    "inline",
+    "opaque",
+    "open",
+    "throws",
+    "transparent",
+    "using",
```
jpassaro: the soft keywords are *probably* not needed in this list, and I'm guessing they furthermore don't belong. still, best that y'all are aware of the distinct keyword types so you can make the correct decision.

modules/codegen/test/src/smithy4s/codegen/internals/RendererSpec.scala
```diff
@@ -484,4 +484,61 @@ final class RendererSpec extends munit.FunSuite {
...
+        |""".stripMargin
+    val contents = generateScalaCode(smithy).values
+    assert(
+      contents.exists(
+        _.contains(
+          "case class Given(_given: Option[String] = None)"
+        )
+      )
+    )
+  }
```
jpassaro, line 543: I was unable to get a working dev environment to run these tests, and took the gamble that they would compile and pass. If not or if changes are requested, I'll do the legwork to get tests running.

modules/codegen/src/smithy4s/codegen/internals/CollisionAvoidance.scala
kubukoz: I'm thinking if they don't produce a listing issue, they can stay ignored for now. Let's see what other maintainers think.

modules/codegen/test/src/smithy4s/codegen/internals/RendererSpec.scala
```diff
@@ -484,4 +484,65 @@ final class RendererSpec extends munit.FunSuite {
       )
     )
   }
+
+  test("collision avoidance sanitizes `import` as a structure member") {
```
kubukoz: suggestion: IMO these tests don't add much value to the existing setup, we cover collision avoidance in a number of ways already. Additionally, we don't compile these generated strings so these tests don't tell us _that_ much.

I'm happy with a PR that just changes the list in the other file.

jpassaro: the fact that it doesn't try to compile this generated code is a reasonable point. if there's strong preference to avoid adding test bulk, I'm happy to roll with it. that said... is there some spec someplace where i can add   `structure Scala3ReservedWords { export: String }` and watch CI fail without the fix?

modules/codegen/src/smithy4s/codegen/internals/CollisionAvoidance.scala
```diff
@@ -279,7 +279,30 @@ private[internals] object CollisionAvoidance {
...
+    "end",
+    "erased",
+    "extension",
+    "infix",
+    "inline",
+    "opaque",
+    "open",
+    "throws",
+    "transparent",
+    "using"
```
jpassaro: this got clobbered by a scalafmt commit so just linking to it again:

https://github.com/disneystreaming/smithy4s/pull/1392#pullrequestreview-1863943494

modules/codegen/test/src/smithy4s/codegen/internals/RendererSpec.scala
Baccata: Yup, you can put them here : https://github.com/disneystreaming/smithy4s/blob/series/0.18/sampleSpecs/reservednames.smithy

modules/codegen/src/smithy4s/codegen/internals/CollisionAvoidance.scala
Baccata: I agree with Jakub

kubukoz: > I'm thinking if they don't produce a listing issue, 

Linting* 😅

jpassaro: removed them

modules/codegen/test/src/smithy4s/codegen/internals/RendererSpec.scala
jpassaro: replaced with an update to reservednames.smithy

jpassaro: I verified the fix locally by adding to `reservednames.smithy`, watching `sbt bootstrapped3/compile` fail without the fix, then watching it succeed with the fix. didn't technically verify via the CI build matrix but I think the local verification should be enough.

jpassaro:
thanks kindly to @kubukoz and @Baccata for thoughtful feedback and @daddykotex for the changelog. I think I've got it all covered now.

daddykotex (approved):
Thanks for the PR

kubukoz (approved):

kubukoz:
Thanks @jpassaro :)

jpassaro:
thanks for the merge and swift release!

PR #1397, opened 2024-02-06
Support unions with status codes

miguel-vila:
Added support for unions that have status codes.
Also, should I also add a case for when the union itself has the trait? In this case, would it override any of the variants in case they would also have it?

### DIFF
--- a/modules/core/src/smithy4s/http/internals/HttpResponseCodeSchemaVisitor.scala
+++ b/modules/core/src/smithy4s/http/internals/HttpResponseCodeSchemaVisitor.scala
@@ -96,7 +96,7 @@ class HttpResponseCodeSchemaVisitor()
       dispatch: Alt.Dispatcher[U]
   ): ResponseCodeExtractor[U] = {
     import HttpResponseCodeSchemaVisitor.VariantResponseCodeExtractor
-    val variants = alternatives.map { alt: Alt[U, _] =>
+    val variants = alternatives.map { (alt: Alt[U, _]) =>
       VariantResponseCodeExtractor.fromAlt(this, alt)
     }
     val allAreNoResponseCode = variants.forall { _.extractor == NoResponseCode }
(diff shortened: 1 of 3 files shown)


modules/core/src/smithy4s/http/internals/HttpResponseCodeSchemaVisitor.scala
```diff
@@ -89,18 +89,36 @@ class HttpResponseCodeSchemaVisitor()
...
+    }
+    val allAreNoResponseCode = variants.forall { _.extractor == NoResponseCode }
+    if (allAreNoResponseCode) {
+      NoResponseCode
+    } else {
+      OptionalResponseCode { (u: U) =>
+        variants.flatMap { case variantExtractor =>
+          variantExtractor.project(u).flatMap(variantExtractor.extractor.apply)
+        }.headOption
+      }
```
miguel-vila, line 110: **thought:** I went for `OptionalResponseCode` althought there are other things we could like return `RequiredResponseCode` if _all_ variants are also required. Not sure what would be the difference.  Also could have implemented the `NoResponseCode` by returning an `OptionalResponseCode` that returns `None`

Baccata:
@miguel-vila I'm afraid that I'm gonna veto this change. 

For starter, the `@httpStatusCode` trait cannot currently be applied on union members at the smithy level, only on structure members. **EDIT** : nor can they be applied on structure members that target union shapes. 

Second : even if we did want to support more flexible semantics what the standard http traits propose in the smithy prelude (which we don't for maintenance-burden reasons), the approach you take in this PR would still not accommodate for the kind of flexibility that a lot of people would want. Namely, a better approach would to tie the responses to union members, and allow for the target shapes of the members in question to carry the usual http-bindings. 

```smithy 
union Alternatives {
   @httpAltCode(200) 
   ok: OkResponse
   @httpAltCode(201)
   created: CreatedResponse
}

operation foo {
   input := {
      @httpAlternatives 
      alternatives : Alternatives
   } 
} 
```

One of the reasons we don't want to diverge from the http-bindings is that for better or worse, keeping close to them allows for our protocols to retain a large overlap with the aws-issued REST protocols, which is beneficial in many ways (in particular because of the fact it makes it simpler to adapt the existing smithy => openapi functionality from the smithy library).

PR #1401, opened 2024-02-12
fix infinite recursion when calling Bijection.identity

benoitlouy:


daddykotex (approved):
thanks, I'm surprised nobody hit that before

Baccata:
An entry in the change log and a unit test (for regression) would have been appreciated, but thank you very much for the fix 👍 

edit : @daddykotex, mind taking care of that when you have some time to spare ?

daddykotex:
I added it to my todos, absolutely

PR #1408, opened 2024-02-20
Add support for @nullable trait in order to support MergePatch usecases

astridej:
This PR is an update of https://github.com/disneystreaming/smithy4s/pull/1151 to take into account the changes to default and required handling that have happened since and to also change the modelling a little in line with what was discussed in the Discord (in particular, modelling required vs nullable as independent of each other). The change was big enough that I ended up writing the code from scratch, but I definitely based it on the linked PR.

With the change, it is possible to mark a field as nullable in order to distinguish between absence and removal of a field. This is modelled in Scala via `Option[Nullable[A]]` - `Nullable` being isomorphic to `Option` and left as a separate class for clarity. Unfortunately, this opens up a can of worms regarding possible combinations with @required and @default, the vast majority of which make no sense but are still theoretically possible. I had to rework `Field.Modifier` to make sure all possible combinations were reflected and handled properly, and although I did my best to keep the resulting complexity from leaking out it does show up in some places 😬

I think I've now got this in roughly the state the original PR was in, so opening a draft PR in order to get feedback while I work on the missing pieces.

- [✅] Added unit-tests (for runtime code)
- [✅] Added bootstrapped code + smoke tests (when the rendering logic is modified)
- [👎] Added build-plugins integration tests (when reflection loading is required at codegen-time)
- [👎] Added alloy compliance tests (when simpleRestJson protocol behaviour is expanded/updated)
- [✅ ] Updated dynamic module to match generated-code behaviour
- [✅ ] Added documentation
- [✅ ] Updated changelog


modules/bootstrapped/src/generated/com/amazonaws/dynamodb/Endpoint.scala
```diff
@@ -25,7 +25,7 @@ object Endpoint extends ShapeTag.Companion[Endpoint] {
 
   implicit val schema: Schema[Endpoint] = struct(
     string.required[Endpoint]("Address", _.address).addHints(smithy.api.Documentation("<p>IP address of the endpoint.</p>")),
-    long.required[Endpoint]("CachePeriodInMinutes", _.cachePeriodInMinutes).addHints(smithy.api.Default(smithy4s.Document.fromDouble(0.0d)), smithy.api.Documentation("<p>Endpoint cache time to live (TTL) value.</p>")),
+    long.required[Endpoint]("CachePeriodInMinutes", _.cachePeriodInMinutes).addHints(smithy.api.Documentation("<p>Endpoint cache time to live (TTL) value.</p>"), smithy.api.Default(smithy4s.Document.fromDouble(0.0d))),
```
astridej: I'm not exactly sure why the hint ordering changed on some of the generated code or whether that's a problem? If it is, I can dig into it.

modules/codegen/src/smithy4s/codegen/internals/IR.scala
```diff
@@ -141,29 +142,51 @@ private[internals] case class StreamingField(
...
-      case Modifier.RequiredMod              => false
-      case Modifier.NoModifier               => true
-    }
+  case class Default(node: Node, typedNode: Option[Fix[TypedNode]])
+  case class Modifier(
+      required: Boolean,
+      nullable: Boolean,
+      default: Option[Default]
+  ) {
+    def typeMod: TypeModification =
```
astridej, line 167: This case class is unfortunately complex, so I tried to make sure most places using it didn't actually need to destructure it. The `typeMod` field is part of that, since a bunch of consumers of this class only care about whether the resulting type is wrapped in an `Option` (or `Nullable`, or `Option[Nullable[...]]`).

I did look into converting the underlying type representation to a `Nullable` before reaching the point where the modifier was generated - in which case in theory nothing here should need to change - but I was scared off by the fact that nothing in that part of the code seemed to be doing modifications of the underlying shape's type. I can have a second look if people think it's better there, though - I'm still not happy with how complex the modifier logic is now. 🤷‍♀️

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -811,17 +808,18 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
-    field match {
-      case field if field.tpe.isResolved && hasModifier =>
-        line"override def getMessage(): $string_ = ${field.name}"
-      case field if field.tpe.isResolved =>
-        line"override def getMessage(): $string_ = ${field.name}.orNull"
-      case field if hasModifier =>
-        line"override def getMessage(): $string_ = ${field.name}.value"
-      case field =>
-        line"override def getMessage(): $string_ = ${field.name}.map(_.value).orNull"
+    val fetchLogic = field.modifier.typeMod match {
```
astridej, line 811: I still need to test that this works the way it should. (Even if part of me wants to say that anyone who annotates a field with `@errorMessage @nullable @required @default` or the like is getting what they deserve 🙈 )

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -1239,17 +1230,19 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
         val fields: List[TypedNode.FieldTN[NodeAndType]] = structFields.map {
-          case Field(_, realName, tpe, Field.Modifier.NoModifier, _) =>
+          case Field(_, realName, tpe, mod, _)
+              if mod.typeMod == Field.TypeModification.None =>
+            val node = map.get(realName).getOrElse {
+              mod.default.get.node
+            } // value or default must be present if type is not wrapped
+            TypedNode.FieldTN.RequiredTN(NodeAndType(node, tpe))
+          case Field(_, realName, tpe, _, _) =>
+            // TODO AJ: anything else need to happen here?
```
astridej: See todo comment - given the extra complexity involving `Nullable`, the current set of `FieldTN` values doesn't seem to accurately reflect the state of the field. That said, everything _seems_ to be working fine and this part of the code wasn't touched further in the original PR, so I'm not sure if I need to update them.

modules/core/src/smithy4s/schema/Schema.scala
```diff
@@ -90,8 +90,11 @@ sealed trait Schema[A]{
   final def biject[B](to: A => B)(from: B => A) : Schema[B] = Schema.bijection(this, to, from)
   final def option: Schema[Option[A]] = Schema.option(this)
 
+  final def nullable: Schema[Nullable[A]] = Nullable.schema(this)
+
   final def isOption: Boolean = this match {
     case _: OptionSchema[_] => true
+    case BijectionSchema(underlying, _) => underlying.isOption // TODO AJ: does this have any side-effects in the other places .isOption is used? is it a breaking change?
```
astridej: See TODO comment...

This is necessary in order to get the new code in `DocumentDecoderSchemaVisitor` and `SchemaVisitorJCodec` (which I copied from the original PR) to work for `Nullable.schema`, but I'm not sure if it breaks anything else. The main alternatives I can see are to drop that check and just always delegate null values to the underlying decoder if a `nullable` hint is present (is that safe?), or define the new schema here instead of as a bijection from option.

modules/bootstrapped/src/generated/com/amazonaws/dynamodb/Endpoint.scala
Baccata: mmm it may depend on non-deterministic stuff (ordering of map and sets). Not a problem unless the CI starts shouting (which it might), in which case we'll force a ShapeId-based ordering or something.

modules/bootstrapped/src/generated/smithy4s/example/PatchableEdgeCases.scala
```diff
@@ -0,0 +1,28 @@
...
+import smithy4s.Hints
+import smithy4s.Nullable
+import smithy4s.Nullable.Null
+import smithy4s.Schema
+import smithy4s.ShapeId
+import smithy4s.ShapeTag
+import smithy4s.schema.Schema.int
+import smithy4s.schema.Schema.struct
+
+final case class PatchableEdgeCases(required: Nullable[Int], requiredDefaultValue: Nullable[Int] = smithy4s.Nullable.Value(3), requiredDefaultNull: Nullable[Int] = Null, defaultValue: Nullable[Int] = smithy4s.Nullable.Value(5), defaultNull: Nullable[Int] = Null)
```
Baccata, line 12: this is great, thank you !

modules/core/src/smithy4s/schema/Schema.scala
Baccata: I don't think `.isOption` is used anymore. It was probably used transiently, but I think the change is fine. 

Probably worth adding a `case RefinementSchema` too, just in case.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
Baccata: 😆 yeah unfortunately this project is very much subject to Murphy's law, so I'm thankful with the additional patterns

modules/core/src/smithy4s/schema/Schema.scala
astridej: It shows up once in a place involving `flexibleCollectionSupport` in the JSON codec ( https://github.com/disneystreaming/smithy4s/blob/aaf8b1f034b7d3f97e6f250c97fa08c517686d28/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala#L895 ) but I'm not sure what the logic there is doing.

And good point on the RefinementSchema - will add that!

modules/codegen/src/smithy4s/codegen/internals/IR.scala
Baccata: I'll give it a think, but worst case scenario I'm not too bothered : the internal representation is implementation detail, and even though the complexity does increase with the "nullable" requirement, it's still acceptable imho

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
Baccata: I think you're not gonna need to do anything else : this code is aimed at rendering smithy trait as smithy4s hints. Smithy traits are expressed using a "canonical" node representation, which doesn't take into consideration any other trait but the `@required` trait. 

This kinda implies that we should probably protect against the application of the `@nullable` trait on shapes that are traits (or used in traits). 

You can avoid worrying about it.

modules/core/src/smithy4s/Nullable.scala
```diff
@@ -0,0 +1,62 @@
...
+ * The goal of this datatype is to offer the ability to distinguish, during
+ * serialisation, between the absence of a field and the nullity of a field.
+ */
+sealed trait Nullable[+A] {
+  def map[B](f: A => B): Nullable[B] = this match {
+    case Value(a) => Value(f(a))
+    case Null     => Null
+  }
+
+  def fold[B](whenValue: A => B, whenNull: => B): B =
```
Baccata: I think it'd be good to match the signature of Option's fold : 

```
def fold[B](whenNull : => B)(whenValue: A => B) : B
```

modules/core/src/smithy4s/Nullable.scala
```diff
@@ -0,0 +1,62 @@
...
+object Nullable {
+
+  def value[A](a: A): Nullable[A] = Value(a)
+
+  def fromOption[A](option: Option[A]) = option match {
+    case Some(a) => Value(a)
+    case None    => Null
+  }
+
+  case class Value[A](a: A) extends Nullable[A]
```
Baccata:
```suggestion
  final case class Value[A](a: A) extends Nullable[A]
```

modules/codegen/src/smithy4s/codegen/internals/IR.scala
Baccata: I tried to think of alternative ways of modelling this but what potentially made the code look nicer (using composable binary flags, for instance) or custom extractors also lose us exhaustive pat-mat, so ... I'm fine with the seemingly ugliness if you can't find a better way of modelling it.

Baccata:
@astridej, one important aspect of this work is gonna be the addition of tests to [this specification](https://github.com/disneystreaming/smithy4s/blob/aaf8b1f034b7d3f97e6f250c97fa08c517686d28/modules/bootstrapped/test/src/smithy4s/schema/FieldSpec.scala#L9). 

Most encoders now use the `foreachUnlessDefault` or `getUnlessDefault`, and we need to ensure that those behave in a somewhat coherent manner with respect to the `nullable` trait. 

I presume that this line may have to change in order to account for the `nullable` trait  

https://github.com/disneystreaming/smithy4s/blob/aaf8b1f034b7d3f97e6f250c97fa08c517686d28/modules/core/src/smithy4s/schema/DefaultValueSchemaVisitor.scala#L97

modules/core/src/smithy4s/Nullable.scala
astridej: Shall do!

Also, just realising - would it make sense to put some type classes for `Nullable` in the `cats` module?

Baccata: I suppose it would 👍

modules/codegen/src/smithy4s/codegen/internals/IR.scala
astridej: Yeah, that was pretty much my conclusion too 😓 . At least `typeMod` got rid of a bunch of the places that would have needed to destructure this case class otherwise.

astridej:
> @astridej, one important aspect of this work is gonna be the addition of tests to [this specification](https://github.com/disneystreaming/smithy4s/blob/aaf8b1f034b7d3f97e6f250c97fa08c517686d28/modules/bootstrapped/test/src/smithy4s/schema/FieldSpec.scala#L9).
> 
> Most encoders now use the `foreachUnlessDefault` or `getUnlessDefault`, and we need to ensure that those behave in a somewhat coherent manner with respect to the `nullable` trait.
> 
> I presume that this line may have to change in order to account for the `nullable` trait
> 
> https://github.com/disneystreaming/smithy4s/blob/aaf8b1f034b7d3f97e6f250c97fa08c517686d28/modules/core/src/smithy4s/schema/DefaultValueSchemaVisitor.scala#L97

@Baccata  I just had a look at this and wrote some tests. I can push them, but I should note that it looks like with the current implementation, the code is actually totally independent of `nullable` and doesn't have to change at all 😄. Digging into it, it seems to come down to two things:

* in this part of the code, `nullable` is handled in a different place from required/default - we first transform the schema into a nullable schema which sets the nullable hint, and then we call `.optional`, `.required`, or `.field` depending on the required/default situation to create the `Field` from it. `getUnlessDefault` can then ignore the nullable trait in the same it can ignore any other type refinement or complex underlying schema for a field. The only thing that could go wrong is that it could screw up on calculating the default, but...
* although there's no special logic for handling `Document.DNull` in the nullable case here, it does the right thing anyway: on receiving null, the logic gets forwarded to `DefaultValueSchemaVisitor`, which attempts to calculate the default value for a schema for `Nullable[T]` and ends up with `Nullable.Null` via the bijection to `Option`, which is what we would've wanted anyway.

If I'm off-base here, let me know!

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
astridej: 😂 OK, added another detailed smoke spec test that goes through the eight possible cases and makes sure it can retrieve the error message.

modules/core/src/smithy4s/schema/Field.scala
```diff
@@ -65,10 +65,8 @@ final case class Field[S, A](
    * Applies a side-effecting lambda if the field is marked as required (using `smithy.api#required`),
    * OR if the field value is different from its default. Otherwise nothing happens.
    */
-  def foreachUnlessDefault(s: S)(f: A => Unit): Unit = {
-    val a = get(s)
-    if (isDefaultValue(a) && !isRequired) () else f(a)
```
Baccata, line 70: actually the reason we didn't factor it was to avoid an un-necessary allocation to `Option`, so I'm gonna ask you to revert (feel free to add a comment to point that out)

modules/bootstrapped/test/src/smithy4s/schema/FieldSpec.scala
```diff
@@ -52,4 +52,26 @@ final class FieldSpec extends FunSuite {
     assertEquals(result.toList, List(Some("test2"), None))
   }
 
+  test(
```
Baccata, line 55: Thank you for the additional tests. I'd like to see one for the following additional cases : 

```scala
val requiredNullable = string.nullable.required[Foo]("a", _.a)
```

and

```scala
case class Bar(
  a: Option[Nullable[String]]
)  

val optionalNullable = string.nullable.optional[Bar]("a", _.a)
```

Baccata:
> Digging into it, it seems to come down to two things: [...] 
> If I'm off-base here, let me know!

I think you may be right, but I'd love for the assumption to be validated with a couple additional [test cases](https://github.com/disneystreaming/smithy4s/pull/1408#discussion_r1497774667).

modules/core/src/smithy4s/schema/Field.scala
astridej: Done, thanks for letting me know!

modules/bootstrapped/test/src/smithy4s/schema/FieldSpec.scala
astridej: Added! I also took the liberty of adding one for `string.optional` (with no default set), because we didn't have one before and I wanted to make sure that the behaviour I was getting on `string.nullable.optional` was the same as the one on `string.optional`.

Baccata: Awesome, thank you very much !

kubukoz:
I love the emojis in the checklist. Going to steal this for my own :P

modules/docs/markdown/04-codegen/01-customisation/13-nullable-values.md
```diff
@@ -0,0 +1,56 @@
+---
+sidebar_label: Nullable Values 
+title: Nullable Values
+---
+
+By default, smithy does not distinguish between a value being absent and it being null - both are translated to `scala.None`, which in turn will be serialised as an absent value.
```
Baccata:
```suggestion
The official smithy toolset does not offer anything to distinguish between an absence of value and a value set to null during (de) serialisation. 
```

Baccata (approved):
This is an amazing PR, both from the added value but also from the fact that it has not required much oversight, and also because it adds some pretty thorough tests. 

I am extremely thankful for this 🥇 

Before we merge, I'd appreciate if @kubukoz or @lewisjkl could offer a second pair of eyes for a review, considering it's a fairly big PR.

astridej:
@Baccata thank you for the kind words! 😊 🥳 I'm glad the PR worked without major changes being needed myself - and that you appreciate my edge-case-driven testing approach :)

And thank you for your quick review and good guidance!

lewisjkl (approved):
This is great work! Thanks for including docs and updating existing docs too. Really appreciate all of it 🙏

PR #1427, opened 2024-03-01
Retain original field ordering in schemas

Baccata:
This changes the rendering logic to retain the original ordering of fields in Structure schemas, thus reflecting the spec more accurately. 

The original ordering of fields is important in some protocols, when positional information has semantic importance. 

NB for reviewers : the commits can be reviewed individually (in particular the first one)

- [x] Updated changelog

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/internals/IR.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/IR.scala
@@ -132,6 +132,7 @@ private[internals] case class Field(
     realName: String,
     tpe: Type,
     modifier: Field.Modifier,
+    originalIndex: Int,
     hints: List[Hint]
 )
 
@@ -193,9 +194,10 @@ private[internals] object Field {
       name: String,
       tpe: Type,
       modifier: Modifier,
+      originalIndex: Int,
       hints: List[Hint] = Nil
   ): Field =
-    Field(name, name, tpe, modifier, hints)
+    Field(name, name, tpe, modifier, originalIndex, hints)
 
 }
 
--- a/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
@@ -737,8 +737,7 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
               if (recursive) line"$recursive_($struct_" else line"$struct_"
             line"${schemaImplicit}val schema: $Schema_[${product.nameRef}] = $definition"
               .args(renderedFields)
-              .block(line"make")
-              .appendToLast(".withId(id).addHints(hints)")
+              .appendToLast("(make).withId(id).addHints(hints)")
               .appendToLast(if (recursive) ")" else "")
           } else {
             val definition =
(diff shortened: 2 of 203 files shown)


modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -698,9 +698,19 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
         renderProtocol(product.nameRef, hints),
         newline,
         renderLenses(product, hints),
+        locally {
+          val params =
+            fields.sortBy(_.originalIndex).map(fieldToRenderLine(_, noDefault = true)).intercalate(Line.comma)
+          val args = fields.map(f => Line(f.name)).intercalate(Line.comma)
+          lines(
+            line"// constructor using the original order from the spec",
+            line"private def make($params): ${product.nameRef} = ${product.nameRef}($args)"
```
kubukoz, line 707: **question:** I think it'd be valuable to have this in the public API, e.g. if someone really really wants to avoid getting bit by a default parameter (I've seen it happen). WDYT?

modules/codegen/src/smithy4s/codegen/internals/IR.scala
```diff
@@ -193,9 +194,10 @@ private[internals] object Field {
       name: String,
       tpe: Type,
       modifier: Modifier,
+      originalIndex: Int,
       hints: List[Hint] = Nil
   ): Field =
-    Field(name, name, tpe, modifier, hints)
+    Field(name, name, tpe, modifier, originalIndex, hints)
```
kubukoz, line 200: **suggestion:** named params

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -727,7 +737,7 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
               if (recursive) line"$recursive_($struct_" else line"$struct_"
             line"${schemaImplicit}val schema: $Schema_[${product.nameRef}] = $definition"
               .args(renderedFields)
-              .block(line"${product.nameRef}.apply")
+              .block(line"make")
```
kubukoz: **nitpick:** I'd try to avoid the block here and pass this in the same line

kubukoz (approved):

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
Baccata: If people care about it, we have rendering options to prevent default parameters from being generated : https://disneystreaming.github.io/smithy4s/docs/codegen/customisation/default-rendering

I think the "correct" course of action would be to generate the case-class without default parameters, retaining the original order, and to create a custom apply method that would have the default parameters. The problem is that it'd also impact extractors, so I don't want to pull that trigger just yet. We could potentially roll it in `0.19`, as people should be more careful about reading release notes. 

I do not however want to expose this `make` method. I've toyed with the idea of actually turning it into a `IndexedSeq[Any] => A`, as it's general and could be captured in a potential `GeneratedStructureCompanion` interface, which could potentially help the Spark usecase. I don't want to make the generated `make` public because we may change it later on.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
Baccata: [f0b79af](https://github.com/disneystreaming/smithy4s/pull/1427/commits/f0b79af38cfbf7f131a19c429b77e611589edb83)

PR #1446, opened 2024-03-15
support for untagged in DocumentDecoderSchemaVisitor

lewisjkl:
There may be a reason why we didn't have this before, but we came across a case that seems to indicate this was overlooked. We already had support for encoding untagged unions from documents, but not decoding them. Feel free to close this if I missed something and there actually is a good reason to omit them.

- [ ] Added documentation
- [x] Updated changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.18.14
+
+* Add support for decoding Document representations of untagged unions
+
 # 0.18.13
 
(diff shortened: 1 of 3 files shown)


kubukoz (approved):

daddykotex (approved):
Again, surprising we just find out that we did not support this!

thanks for the fix

Baccata (approved):

PR #1454, opened 2024-03-19
Validated newtypes

### LINKED ISSUE #966: SUGGESTION:  guidance for write-side validation of constraints
disclaimer: We're only using smithy4s as an IDL for events so this description is written from that point of view and may need translation in your head to make sense from a more service oriented perspective.

It would be really nice if there was an on rails way to check that smithy case classes have been constructed such that they pass all the constraints defined on the corresponding IDL. I only just realised it wasn't happening when our smithy defined events started failing validation on rehydration. My team will probably solve the problem by writing scalacheck tests that do a round trip on all our event structures (ie check that they can be turned into JSON and back again without erroring). It would be nice if there was an idiomatic way to check at runtime rather than relying on test exhaustivity but perhaps it isn't practical. Nevertheless @kubukoz suggested raising this issue as a placeholder to discuss.

denisrosca:
Add an option to render constrained newtypes over Smithy primitives as "validated" newtypes.

Partially implements #966 

- [x] Add validated newtypes
- [x] Make validated newtypes rendering opt-in via metadata
- [x] Handle primitives

- [x] Added documentation
- [x] Updated changelog

### DIFF
--- a/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
+++ b/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
@@ -19,7 +19,10 @@ package smithy4s.codegen
 import sbt.Keys._
 import sbt.util.CacheImplicits._
 import sbt.{fileJsonFormatter => _, _}
-import scala.util.{Success, Try}
+
+import scala.util.Success
+import scala.util.Try
+
 import JsonConverters._
 
 object Smithy4sCodegenPlugin extends AutoPlugin {
@@ -269,7 +272,8 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
           cacheFactory.make("smithy4sGeneratedSmithyFilesOutput")
         ) { case (changed, prevResult) =>
           if (changed || prevResult.isEmpty) {
-            val file = (config / smithy4sGeneratedSmithyMetadataFile).value
+            val file =
+              (config / smithy4sGeneratedSmithyMetadataFile).value
             IO.write(
               file,
               s"""$$version: "2"
--- a/modules/codegen/src/smithy4s/codegen/internals/CodegenImpl.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/CodegenImpl.scala
@@ -217,6 +217,7 @@ private[codegen] object CodegenImpl { self =>
       AwsConstraintsRemover.name :+
       AwsStandardTypesTransformer.name :+
       OpenEnumTransformer.name :+
-      KeepOnlyMarkedShapes.name
+      KeepOnlyMarkedShapes.name :+
+      ValidatedNewtypesTransformer.name
 
 }
--- a/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
@@ -187,6 +187,8 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
       renderUnion(shapeId, union.nameRef, alts, mixins, recursive, hints)
     case ta @ TypeAlias(shapeId, _, tpe, _, recursive, hints) =>
       renderNewtype(shapeId, ta.nameRef, tpe, recursive, hints)
+    case vta @ ValidatedTypeAlias(shapeId, _, tpe, recursive, hints) =>
+      renderValidatedNewtype(shapeId, vta.nameRef, tpe, recursive, hints)
     case enumeration @ Enumeration(shapeId, _, tag, values, hints) =>
       renderEnum(shapeId, enumeration.nameRef, tag, values, hints)
   }
@@ -271,8 +273,11 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
 
   def renderPackageContents: Lines = {
     val typeAliases = compilationUnit.declarations
-      .collect { case TypeAlias(_, name, _, _, _, hints) =>
-        (name, hints)
+      .collect {
+        case TypeAlias(_, name, _, _, _, hints) =>
+          (name, hints)
+        case ValidatedTypeAlias(_, name, _, _, hints) =>
+          (name, hints)
       }
       .sortBy(_._1)
       .map { case (name, hints) =>
@@ -1344,6 +1349,50 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     )
   }
 
+  private def renderValidatedNewtype(
+      shapeId: ShapeId,
+      name: NameRef,
+      tpe: Type,
+      recursive: Boolean,
+      hints: List[Hint]
+  ): Lines = {
+    val validator = {
+      val tags = hints.collect { case t: Hint.Constraint => t }
+      tags match {
+        case h :: tail =>
+          (
+            line".validating(${renderNativeHint(h.native)})" +:
+              tail.map { tag => line".alsoValidating(${renderNativeHint(tag.native)})" }
+          ).intercalate(Line.empty)
+        case _ => Line.empty
+      }
+    }
+
+    val definition =
+      if (recursive) line"$recursive_("
+      else Line.empty
+    val trailingCalls =
+      line".withId(id).addHints(hints)${renderConstraintValidation(hints)}"
+    val closing = if (recursive) ")" else ""
+    lines(
+      documentationAnnotation(hints),
+      deprecationAnnotation(hints),
+      obj(name, line"$ValidatedNewtype_[$tpe]")(
+        renderId(shapeId),
+        renderHintsVal(hints),
+        line"val underlyingSchema: $Schema_[$tpe] = ${tpe.schemaRef}$trailingCalls",
+        lines(
+          line"val validator: $Validator_[$tpe, $name] = $Validator_.of[$tpe, $name]($Bijection_[$tpe, $name](_.asInstanceOf[$name], value(_)))$validator"
+        ),
+        lines(
+          line"implicit val schema: $Schema_[$name] = ${definition}validator.toSchema(underlyingSchema)$closing"
+        ),
+        line"@inline def apply(a: $tpe): Either[String, $name] = validator.validate(a)",
+        renderTypeclasses(hints, name)
+      )
+    )
+  }
+
   private implicit class OperationExt(op: Operation) {
     def renderArgs =
       if (op.input == Type.unit) Line.empty
@@ -1400,6 +1449,8 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
             _,
             false
           ) =>
+        NameRef(ns, s"$name.schema").toLine
+      case Type.ValidatedAlias(ns, name, _) =>
         NameRef(ns, s"$name.schema").toLine
       case Type.Alias(ns, name, _, _) =>
         NameRef(ns, s"$name.underlyingSchema").toLine
@@ -1549,6 +1600,14 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
           false -> text
         else
           false -> line"${ref.show}($text)"
+      })
+    case ValidatedNewTypeTN(ref, target) =>
+      Reader(topLevel => {
+        val (wroteCollection, text) = target.run(topLevel)
+        if (wroteCollection && !topLevel)
+          false -> text
+        else
+          false -> line"${ref.show}.unsafeApply($text)"
       })
 
     case AltTN(ref, altName, AltValueTN.TypeAltTN(alt)) =>
--- a/modules/core/src-2/ValidatedNewtype.scala
+++ b/modules/core/src-2/ValidatedNewtype.scala
@@ -1,0 +1,56 @@
+/*
+ *  Copyright 2021-2024 Disney Streaming
+ *
+ *  Licensed under the Tomorrow Open Source Technology License, Version 1.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     https://disneystreaming.github.io/TOST-1.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s
+
+abstract class ValidatedNewtype[A] extends HasId { self =>
+  // This encoding originally comes from this library:
+  // https://github.com/alexknvl/newtypes#what-does-it-do
+  type Base
+  trait _Tag extends Any
+  type Type <: Base with _Tag
+
+  @inline def apply(a: A): Either[String, Type]
+
+  def schema: Schema[Type]
+
+  @inline final def unsafeApply(a: A): Type = apply(a) match {
+    case Right(value) => value
+    case Left(error)  => throw new IllegalArgumentException(error)
+  }
+
+  @inline final def value(x: Type): A =
+    x.asInstanceOf[A]
+
+  implicit final class Ops(val self: Type) {
+    @inline final def value: A = ValidatedNewtype.this.value(self)
+  }
+
+  implicit val tag: ShapeTag[Type] = new ShapeTag[Type] {
+    def id: ShapeId = self.id
+    def schema: Schema[Type] = self.schema
+  }
+
+  def unapply(t: Type): Some[A] = Some(t.value)
+
+  object hint {
+    def unapply(h: Hints): Option[Type] = h.get(tag)
+  }
+}
+
+object ValidatedNewtype {
+  private[smithy4s] trait Make[A, B] extends Bijection[A, B]
+}
--- a/modules/core/src-2/ValidatedNewtype.scala
+++ b/modules/core/src-2/ValidatedNewtype.scala
@@ -16,32 +16,17 @@
 
 package smithy4s
 
-abstract class ValidatedNewtype[A] extends HasId { self =>
-  // This encoding originally comes from this library:
-  // https://github.com/alexknvl/newtypes#what-does-it-do
-  type Base
-  trait _Tag extends Any
-  type Type <: Base with _Tag
+abstract class ValidatedNewtype[A] extends AbstractNewtype[A] { self =>
 
   @inline def apply(a: A): Either[String, Type]
-
-  def schema: Schema[Type]
 
   @inline final def unsafeApply(a: A): Type = apply(a) match {
     case Right(value) => value
     case Left(error)  => throw new IllegalArgumentException(error)
   }
 
-  @inline final def value(x: Type): A =
-    x.asInstanceOf[A]
-
   implicit final class Ops(val self: Type) {
     @inline final def value: A = ValidatedNewtype.this.value(self)
-  }
-
-  implicit val tag: ShapeTag[Type] = new ShapeTag[Type] {
-    def id: ShapeId = self.id
-    def schema: Schema[Type] = self.schema
   }
 
   def unapply(t: Type): Some[A] = Some(t.value)
--- a/modules/core/src/smithy4s/Hints.scala
+++ b/modules/core/src/smithy4s/Hints.scala
@@ -47,6 +47,7 @@ trait Hints {
   final def has[A](implicit key: ShapeTag[A]): Boolean = this.get[A].isDefined
   final def get[A](key: ShapeTag.Has[A]): Option[A] = get(key.getTag)
   final def get[T](nt: Newtype[T]): Option[nt.Type] = get(nt.tag)
+  final def get[T](vnt: ValidatedNewtype[T]): Option[vnt.Type] = get(vnt.tag)
   final def filter(predicate: Hint => Boolean): Hints =
     Hints.fromSeq(all.filter(predicate).toSeq)
   final def filterNot(predicate: Hint => Boolean): Hints =
--- a/modules/core/src/smithy4s/Hints.scala
+++ b/modules/core/src/smithy4s/Hints.scala
@@ -46,7 +46,9 @@ trait Hints {
   def get[A](implicit key: ShapeTag[A]): Option[A]
   final def has[A](implicit key: ShapeTag[A]): Boolean = this.get[A].isDefined
   final def get[A](key: ShapeTag.Has[A]): Option[A] = get(key.getTag)
-  private[smithy4s] final def get[T](nt: Newtype[T]): Option[nt.Type] = get(nt.tag)
+  private[smithy4s] final def get[T](nt: Newtype[T]): Option[nt.Type] = get(
+    nt.tag
+  )
   final def get[T](nt: AbstractNewtype[T]): Option[nt.Type] = get(nt.tag)
   final def filter(predicate: Hint => Boolean): Hints =
     Hints.fromSeq(all.filter(predicate).toSeq)
--- a/modules/core/src/smithy4s/Hints.scala
+++ b/modules/core/src/smithy4s/Hints.scala
@@ -46,8 +46,7 @@ trait Hints {
   def get[A](implicit key: ShapeTag[A]): Option[A]
   final def has[A](implicit key: ShapeTag[A]): Boolean = this.get[A].isDefined
   final def get[A](key: ShapeTag.Has[A]): Option[A] = get(key.getTag)
-  final def get[T](nt: Newtype[T]): Option[nt.Type] = get(nt.tag)
-  final def get[T](vnt: ValidatedNewtype[T]): Option[vnt.Type] = get(vnt.tag)
+  final def get[T](nt: AbstractNewtype[T]): Option[nt.Type] = get(nt.tag)
   final def filter(predicate: Hint => Boolean): Hints =
     Hints.fromSeq(all.filter(predicate).toSeq)
   final def filterNot(predicate: Hint => Boolean): Hints =
--- a/modules/core/src/smithy4s/Hints.scala
+++ b/modules/core/src/smithy4s/Hints.scala
@@ -46,6 +46,7 @@ trait Hints {
   def get[A](implicit key: ShapeTag[A]): Option[A]
   final def has[A](implicit key: ShapeTag[A]): Boolean = this.get[A].isDefined
   final def get[A](key: ShapeTag.Has[A]): Option[A] = get(key.getTag)
+  private[smithy4s] final def get[T](nt: Newtype[T]): Option[nt.Type] = get(nt.tag)
   final def get[T](nt: AbstractNewtype[T]): Option[nt.Type] = get(nt.tag)
   final def filter(predicate: Hint => Boolean): Hints =
     Hints.fromSeq(all.filter(predicate).toSeq)
--- a/modules/core/src/smithy4s/Validator.scala
+++ b/modules/core/src/smithy4s/Validator.scala
@@ -1,0 +1,70 @@
+/*
+ *  Copyright 2021-2024 Disney Streaming
+ *
+ *  Licensed under the Tomorrow Open Source Technology License, Version 1.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     https://disneystreaming.github.io/TOST-1.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s
+
+sealed trait Validator[A, B] {
+  def validate(value: A): Either[String, B]
+
+  def toSchema(a: Schema[A]): Schema[B]
+
+  def alsoValidating[C](constraint: C)(implicit
+      ev: RefinementProvider.Simple[C, A]
+  ): Validator[A, B]
+}
+
+object Validator {
+
+  def of[A, B](bijection: Bijection[A, B]): ValidatorBuilder[A, B] =
+    new ValidatorBuilder[A, B](bijection)
+
+  final class ValidatorBuilder[A, B] private[smithy4s] (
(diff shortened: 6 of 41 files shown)


Baccata:
Hey @denisrosca , thanks for looking into this. 

However : validated newtypes should not be rendered by default, for the simple reason that it'd break a lot of user code and that's not really acceptable now that smithy4s has received a reasonable amount of adoption. So it should be an opt-in, controlled by means of smithy metadata (with a key prefixed with`smithy4s`)

denisrosca:
> Hey @denisrosca , thanks for looking into this.
> 
> However : validated newtypes should not be rendered by default, for the simple reason that it'd break a lot of user code and that's not really acceptable now that smithy4s has received a reasonable amount of adoption. So it should be an opt-in, controlled by means of smithy metadata (with a key prefixed with`smithy4s`)

Working on that 👍 
Just wanted to create the PR as early as possible to get some 👀 on the newtype encoding.

denisrosca:
@Baccata @kubukoz 
I've hit a bit of a problem after adding the metadata conditional.

The main issue is that now the same smithy source file can be rendered in different ways depending if the metadata flag is set to true or false.

Consider this smithy definition:
```smithy
namespace newtypes.validated

use smithy4s.meta#unwrap
use alloy#simpleRestJson

@length(min: 1, max: 10)
string ValidatedCity

@length(min: 1, max: 10)
string ValidatedName

@unwrap
@length(min: 1, max: 10)
string ValidatedCountry

structure Person {
  @httpLabel
  @required
  name: ValidatedName

  @httpQuery("town")
  town: ValidatedCity

  @httpQuery("country")
  country: ValidatedCountry
}

```

which gets rendered to:

```scala
final case class Person(name: ValidatedName, town: Option[ValidatedCity] = None, country: Option[String] = None)

object Person extends ShapeTag.Companion[Person] {
  val id: ShapeId = ShapeId("newtypes.validated", "Person")

  implicit val schema: Schema[Person] = struct(
    ValidatedName.schema.required[Person]("name", _.name).addHints(smithy.api.HttpLabel()),
    ValidatedCity.schema.optional[Person]("town", _.town).addHints(smithy.api.HttpQuery.unsafeApply("town")),
    ValidatedCountry.underlyingSchema.optional[Person]("country", _.country).addHints(smithy.api.HttpQuery.unsafeApply("country")),
  ){
    Person.apply
  }.withId(id).addHints(hints)
}
```

This fails to compile because `smithy.api#httpQuery` is defined [like this](https://github.com/smithy-lang/smithy/blob/7c9e134eaf83f9ebaa8082e7ae9e0287cd5d2787/smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy#L938-L949):
```smithy
@length(min: 1)
string httpQuery
```

which when converted from smithy to IR is interpreted as a **validated** newtype. As a result the hint in `Person.schema` is rendered via a call to `smithy.api.HttpQuery.unsafeApply` which leads to a compilation error:

```
[info] [error] /private/var/folders/_z/c52c1q4n0fj6617nkht2pgcr0000gn/T/sbt_59b16f00/target/scala-2.13/src_managed/main/scala/newtypes/validated/Person.scala:18:89: value unsafeApply is not a member of object smithy.api.HttpQuery
[info] [error]     ValidatedCity.schema.optional[Person]("town", _.town).addHints(smithy.api.HttpQuery.unsafeApply("town")),
```


As a workaround I'm [excluding](https://github.com/disneystreaming/smithy4s/pull/1454/commits/cf9ec8eb2df2b3e91ee798823f943dc632297877#diff-3d87f30f2b03db7c8e012d2c726b1e1a0f1227fe70d2e5c24166a9046c4428a5R627-R628) the `smithy.api` namespace from the validated newtypes feature, but that doesn't seem like a bullet proof solution.

kubukoz:
I see. This is probably not specific to `smithy.api`, but a symptom of a more general problem: we should make sure that we refer to newtypes in the same way they were generated, even across codegen runs (e.g. smithy4s runs its own, then you have a library on top, then you have the user code).

If the default is "false" (unvalidated newtypes), then you can only go from `false` (nothing) to `true` (by setting the metadata key). Perhaps keeping the "previous value" in `smithy4sGenerated` metadata would be the way?

Shapes from a known previously-generated namespace would be referred to in the style they were generated, and the shapes in the currently-being-generated compilation unit would use the metadata key.

Regardless of whether we do it this way or another, I think `smithy.api` could get the validated treatment from day 1.

modules/core/src-3/NewtypeValidated.scala
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+package smithy4s
+
+abstract class NewtypeValidated[A] extends HasId { self =>
+  opaque type Type = A
+
+  def apply(a: A): Either[String, Type]
+
+  def unsafeApply(a: A): Type = a
```
kubukoz: **question:** (for discussion): should we make this throw in case of invalid values? We could also have both variants available, although the "just wrap, not questions asked" one should be so obscure that nobody will accidentally use it.

Baccata:
If anything, this is yet another argument for rendering dynamic bindings instead of static ones. As much as I hate to say it, the arguments in favour of dynamic bindings are starting to largely outweigh the ones in favour of static binding. 

> Regardless of whether we do it this way or another, I think smithy.api could get the validated treatment from day 1.

I disagree (somewhat heavily). I appreciate that validated newtypes may be a desirable feature for users, but it's also not something I want to deal when writing unit tests in smithy4s that involve hints, and arguably the maintainers of smithy4s are the ones who are most expose to the shapes coming from the "smithy.api" namespace ... unless you could macro-ify the validation when dealing with literals. 

Putting aside the "should we just give up and render dynamic bindings", what we could do is create a meta-trait that would trigger the render of newtypes as validated. The metadata flag would indicate that a model preprocessor should be run to add the meta trait to all relevant shapes. The meta-trait would be stored in the resource of published artifacts, alongside the generated code, and could be inspected by downstream code-gen runs (unlike smithy4s-prefixed metadata, which get eluded).

kubukoz:
> it's also not something I want to deal when writing unit tests in smithy4s that involve hints

We discussed this briefly with Denis, I suggested to try and see how much stuff breaks: if it's more than a handful of locations let's keep them as-is, but if it's just a few I'd say `unsafeApply` is perfectly fine. Let's be pragmatic ;)

> The metadata flag would indicate that a model preprocessor should be run to add the meta trait to all relevant shapes

hmm, so that way users could still apply the meta-trait on a case-by-case basis even if metadata is `false` / undefined. That'd be cool, I think.

> The meta-trait would be stored in the resource of published artifacts, alongside the generated code, and could be inspected by downstream code-gen runs (unlike smithy4s-prefixed metadata, which get eluded).

Not sure what you mean here - I was thinking of using the `smithy4sGenerated` metadata, which does get inspected by downstream codegen runs.

Baccata:
> Not sure what you mean here - I was thinking of using the smithy4sGenerated metadata, which does get inspected by downstream codegen runs.

Well, thing is in order to use the generated metadata, you'd basically have to store a `Map[ShapeId, Something]`, which is somewhat redundant where you could have a `@something` trait instead, which you could inspect from the shapes itself without having to thread the map where it needs to be inspected. It's arguably more idiomatic to smithy to use traits to store metadata related to specific shapes. 

>  but if it's just a few I'd say unsafeApply is perfectly fine

To be transparent, my concern also has to do with the following aspects : 

1. binary compatibility. If you want smithy.api and alloy shapes to be validated newtypes, it'd unavoidably means moving this PR to 0.19, as it'll be breaking bincompat of generated traits. 
2. convenience of using hints in annotations. This concern is tied to my prototype https://github.com/Baccata/smithy4s-deriving. The inconvenience of validated types in hints would then be felt by end users.

kubukoz:
> moving this PR to 0.19

It's already on 0.19 :)

Baccata:
My bad, haha

denisrosca:
> Putting aside the "should we just give up and render dynamic bindings", what we could do is create a meta-trait that would trigger the render of newtypes as validated. The metadata flag would indicate that a model preprocessor should be run to add the meta trait to all relevant shapes. The meta-trait would be stored in the resource of published artifacts, alongside the generated code, and could be inspected by downstream code-gen runs (unlike smithy4s-prefixed metadata, which get eluded).

I'm currently exploring this option and have a question about how the `metadata flag` would be passed to the generator. 

I have a `ValidatedNewtypesTransformer` model transformer that, when enabled, adds a new `@validateNewtypeConstraints` (name is still up for debate) trait to the relevant shapes.

My initial attempt was to basically do the same thing we do for `smithy4sRenderOptics` (i.e. have a setting in the build plugin that gets written to `generated-metadata.smithy`), but that presents a bit of a problem. 
The issue is that the decision if `ValidatedNewtypesTransformer` should be executed needs to happen before loading the smithy model, but that information is in `generated-metadata.smithy` which is read by the model loader.

What do you think about making this setting a full-blown generator argument instead of passing it via the generated metadata file?

Baccata:
> What do you think about making this setting a full-blown generator argument instead of passing it via the generated metadata file? 

You make a good point, it'd be silly to load the model twice. However, I'd like to avoid multiplying codegen arguments which need to be forwarded to 3 different "front-ends" (SBT/Mill/cli), in favour of allowing some level of configuration via the reading of a `smithy-build.json` file. Interestingly, there was a discussion this week [about this](https://discord.com/channels/1045676621761347615/1045676622491164724/1227171001003933736).

kubukoz:
@Baccata I had a brief discussion with Denis, we're thinking to "just":

- make the transformer run unconditionally (like e.g. `AwsConstraintsRemover`), but
- make it check the metadata key before applying any changes

That should resolve the double-loading problem, and the configuration continues to live in the Smithy files. Did we get this right?

Baccata:
That's sensible, actually !

modules/core/src-2/NewtypeValidated.scala
```diff
@@ -0,0 +1,75 @@
...
+      override def make(c: C): Refinement.Aux[C, Type, Type] =
+        ev.make(c).imapFull[Type, Type](asBijectionUnsafe, asBijectionUnsafe)
+    }
+
+  protected val validators: List[A => Either[String, A]]
+
+  protected def validateInternal[C](c: C)(value: A)(implicit
+      ev: RefinementProvider.Simple[C, A]
+  ): Either[String, A] =
+    ev.make(c).apply(value)
```
Baccata: I'm not liking this. For starters, the `validateInternal` re-instantiates the refinement on a per-element basis. Also, the fact that a validation may be composed of several validators is an implementation detail that should not appear in the interface of `NewTypeValidated`

You could create a `smithy4s.Validator` interface that `Refinement` would extend, and enrich it with a composition method (andThen). 

The renderer would be responsible for rendering something that could look like 

```scama
val validator : Validator[A] = Validator.of[A].validating(C1).and(C2).and(C3).build()
def apply(a: A) : Either[String, Type] = {
   validator(a) 
}
```

sampleSpecs/validated-newtype.smithy
```diff
@@ -0,0 +1,19 @@
+$version: "2"
+
+namespace smithy4s.example
+
+use smithy4s.meta#validateNewtypeConstraints
+
+@length(min: 1)
+@pattern("[a-zA-Z0-9]+")
+@validateNewtypeConstraints
```
Baccata: maybe just `validatedNewtype` ?

denisrosca: Done

modules/core/src-2/NewtypeValidated.scala
denisrosca: This is done now, could you please take another look?
Let me know if matches what you had in mind.

modules/core/src/smithy4s/Validator.scala
```diff
@@ -0,0 +1,61 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s
+
+trait Validator[A, B] {
+  def validate(value: A): Either[String, B]
```
kubukoz, line 20: **question:** given that a validator can potentially include multiple refinements, should we support `EitherNel` / `Either[::[String]` (due to the lack of a better NEL in the stdlib) from day 1, with error accumulation? A fail-fast alternative can also be given (with the current shape), for convenience.

modules/core/src/smithy4s/Validator.scala
```diff
@@ -0,0 +1,61 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s
+
+trait Validator[A, B] {
```
kubukoz: **question:** can/should we make this sealed, to allow ourselves some API changes? AFAIU users aren't expected to instantiate these, apart from `of` and the combinators like `validating`.

modules/core/src/smithy4s/Validator.scala
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  private[smithy4s] class ValidatorBuilder[A, B] private (
+      first: Refinement.Aux[_, A, A],
+      other: List[Refinement.Aux[_, A, A]],
+      bijection: Bijection[A, B]
+  ) {
+
+    def this(ref: Refinement.Aux[_, A, A], bijection: Bijection[A, B]) =
+      this(ref, List.empty, bijection)
```
kubukoz: **nitpick:** I'd replace this auxiliary constructor with an explicit parameter in the use site (it's just the one in `PartialValidatorBuilder`, right?

modules/core/src/smithy4s/Validator.scala
kubukoz, line 1: **issue:** commenting on a random spot, but we need a CHANGELOG entry. For Dynamic, I don't think we need to do anything but correct me if I'm wrong please @Baccata

modules/codegen/src/smithy4s/codegen/internals/CodegenImpl.scala
```diff
@@ -175,15 +175,17 @@ private[codegen] object CodegenImpl { self =>
...
 
     Node.prettyPrintJson(ModelSerializer.builder().build.serialize(model))
   }
 
-  private def withAwsTypeTransformer(transformers: List[String]): List[String] =
-    transformers :+ AwsConstraintsRemover.name :+ AwsStandardTypesTransformer.name :+ OpenEnumTransformer.name
+  private def withMandatoryTransformers(
+      transformers: List[String]
+  ): List[String] =
+    transformers :+ AwsConstraintsRemover.name :+ AwsStandardTypesTransformer.name :+ OpenEnumTransformer.name :+ ValidatedNewtypesTransformer.name
```
kubukoz: **nitpick:** I'd split this as

```suggestion
    transformers :+ 
      AwsConstraintsRemover.name :+ 
      AwsStandardTypesTransformer.name :+ 
      OpenEnumTransformer.name :+ 
      ValidatedNewtypesTransformer.name
```

so that future additions don't cause a diff for the whole thing

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -252,36 +260,51 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
...
+                       |metadata smithy4sRenderOptics = $shouldGenerateOptics
+                       |metadata smithy4sRenderValidatedNewtypes = $shouldRenderValidatedNewtypes
+                       |""".stripMargin
+                  )
+                  Seq(file)
+                } else {
+                  prevResult.get
+                }
+              }
+              lastOutput(changed)
```
kubukoz: **suggestion:** I'd extract some of this to avoid the current nesting level, which is becoming a bit unwieldy

modules/core/src/smithy4s/Validator.scala
denisrosca: I thought about this, but decided against it initially due to a lack of a `NonEmptyList`. Forcing the caller to deal with a result of `Left(Nil)` felt kinda awkward, even though it would technically never happen.

modules/core/src/smithy4s/Validator.scala
denisrosca: Done.

modules/core/src/smithy4s/Validator.scala
denisrosca: Added an entry to the CHANGELOG.

modules/core/src-3/NewtypeValidated.scala
denisrosca: Changed this to throw an exception. Decided against having an "obscure, unsafe-not-even-throwing" method. If anyone needs that behavior, they could just catch the exception and ignore it?

kubukoz: yeah, I think that's fair

modules/core/src/smithy4s/Validator.scala
Baccata: I agree with Denis. Generally speaking, and for better-or-worse, smithy4s does favour fail-fast as opposed to accumulation, so the decision is in line with the philosophy of the lib.

kubukoz: it does... but I do recall us discussing the possibility of allowing accumulation :)

anyway, it's a sealed trait so we can extend it later.

modules/core/src/smithy4s/Validator.scala
```diff
@@ -0,0 +1,64 @@
...
+sealed trait Validator[A, B] {
+  def validate(value: A): Either[String, B]
+}
+
+object Validator {
+
+  def of[A, B]: PartialValidatorBuilder[A, B] =
+    new PartialValidatorBuilder[A, B]
+
+  private[smithy4s] class PartialValidatorBuilder[A, B] {
```
Baccata: this always throws me off, because I'm not quite sure it's possible to call its methods from outside the smithy4s packages. 

I'd rather make the constructor private : `final class PartialValidatorBuilder[A, B] private ()`

modules/core/src/smithy4s/Validator.scala
```diff
@@ -0,0 +1,64 @@
...
+
+    def validating[C](constraint: C)(implicit
+        ev: RefinementProvider.Simple[C, A]
+    ): ValidatorBuilder[A, B] =
+      new ValidatorBuilder[A, B](first, other :+ ev.make(constraint), bijection)
+
+    def build: Validator[A, B] = new ComposedValidator(first, other, bijection)
+  }
+
+  private final class ComposedValidator[A, B](
```
Baccata: this class and `ValidatorBuilder` have exactly the same data structure. There's no reason to separate the `Validator` with its `Builder` at all then, as you could very well just have a `def alsoValidate[C](constraint: C)(implicit ...) : ValidatorBuilder[A, B]` directly on the `Validator` interface.

modules/docs/markdown/04-codegen/01-customisation/13-validated-newtypes.md
```diff
@@ -0,0 +1,66 @@
...
+---
+
+As of version `0.19.x` Smithy4s has the ability to render constrained newtypes over Smithy primitives as
+"validated" classes in the code it generates. In practice this means that a newtype will now have an 
+`apply` method that returns either a validated value or an error.
+
+If you're using Smithy4s via `mill` or `sbt`, then you can enable this functionality with the following keys:
+
+* in mill, task: `def smithy4sRenderValidatedNewtypes = true`
+* in sbt, setting: `smithy4sRenderValidatedNewtypes := true`
```
Baccata: Not sure I was totally clear, but I **do not want** for build-tool plugins (SBT, mill, CLI) to expose settings that modify the rendering.

modules/codegen/src/smithy4s/codegen/internals/CodegenRecord.scala
```diff
@@ -20,9 +20,11 @@ import software.amazon.smithy.model.Model
 import software.amazon.smithy.model.node.Node
 
 import scala.jdk.CollectionConverters._
+import scala.jdk.OptionConverters._
 
 private[internals] final case class CodegenRecord(
-    namespaces: List[String]
+    namespaces: List[String],
+    validatedNewtypes: Option[Boolean]
```
Baccata, line 27: If this is intended to be some configuration, this is not the right place for it. That'd be here : 

See https://github.com/disneystreaming/smithy4s/blob/89e88c972174fa0a3b2853901a52a5a6a4662c6d/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala#L43-L46

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
Baccata: I don't want build-plugin settings to control rendering behaviour. 

The only reason why there's. something for wildcard arguments is that this setting is inherently tied to the scala version of the project the plugin is applied. 

The optics one is a mis-hap that I'd like to actually remove in 0.19

modules/core/src-2/NewtypeValidated.scala
Baccata: It's pretty close. See my comments on the file in question.

modules/docs/markdown/04-codegen/01-customisation/13-validated-newtypes.md
denisrosca: Got it. This was a misunderstanding on my part and I followed the optics example 😅 .
I'll remove the build tool related stuff.

denisrosca: This option is gone now.

modules/core/src/smithy4s/Validator.scala
denisrosca: Reworked this a bit and got rid of the multiple "builders" and added, as suggested, a new `alsoValidate` method.

modules/core/src/smithy4s/Validator.scala
denisrosca: Done.

modules/codegen/src/smithy4s/codegen/internals/CodegenRecord.scala
denisrosca: This is not meant as "configuration", but more of a record of how namespaces where previously generated.

I'm using this in the [ValidatedNewtypesTransformer](https://github.com/disneystreaming/smithy4s/pull/1454/files#diff-ecaff2416dbb6664c0ec30aa9a980b44003a627cbe092121fcad46fe481d1abeR43-R62) to make sure that namespaces that were previously generated using this new feature have the proper `@validateNewtype` trait applied.

Does that make sense?

Baccata: I see. Yeah I guess that works 👍. The approach I would have preferred would have been to generate an `apply @validatedNewType` statement in a bespoke file that would get packaged with the artifact, but I think your approach is sound too, and arguably requires less work.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -1338,6 +1343,50 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
+    val closing = if (recursive) ")" else ""
+    lines(
+      documentationAnnotation(hints),
+      deprecationAnnotation(hints),
+      obj(name, line"$NewtypeValidated_[$tpe]")(
+        renderId(shapeId),
+        renderHintsVal(hints),
+        line"val underlyingSchema: $Schema_[$tpe] = ${tpe.schemaRef}$trailingCalls",
+        lines(
+          line"implicit val schema: $Schema_[$name] = $definition$bijection_(underlyingSchema, asBijectionUnsafe)$closing"
```
Baccata: You're gonna hate me, but the schema of `NewtypeValidated` should not be a bijection, it should be a refinement (or a series of refinements)

modules/core/src-2/NewtypeValidated.scala
```diff
@@ -0,0 +1,62 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s
+
+abstract class NewtypeValidated[A] extends HasId { self =>
```
Baccata: I think `ValidatedNewtype` sounds better (as in, it's correct English)

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
Baccata: To be more precise : the schema should not call upon `asBijectionUnsafe`. 

I think the easiest way to achieve this would be to add a `def toSchema(a: Schema[A]) : Schema[B]` to the `Validator[A, B]` interface, because the `ValidatorImpl` keeps everything that you'll need to formulate the correct schema : a list of refinements and a bijection to apply at the end.

denisrosca: Just to make sure that I understand what the expected the implementation of `toSchema` to be. I can think of something like this:

```scala
override def toSchema(a: Schema[A]): Schema[B] = 
  refinements.foldLeft(a) { (schema, refinement) => 
    schema.refined[B](refinement.constraint)
  }.biject(bijection) // problem here is that the bijection is still of the "unsafe kind"
```

and it would be invoked like so:

```scala
// ValidatedString

val underlyingSchema: Schema[String] = string.withId(id).addHints(hints).validated(smithy.api.Length(min = Some(1L), max = None))

val validator: Validator[String, ValidatedString] = Validator.of[String, ValidatedString].validating(smithy.api.Length(min = Some(1L), max = None))

val schema: Schema[ValidatedString] = validator.toSchema(string.withId(id).addHints(hints))
```

Is that close to what you had in mind?

modules/core/src-2/NewtypeValidated.scala
denisrosca: Done

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
Baccata: Sorry, I've been quite busy and had forgotten to reply to this.

Yes, that's pretty much what I had in mind 👍  except that it should be 

``` 
refinements.foldLeft(a) { (schema, refinement) 
   schema.refined[A](refinement.constraint) 
}.biject(bijection) 
```

NB : the `bijection` here is not unsafe, in that it doesn't throw. It's named "unsafe" because it circumvents validation. 

The point of doing things this way is that at the Schema level, all validation should be encompassed as a refinement schema, not as a throwing bijection.

modules/core/src-2/ValidatedNewtype.scala
```diff
@@ -0,0 +1,62 @@
...
+  def schema: Schema[Type]
+
+  implicit val tag: ShapeTag[Type] = new ShapeTag[Type] {
+    def id: ShapeId = self.id
+    def schema: Schema[Type] = self.schema
+  }
+
+  def unapply(t: Type): Some[A] = Some(t.value)
+
+  protected implicit val asBijectionUnsafe: Bijection[A, Type] =
```
Baccata: Should probably not be implicit

modules/core/src/smithy4s/Validator.scala
```diff
@@ -0,0 +1,59 @@
...
+}
+
+object Validator {
+
+  def of[A, B]: ValidatorBuilder[A, B] =
+    new ValidatorBuilder[A, B]
+
+  final class ValidatorBuilder[A, B] private[smithy4s] () {
+    def validating[C](constraint: C)(implicit
+        bijection: Bijection[A, B],
```
Baccata: The bijection should not be provided implicitly here (it should be an explicit parameter), maybe moving it to the `of` method.

modules/core/src/smithy4s/Hints.scala
```diff
@@ -47,6 +47,7 @@ trait Hints {
   final def has[A](implicit key: ShapeTag[A]): Boolean = this.get[A].isDefined
   final def get[A](key: ShapeTag.Has[A]): Option[A] = get(key.getTag)
   final def get[T](nt: Newtype[T]): Option[nt.Type] = get(nt.tag)
+  final def get[T](vnt: ValidatedNewtype[T]): Option[vnt.Type] = get(vnt.tag)
```
Baccata: possibly worth making a common supertype for `Newtype` and `ValidatedNewtype` to avoid the duplication of method here

Baccata:
It's getting close 👍 just wanted to say that your efforts are much appreciated, and apologise for how long it takes me to review

modules/core/src/smithy4s/Validator.scala
denisrosca: Done.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
denisrosca: Done

modules/core/src-2/ValidatedNewtype.scala
denisrosca: Removed

modules/core/src/smithy4s/Hints.scala
denisrosca: Changing this breaks binary compatibility (now that the branch is rebased on 0.18).

kubukoz: Looks like it's just this one method:

```
[error] [core3]  * method get(smithy4s.Newtype)scala.Option in interface smithy4s.Hints does not have a correspondent in current version
```

can we add it back and make it package-private+deprecated, for example? That could possibly let us move forward.

Baccata (approved):
Pre-emptively approving, as only scalafmt needs running I think. 

Thank you so much @denisrosca for putting in the work. I know it's been a lengthy endeavour, but it's very appreciated 👍

kubukoz (approved):

kubukoz:
well, this is fun - I can't reproduce the failure locally. Merged `series/0.18` into this one to see if that helps reproduce it, but no dice. Shall get back to it after the weekend...

kubukoz:
🎉

PR #1455, opened 2024-03-20
Protobuf support

Baccata:
This adds a `smithy4s-protobuf` module, which provides support for protobuf serialisation following the [alloy semantics](https://github.com/disneystreaming/alloy/blob/main/docs/serialisation/protobuf.md). 

The module depends on the official `protobuf-java` library, which provides low-level primitives for parsing protobuf from raw bytes. On scala-js/scala-native, the module depends on `protobuf-runtime-scala`, a pure-scala implementation of `protobuf-java` provided by the awesome [ScalaPB library](https://github.com/scalapb/ScalaPB). 

The implementation is tested against ScalaPB. The tests involve a conversion from smithy to the proto IDL, performed by means of smithy-translate. 

Additionally, this PR adds a documentation page, summarising various serialisation constructs provided by smithy4s out-of-the-box (including Document/JSON/XML/Protobuf). This documentation is by no mean exhaustive but should be helpful to users. 

- [x] Added documentation
- [x] Updated changelog


build.sbt
```diff
@@ -849,6 +882,19 @@ lazy val bootstrapped = projectMatrix
...
+      genSmithyResources(Compile).taskValue
+      (Compile / PB.generate).value
+    },
+    Compile / PB.protoSources ++= Seq(
+      exampleGeneratedResourcesOutput.value
+    ),
+    Compile / PB.targets := Seq(
+      scalapb.gen() -> (Compile / sourceManaged).value / "scalapb"
+    ),
     Test / fork := true,
```
kubukoz: **nitpick:** caught my eye randomly: we're setting `Test/fork` to true unconditionally here, but in the protobuf module it's only set for the JVM. Any clue as to why this line doesn't cause problems? (IIRC JS will complain if you keep forking on)

modules/docs/markdown/02.1-serialisation/01-serialisation.md
```diff
@@ -0,0 +1,118 @@
...
+It is worth noting that, although `Document` is isomorphic to a JSON ADT, its `.toString` is not valid JSON. Likewise, the `smithy4s-core` module does not contain logic to parse JSON strings into Documents. In order to read/write Documents from/to JSON strings, you need the `smithy4s-json` module. The `smithy4s.json.Json` entry-point contains methods that work with Documents.
+
+### JSON
+
+The `smithy4s-json` module provides [jsoniter-based](https://github.com/plokhotnyuk/jsoniter-scala) encoders/decoders that can read/write generated data-types from/to JSON bytes/strings, without an intermediate JSON ADT. The performance of this module is very competitive are [very competitive](https://plokhotnyuk.github.io/jsoniter-scala/) in the Scala ecosystem.
+
+This module is provided at the following coordinates :
+
+```
+sbt : "com.disneystreaming.smithy4s"  % "smithy4s-json" % "@VERSION@"
```
kubukoz: **issue:** `%%` instead of `%`, also there are extra spaces.

suggestion: UI-wise I think we could improve upon these code snippets by splitting them in two, in addition we could use more advanced Docusaurus features such as [mdx tabs](https://docusaurus.io/docs/markdown-features/tabs). They're nice in the fact that you can make them synchronized, so that picking Mill once updates all the visible tab groups. Example: https://scala-php.org/docs/getting-started/ (sbt/mill tab groups are synced)

modules/docs/markdown/02.1-serialisation/01-serialisation.md
```diff
@@ -0,0 +1,118 @@
...
+import smithy4s.protobuf.Protobuf
+
+val personCodec = Protobuf.codecs.fromSchema(Person.schema)
+val personBytes = personCodec.writeBlob(Person(name = "John Doe"))
+val maybePerson = personCodec.readBlob(personBytes)
+```
+
+By default, `smithy4s-protobuf` abides by the semantics of :
+
+* [alloy protobuf traits](https://github.com/disneystreaming/alloy/blob/main/docs/serialisation/protobuf.md). These semantics are the exact same semantics that [smithy-translate](https://github.com/disneystreaming/smithy-translate) uses to translate smithy to protobuf. This implies that the Smithy4s protobuf codecs are compatible with the codecs of other protobuf tools, generated from the .proto files resulting from running smithy through smithy-translate. In short, Smithy4s and [ScalaPB](https://github.com/scalapb/ScalaPB) can talk to each other.
```
kubukoz: **suggestion:** should we expand "smithy4s and scalapb can talk to each other" to "smithy4s, scalapb and other Protobuf implementations can talk to each other"? It's implied I think, but being explicit couldn't hurt.

modules/protobuf/src/main/scala/smithy4s/protobuf/Protobuf.scala
```diff
@@ -0,0 +1,25 @@
...
+ *  limitations under the License.
+ */
+
+package smithy4s.protobuf
+
+import smithy4s.schema._
+
+object Protobuf {
+
+  def codecs: CachedSchemaCompiler[ProtobufCodec] = ProtobufCodec
```
kubukoz, line 23: **question:** shouldn't this be a val? It seems to be the case for xml/json.

modules/protobuf/src/main/scala/smithy4s/protobuf/internals/Wire.scala
```diff
@@ -0,0 +1,39 @@
...
+    (protoIndex << 3) | wireType
+
+  def decodeTag(byte: Byte): (Int, Int) = {
+    val m = byte & 0xff
+    (m >> 3, m & 0x7)
+  }
+
+  // See https://protobuf.dev/programming-guides/encoding/#structure
+  //
+  // NB : not ascribing types to get special encoding in bytecode.
```
kubukoz, line 31: **question:** (optional) - I'd be interested in the particular encoding we're aiming for. Also, would `final val` be of any help here? (hoping for inlining)

kubukoz:
Looks good UX-wise. Do we have a plan to support full-blown gRPC as well?

modules/protobuf/src/main/scala/smithy4s/protobuf/internals/Wire.scala
Baccata: I think I was going for inlined constant, but it's been ages and you're absolutely right that I'm missing the final.

modules/docs/markdown/02.1-serialisation/01-serialisation.md
Baccata: I like the tabs idea, but let's defer it to a next PR to do it more widely. I'll create an issue

Baccata: https://github.com/disneystreaming/smithy4s/issues/1471

modules/protobuf/src/main/scala/smithy4s/protobuf/Protobuf.scala
Baccata: Not necessarily, because `ProtobufCodec` itself is a singleton and this is just forwarding the reference.

build.sbt
Baccata: [45a100a](https://github.com/disneystreaming/smithy4s/pull/1455/commits/45a100a19020e53c92007f51dcd709bfaf9582b7)

modules/docs/markdown/02.1-serialisation/01-serialisation.md
Baccata: [e1bccc0](https://github.com/disneystreaming/smithy4s/pull/1455/commits/e1bccc0f340772c4356f2f77a893094e80f67ed5)

Baccata:
> Do we have a plan to support full-blown gRPC as well 

Not at this point. It'd be really easy to do with http4s-grpc (which is still reasonably young) or with the official grpc lib (which is java only), but I don't currently want to a assume the maintenance of a gRPC module in open-source. So it's left as an exercise to the user 😄

modules/protobuf/src/main/scala/smithy4s/protobuf/internals/Wire.scala
Baccata: [9334fe5](https://github.com/disneystreaming/smithy4s/pull/1455/commits/9334fe568a80d34a2b356cb3a4615ea26a0b6fb5)

kubukoz:
I call dibs on the blogpost then. @velvetbaldmime don't you dare 🤣

modules/protobuf/src/main/scala/smithy4s/protobuf/Protobuf.scala
kubukoz: hmmmmm fair nuf

modules/docs/markdown/02.1-serialisation/01-serialisation.md
kubukoz: SGTM

modules/protobuf/src/main/scala/smithy4s/protobuf/ProtobufCodec.scala
```diff
@@ -0,0 +1,103 @@
...
+import java.io.InputStream
+import java.io.OutputStream
+
+// scalafmt: {maxColumn = 120}
+trait ProtobufCodec[A] {
+  def writeBlob(a: A): Blob
+  final def readBlob(blob: Blob): Either[ProtobufReadError, A] = try {
+    Right(unsafeReadBlob(blob))
+  } catch {
+    case e: Throwable => Left(ProtobufReadError(e))
```
kubukoz: **suggestion:** catch only NonFatal

modules/protobuf/src/main/scala/smithy4s/protobuf/internals/TaggedCodec.scala
```diff
@@ -0,0 +1,555 @@
...
+  final case class ScalarFieldCodec[A](
+      underlying: ScalarCodec[A]
+  ) extends TaggedCodec[A] {
+    def wireType: Int = underlying.wireType
+    def isPrimitive: Boolean = true
+    def oneOfTags: Option[Seq[Int]] = None
+    def write(protoIndex: Int, a: A, os: CodedOutputStream): Unit =
+      underlying.writeTag(protoIndex, a, os)
+    def prepareWrite(protoIndex: Int, a: A): WriteNode = {
+      if (a == underlying.zero) {
```
kubukoz: **nitpick:** 
```suggestion
      if (underlying.isZero(a)) {
```

kubukoz: there's another one in this file

modules/protobuf/src/main/scala/smithy4s/protobuf/internals/TaggedCodecSchemaVisitor.scala
```diff
@@ -0,0 +1,285 @@
...
+      shapeId: ShapeId,
+      hints: Hints,
+      tag: EnumTag[E],
+      values: List[EnumValue[E]],
+      total: E => EnumValue[E]
+  ): TaggedCodec[E] = tag match {
+    case ClosedIntEnum | ClosedStringEnum =>
+      val allIndexed = values.forall(_.hints.has(ProtoIndex))
+      val indexedValues = values.zipWithIndex.map { case (ev, index) =>
+        val protoIndex = if (allIndexed) ev.hints.get(ProtoIndex).map(_.value).getOrElse(index) else index
```
kubukoz: **question:** I'm curious about this pattern, it's applied in other kinds of schemas as well.

IIUC the whole conditional is functionally equivalent to the `true` case, so it appears to be an optimization (if everything is indexed we don't have to go through hints at all).
If that's the case, is it really a significant gain, especially given this happens during schema compilation and not on the hot path?

modules/docs/markdown/02.1-serialisation/01-serialisation.md
```diff
@@ -0,0 +1,156 @@
...
+val personBytes = personCodec.writeBlob(Person(name = "John Doe"))
+val maybePerson = personCodec.readBlob(personBytes)
+```
+
+By default, `smithy4s-protobuf` abides by the semantics of :
+
+* [alloy protobuf traits](https://github.com/disneystreaming/alloy/blob/main/docs/serialisation/protobuf.md). These semantics are the exact same semantics that [smithy-translate](https://github.com/disneystreaming/smithy-translate) uses to translate smithy to protobuf. This implies that the Smithy4s protobuf codecs are compatible with the codecs of other protobuf tools, generated from the .proto files resulting from running smithy through smithy-translate. In short, Smithy4s and [ScalaPB](https://github.com/scalapb/ScalaPB) can talk to each other : the ScalaPB codecs generated from protobuf after a translation from smithy are able to decode binary data produced by Smithy4s protobuf codecs (and vice versa).
+
+
+```
```
kubukoz, line 121: **suggestion:** please add a code comment with the name of the tool used to generate this, so that we can easily update it in the future if needed.

Also, Docusaurus has Mermaid support built in, apparently - https://docusaurus.io/docs/next/markdown-features/diagrams - I'd rather use that instead, but consider this a nitpick as the diagram it fairly straightforward anyway.

Baccata: I looked at mermaid and didn't find a type of diagram that was just a few boxes and arrows. Maybe I didn't look hard enough

modules/protobuf/src/main/scala/smithy4s/protobuf/internals/TaggedCodecSchemaVisitor.scala
Baccata: It's not really an optimisation, the goal is disregard the manually added indexes if only a subset of the fields are annotated (which should be caught at build time in the nominal case). 

Arguably this falls under the category of "how much effort should we put into validating the schemas at the level of smithy4s ?", and that's a bit of a gray area. Maybe I should throw instead ... ?

Baccata: After some thoughts, I decided to throw : [dc23a16](https://github.com/disneystreaming/smithy4s/pull/1455/commits/dc23a16f2084dd2eb926fdcb5f999297209d1c51)

modules/protobuf/src/main/scala/smithy4s/protobuf/internals/TaggedCodec.scala
Baccata: [c0cc265](https://github.com/disneystreaming/smithy4s/pull/1455/commits/c0cc265bb948ee389361829647eefdb363c8bd32)

modules/protobuf/src/main/scala/smithy4s/protobuf/ProtobufCodec.scala
Baccata: [c0cc265](https://github.com/disneystreaming/smithy4s/pull/1455/commits/c0cc265bb948ee389361829647eefdb363c8bd32)

kubukoz (approved):

PR #1456, opened 2024-03-20
Prevent documentation for streaming members

daddykotex:

- [x] Updated changelog

You can see from the [first commit](https://github.com/disneystreaming/smithy4s/commit/5d6e486a38dd8292c90b6d817cc53eb6845f81ea) that originally, codegen would write scaladocs in irrelevant places

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,7 @@
 # 0.19.0
 
-<<<<<<< HEAD
-* Reworked enumerations / the EnumerationSchema to eliminate a OOM pitfall and improve ergonomics of SchemaVisitor
-=======
+## Reworked enumerations / the EnumerationSchema to eliminate a OOM pitfall and improve ergonomics of SchemaVisitor
+
 ## Remove localhost from default URI in [#1341](https://github.com/disneystreaming/smithy4s/pull/1341)
 
 Previously, URIs constructed with a base URI of `/` would have `localhost` as the host. In some cases, that may not be desirable, such as in the case of frontend clients that want to reuse the window's origin. This is now fixed: hostnames are optional in the smithy4s URI model, and default to `None`.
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
 # 0.19.0
 
-<<<<<<< HEAD
-* Reworked enumerations / the EnumerationSchema to eliminate a OOM pitfall and improve ergonomics of SchemaVisitor
-=======
+## Documentation fix
+
+Prevent documentation from being generated for case class when the field are not generated because they're annotated with `@streaming`
```
Baccata, line 5: Instead of removing it altogether, could we make a separate paragraph out of it in the docs of the holding operation ?

daddykotex: sure, I could enhance this part [here](https://github.com/disneystreaming/smithy4s/pull/1456/commits/5d6e486a38dd8292c90b6d817cc53eb6845f81ea#diff-34e511a0fb4161255b724d9a8634097950f7189fd341b25c2eb061ef7b045b13R18-R20) that was generated originally

👍

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -932,13 +930,69 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+              } yield s"This operation uses @streaming on the input (${inMem})."
+              maybeDoc.toList
+            case (None, Some(out)) =>
+              val maybeDoc = for {
+                outMem <- out
+                  .asMemberShape()
+                  .asScala
+                  .map(_.getMemberName)
+              } yield s"This operation uses @streaming on the output (${outMem})."
+              maybeDoc.toList
```
kubukoz, line 989: **question:** shall we have some examples of this in `bootstrapped`?

daddykotex: we have examples for input, output but not both, should I add  one for bidirectional streaming?

kubukoz: ugh actually I'm just blind/dumb and missed these examples. I don't think it's a must to have the one with both

kubukoz: that said, it might be good to have such an example regardless of this documentation PR.

Baccata (approved):

PR #1468, opened 2024-04-03
@only trait that allows keeping only some operations from services (e.g. AWS specs)

keynmol:
PR TODO:
- [x] Implement transitive shape removal logic
- [x] Add documentation
- [x] Remove logging

- [x] Added documentation
- [x] Updated changelog

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/transformers/KeepOnlyMarkedShapes.scala
+++ b/modules/codegen/src/smithy4s/codegen/transformers/KeepOnlyMarkedShapes.scala
@@ -1,0 +1,113 @@
+/*
+ *  Copyright 2021-2024 Disney Streaming
+ *
+ *  Licensed under the Tomorrow Open Source Technology License, Version 1.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     https://disneystreaming.github.io/TOST-1.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.codegen.transformers
+
+import software.amazon.smithy.build.ProjectionTransformer
+import software.amazon.smithy.build.TransformContext
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.shapes._
+import scala.jdk.CollectionConverters._
+import scala.jdk.OptionConverters._
+import software.amazon.smithy.model.neighbor._
+
+class KeepOnlyMarkedShapes extends ProjectionTransformer {
+
+  override def getName: String = KeepOnlyMarkedShapes.name
+
+  private val annotationName = "smithy4s.meta#only"
+  private val protectedNamespaces = Set("smithy.api")
+
+  private def log(msg: Any) = {
+    // Uncomment to actually see the logs
+    // System.err.println(
+    //   s"[KeepOnlyMarkedShapes]: $msg"
+    // )
+
+    ()
+  }
+
+  override def transform(ctx: TransformContext): Model = {
+    val operationsToKeepBuilder = collection.mutable.Set.empty[ShapeId]
+    val operationsToRemoveBuilder = collection.mutable.Set.empty[ShapeId]
+
+    val taggedOperations = ctx
+      .getModel()
+      .getShapesWithTrait(ShapeId.from(annotationName))
+      .asScala
+      .flatMap(_.asOperationShape().toScala.map(_.getId()))
+
+    operationsToKeepBuilder ++= taggedOperations
+
+    val model = ctx.getModel()
+
+    model
+      .getServiceShapes()
+      .asScala
+      .foreach { service =>
+        val operationIds = service.getOperations().asScala.toSet
+        val serviceTaggedOperations =
+          operationIds.filter(taggedOperations.contains)
+        log(s"${service.getId()}: ${serviceTaggedOperations}")
+
+        if (serviceTaggedOperations.nonEmpty) {
+          operationsToRemoveBuilder ++= operationIds -- serviceTaggedOperations
+        } else {
+          // pretend that all service's operations were tagged
+          operationsToKeepBuilder ++= operationIds
+        }
+      }
+
+    log(s"Operations to keep: $operationsToKeepBuilder")
+    log(s"Operations to remove: $operationsToRemoveBuilder")
+
+    val walker = new Walker(model)
+    val removalCandidates = operationsToRemoveBuilder.flatMap { id =>
+      model.getShape(id).toScala.toSet.flatMap { (shape: Shape) =>
+        val closure = walker
+          .walkShapeIds(shape)
+          .asScala
+          .filterNot(ns => protectedNamespaces.contains(ns.getNamespace()))
+
+        log(s"Removed operation $id closure: $closure")
+
+        closure
+      }
+    }
+    log(s"Removal candidates (before filtering): $removalCandidates")
+
+    operationsToKeepBuilder.foreach { oid =>
+      model.getShape(oid).toScala.foreach { shape =>
+        val closure = walker.walkShapeIds(shape).asScala
+        log(s"Kept operation $oid closure: $closure")
+        removalCandidates --= closure
+      }
+    }
+
+    log(s"Removal candidates (after filtering): $removalCandidates")
+
+    ctx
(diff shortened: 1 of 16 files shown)


modules/codegen/test/src/smithy4s/codegen/internals/TestUtils.scala
```diff
@@ -101,4 +102,29 @@ object TestUtils {
       .unwrap()
   }
 
+  def loadAndValidateModel(namespaces: String*): Model = {
```
keynmol, line 104: Using `loadModel`, which disabled validations, led to unpredictable results - like shapes not being correctly handled by walker, incorrect output/input shapes (i.e. allowing booleans to be naked inputs), etc.

I'm not sure what the original rationale for this was, but it affected the behaviour of tests so severely that I decided to stop guessing what went wrong and just fix the test specs instead.

modules/codegen/test/src/smithy4s/codegen/internals/TestUtils.scala
```diff
@@ -101,4 +102,29 @@ object TestUtils {
...
+
+    namespaces
+      .foldLeft(assembler) { case (a, model) =>
+        a.addUnparsedModel(s"test-${model.hashCode}.smithy", model)
+      }
+      .assemble()
+      .unwrap()
+  }
+
+  def serializeModel(model: Model): String = {
```
kubukoz: **question:** is this used? Text search tells me no

modules/docs/markdown/03-protocols/03-aws/04-code-size.md
```diff
@@ -0,0 +1,88 @@
+---
+sidebar_label: Managing code size
```
kubukoz: **suggestion:** perhaps this should be moved to codegen customizations, rather than stay in protocol-specific land? IMO it's useful beyond AWS

kubukoz: or to Guides. But then a more dry-style documentation page would be nice to have in `codegen`, imo

modules/docs/markdown/03-protocols/03-aws/04-code-size.md
```diff
@@ -0,0 +1,88 @@
...
+
+```scala
+    smithy4sAwsSpecs ++= Seq(AWS.comprehend)
+```
+
+As you run `compile` in your build tool, you can see that there are 428 files being compiled. Uh-oh.
+
+Those 428 files cover the entirety of Comprehend's [84 operations](https://docs.aws.amazon.com/comprehend/latest/APIReference/API_Operations.html) and all the dataclasses supporting them. But we don't need all of these operations - we only care about [DetectSentiment](https://docs.aws.amazon.com/comprehend/latest/APIReference/API_DetectSentiment.html).
+
+### @only annotation
```
kubukoz: **issue:** needs backticks

modules/docs/markdown/03-protocols/03-aws/04-code-size.md
```diff
@@ -0,0 +1,88 @@
...
+
+You start with the [smithy4s.g8 template](https://github.com/disneystreaming/smithy4s.g8/), and, following the [documentation](01-aws.md), instruct Smithy4s plugin to generate you a client SDK for Comprehend:
+
+```scala
+    smithy4sAwsSpecs ++= Seq(AWS.comprehend)
+```
+
+As you run `compile` in your build tool, you can see that there are 428 files being compiled. Uh-oh.
+
+Those 428 files cover the entirety of Comprehend's [84 operations](https://docs.aws.amazon.com/comprehend/latest/APIReference/API_Operations.html) and all the dataclasses supporting them. But we don't need all of these operations - we only care about [DetectSentiment](https://docs.aws.amazon.com/comprehend/latest/APIReference/API_DetectSentiment.html).
```
kubukoz: **nitpick:** 
```suggestion
Those 428 files cover the entirety of Comprehend's [84 operations](https://docs.aws.amazon.com/comprehend/latest/APIReference/API_Operations.html) and all the datatypes supporting them. But we don't need all of these operations - we only care about [DetectSentiment](https://docs.aws.amazon.com/comprehend/latest/APIReference/API_DetectSentiment.html).
```

modules/docs/markdown/03-protocols/03-aws/04-code-size.md
```diff
@@ -0,0 +1,88 @@
...
+The exact semantics of the annotation are as follows:
+
+- It can only be applied to operations
+- For any service that has operations tagged with this annotation, the rest of the operations will be **removed**, along with all the shapes (structures, aliases, unions, enums, etc.) that those operations reference, as long as those shapes are deemed **removable** (see below)
+- A shape is deemed **removable** if it's only reachable from the removed operations (either directly or transitively)
+- Services that don't have *any* operations tagged with this annotation are assumed to have *all* of their operations tagged implicitly (so that all the operations are kept)
+
+The key mechanism that allows this operation to be usable with AWS client generation is that we can [apply annotations post factum on shapes defined elsewhere](https://disneystreaming.github.io/smithy4s/docs/guides/model-preprocessing#note-on-third-party-models).
+
+So to only preserve `DetectSentiment` operation, we can create a Smithy file in our project with these contents:
```
kubukoz: **issue:** this says "only `DetectSentiment` but you're applying `only` on two ops. Which is it?

modules/protocol/resources/META-INF/smithy/smithy4s.meta.smithy
```diff
@@ -11,6 +11,9 @@ metadata suppressions = [
 
 namespace smithy4s.meta
 
+@trait(selector: ":is(service, operation)")
```
kubukoz: **question:** the doc page said it's for operations only (line 46 of `04-code-size.md` at this moment). Should this be updated?

modules/protocol/resources/META-INF/smithy/smithy4s.meta.smithy
```diff
@@ -11,6 +11,9 @@ metadata suppressions = [
...
+structure only {}
```
kubukoz, line 15: **nitpick:** if we're sticking to "just operations", I'd call it sth like `keepOperation`. I'm ok with the current name though

modules/protocol/src/smithy4s/meta/OnlyTrait.java
```diff
@@ -0,0 +1,47 @@
...
+
+package smithy4s.meta;
+
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.traits.AnnotationTrait;
+import software.amazon.smithy.model.traits.AbstractTrait;
+
+public class OnlyTrait extends AnnotationTrait {
```
kubukoz, line 25: **nitpick:** side note, these can now be generated - https://github.com/smithy-lang/smithy/pull/2074 - perhaps now would be a good time to try it out and compare?

Baccata (approved):

modules/docs/markdown/03-protocols/03-aws/04-code-size.md
Baccata: That's a fair point, I'd support moving under the "code-generation/customizations" section.

modules/protocol/resources/META-INF/smithy/smithy4s.meta.smithy
Baccata: Yes

modules/protocol/resources/META-INF/smithy/smithy4s.meta.smithy
Baccata: We may expand it later on with additional uses, so I'm happy with the name.

modules/protocol/src/smithy4s/meta/OnlyTrait.java
Baccata: Not necessary.

keynmol:
Thanks for review folks, let me know if this is good to go.

modules/codegen/src/smithy4s/codegen/transformers/KeepOnlyMarkedShapes.scala
```diff
@@ -0,0 +1,113 @@
...
+import software.amazon.smithy.model.neighbor._
+
+class KeepOnlyMarkedShapes extends ProjectionTransformer {
+
+  override def getName: String = KeepOnlyMarkedShapes.name
+
+  private val annotationName = "smithy4s.meta#only"
+  private val protectedNamespaces = Set("smithy.api")
+
+  private def log(msg: Any) = {
```
kubukoz: **suggestion:** make it by-name so that we don't compute `toString` of large sets for no reason

kubukoz (approved):

PR #1470, opened 2024-04-03
Add functionality to recreate a smithy model from smithy4s schemas/services 

lewisjkl:
Adds the capability to translate a smithy4s service or schema to the corresponding smithy model object (from the smithy model library). This lives in the dynamic module and will be useful for testing and other purposes in the future.

- [x] Added documentation
- [x] Updated changelog


lewisjkl:
Will update changelog and add documentation tomorrow, but wanted to put this up so people could take a look at what I have so far.

modules/dynamic/src-jvm/DynamicSchemaIndexPlatform.scala
```diff
@@ -16,28 +16,99 @@
...
-    val flattenedModel =
-      ModelTransformer.create().flattenAndRemoveMixins(model);
-    val node = ModelSerializer.builder().build.serialize(flattenedModel)
-    val document = NodeToDocument(node)
-    smithy4s.Document
-      .decode[smithy4s.dynamic.model.Model](document)
-      .map(load(_)) match {
-      case Left(error)  => throw error
-      case Right(value) => value
+  def toSmithyModel: Model = {
```
kubukoz, line 36: **question:** has it been considered that we'd convert to the `dynamic.Model` representation instead? That one can still be loaded into a `Model.assembler` and serialized to the IDL, but it'd be a cross-platform representation.

Baccata: We've considered it and decided to punt it to a later time, as it'd be more work.

Baccata (approved):

PR #1486, opened 2024-04-12
Switch out uPickle for Circe in codegen

astridej:
As discussed in the Discord, this is a preliminary change for being able to support smithy-build.json as input configuration objects; parsing these in a future-proof way is relatively complicated and difficult to support in uPickle without hand-writing the entirety of the parsing logic. Since Circe is already used elsewhere in the project, we replace the uPickle with Circe, including the implementation of creating the smithy-build.json file from settings.

Note that the special requirement of merging two JSON files while concatenating and deduplicating their arrays is not supported out of the box by Circe's deepMerge and still has to be handled manually, but at least the concatenation part has an option issue ( https://github.com/circe/circe/issues/271 )

- [✅ ] Added unit-tests (for runtime code)
- [🚫] Added bootstrapped code + smoke tests (when the rendering logic is modified)
- [🚫] Added build-plugins integration tests (when reflection loading is required at codegen-time)
- [🚫] Added alloy compliance tests (when simpleRestJson protocol behaviour is expanded/updated)
- [🚫] Updated dynamic module to match generated-code behaviour
- [🚫] Added documentation
- [🚫] Updated changelog

### DIFF
--- a/modules/codegen/test/src/smithy4s/codegen/internals/SmithyBuildSpec.scala
+++ b/modules/codegen/test/src/smithy4s/codegen/internals/SmithyBuildSpec.scala
@@ -30,17 +30,17 @@ final class SmithyBuildSpec extends munit.FunSuite {
     assertEquals(
       actual,
       """|{
-         |    "version": "1.0",
-         |    "imports": [
+         |    "version" : "1.0",
+         |    "imports" : [
          |        "src/"
          |    ],
-         |    "maven": {
-         |        "dependencies": [
+         |    "maven" : {
+         |        "dependencies" : [
          |            "dep"
          |        ],
-         |        "repositories": [
+         |        "repositories" : [
          |            {
-         |                "url": "repo"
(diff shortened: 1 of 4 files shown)


modules/codegen/test/src/smithy4s/codegen/internals/SmithyBuildSpec.scala
```diff
@@ -30,17 +30,17 @@ final class SmithyBuildSpec extends munit.FunSuite {
     assertEquals(
       actual,
       """|{
-         |    "version": "1.0",
-         |    "imports": [
+         |    "version" : "1.0",
```
astridej, line 33: uPickle and Circe seem to have slightly different opinions on suitable JSON whitespace. I could also change the tests to do a whitespace-ignoring compare to avoid them breaking with changes like this, but how often do we expect to swap out the JSON parsing library? (Famous last words...)

Baccata (approved):

PR #1487, opened 2024-04-12
Prevent default values from being rendered for refined fields

Baccata:
Fixes a bug that would happen when a refinement trait would be used on a member rather than a top-level shape, that would allow for default values of the wrong type to be rendered for the case class field.

This corrects the behaviour by preventing the rendering of default values altogether for case-class fields that are refined

- [x] Added bootstrapped code 
- [x] Updated changelog


sampleSpecs/refined.smithy
```diff
@@ -86,22 +79,24 @@ string Name
...
+    age: Age
+    personAge: PersonAge = 1
+    @required
+    requiredAge: Age
+    fancyList: FancyList
+    unwrappedFancyList: UnwrappedFancyList
+    name: Name
+    dogName: DogName
+    @ageFormat
+    inlineFieldConstraint: Integer = 1
```
kubukoz, line 91: **question:** perhaps we should validate-prevent having defaults on fields with refinements?

also, how are we handling defaults on the target shape (say, `@nameFormat @default("") string DogName`)?

Baccata: > perhaps we should validate-prevent having defaults on fields with refinements? 

I don't want to go that far. The way I see it : defaults are inherently more important in the context of deserialisation than in the context of case-class instantiation, and we already have users mixing refinements and defaults. Besides, the bug this PR fixes only happened when the refinement constraint is added on a field. If the refinement was added on a simple shape, and a field was targeting that refined shape and had a default, it was already working exactly the same way (ie the default parameter on the case class was already eluded). 

> also, how are we handling defaults on the target shape (say, @nameFormat @default("") string DogName)?

We don't need to do anything specific, this basically constraint every field targeting `DogName` to either repeat the default (`name: DogName = ""`) or erase it (`name: DogName = null`).

kubukoz (approved):

Issue #1478, opened 2024-04-10
sbt plugin: allow configuring openapi spec generation


yoohaemin:
I have a problem that is solved by calling `jsonAdd` on openapi generation. It’s the same as this: https://github.com/smithy-lang/smithy/issues/834

It’s easy if I can control the arguments passed to the following method:

https://github.com/disneystreaming/alloy/blob/fd6b8b067b69e3033c288bb665de8054108bf747/modules/openapi/src/alloy/openapi/package.scala#L33

However, the current sbt plugin doesn’t provide a way to configure this; it simply calls the generate method that doesn’t take any configuration arguments:

https://github.com/disneystreaming/smithy4s/blob/ee9abfb9430879eeabc18dd9edab188f7b32c03d/modules/codegen/src/smithy4s/codegen/internals/CodegenImpl.scala#L67

Would it be possible that the sbt plugin expose a key to transform this configuration? 

In the meantime, because it’s quite difficult to modify the build process from our sbt build, I am thinking of re-parsing and transform the outputted json...

Baccata:
In a very timely fashion, @astridej and I have been [discussing](https://discord.com/channels/1045676621761347615/1045676622491164724/1227171001003933736) at how to best solve this. 

Can't guarantee a timeframe but it's being looked at.

yisraelU:
had a similar such desire to add a server list. TY

kubukoz:
Is this fixed in #1492?

kubukoz:
@yoohaemin I'm closing this as it's either fixed in #1492 or not needed anymore (which is fine by me!) :) 

Please reopen / ping us if this is still unresolved and you'd like to help.

yoohaemin:
It's fixed! Sorry, I was meaning to come back to this but I forgot. Thank you very much.

PR #1492, opened 2024-04-17
Add smithy build openapi support

astridej:
As discussed in Discord -

This adds support for supplying a `smithy-build.json` during codegen, either by CLI or by either the SBT or Mill plugins. Initially, the only thing in this file that is supported is the OpenAPI plugin, but in future more features can be parsed and handled in codegen if so desired. Docs have been added that explain this.

This would solve #1478 .

- [✅ Added unit-tests (for runtime code) _one in the codegen module_
- [👎] Added bootstrapped code + smoke tests (when the rendering logic is modified)
- [✅] Added build-plugins integration tests (when reflection loading is required at codegen-time)
- [👎 ] Added alloy compliance tests (when simpleRestJson protocol behaviour is expanded/updated)
- [👎 ] Updated dynamic module to match generated-code behaviour
- [✅] Added documentation
- [✅] Updated changelog


modules/codegen/src/smithy4s/codegen/internals/CodegenImpl.scala
```diff
@@ -24,13 +24,17 @@ import smithy4s.codegen.transformers._
...
 import software.amazon.smithy.model.shapes.ModelSerializer
+import software.amazon.smithy.openapi.OpenApiConfig
 
 import scala.jdk.CollectionConverters._
 import software.amazon.smithy.model.transform.ModelTransformer
 
 private[codegen] object CodegenImpl { self =>
 
   def generate(args: CodegenArgs): CodegenResult = {
+    val smithyBuild = args.smithyBuild
```
astridej, line 35: I tried to also handle the existing fields we already had in `SmithyBuild` here (`imports`, `dependencies` and `repositories`) but ran into problems with it, probably because I don't actually understand how the `ModelLoader` works particularly well. 😅 Especially since the PR was big enough already, I figured I'd just sent one implementation that acts on nothing but the OpenApiConfig and one can always iterate further from there.

modules/codegen/src/smithy4s/codegen/internals/SmithyBuild.scala
```diff
@@ -17,25 +17,87 @@
...
+}
+
 private[codegen] object SmithyBuild {
-  implicit val codecs: Codec[SmithyBuild] = deriveCodec
-  def writeJson(sb: SmithyBuild): String = sb.asJson.spaces4
+  // automatically map absence of value to empty Seq for ease of use
+  implicit def optionalSeqDecoder[T](implicit
+      base: Decoder[T]
+  ): Decoder[Seq[T]] =
+    Decoder.decodeOption(Decoder.decodeSeq[T]).map(_.getOrElse(Seq.empty))
```
astridej: I had to decide between having a semantically correct model that's easy to use with complex parsing logic, or a model closer to the actual SmithyBuild JSON file in structure that was easy to parse but hard to use. I opted for the first, which means:
* custom `Seq` decoder so the model doesn't have to use `Option[Seq[T]]` to handle missing fields (I believe circe-generic-extras actually has a configuration option for this, but it doesn't seem to be available in all the Scala versions we have here so I went for the homebrewed version)
* modelling the plugins as a sequence, with a hand-written decoder that reads an object of the form
```
{
   "openapi": (openapi config structure goes here),
   "other-plugin": (other plugin structure goes here)
}
```
and looks up the appropriate decoder for each plugin. (Although right now you could model this as a case class with a field `openApi` instead, plugin names can be dynamic so that version wouldn't be future-proof).

Let me know if you think this is too much Circe magic.

modules/codegen/src/smithy4s/codegen/internals/SmithyBuild.scala
```diff
@@ -17,25 +17,87 @@
...
+          keys.toList.traverse(key => c.get(key)(SmithyBuildPlugin.decode(key)))
+      }
+    }
+    .map(_.getOrElse(Seq.empty))
+
+  /* Class containing only the subset of the smithy-build.json properties that need
+   * to be serialized when creating a smithy-build.json file. Allows us to skip
+   * things that are both unnecessary and very complicated to serialize,
+   * such as OpenApiConfig.
+   */
```
astridej, line 71: See comment - as discussed in the Discord, Smithy doesn't seem to think we'd ever need to serialize `OpenApiConfig` objects back to JSON. To be entirely fair to them, we _don't_ - the `writeJson` logic is about creating a `smithy-build.json` file from our SBT/Mill settings, and the openApi configuration isn't one of them. Still, it means that we can't use the same model for read and write.

modules/docs/markdown/06-guides/smithy-build-config.md
```diff
@@ -0,0 +1,47 @@
+---
+sidebar_label: Smithy build config
+title: Smithy Build Configuration
+---
+
+## Introduction
+
+Smithy provides the ability to configure the Smithy build and output by a [smithy-build configuration file](https://smithy.io/2.0/guides/smithy-build-json.html#smithy-build-json). As smithy4s uses its own build logic, it generally loads its configuration from elsewhere. However, limited support for build customization using a Smithy build configuration file is available. In particular, the [OpenAPI plugin](https://smithy.io/2.0/guides/model-translations/converting-to-openapi.html) can be used to customize the OpenAPI generation.
```
astridej, line 8: Questions:
* Does this place make sense for docs, or is there any existing section it could be added to? Nothing seemed to really fit.
* Would it make sense to mention the smithy-build JSON _generation_ capabilities here as well?
* Any other suggestions on how to better document this feature?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Add support for converting smithy4s services and schemas to smithy models
 * Add `smithy4s.meta#only` annotation allowing to filter operations in
 services, intended to reduce the amount of code generated from AWS specs
+* Add support for passing custom OpenAPI config via a `smithy-build.json` file
```
Baccata: you'll have to sync with `series/0.18` again, we've released 0.18.16 already

modules/codegen-plugin/src/sbt-test/codegen-plugin/smithy-build/smithy-build.json
```diff
@@ -0,0 +1,12 @@
+{
+  "version": "1.0",
+  "plugins": {
+    "openapi": {
+      "service": "smithy4s.example#ObjectService",
```
Baccata, line 5: our of curiosity, does this imply that the openapi-translation of only one service can be configured this way ?

modules/codegen/src/smithy4s/codegen/internals/CodegenImpl.scala
Baccata: that's entirely fine imho : in the context of build-plugins (which constitute the large majority of the smithy4s usage), this information is transmitted over by the build-tool anyway, which should act as the source of truth for this.

modules/codegen/src/smithy4s/codegen/internals/SmithyBuild.scala
Baccata: Nah I'm fine with it, thank you 👍

modules/codegen/src/smithy4s/codegen/internals/SmithyBuild.scala
```diff
@@ -17,25 +17,87 @@
...
 private[codegen] object SmithyBuild {
-  implicit val codecs: Codec[SmithyBuild] = deriveCodec
-  def writeJson(sb: SmithyBuild): String = sb.asJson.spaces4
+  // automatically map absence of value to empty Seq for ease of use
+  implicit def optionalSeqDecoder[T](implicit
+      base: Decoder[T]
+  ): Decoder[Seq[T]] =
+    Decoder.decodeOption(Decoder.decodeSeq[T]).map(_.getOrElse(Seq.empty))
+
+  implicit val pathDecoder: Decoder[os.FilePath] =
```
Baccata, line 50: Could you make these implicit values private, please ?

modules/codegen/src/smithy4s/codegen/internals/SmithyBuild.scala
```diff
@@ -45,3 +107,27 @@ private[internals] final case class SmithyBuildMavenRepository(
...
+      )
+  }
+
+  case class OpenApi(config: OpenApiConfig) extends SmithyBuildPlugin
+
+  object OpenApi {
+    implicit val decoder: Decoder[OpenApi] = Decoder[JsonObject].emapTry {
+      obj =>
+        Try {
+          val config = OpenApiConfig.fromNode(Node.parse(obj.toJson.noSpaces))
```
Baccata: This certainly works, but I'd rather reconstruct a node by using a Json folder from circe instead of going via the strings.

modules/docs/markdown/06-guides/smithy-build-config.md
Baccata: > Does this place make sense for docs, or is there any existing section it could be added to? Nothing seemed to really fit.

Yeah I think it's fine. The `Guides` section is  a bit of a misc anyway. 

> Would it make sense to mention the smithy-build JSON generation capabilities here as well? 

I think so ? Even if we end up repeating information

modules/codegen-plugin/src/sbt-test/codegen-plugin/smithy-build/smithy-build.json
astridej: The `alloy` conversion logic completely ignores whatever service is set here and will apply [the same OpenAPI config to all services](https://github.com/disneystreaming/alloy/blob/main/modules/openapi/src/alloy/openapi/package.scala#L79-L81). If we wanted to do anything involving service-based filtering, that would have to change first. For now, it's probably worth mentioning this and other Alloy OpenAPI overrides in the docs - I can add a bit about that.

That said, good point that if alloy _did_ check the service name, the current implementation would only allow for one service's OpenAPI to be configured. Checking the Smithy docs, it looks like I overlooked [support for an optional artifact ID after the plugin name](https://smithy.io/2.0/guides/smithy-build-json.html#plugin-id). If I fix that, you'd be able to include separate plugins `openapi::foo` vs `openapi::bar`, but you'd need a change on the alloy side to be able to actually pass multiple of them and have them be respected.

modules/codegen/src/smithy4s/codegen/internals/SmithyBuild.scala
astridej: Yeah, I was not really happy about this way of wiring up the smithy `Node` with circe JSON. I'll have a look at the fold logic and see if I can get it to work without going all the way back to the string level.

astridej: Done! Number conversion was a little awkward but I _think_ this should minimize any errors introduced by unnecessary rounding.

modules/codegen-plugin/src/sbt-test/codegen-plugin/smithy-build/smithy-build.json
Baccata: Alright, that's somewhat disappointing. If it comes to that, I suppose we could codify our own plugin semantics (which could map a unique configuration to services). 

Regardless, this doesn't take away from the fact that this PR is a decent step forward 👍

Baccata (approved):
Happy with the changes. @kubukoz, if you find the time, would you mind reviewing too ?

kubukoz:
> Happy with the changes. @kubukoz, if you find the time, would you mind reviewing too ?

I'll try closer to the weekend 🥺

Baccata:
@astridej, mind syncing the PR ? I'll merge right after

astridej:
@Baccata Should be up-to-date now!

PR #1499, opened 2024-04-22
Improve SBT caching during code generation

### LINKED ISSUE #1495: Default value of `smithy4sOutputDir` can break codegen cache
## TL;DR
Can we please change:
```scala
config / smithy4sOutputDir := (config / sourceManaged).value / "scala"
```
to something like:
```scala
config / smithy4sOutputDir := (config / sourceManaged).value / "scala" / "smithy4s"
```

Or exclude `codegenArgs.output` from cache altogether?  

## Problem
I'm working on an sbt plugin that uses smithy4s, but also generates some of it's own managed sources. Smithy4s produces files to smithy4sOutputDir which by default is defined as `smithy4sOutputDir := (config / sourceManaged).value / "scala"`. This value is then later used [by Smithy4sCodegenPlugin](https://github.com/disneystreaming/smithy4s/blob/7042418b036d85aa97120ba9b9421d22b5a3b93b/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala#L426) as a part of `codegenArgs` used to produce files. The entire `CodegenArgs` object is used as a cache key and thus the `output` counts in. That wouldn't be a problem because the path would just remain cached as a `String`, but smithy4s also overrides the [json format for os.Path](https://github.com/disneystreaming/smithy4s/blob/7042418b036d85aa97120ba9b9421d22b5a3b93b/modules/codegen-plugin/src/smithy4s/codegen/JsonConverters.scala#L33) in a way that calculates the hash of the entire directory content. It isn't a problem as long as smithy4s is the only source generator. When another plugin adds managed sources the hashes change and we stop hitting cache. In some cases this might lead to repeated very long compilation times.

Perhaps it's worth reconsidering if smithy4s codegen cache should rely on the contents of `source_managed` and `resource_managed` at all, since it seems easy to invalidate the cache. 

It's also not obvious to the end user why the cache got invalidated. In my debugging case I would be very happy if the plugin told me why it decided to rerun codegen even though smithy sources remained unchanged.

## Solution

I think adding another segment in the path should solve this and save fellow sbt plugin authors some headache. This way other plugins can either use the root `sourceManaged / scala` directory or anything different than `sourceManaged / scala / smithy4s`.

The other option is to modify `JsonConverters.codegenArgsIso` in a way that it doesn't serialize `output`.

majk-p:
Closes #1495.

@kubukoz's comment:

- adds a `smithy4s/` prefix to generated paths under `src_managed/main/scala` to avoid clashing with other codegens (in the default config for the generated code output)
- changes the hashing of paths for the output directory to be string-based (`os.Path`), while inputs are still hashed based on contents (`PathRef`)

- [ ] Added documentation
- [x] Updated changelog


modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -443,14 +443,16 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
       ) {
         Function.untupled {
           Tracked.lastOutput[(Boolean, CodegenArgs), Seq[File]](
-            s.cacheStoreFactory.make("output")
+            s.cacheDirectory / "smithy4s-output"
           ) { case ((inputChanged, args), outputs) =>
             if (inputChanged || outputs.isEmpty) {
+              s.log.debug("Regenerating managed sources")
```
majk-p, line 462: For local debugging I also used json converter to serialize `codegenArgs` and wrote then to a tmp file for comparison. Do you think this is an useful thing to add upstream?

modules/codegen-plugin/src/smithy4s/codegen/JsonConverters.scala
```diff
@@ -61,13 +61,17 @@ private[smithy4s] object JsonConverters {
     )
 
   // format: off
-  type GenTarget = List[os.Path] :*: os.Path :*: os.Path :*: Set[FileType] :*: Boolean:*: Option[Set[String]] :*: Option[Set[String]] :*: List[String] :*: List[String] :*: List[String] :*: List[os.Path] :*: LNil
+  type GenTarget = List[os.Path] :*: String :*: String :*: Set[FileType] :*: Boolean:*: Option[Set[String]] :*: Option[Set[String]] :*: List[String] :*: List[String] :*: List[String] :*: List[os.Path] :*: LNil
```
kubukoz: **suggestion:** I would rather restore `os.Path` here, and make `os.Path`'s format based on the toString. For cases where we do want to use `os.Path` as a checksum, IMO let's newtype it - that way we'll get closer to Mill's `PathRef`, which is the inspiration for our checksumming of paths.

Most importantly, this would save us the trouble of rendering and re-parsing the path. "stringly typed" isn't great

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
kubukoz: I think logging the params here is OK (on `debug`)

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -403,8 +403,8 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
       (inputDirs ++ generatedFiles)
         .filter(_.exists())
         .toList
```
kubukoz, line 410: **question:** what about Mill? I think it has the same problem:

```scala
def smithy4sOutputDir: T[PathRef] = T {
  PathRef(T.ctx().dest / "scala")
}

def smithy4sResourceOutputDir: T[PathRef] = T {
  PathRef(T.ctx().dest / "resources")
}
```

cached `PathRef` state will be invalidated if the contents change. It should probably be just `os.Path` in these settings (this might have to wait for 0.19 because we'd change public type signatures).

Also, even today this probably isn't idiomatic Mill: `T{...}` doesn't get recomputed, but it's wrapping a mutable input here (`PathRef`). Should probably be `T.input`, but then you'd see the problem from #1495 quite prominently in Mill.

Baccata: Nah, mill is fine. For starters, downstream tasks need to know when their upstreams change, that's what the pathref is for. 

Secondly, the dest call means that the location is specific to the task, which implies a separation between different plugins, unlike what we had in SBT

modules/codegen-plugin/src/smithy4s/codegen/JsonConverters.scala
Baccata: Agreed

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
kubukoz: > Secondly, the dest call means that the location is specific to the task, which implies a separation between different plugins, unlike what we had in SBT

I agree that it does have the separation, it's not what I'm arguing.

> downstream tasks need to know when their upstreams change, that's what the pathref is for.

yeah, and I'm saying that changes to the upstream (`smithy4sOutputDir`) don't invalidate the downstream _because_ the output dir is a Target:

```scala
import $ivy.`com.disneystreaming.smithy4s::smithy4s-mill-codegen-plugin::0.18.16`

import smithy4s.codegen.mill._
import mill._
import mill.scalalib._

object service extends Smithy4sModule {

  def millSourcePath = os.pwd

  def scalaVersion = "3.3.3"

  def foo = T {
    T.ctx().log.info(s"recomputing foo at ${java.time.Instant.now()}!")
    os.walk(smithy4sOutputDir().path).size
  }

}
```

demo: the actual file count changes, but not the one `foo` sees (because `foo` doesn't re-run, because its input, `smithy4sOutputDir`, didn't change, because it memoizes the path):

image

project here: https://github.com/kubukoz/demos/tree/mill-smithy4s-demo-pathref

Demo without `foo`: https://asciinema.org/a/qSS09OZIRlUi7e5lnMInFzfOW

so to sum up, I think the Mill plugin is in the limbo of not deciding whether regenerating code: should result in invalidating the cache of the output path (`PathRef` would do that), or not (`T{...}` doesn't invalidate).

For what it's worth, just slapping on `T.input` will technically help tasks like `foo`, but it's not a full solution as it will also break the plugin. To see how:

1. Make `smithy4sOutputDir` a `T.input`, publish snapshot
2. run `service.generatedSources` (this will correctly create files in the output path)
3. run `service.smithy4sOutputDir`
4. look at the output path

You'll find that it has been removed, and with it the generated sources.

---

A proper solution in my mind looks more like this:

1. turn `smithy4sOutputDir` into an `os.Path`, mirroring what Michał is doing in this PR
6. anyone who wishes to be invalidated when there are output changes will wrap that task in a `PathRef` in their own `T.input`

so:

```scala
//smithy4s-mill
def smithy4sOutputDir: T[os.Path] = T { ... }

//
//
// use-site of whatever wants to be notified of the output dir's current state
def smithy4sOutputPath: T[PathRef] = T.input { PathRef(smithy4sOutputDir()) }

// example task using the thing
def foo = T {
    //this now gets re-run every time generated sources are changed
    os.walk(smithy4sOutputPath().path).size
}

```

Baccata: Oh, you're right.  It'd be worth asking the mill folks what they think.

Like, ideally, the codegen task would only every be able to produce code in its own dest, sandbox style, and you'd be able to inspect the path to the dest folder of a task statically without actually running it. Doing that is really important in the context of scala LSP, where you shouldn't have to run anything to know the source directories of a module.

Anyway, that's the rationale for having the smithy4s output path as a separate task : we shouldn't have to run the codegen to know where it outputs things. 

But you are correct, the PathRef-returning task should be an input rather than a target.

modules/codegen-plugin/src/smithy4s/codegen/JsonConverters.scala
majk-p: Done, please let me know if that's what you meant

modules/codegen-cli/src/smithy4s/codegen/cli/CodegenCommand.scala
```diff
@@ -116,7 +117,7 @@ object CodegenCommand {
             defaultDependencies ++ dependencies.getOrElse(List.empty)
           }
           CodegenArgs(
-            specsArgs,
+            specsArgs.map(PathRef(_)),
```
kubukoz: **thought:** / issue: I thought we'd keep `CodegenArgs` as-is, given that `PathRef`'s hashing is used by the build tool, and not the underlying codegen. i.e. you'd unwrap PathRef in `codegenArgsIso`

Baccata: Agreed

Baccata:
Don't forget to sign the CLA

modules/codegen-cli/src/smithy4s/codegen/cli/CodegenCommand.scala
majk-p: Fair point, I've restored the public API for `CodegenArgs`. `PathRef` is only used for serialization now

Baccata: @majk-p, did you push your latest commit ? I'm not seeing it.

lefou:
Since it was requested in Mill's chat room, here is some input regarding Mill's cache. I have to admit, I didn't review the full PR and all comments, but from what I digested, here are my notes:

* If you return a `PathRef` that points to somewhere outside of Mill's cache, there is a chance Mill will miss it's later invalidation due to removal or change. Therefore in Mill 0.11 we introduced the `revalidate` flag which you can enable with `PathRef.withRevalidateOnce`. When enabled and before re-using a targets cache, Mill will re-calculate the `PathRef.sig` and will only use the cached value if the `sig` is still identical. I think you should use this flag since you don't have exclusive control over the output directory.

* Using an `T.input` or a `PathRef` in general for a path you intend to change from a downstream target, will most likely cause unnecessary re-evaluation. Since you intend to generate something into that path, it's `sig` will change. If you're not interested in the content of a path, which is the case for config values, the path itself should be enough.

modules/codegen-cli/src/smithy4s/codegen/cli/CodegenCommand.scala
majk-p: I didn't :facepalm: Pushed now

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -403,28 +403,30 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
...
     val localJars =
-      (conf / smithy4sAllDependenciesAsJars).value.map(os.Path(_)).toList
+      (conf / smithy4sAllDependenciesAsJars).value.toList.sorted
+        .map(p => os.Path(p))
     val res =
       (conf / resolvers).value.toList.collect { case m: MavenRepository =>
         m.root
       }
-    val transforms = (conf / smithy4sModelTransformers).value
+    val transforms = (conf / smithy4sModelTransformers).value.sorted
```
Baccata: this changes the semantics. We do not want to sort the transformers, the ability for users to control their order is important.

Baccata (changes requested):

modules/codegen/src/smithy4s/codegen/PathRef.scala
```diff
@@ -0,0 +1,19 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.codegen
+
+final case class PathRef(underlying: os.Path)
```
Baccata: this should not be part of the `codegen` module, but rather the sbt module.

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -403,28 +403,30 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
       (inputDirs ++ generatedFiles)
         .filter(_.exists())
         .toList
-    val outputPath = (conf / smithy4sOutputDir).value
-    val resourceOutputPath = (conf / smithy4sResourceDir).value
+    val outputPath = (conf / smithy4sOutputDir).value / "smithy4s"
+    val resourceOutputPath = (conf / smithy4sResourceDir).value / "smithy4s"
```
Baccata: the changes to the target directories is the reason why the scripted SBT tests are failing. You're gonna have to amend a bunch of files called `test` in the direct subfolders of https://github.com/disneystreaming/smithy4s/tree/series/0.18/modules/codegen-plugin/src/sbt-test/codegen-plugin

modules/codegen/src/smithy4s/codegen/PathRef.scala
majk-p: Sure, will move it

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
majk-p: That's what I'm on right now. It takes a while to run those + I can't make `multimodule` case run yet, but I'll get there :crossed_fingers:

Baccata: you can run them individually if it helps

majk-p: I went through the scripted tests and adjusted the paths. Here's what seems to work and what not
```
✅ aws-newtype-flatten
✅ aws-open-enums
✅ aws-specs
✅ custom-settings
✅ defaults
✅ dependencies-only
✅ extra-configs
❌ multimodule
✅ multimodule-aws
✅ multimodule-no-compile
❌ multimodule-staged
✅ optics-generate
✅ protobuf
✅ scala3
✅ smithy-rules
❌ wildcard-config
```
This made me question the part about getting rid of hashing generated sources contents. If we have a look at the last test for example, here's what it does:
```
> ++2.13.10 root/compile
$ exists target/scala-2.13/src_managed/main/smithy/generated-metadata.smithy
$ exists target/scala-2.13/resource_managed/main/smithy4s/META-INF/smithy/generated-metadata.smithy

> ++3.3.0 root/compile
$ exists target/scala-3.3.0/src_managed/main/smithy/generated-metadata.smithy
$ exists target/scala-3.3.0/resource_managed/main/smithy4s/META-INF/smithy/generated-metadata.smithy

# ensure metadata file is re-generated after deleting
$ delete target/scala-2.13/src_managed/main/smithy/generated-metadata.smithy
$ delete target/scala-3.3.0/resource_managed/main/smithy4s/META-INF/smithy/generated-metadata.smithy
> ++3.3.0 root/compile
$ exists target/scala-3.3.0/src_managed/main/smithy/generated-metadata.smithy
$ exists target/scala-3.3.0/resource_managed/main/smithy4s/META-INF/smithy/generated-metadata.smithy
```
Why it does `delete target/scala-2.13/src_managed/main/smithy/generated-metadata.smithy` is unclear to me (i'd expect 3.3.0 there) but since my version doesn't do hashing and only remembers the path, it would not re-generate the sources in case they were modified by hand.

I think we need to decide if this fine or not. Without hashing managed sources and resources won't be regenerated which I think users will find confusing. Should I restore hashing and only stick with having the dedicated smithy4s managed directories?

kubukoz: I think we can't just change the base path of `resources` because that messes with SPI and smithy's manifest (it has to be `META-INF/smithy/manifest`), that's why `multimodule` is failing (downstreams don't see shapes from upstreams)

kubukoz: Changing this (removing the `smithy4s` nesting in resources) makes most tests pass, but the one you've posted is still problematic. Did that in 99254fe1bca4f01fa7e7fd0cb980237c76d07fa7

IMO manually touching target files means you're on your own. Instead of this `delete`, we could do a `clean` or just skip that check entirely.

Baccata: Logically I agree, but I wonder why the test was added in the first place ... Let's exhume whatever PR added it, see if we can get a hint

kubukoz: I'm looking into it.

modules/codegen-plugin/src/smithy4s/codegen/JsonConverters.scala
kubukoz: yeah

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
kubukoz: The test was added in https://github.com/disneystreaming/smithy4s/pull/736/.

I think it was just added as what seemed intuitive at the time. Removed these lines in 70709ae596a3613d6ad3aeac4445de70ea820ea2

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
kubukoz: moved to new ticket: https://github.com/disneystreaming/smithy4s/issues/1529

modules/codegen/src/smithy4s/codegen/PathRef.scala
kubukoz: acfd9ddcef09095f154acdb34abe202543da5483

kubukoz:
LGTM (@Baccata, you haven't committed here so I'll leave it for you to approve if, it looks OK now)

Baccata (approved):

PR #1508, opened 2024-05-03
Use ListSet for repos, imports and dependencies

daddykotex:
The reason we saw duplicates is because we have multiple modules and they all inherit from the project plugin that is enabled. Because we use `Seq`, we just keep on adding new elements. Using a `Set` ensure we only have uniques in the end.

When running the command twice, the duplicates are cleared. This happens because the deep merge algorithm calls `distinct` when it merge arrays.

- [ ] Added documentation
- [x] Updated changelog

### DIFF
--- a/modules/codegen-plugin/src/smithy4s/codegen/GenerateSmithyBuild.scala
+++ b/modules/codegen-plugin/src/smithy4s/codegen/GenerateSmithyBuild.scala
@@ -19,16 +19,17 @@ package smithy4s.codegen
 import sbt._
 import sbt.Keys._
 import Smithy4sCodegenPlugin.autoImport._
+import scala.collection.immutable.ListSet
 
 private final case class SmithyBuildData(
-    imports: Set[String],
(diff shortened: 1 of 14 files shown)


modules/codegen-plugin/src/smithy4s/codegen/GenerateSmithyBuild.scala
```diff
@@ -21,14 +21,14 @@ import sbt.Keys._
 import Smithy4sCodegenPlugin.autoImport._
 
 private final case class SmithyBuildData(
-    imports: Seq[String],
-    deps: Seq[String],
-    repos: Seq[String]
+    imports: Set[String],
+    deps: Set[String],
+    repos: Set[String]
```
Baccata: I'd rather we used `scala.collection.immutable.ListSet` instead. The ordering (in particular for the repos) is significant

Baccata (approved):

Issue #1457, opened 2024-03-25
`@adt`, mixins and optionality changes don't play well together


kubukoz:
If you apply a mixin to a shape, and then make an inherited field required, smithy4s will not generate an `extends` with that mixin.

However, if such a shape is part of and `@adt` union, the union still has `extends`. This results in a compilation error:

```smithy
$version: "2"

namespace input

@mixin
structure HasName {
    name: String
}

structure Person with [HasName] {
    @required
    $name
}

@smithy4s.meta#adt
union U {
    p: Person
}
```

```
[error] /Users/kubukoz/projects/demos/target/scala-2.13/src_managed/main/scala/input/U.scala:31:27: incompatible type in overriding
[error] def name: Option[String] (defined in trait HasName);
[error]  found   : String
[error]  required: Option[String]
[error]   final case class Person(name: String) extends U {
[error]                           ^
[error] one error found
```

Checked in 0.18.15.

kubukoz:
Slightly related / worth considering the relation to: #1312

Baccata:
Don't have much thoughts about it besides "let's fix it I guess ?"

kubukoz:
yeah, I think we're gonna have to extend #425.

kubukoz:
Looks like it's closed in #1510

PR #1510, opened 2024-05-05
Filter mixins that have different types

daddykotex:
Similar to what was done in https://github.com/disneystreaming/smithy4s/pull/425, I've filtered out the
mixins that have different type than the final shape they're
applied to.

As such, in the example above, `HasName` has a optional name
member where as `Person` inherits the mixin but makes the member
required. The filtering catches that and as a result the mixin is
removed from the list of mixins to add on `Person`.

This behaviour aligns with the behaviour that happens when no `@adt`
trait is added and `Person` is code generated on it's own. In both
cases, now, the trait `HasName` is not extended because the types of
`name` don't align.

- [ ] Added documentation
- [x] Updated changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 0.18.17
 
 * Constraints applied to list or map members are now correctly rendered in the generated code.
+* Fix a bug when using `adt` with mixins, see #1457
 
 # 0.18.16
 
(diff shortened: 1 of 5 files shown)


Baccata (approved):

Baccata:
@daddykotex thanks for this. Please sync the PR and I'll get it merged

kubukoz:
This closes #1457, right? @daddykotex

daddykotex:
Yup my bad.

kubukoz:
No problem at all! Thanks for the fix :)

PR #1511, opened 2024-05-06
fix: use default values for optional httpPayload if there is no body

### LINKED ISSUE #1502: Support `@default` on payloads?
This may be controversial...

I think `@default` should be used on payloads, i.e. if the HTTP request/response body is empty, we should attempt to decode it from something like `defaultValue.orElse(emptyObject)`.

It appears that we're currently only trying to decode from an empty body:

(this is from a reproduction posted below)

```
[info] 400 Bad Request: Succeeded: {"payload":{"path":".","expected":"string","message":"expected '\"', offset: 0x00000000, buf:\n+----------+-------------------------------------------------+------------------+\n|          |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f | 0123456789abcdef |\n+----------+-------------------------------------------------+------------------+\n| 00000000 | 7b 7d                                           | {}               |\n+----------+-------------------------------------------------+------------------+"}}
[info] DEBUG: Errored: HttpPayloadError(., expected = string, message=expected '"', offset: 0x00000000, buf:
[info] +----------+-------------------------------------------------+------------------+
[info] |          |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f | 0123456789abcdef |
[info] +----------+-------------------------------------------------+------------------+
[info] | 00000000 | 7b 7d                                           | {}               |
[info] +----------+-------------------------------------------------+------------------+)
```

To reproduce:

```smithy
$version: "2"

namespace hello

use alloy#simpleRestJson

@simpleRestJson
service WeatherService {
    operations: [PostWeather]
}

@http(method: "POST", uri: "/weather")
operation PostWeather {
    input := {
        @required
        @httpPayload
        city: String = "strange town"
    }
    output := {
        @httpPayload
        weather: String = "good weather"
    }
}

```

```scala
package hello

import cats.effect.IO
import cats.effect.IOApp
import hello.WeatherService
import org.http4s.HttpApp
import org.http4s.Method
import org.http4s.Request
import org.http4s.Response

  def serverDemo = SimpleRestJsonBuilder
    .routes(
      new WeatherService.Default[IO](IO.stub) {}: WeatherService[IO]
    )
    .resource
    .flatMap { routes =>
      EmberServerBuilder
        .default[IO]
        .withHttpApp(routes.orNotFound)
        .build
    }
    .use { server =>
      EmberClientBuilder.default[IO].build.use { client =>
        client
          .run(
            Request[IO](method = Method.POST, uri = server.baseUri / "weather")
          )
          .use { response =>
            response.bodyText.compile.string
              .debug(response.status.toString())
              .void
          }
      }
    }
    .void

  def clientDemo = SimpleRestJsonBuilder(WeatherService)
    .client(
      Client.fromHttpApp(HttpApp.pure[IO](Response[IO]()))
    )
    .resource
    .use(_.postWeather("NYC"))
    .debug()
    .attempt
    .void

  def run = serverDemo *> clientDemo
}

```

Full repro: https://github.com/kubukoz/demos/tree/smithy4s-required-payloads

ghostbuster91:
Related PR in alloy: https://github.com/disneystreaming/alloy/pull/163
This closes https://github.com/disneystreaming/smithy4s/issues/1502

TODO:
- [x] Update alloy dependency once released

- [ ] Added documentation
- [ ] Updated changelog


modules/core/src/smithy4s/http/HttpUnaryClientCodecs.scala
```diff
@@ -119,19 +119,10 @@ object HttpUnaryClientCodecs {
...
-            case None =>
-              new HttpRequest.Writer[Blob, A] {
-                def write(request: HttpRequest[Blob], value: A): HttpRequest[Blob] =
-                  if (request.body.isEmpty) request
-                  else request.withContentType(requestMediaType)
-              }
+          new HttpRequest.Writer[Blob, A] {
+            def write(request: HttpRequest[Blob], value: A): HttpRequest[Blob] =
+              if (request.body.isEmpty) request
+              else request.withContentType(requestMediaType)
```
ghostbuster91, line 125: **thought:** I am not convinced about this change. 
I added following test in alloy:
```smithy
apply HttpPayloadWithDefault @httpRequestTests([{
    id: "SimpleRestJsonSomeHttpPayloadWithDefault"
    documentation: "Pass simple string value as is if payload provided"
    protocol: simpleRestJson
    method: "PUT"
    uri: "/httpPayloadWithDefault"
    body: "\"custom value\""
    headers: {
        "Content-Type": "application/json"
    }
    requireHeaders: ["Content-Length"]
    params: {
        nested: "custom value"
    }
    bodyMediaType: "application/json"
}])
```
and without this change the test fails in one case:
```
[error] [http4s] - alloy.test#HttpPayloadWithDefault(client|Request): SimpleRestJsonSomeHttpPayloadWithDefault 3ms
[error] [http4s]   Header Content-Type has value `text/plain` but expected `application/json` (modules/tests/src/smithy4s/tests/ProtocolComplianceSuite.scala:183)
[error] [http4s]
[error] [http4s]     [INFO] 15:03:03 [ProtocolComplianceSuite.scala:173] Pass simple string value as is if payload provided
[error] [http4s]
```

ghostbuster91:
This currently fails with the linked version of alloy due to:
```
[info] [http4s] smithy4s.http4s.SimpleRestJsonComplianceSuite
[error] [http4s] - alloy.test#HttpPayloadWithDefault(client|Request): SimpleRestJsonSomeHttpPayloadWithDefault 5ms
[error] [http4s]   Header Content-Type has value `text/plain` but expected `application/json` (modules/tests/src/smithy4s/tests/ProtocolComplianceSuite.scala:180)
[error] [http4s]
[error] [http4s]     [INFO] 17:34:32 [ProtocolComplianceSuite.scala:170] Pass simple string value as is if payload provided
[error] [http4s]
[error] [http4s] - alloy.test#HttpPayloadWithDefault(client|Request): SimpleRestJsonNoneHttpPayloadWithDefault 5ms
[error] [http4s]   Header Content-Type has value `text/plain` but expected `application/json` (modules/tests/src/smithy4s/tests/ProtocolComplianceSuite.scala:180)
[error] [http4s]
[error] [http4s]     [INFO] 17:34:32 [ProtocolComplianceSuite.scala:170] Use default value when there is no payload
[error] [http4s]
```

modules/core/src/smithy4s/http/HttpMediaType.scala
```diff
@@ -36,10 +36,10 @@ object HttpMediaType extends Newtype[String] {
 
     override def default[A]: Option[String] = None
 
-    private def stringMediaType(hints: Hints): String =
+    private def stringMediaType(hints: Hints): Option[String] =
       hints.get(smithy.api.MediaType) match {
-        case Some(mediaType) => mediaType.value
-        case None            => "text/plain"
+        case Some(mediaType) => Some(mediaType.value)
+        case None            => None
```
ghostbuster91: **thought** I am not sure about this fix but I couldn't think of anything better. We cannot render the the `smithy#mediaType` hint on the plain string type.

build.sbt
```diff
@@ -870,7 +870,8 @@ lazy val complianceTests = projectMatrix
         Dependencies.Pprint.core.value,
         Dependencies.Fs2Data.xml.value
       )
-    }
+    },
```
kubukoz: **thought:** I like trailing commas but I don't think we've configured them, so this may have to be reverted

ghostbuster91: done

modules/core/src/smithy4s/http/HttpMediaType.scala
kubukoz: I'm not that familiar with this component of the project, but could this be related to `withRawStringsAndBlobsPayloads`? @Baccata's input would be appreciated here

kubukoz:
Have you checked this with the latest changes on the alloy side (content type related)?

ghostbuster91:
> Have you checked this with the latest changes on the alloy side (content type related)?

Yes, twice.

modules/core/src/smithy4s/http/HttpMediaType.scala
Baccata: In theory this code is not exercised in the context of `SimpleRestJson`, and it's important to retain the existing behaviour, as raw strings and blobs implies that strings ought to be encoded as `text/plain`. 

Without running the tests to see what fails without this change, I'm gonna make an educated guess and assume that a more correct fix would be to condition [this line](https://github.com/disneystreaming/smithy4s/blob/7cc1aa9a1b34d30b90308eedd1edb7a7ddd7e1aa/modules/core/src/smithy4s/http/HttpUnaryClientCodecs.scala#L122) on the `rawStringsAndBlobPayloads` boolean.

Baccata: @ghostbuster91, in case you had missed my comment above ☝️

ghostbuster91: Thanks, I remember about it. I am going to look into this now.

kubukoz:
Scripted test failing due to a mismatch of the Alloy version btw

kubukoz (approved):

Baccata (approved):

PR #1513, opened 2024-05-09
Support for sparse collections in query params

### LINKED ISSUE #1490: Incorrect/strange handling of query parameters without a value
Consider the following URI:

```
?foo=bar&foo&foo=baz
```

On http4s level, the `query.pairs` contains the whole truth: `Vector((foo,Some(bar)), (foo,None), (foo,Some(baz)))`.

However, in the conversion to `smithy4s.http.HttpUri`, we're converting all `None`s to `true`. In 0.18 (0.19 is so far the same):

https://github.com/disneystreaming/smithy4s/blob/b6c789785d6085161de9d1d42469783696360112/modules/http4s-kernel/src/smithy4s/http4s/kernel/package.scala#L187-L196

The outcome of this is that, if you have a `@httpQuery @sparse` list of strings, instead of Nones you'll see `Some(true)`.

IMO this is an issue - we're discarding information about the `None` too soon.

I assume this is not changeable in 0.18 because we'd have to change the URI model:

https://github.com/disneystreaming/smithy4s/blob/b6c789785d6085161de9d1d42469783696360112/modules/core/src/smithy4s/http/HttpUri.scala#L27

unless... we use nulls. Which is probably worse than waiting for 0.19.

denisrosca:
Closes #1490 

- [x] Add support for sparse lists in query parameters
- [ ] Add support for sparse maps in query parameters
- [x] Updated changelog


modules/http4s/test/src/smithy4s/http4s/SparseQueryParametersSuite.scala
```diff
@@ -0,0 +1,109 @@
...
+import io.circe.Json
+import org.http4s._
+import org.http4s.circe.CirceInstances
+import org.http4s.client.Client
+import org.http4s.implicits._
+
+import smithy4s.example.{ServiceWithSparseQueryParams, SparseQueryOutput}
+import weaver._
+
+object SparseQueryParametersSuite extends SimpleIOSuite with CirceInstances {
```
Baccata, line 14: I'd rather not have a bespoke http4s test for this. I'd rather the test was encoded as a compliance test in alloy instead, but that could happen in later PR

modules/core/src/smithy4s/http/HttpUri.scala
```diff
@@ -24,7 +24,7 @@ final case class HttpUri(
       * A sequence of URL-decoded URI segment.
       */
     path: IndexedSeq[String],
-    queryParams: Map[String, Seq[String]],
+    queryParams: Map[String, Seq[Option[String]]],
```
Baccata: I think we should take advantage of the fact we're changing the model to change it to `queryParams: IndexedSeq[(String, Option[String])]`

modules/core/src/smithy4s/http/internals/MetaDecode.scala
```diff
@@ -141,6 +183,17 @@ private[http] sealed abstract class MetaDecode[+A] {
               // TODO add a specialised case for this
               putField(f(statusCode.toString))
           }
+      case (StatusCodeBinding, OptionalStringValueMetaDecode(f)) =>
+        (metadata, putField) =>
+          metadata.statusCode match {
+            case None =>
+              sys.error(
```
Baccata: This should throw a `MetadataError.MissingValueError`

modules/core/src/smithy4s/http/internals/package.scala
```diff
@@ -75,18 +75,23 @@ package object internals {
...
-        }
+        query
+          .split("&")
+          .toList
+          .foldLeft(Map.empty[String, Seq[Option[String]]]) {
+            case (acc, param) =>
+              val (k, v) = param.split("=", 2) match {
+                case Array(key) => (key, None)
+                case Array(key, "") =>
+                  (key, None) // FIXME: denisrosca not sure about this
```
Baccata: probably not needed.

modules/http4s/test/src/smithy4s/http4s/SparseQueryParametersSuite.scala
denisrosca: I tried adding a compliance test in [disneystreaming/alloy#172](https://github.com/disneystreaming/alloy/pull/172/files#diff-21faf7d6abfa2108cb58da42f60f7405cac626e53df890b273b00cd74ef6dadbR9-R27) , but for some reason that I can't comprehend it fails:

```sbt
[info] [http4s] smithy4s.http4s.SimpleRestJsonComplianceSuite
...
[info] [http4s] + alloy.test#SparseQueryParam(client|Request): SparseQueryParam 676ms
[info] [http4s] - alloy.test#SparseQueryParam(server|Request): SparseQueryParam 202ms
[info] [http4s] *************FAILURES**************
[info] [http4s] smithy4s.http4s.SimpleRestJsonComplianceSuite
[error] [http4s] - alloy.test#SparseQueryParam(server|Request): SparseQueryParam 202ms
[error] [http4s]    the result value: Array(List(Some(value = "bar"), Some(value = ""), Some(value = "baz"), Some(value = ""))) was not equal to the expected TestCase value Array(List(Some(value = "bar"), None, Some(value = "baz"), Some(value = ""))). (modules/tests/src/smithy4s/tests/ProtocolComplianceSuite.scala:183)
```

The client part passes, but the server one doesn't and I just can't wrap my head around why 🤯

modules/core/src/smithy4s/http/HttpUri.scala
denisrosca: I changed it to `queryParams:IndexedSeq[(String, Option[String])]`, but only in `HttpUri` and left other places like `Metadata` as a map.

modules/compliance-tests/src/smithy4s/compliancetests/internals/ServerHttpComplianceTestCase.scala
```diff
@@ -66,13 +68,33 @@ private[compliancetests] class ServerHttpComplianceTestCase[
...
+          // passes.
+          // Previously this was done in `parseQueryParams`.
+          case (k, Vector(None)) => k -> Vector(Some(""))
+          case kv                => kv
+        }
+        .toVector
+        .flatMap { case (k, v) =>
+          v.map(k -> _)
+        }
+
```
denisrosca, line 92: This is a bit of a hack to get the compliance tests passing, but will require some discussion.

As far as I can tell, this [aws restJson compliance test](https://github.com/smithy-lang/smithy/blob/6c42bc9d60a681e63c66f4cde33d7a189a1ff9a6/smithy-aws-protocol-tests/model/restJson1/http-query.smithy#L476-L489) expects a query of `example.com/?foo` to be treated as `foo=Some("")` instead of `foo=None` which is a bit problematic for this change.

modules/core/src/smithy4s/http/internals/package.scala
```diff
@@ -75,18 +75,21 @@ package object internals {
...
-            acc.updated(k, acc.getOrElse(k, Seq.empty) :+ v)
-        }
+        query
+          .split("&")
+          .toList
+          .foldLeft(Map.empty[String, Seq[Option[String]]]) {
+            case (acc, param) =>
+              val (k, v) = param.split("=", 2) match {
+                case Array(key)        => (key, None)
+                case Array(key, value) => (key, Some(value))
```
denisrosca, line 89: This bit here means that we treat `?foo=` differently to `?foo` which is probably not correct?

denisrosca:
This is now woefully out of date.

PR #1514, opened 2024-05-09
Fix shape equality in ADT validator

kubukoz:
It was causing false negatives (`doesMemberTargetAdtShape` being false when it should be `true`) for some shapes, when a lot of shapes were added in an unrelated namespace.

Some `ShapeId` constructors use a cache internally, which explains why it'd sometimes still match instance equality, but it's not a mechanism that should be relied upon, and [it will never be one](https://github.com/smithy-lang/smithy/issues/2242).

I've also checked for any other usages of `==` in our Java sources and they are all about primitives/enums, so it's fine.

Not adding tests due to the unpredictable nature of the issue.

- :x: Added unit-tests (for runtime code)
- :x: Added bootstrapped code + smoke tests (when the rendering logic is modified)
- :x: Added build-plugins integration tests (when reflection loading is required at codegen-time)
- :x: Added alloy compliance tests (when simpleRestJson protocol behaviour is expanded/updated)
- :x: Updated dynamic module to match generated-code behaviour
- :x: Added documentation
- [x] Updated changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.18.18
+
+* Fix an issue in the ADT trait validators that would sometimes fail validation while it shouldn't.
+
 # 0.18.17
 
 * Constraints applied to list or map members are now correctly rendered in the generated code.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 0.18.18
 
-* Fix an issue in the ADT trait validators that would sometimes fail validation while it shouldn't.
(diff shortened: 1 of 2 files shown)


kubukoz:
update:

image

the ShapeId factory/cache has a size bound, so I'm pretty sure the users who bumped into this were simply adding enough new shapes to go over this limit before the `@adt`-marked shapes had their ShapeIds created - causing the latter to never get cached, and thus never have referential equality match.

Baccata (approved):

PR #1519, opened 2024-05-13
EncoderK: use type member for `Result`

kubukoz:
~~Targetting 0.18 as it's only source-breaking (like `cats.Parallel` if you're old enough to remember that).~~

- [ ] Added documentation
- [x] Updated changelog

### DIFF
--- a/modules/cats/src/smithy4s/interopcats/SchemaVisitorHash.scala
+++ b/modules/cats/src/smithy4s/interopcats/SchemaVisitorHash.scala
@@ -123,7 +123,8 @@ final class SchemaVisitorHash(
     }
 
     // The encoded form that Hash works against for the eqV method, is a partially-applied curried function.
-    implicit val encoderKInstance = new EncoderK[AltHash, (U => Boolean, Int)] {
+    implicit val encoderKInstance: EncoderK[AltHash] = new EncoderK[AltHash] {
+      type Result = (U => Boolean, Int)
       def apply[A](fa: AltHash[A], a: A): (U => Boolean, Int) = {
         ((u: U) => fa.eqv(a, u), fa.hash(a))
       }
--- a/modules/core/src/smithy4s/capability/EncoderK.scala
+++ b/modules/core/src/smithy4s/capability/EncoderK.scala
@@ -24,7 +24,8 @@ package smithy4s.capability
   * to pre-compile codecs for each union member, and dispatch union instances
   * to a specific codec.
   */
-trait EncoderK[F[_], Result] extends Contravariant[F] {
+trait EncoderK[F[_]] extends Contravariant[F] {
+  type Result
   def apply[A](fa: F[A], a: A): Result
   def absorb[A](f: A => Result): F[A]
   def contramap[A, B](fa: F[A])(f: B => A): F[B] =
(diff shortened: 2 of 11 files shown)


modules/cats/src/smithy4s/interopcats/SchemaVisitorHash.scala
```diff
@@ -127,7 +127,8 @@ final class SchemaVisitorHash(
     }
 
     // The encoded form that Hash works against for the eqV method, is a partially-applied curried function.
-    implicit val encoderKInstance = new EncoderK[AltHash, (U => Boolean, Int)] {
+    implicit val encoderKInstance: EncoderK[AltHash] = new EncoderK[AltHash] {
```
kubukoz, line 126: **thought:** this is a local instance so I didn't annotate it with `Aux`, it'd be useless outside smithy4s

modules/core/src/smithy4s/codecs/Writer.scala
```diff
@@ -110,8 +110,9 @@ object Writer {
   def noop[Message]: Writer[Message, Any] = (message, _) => message
 
   // format: off
-  implicit def writerEncoderK[Message]: EncoderK[Writer[Message, *], Message => Message] =
-    new EncoderK[Writer[Message, *], Message => Message] {
+  implicit def writerEncoderK[Message]: EncoderK.Aux[Writer[Message, *], Message => Message] =
```
kubukoz: **thought:** for public instances like this one, I left `Aux` so that the precise type is still available.

modules/core/src/smithy4s/capability/EncoderK.scala
```diff
@@ -24,16 +24,21 @@ package smithy4s.capability
   * to pre-compile codecs for each union member, and dispatch union instances
   * to a specific codec.
   */
-trait EncoderK[F[_], Result] extends Contravariant[F] {
+trait EncoderK[F[_]] extends Contravariant[F] {
+  type Result
```
kubukoz, line 28: issue: missing changelog entry, I'll add that.

modules/core/src/smithy4s/capability/EncoderK.scala
```diff
@@ -24,16 +24,21 @@ package smithy4s.capability
...
   def apply[A](fa: F[A], a: A): Result
   def absorb[A](f: A => Result): F[A]
   def contramap[A, B](fa: F[A])(f: B => A): F[B] =
     absorb[B](b => apply[A](fa, f(b)))
 }
 
 object EncoderK {
-  implicit def encoderKForFunction[B]: EncoderK[* => B, B] =
-    new EncoderK[* => B, B] {
+  // todo: do we even need this? Arguably you're only going to have one instance per F, and the Result is an implementation detail
```
kubukoz: gotta answer this question

Baccata:
Please target `0.19`, even if it's "only" source-breaking, it's a public construct and I wouldn't want to have it broken in a patch release.

modules/core/src/smithy4s/capability/EncoderK.scala
Baccata: let's not add it until we have a usecase for it

Baccata:
Overall I think we can roll with it (in 0.19)

modules/core/src/smithy4s/capability/EncoderK.scala
kubukoz: users might ;)

I guess we can get rid of `Aux` and remove the type member from public signatures, if someone really needs that type they can downcast while we put the types back.

Baccata (approved):

PR #1522, opened 2024-05-15
Adds Handlers constructs and functions

Baccata:
This adds functionality to facilitate the decoupled implementation of individual operations. Using these constructs allow to split the dependencies needed for the implementation of each operation, and move the coupling closer to the "end-of-the-world".

- [x] Added documentation
- [x] Updated changelog


modules/core/src/smithy4s/EndpointHandler.scala
```diff
@@ -0,0 +1,135 @@
...
+  protected[smithy4s] def lift[Alg[_[_, _, _, _, _]]](
+      service: Service.Aux[Alg, Op]
+  ): PolyFunction5[Op, Kind5[F]#optional]
+
+  final def asService[Alg[_[_, _, _, _, _]]](
+      service: Service.Aux[Alg, Op]
+  ): AsService[Alg, F] =
+    new EndpointHandler.AsServiceImpl[Alg, Op, F](this, service)
+
+  final def orElse(other: EndpointHandler[Op, F]): EndpointHandler[Op, F] =
```
kubukoz: **nitpick:** `orElse` is used in cats/CE for error handling and could be confusing, perhaps we can find a name that isn't, e.g. just `or`?

other ideas off the top of my head: `merge`, `add`, `combine`

modules/core/src/smithy4s/EndpointHandler.scala
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    def partial: Alg[Kind5[F]#optional] =
+      service.fromPolyFunction[Kind5[F]#optional](handler.lift(service))
+  }
+
+  private[smithy4s] def combineAll[Op[_, _, _, _, _], F[_, _, _, _, _]](
+      handlers: EndpointHandler[Op, F]*
+  ): EndpointHandler[Op, F] =
+    Combined(handlers.toVector)
```
kubukoz: **suggestion:** should we account for the possibility that some of these handlers are combined already?

```suggestion
  private[smithy4s] def combineAll[Op[_, _, _, _, _], F[_, _, _, _, _]](
      handlers: EndpointHandler[Op, F]*
  ): EndpointHandler[Op, F] =
    handlers.foldLeft(Combined(Vector()))(_ orElse _)
```

kubukoz:
I like that this is lightweight on changes to existing code

modules/core/src/smithy4s/EndpointHandler.scala
Baccata: [e813d05](https://github.com/disneystreaming/smithy4s/pull/1522/commits/e813d05d519218133d0263383059ea7bb7fd6740)

kubukoz (approved):

PR #1526, opened 2024-05-16
Change Blob equals semantics + fix tests

### LINKED ISSUE #1521: Blob equals issue
Currently `Blob` has three variants, two of which define `equals` in the following way:

```scala
    override def equals(other: Any): Boolean = {
      other.isInstanceOf[ByteBufferBlob] &&
      buf.compareTo(other.asInstanceOf[ByteBufferBlob].buf) == 0
    }
``` 
meaning that `ArraySliceBlob` will never be equal to `ByteBufferBlob`.

In addition to that, `QueueBlob` does not define `equals` and `hashCode` at all.

msosnicki:

fixes #1521

- [ ] Added documentation
- [x] Updated changelog


modules/core/src/smithy4s/Blob.scala
```diff
@@ -173,10 +173,12 @@ object Blob {
     override def size: Int = buf.remaining()
     override def hashCode = buf.hashCode()
 
-    override def equals(other: Any): Boolean = {
-      other.isInstanceOf[ByteBufferBlob] &&
-      buf.compareTo(other.asInstanceOf[ByteBufferBlob].buf) == 0
-    }
+    override def equals(other: Any): Boolean =
```
Baccata: If the point of the PR is to allow Blob constructed from different manners to be equivalent to one another, then the `equals` method should be implemented at the level of the trait, not in each component. 

Moreover, `hashCode` and `equals` should have the same coherence : if you're gonna treat differently structured Blobs are equivalent, then they need to have the same hashCode, therefore the `hashCode` implementation should also move to the trait. 

You can have a look at [fs2.Chunk](https://github.com/typelevel/fs2/blob/f12f5d2fb565dd1c1e7208060bf30ef3b35ace82/core/shared/src/main/scala/fs2/Chunk.scala#L547-L559), which the `Blob` datatype is largely inspired from.

msosnicki: I was thinking about having `equals` in the trait, but never seen such definition and felt something might be off with it. The implementations are the same though, so it makes sense: if I define those two as top level methods, it will automatically eliminate the hashCode issue that I missed here.

Baccata (approved):

PR #1531, opened 2024-05-18
Make examples in design section compile

lenguyenthanh:

- [ ] Added documentation
- [ ] Updated changelog


kubukoz (approved):

kubukoz:
Thanks! It'd be cool to have these actually go through mdoc, but I don't think it's necessary. The design pages are more like blogposts, which are often difficult to mdoc-ify because we keep rewriting some definitions.

PR #1551, opened 2024-05-29
Allow customize sbt opts

lenguyenthanh:
This probably unnecessary, but `sbt` always crash on my machine with the current `.sbtopts`. I have to increase max memory to `16G` (can be less but 6G is not enough).


Baccata:
Mmm, I'm not against adding .sbtopts to `.gitignore` and removing the existing file, but I'd rather we used top-level [environment variables](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#env) in the ci-workflow than having a step copying the sbtopts file.

kubukoz:
I don't have a strong opinion, but personally I've been passing `-J-Xmx16g` explicitly for my local usage.

lenguyenthanh:
added
```yml
env:
  SBT_MAX_HEAP_SIZE: 6G # Max heap size for sbt
```
and use it every where

Baccata:
@lenguyenthanh sorry for the churn, but I actually meant using the `SBT_OPTS` environment variable, to avoid repeating the change in many steps. See here : https://www.scala-sbt.org/1.x/docs/Command-Line-Reference.html#sbt+JVM+options+and+system+properties

lenguyenthanh:
no problem, I'll update it, when I can reach my computer.

lenguyenthanh:
added `SBT_OPTS` env variables with the same values as `.sbtopts.example`.

Baccata (approved):

Baccata:
Thank you !

PR #1553, opened 2024-05-30
Support parsing timestamps without seconds

dhpiggott:

- [ ] Added documentation
- [x] Updated changelog


modules/core/src-js/smithy4s/Timestamp.scala
```diff
@@ -290,70 +290,78 @@ object Timestamp {
```
dhpiggott, line 290: These changes are the same as the JVM changes. See the notes on it.

modules/core/src-jvm-native/smithy4s/Timestamp.scala
```diff
@@ -273,70 +273,81 @@ object Timestamp extends TimestampCompanionPlatform {
...
+      minute
     }
     val second = {
-      val ch0 = s.charAt(pos)
-      val ch1 = s.charAt(pos + 1)
-      if (ch0 < '0' || ch0 > '5' || ch1 < '0' || ch1 > '9') error()
-      pos += 2
-      ch0 * 10 + ch1 - 528 // 528 == '0' * 11
+      val separator = s.charAt(pos)
+      if (separator == ':' && pos + 3 < len) {
```
dhpiggott: This is the main change - we use the presence of `:` after minutes to inform second parsing. That's why this check now happens after we've finished parsing minutes, and why I changed the validation for the preceding terms for consistency.

modules/core/src-jvm-native/smithy4s/Timestamp.scala
```diff
@@ -273,70 +273,81 @@ object Timestamp extends TimestampCompanionPlatform {
...
-      val ch2 = s.charAt(pos + 2)
+      val separator = s.charAt(pos)
+      if (separator != '-' || pos + 3 > len) error()
+      val ch0 = s.charAt(pos + 1)
+      val ch1 = s.charAt(pos + 2)
       val day = ch0 * 10 + ch1 - 528 // 528 == '0' * 11
       if (
-        ch0 < '0' || ch0 > '3' || ch1 < '0' || ch1 > '9' || day == 0 ||
-        (day > 28 && day > maxDayForYearMonth(year, month)) || ch2 != 'T'
+        ch0 < '0' || ch0 > '3' || ch1 < '0' || ch1 > '9' || day < 1 ||
```
dhpiggott: I don't think the day can be negative, so I changed this.

modules/core/src-jvm-native/smithy4s/Timestamp.scala
```diff
@@ -273,70 +273,81 @@ object Timestamp extends TimestampCompanionPlatform {
...
-      val ch1 = s.charAt(pos + 1)
-      val ch2 = s.charAt(pos + 2)
+      val separator = s.charAt(pos)
+      if (separator != 'T' || pos + 3 > len) error()
+      val ch0 = s.charAt(pos + 1)
+      val ch1 = s.charAt(pos + 2)
       val hour = ch0 * 10 + ch1 - 528 // 528 == '0' * 11
       if (
-        ch0 < '0' || ch0 > '2' || ch1 < '0' || ch1 > '9' || hour > 23 || ch2 != ':'
+        ch0 < '0' || ch0 > '2' || ch1 < '0' || ch1 > '9' || hour < 0 || hour > 23
```
dhpiggott: I don't think the hour can be negative, so I added this.

modules/core/src-jvm-native/smithy4s/Timestamp.scala
```diff
@@ -273,70 +273,86 @@ object Timestamp extends TimestampCompanionPlatform {
...
-      val ch1 = s.charAt(pos + 1)
-      val ch2 = s.charAt(pos + 2)
-      if (ch0 < '0' || ch0 > '5' || ch1 < '0' || ch1 > '9' || ch2 != ':')
+      val separator = s.charAt(pos)
+      if (separator != ':' || pos + 3 > len) error()
+      val ch0 = s.charAt(pos + 1)
+      val ch1 = s.charAt(pos + 2)
+      val minute = ch0 * 10 + ch1 - 528 // 528 == '0' * 11
+      if (
+        ch0 < '0' || ch0 > '5' || ch1 < '0' || ch1 > '9' || minute < 0 || minute > 59
```
dhpiggott: Added bounds.

modules/core/src-jvm-native/smithy4s/Timestamp.scala
```diff
@@ -273,70 +273,86 @@ object Timestamp extends TimestampCompanionPlatform {
...
-      if (ch0 < '0' || ch0 > '5' || ch1 < '0' || ch1 > '9') error()
-      pos += 2
-      ch0 * 10 + ch1 - 528 // 528 == '0' * 11
+      val separator = s.charAt(pos)
+      if (separator == ':' && pos + 3 < len) {
+        val ch0 = s.charAt(pos + 1)
+        val ch1 = s.charAt(pos + 2)
+        val second = ch0 * 10 + ch1 - 528 // 528 == '0' * 11
+        if (
+          ch0 < '0' || ch0 > '5' || ch1 < '0' || ch1 > '9' || second < 0 || second > 59
```
dhpiggott: Added bounds.

dhpiggott:
Hello @plokhotnyuk, I believe you're the expert in this area - would you mind taking a look at this change?

modules/core/src-jvm-native/smithy4s/Timestamp.scala
dhpiggott: If in future we needed to make minutes optional, or hours, it would be a relatively straightforward change - we'd just need to add the equivalent check in those blocks and nest the subsequent minutes/seconds computations within that condition.

kubukoz (approved):

modules/core/src-js/smithy4s/Timestamp.scala
```diff
@@ -290,70 +290,86 @@ object Timestamp {
...
-      ch0 * 10 + ch1 - 528 // 528 == '0' * 11
+      minute
     }
     val second = {
-      val ch0 = s.charAt(pos)
-      val ch1 = s.charAt(pos + 1)
-      if (ch0 < '0' || ch0 > '5' || ch1 < '0' || ch1 > '9') error()
-      pos += 2
-      ch0 * 10 + ch1 - 528 // 528 == '0' * 11
+      val separator = s.charAt(pos)
```
Baccata, line 351: I believe most of the changes before this line are unnecessary to the task at hand. Considering this function was written with the philosophy of **absolute maximum performance** by @plokhotnyuk (who's way, WAY more knowledgeable than me when it comes to low-level optimisation on JVM), I want to be respectful of his contributions and will ask that any change that are not strictly necessary to addressing the problem are reverted, unless Andriy himself approves the changes. 

In the context of this function (and the others he's written), his input on the topic trumps any subjective argument any of us may have.

Baccata (changes requested):

modules/core/src-js/smithy4s/Timestamp.scala
dhpiggott: I recognise the changes in question are not 100% necessary to the achieve the original goal of the PR, however two changes I did have to make led to me believing that these additional changes resulted in a more coherent function without introducing any significant performance losses:

1. Alteration/removal of the minimum total length check: I had to either reduce the operand, which would have made the check less useful (since there is a broader range of valid lengths now), or decompose it into checks which run in series and behave relative to the current position. I chose the latter and believe it's a good compromise between safety/readability, and performance, though it does result in a few more additions and comparisons. If we choose to go back to the previous approach I would be curious what the value of retaining the length check at all is, and if there's a reason not to just remove it entirely?

2. Changing the minute parser to remove the `ch2 != ':'` check: I had to remove that term from the checks run within the minute parser, and instead perform it in the second parser to determine whether seconds are expected. Having done that, in the interest of coherence, I changed the preceding blocks. If this has any impact on performance, I don't see how, and would certainly appreciate an explanation if it does - it's the same checks that ran before that run, it's just that they've all shifted down a block for consistency.

@plokhotnyuk, Oli and I discussed this offline earlier and we believe the priority on the request that resulted in this change is such that we can wait until Monday before we ought to ship a fix. I proposed to Oli that we defer to your expertise here rather than me immediately reverting these changes, because I'd rather not do so against my own intuitions - to me that would be a missed opportunity to both a) learn something and b) ship the best version we can. So if you have chance to share your perspective I would really appreciate it :)

plokhotnyuk: Changes looks perfect! 

Sorry for being here too late :)

dhpiggott: Thanks! Just to check, which version do you think we should proceed with - before or after https://github.com/disneystreaming/smithy4s/pull/1553/commits/a161b9e5fa64ec6956d27795a2d2a9f3f478248e?

plokhotnyuk: Minimized (latest) changes are preferable because they save a few CPU cycles in case of errors.

dhpiggott: Cool, thanks.

Baccata (approved):

PR #1557, opened 2024-06-03
Timestamp defaults

### LINKED ISSUE #1535: @default support for timestamps
Currently, the `@default` trait on `Timestamp` fields is not supported ([source](https://github.com/disneystreaming/smithy4s/blob/89e88c972174fa0a3b2853901a52a5a6a4662c6d/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala#L1420))

smithy supports setting the default value that depend on `@timestampFormat`, it would be nice if the codegen supported that too.

I can look into it if it's useful to have this feature

msosnicki:

- [ ] Added documentation
- [x] Updated changelog

fixes #1535


modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -1377,61 +1393,89 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
+      case (node, Primitive.Blob) if node == Node.nullNode =>
+        TypedNode.PrimitiveTN(Primitive.Blob, Array.empty[Byte])
+      case (node, Primitive.Bool) if node == Node.nullNode =>
+        TypedNode.PrimitiveTN(Primitive.Bool, false)
+      case timestamp @ (node, Primitive.Timestamp) =>
+        val format = traits
+          .collectFirst { case t: TimestampFormatTrait =>
+            t.getFormat()
+          }
+          .getOrElse(TimestampFormatTrait.Format.DATE_TIME)
```
kubukoz: **issue:** the default should probably be `EPOCH_SECONDS`, it seems to be the case in e.g. AWS's protocol compliance tests.

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -1377,61 +1393,89 @@ private[codegen] class SmithyToIR(model: Model, namespace: String) {
...
-    case (node, Primitive.Blob) if node == Node.nullNode =>
-      TypedNode.PrimitiveTN(Primitive.Blob, Array.empty[Byte])
-    case (node, Primitive.Bool) if node == Node.nullNode =>
-      TypedNode.PrimitiveTN(Primitive.Bool, false)
-    case (node, Primitive.Timestamp) if node == Node.nullNode =>
-      TypedNode.PrimitiveTN(
-        Primitive.Timestamp,
-        java.time.Instant.ofEpochSecond(0)
+      p: Primitive,
+      traits: List[Trait]
```
Baccata: let's avoid passing the trait and just do something like : `parseEpoch(node).orElse(parseDateTime(node)).orElse(parseHttpDate(node))`

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Thank you!
 # 0.18.21
```
kubukoz, line 11: thought: we're gonna have to move it to 0.18.22

lenguyenthanh: that seems correct, as `0.18.21` is already published

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
msosnicki: Changed, for some reason I was sure that the `date-time` is default, but that's incorrect, just verified with encoders.

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
msosnicki: The `node` that is passed here lacks this information, it is just the value of the `@default` trait. Don't like the `traits` being passed particularly too, but wasn't sure what is the best way to pass the `@timestampFormat`, given that this is used in other places (and other types) too.

CHANGELOG.md
msosnicki: Updated

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
Baccata: My point is that it's okay to not pass the timestamp format and just try all officially supported formats in sequence. I think that timestampFormat is probably the only trait that is taken into consideration by the canonical node representation of data in smithy, and given that the data is already validated by the smithy layer, it's fine to just assume that one of the formats we know is supported will be successful when decoding the value.

Baccata (approved):

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
kubukoz: it used to be :)

kubukoz (approved):

PR #1558, opened 2024-06-04
Use protoc from nix when available

ghostbuster91:

- [ ] Added documentation
- [ ] Updated changelog


flake.nix
```diff
@@ -25,6 +25,8 @@
 
           shellHook = ''
             echo "$welcomeMessage"
+            # TODO use pkgs.lib.getExe once nixpkgs are updated
+            export PROTOC_PATH=${pkgs.protobuf3_21}/bin/protoc
```
ghostbuster91: **thought:** can't use this rn because the definition of `protobuf3_21` in the pinned version of nixpkgs has incorrect metadata and it causes the path to be derived incorrectly. This has been already fixed in the newer version of nixpkgs.

I didn't update nixpkgs because the `nodejs_16` is EOL and is no longer available. Let me know if you don't mind using newer version like 18.

kubukoz: I think we should bump to whatever major Node is being used by GHA

ghostbuster91: It is not about GHE itself but rather the image smithy4s uses for the runner. In this case that is the latest LTS ubuntu which at the moment is 22.04. It uses nodejs 18.x as documented [here](https://github.com/actions/runner-images/blob/ubuntu22/20240526.1/images/ubuntu/Ubuntu2204-Readme.md#language-and-runtime)

flake.nix
```diff
@@ -2,29 +2,30 @@
...
     flake-utils.lib.eachDefaultSystem (system:
       let
         pkgs = import nixpkgs { inherit system; };
-        shellPackages = [
-          "jre"
-          "sbt"
-          "nodejs-16_x"
-          "yarn"
+        shellPackages = with pkgs; [
+          temurin-jre-bin-17
```
ghostbuster91, line 10: **thought:** I replaced default `jre` with the same one that is used on the [CI](https://github.com/disneystreaming/smithy4s/blob/series/0.18/.github/workflows/ci.yml#L63)

flake.nix
```diff
@@ -2,29 +2,30 @@
...
+          nodejs-18_x
+          yarn
+          (pkgs.sbt.override { jre = pkgs.temurin-jre-bin-17; })
```
ghostbuster91, line 13: **thought:** I had to override the `jre` version used by this particular derivation of `sbt` (which is still 1.8.x version) as it was using too recent version of `jre` - 21.

I then decided to completely drop the idea of `shellInputs` being just a list of strings since this required the packages to be defined in two places without any particular benefit. Also the errors from nix evaluator were harder to debug when using strings.

flake.nix
kubukoz: ok good to know, 18 sounds good then

flake.nix
```diff
@@ -2,29 +2,30 @@
...
             Welcome to the smithy4s Nix shell! 👋
             Available packages:
-            ${builtins.concatStringsSep "\n" (map (n : "- ${n}") shellPackages)}
+            ${builtins.concatStringsSep "\n" (map (n : "- ${n.name}") shellPackages)}
           '';
 
           shellHook = ''
             echo "$welcomeMessage"
+            export PROTOC_PATH=${pkgs.lib.getExe pkgs.protobuf3_21}
           '';
```
kubukoz: **suggestion:** 
```suggestion
          shellHook = ''
            echo "$welcomeMessage"
          '';
          PROTOC_PATH = pkgs.lib.getExe pkgs.protobuf3_21;
```

ghostbuster91: I knew there was a better way to do this, thanks!

flake.nix
```diff
@@ -2,31 +2,31 @@
...
             Available packages:
-            ${builtins.concatStringsSep "\n" (map (n : "- ${n}") shellPackages)}
+            ${builtins.concatStringsSep "\n" (map (n : "- ${n.name}") shellPackages)}
           '';
 
           shellHook = ''
             echo "$welcomeMessage"
           '';
-          LD_LIBRARY_PATH = "${pkgs.stdenv.cc.cc.lib}/lib";
+          PROTOC_PATH = pkgs.lib.getExe pkgs.protobuf3_21;
```
ghostbuster91: `LD_LIBRARY_PATH` is actually another way to fix this, but I think `PROTOC_PATH` is more appropriate.

kubukoz (approved):

flake.nix
kubukoz: **suggestion:** extract a variable for the particular protobuf package, so that the PROTOC_PATH is in sync with what's in `nativeBuildInputs`

ghostbuster91: good idea, done.

PR #1559, opened 2024-06-10
Log the args diff when re-running the smithy4s codegen

majk-p:
This PR introduces a change in Smithy4s sbt codegen plugin that would allow users to see what caused the codegen to re-execute. It is done by replacing `sbt.util.Tracked.inputChanged` with a similar implementation that keeps the entire codegen args serialized in cache, and when codegen needs to be rerun, plugin will print the diff of the values to debug log. The diff is calculated using [munit-diff](https://scalameta.org/munit/blog/2024/05/22/release-1.0.0.html#diff-module-extracted-to-a-separate-module).

When running `sbt` with `set logLevel := Level.Debug`, user will notice a diff like this when codegen needs to rerun:

obraz

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/modules/codegen-plugin/src/smithy4s/codegen/CachedTask.scala
+++ b/modules/codegen-plugin/src/smithy4s/codegen/CachedTask.scala
@@ -24,7 +24,7 @@ import sbt.util.Logger
 import sjsonnew._
 import sjsonnew.support.murmurhash.Hasher
 import sjsonnew.support.scalajson.unsafe.Converter
-import sjsonnew.support.scalajson.unsafe.PrettyPrinter
+import sjsonnew.support.scalajson.unsafe.{PrettyPrinter => prettify}
 
 import scala.util.Try
 
@@ -37,8 +37,9 @@ private[codegen] object CachedTask {
       f: (Boolean, I) => O
   ): I => O = { in =>
     def debug(str: String): Unit = logger.debug(s"[smithy4s] $str")
+
     val previousValue = Try(store.read[ValueAndHash[I]]()).toOption
-    val newValueHash = Hasher.hash(in).toOption.getOrElse(-1)
+    val newValueHash = hash(in)
     store.write[ValueAndHash[I]]((in, newValueHash))
 
     previousValue match {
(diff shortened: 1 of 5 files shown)


project/Dependencies.scala
```diff
@@ -139,14 +139,18 @@ object Dependencies {
...
-  class MunitCross(munitVersion: String) {
+  class MunitCross(val munitVersion: String) {
     val core: Def.Initialize[ModuleID] =
       Def.setting("org.scalameta" %%% "munit" % munitVersion)
     val scalacheck: Def.Initialize[ModuleID] =
       Def.setting("org.scalameta" %%% "munit-scalacheck" % munitVersion)
   }
   object Munit extends MunitCross("0.7.29")
   object MunitMilestone extends MunitCross("1.0.0-M6")
+  object MunitV1 extends MunitCross("1.0.0") {
```
Baccata, line 150: may as well use 1.0.0 everywhere (or at the very least replace the milestone with it)

majk-p: Tried that, but 1.0.0 is not published for native 0.4.x. This would involve upgrading native to 0.5.x and I'm not sure about cats-effect support there.

modules/codegen-plugin/src/smithy4s/codegen/CachedTask.scala
```diff
@@ -0,0 +1,72 @@
...
+      case None =>
+        debug(
+          "Could not read previous value from inputs, smithy4s codegen needs to be executed."
+        )
+        store.write[I](in)
+        f(true, in)
+
+      case Some(oldValue) =>
+        (serializeCodegenArgs(oldValue), serializeCodegenArgs(in)) match {
+          case (Some(oldArgs), Some(newArgs)) if (oldArgs != newArgs) =>
```
majk-p: `if (oldArgs != newArgs)` string comparison is not enough, I'll switch to by-hash check

Baccata: yeah I'd rather you compared the json payloads than the json strings

Baccata:
out of curiosity, what's the motivation for this change ?

majk-p:
> out of curiosity, what's the motivation for this change ?

With the extensive use of smithy4s at $work, both my coworkers and myself we have noticed unexpected re-compilation of smithy sources. I'm not sure what caused it yet, I had no luck reproducing it with small projects, but it seems to happen in more complex sbt setups.

A way to debug when and what exactly triggers smithy4s code generation would be a very helpful tool to proceed with the investigation.

modules/codegen-plugin/src/smithy4s/codegen/CachedTask.scala
majk-p: That's done now. We'll first compare `JValue`s - if they differ we're able to produce the diff and log it. If arguments match structurally, hashes are checked to catch changes to the underlying files.

Baccata (approved):

PR #1566, opened 2024-07-05
remove smithy4sRenderOptics setting from plugins 

### LINKED ISSUE #1537: (0.19) Remove optics settings from build tools
https://disneystreaming.github.io/smithy4s/docs/codegen/customisation/optics/

```
If you're using Smithy4s via mill or sbt, then you can enable this functionality with the following keys:

    in mill, task: def smithy4sRenderOptics = true
    in sbt, setting: smithy4sRenderOptics := true
```

Citing @Baccata, we don't want to maintain such configuration on the build tool level, it should rather be configurable via Smithy metadata in additional Smithy source files.

In this case, it already is:

```smithy
metadata smithy4sRenderOptics = true
```

so it should be a straightforward removal.

This will possibly break existing usage, so it should target the `series/0.19` branch.

yisraelU:
and the corresponding sbt-test module
closes #1537 

- [x] Added documentation
- [x] Updated changelog


kubukoz:
Do we still have tests that utilize the optics being generated conditionally?

yisraelU:
> Do we still have tests that utilize the optics being generated conditionally?

I thought so , I will verify

Baccata (approved):

kubukoz:
Confirmed, we have:

- tests in `renderer` that assert on the exact rendered code (btw it would be a nice usecase for https://github.com/siriusxm/snapshot4s), these ensure that we don't just generate optics for everything
- tests that involve using actual generated optics on the types that got them, these ensure that the config isn't ignored

so I think we're good.

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -278,7 +272,6 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
               file,
               s"""$$version: "2"
                  |metadata smithy4sWildcardArgument = "$wildcardArg"
```
kubukoz, line 274: thought: I would consider dropping `smithy4sWildcardArgument` too, we should use `*` and ask people to enable `-Xsource:3` or `-Xsource:3-cross`.

kubukoz (approved):

Issue #1438, opened 2024-03-12
`HttpPayloadError` in server logic gets turned into a 400


kubukoz:
Similar to #1256, but not a regression from 0.17: it appears to have been here all along.

Consider this case: a smithy4s server wrapping a smithy4s client (doesn't have to be the same Service, but it can be, e.g. in a proxy scenario).

Now here's what may happen:

1. The upstream service fails with a non-parseable response. This is normal for uncaught exceptions, or services that aren't using smithy4s.
2. The smithy4s client will rightfully fail its effect with a `HttpPayloadError`.
3. The smithy4s server, instead of re-throwing said error, will return a 400 with a JSON body (the serialized form of `HttpPayloadError`).

This appears incorrect: the error is clearly not a problem of the ultimate caller of the service: ideally, the smithy4s server here would return a 500 (the serialization of which is up for debate). At the moment, it makes it seem like the client has made a mistake in its request. This may be very difficult to debug.

Reproduction: this prints a 400 response with a JSON body. I would expect it to fail on IO level.

```scala
//> using dep "com.disneystreaming.smithy4s::smithy4s-http4s:0.18.12"
//> using dep "com.disneystreaming.smithy4s::smithy4s-tests:0.18.12"
//> using option "-Wunused:imports"
import cats.effect.IO
import cats.effect.IOApp
import cats.implicits._
import org.http4s.Request
import org.http4s.implicits._
import smithy4s.example.PizzaAdminService
import smithy4s.http4s.SimpleRestJsonBuilder
import org.http4s.client.Client
import cats.effect.kernel.Resource
import org.http4s.Response
import org.http4s.Status

object Main extends IOApp.Simple {

  val route =
    SimpleRestJsonBuilder
      .routes(
        SimpleRestJsonBuilder(PizzaAdminService)
          .client[IO](
            Client(_ =>
              Resource.pure(
                Response(
                  status = Status.InternalServerError,
                  body = fs2.Stream.emits("goodbye world".getBytes),
                )
              )
            )
          )
          .make
          .toTry
          .get
      )
      .make
      .toTry
      .get

  override def run: IO[Unit] = route
    .run(Request(uri = uri"/health"))
    .value
    .debug()
    .flatMap(_.traverse_(resp => resp.bodyText.compile.string.debug()))

}
```

I believe this is due to this line:
[...]

PR #1570, opened 2024-07-24
Issue-1438: Fix upstream service error represented as 400

GaryAghedo:
Issue -https://github.com/disneystreaming/smithy4s/issues/1438

**Issue description:**
When an upstream service fails with a non-parseable response (e.g., due to an internal server error), the Smithy4s client wraps this failure in an `HttpPayloadError`. Currently, the Smithy4s server interprets this `HttpPayloadError` as a client-side error and responds with a `400 BadRequest`. This behaviour incorrectly suggests that the error originated from the client's request, making it difficult to debug actual upstream service failures. The correct behaviour should be for the server to respond with a `500 InternalServerError`, indicating an issue with the upstream service rather than the client's request.

**Solution:**

Define a new error type `RawErrorResponse`:

Created a new error type `RawErrorResponse` to represent detailed errors, including a failed decode attempt, originating from upstream services.
Update `HttpResponse`:

Updated `HttpResponse` to handle `RawErrorResponse` by including detailed information about the failed decode attempt when appropriate.

**Test**
Added a test case to `PizzaSpec` to validate the server's response when a `RawErrorResponse` is raised, ensuring that it returns a 500 status code with detailed information about the failed decode attempt.

- [x] Added documentation
- [x] Updated changelog


modules/core/src/smithy4s/http/HttpResponse.scala
```diff
@@ -239,13 +239,39 @@ object HttpResponse {
...
                   F.raiseError(
-                    smithy4s.http.UnknownErrorResponse(
-                      response.statusCode,
-                      response.headers,
-                      bodyBlob.toUTF8String
+                    smithy4s.http.RawErrorResponse(
+                      code = response.statusCode,
+                      headers = response.headers,
+                      body = bodyBlob.toUTF8String,
+                      failedDecodeAttempt = Some(
```
Baccata: this should be None.

modules/core/src/smithy4s/http/HttpContractError.scala
```diff
@@ -71,7 +70,7 @@ object HttpPayloadError {
 }
 
 sealed trait MetadataError extends HttpContractError {
-  import MetadataError._
+  import MetadataError.*
```
Baccata: the changes in this file ought to be reversed.

modules/core/src/smithy4s/http/HttpUnaryServerCodecs.scala
```diff
@@ -195,8 +195,10 @@ object HttpUnaryServerCodecs {
 
           def encodeOutput(o: O) = F.map(base)(outputW.write(_, o))
           def encodeError(e: E) = F.map(base)(errorW.write(_, e))
-          def httpContractErrorEncoder(e: HttpContractError) =
+          def httpContractErrorEncoder(e: HttpContractError) = {
```
Baccata: the changes in this file ought to be reversed.

modules/core/src/smithy4s/http/HttpResponse.scala
```diff
@@ -239,13 +239,39 @@ object HttpResponse {
             val strictResponse = response.copy(body = strictBody)
             F.flatMap(discriminate(strictResponse)) { discriminator =>
               select(discriminator) match {
-                case Some(decoder) => decoder.decode(strictResponse)
+                case Some(decoder) =>
+                  F.flatMap(F.handleErrorWith(decoder.decode(strictResponse)) {
```
Baccata: un-necessary flatmap

modules/core/src/smithy4s/http/RawErrorResponse.scala
```diff
@@ -0,0 +1,22 @@
...
+    code: Int,
+    headers: Map[CaseInsensitive, Seq[String]],
+    body: String,
+    failedDecodeAttempt: Option[FailedDecodeAttempt]
+) extends Throwable {
+  override def getMessage(): String = {
+    val baseMessage = s"status $code, headers: $headers, body:\n$body"
+    failedDecodeAttempt match {
+      case Some(attempt) =>
+        baseMessage + s"\nFailedDecodeAttempt:\n  discriminator: ${attempt.discriminator}\n  contractError: ${attempt.contractError}"
```
Baccata: let's use a proper multi-line string here. Also, use `attempt.contractError.getMessage` instead

kubukoz:
@GaryAghedo for these errors:

> There are files without headers!

You should be able to generate the missing headers with `sbt headerCreateAll`.

GaryAghedo:
Hey @Baccata.
PR ready to be reviewed thanks

modules/core/src/smithy4s/http/RawErrorResponse.scala
```diff
@@ -0,0 +1,43 @@
...
+             |FailedDecodeAttempt:
+             |  discriminator: ${attempt.discriminator}
+             |  contractError: ${attempt.contractError.getMessage}
+               """.stripMargin
+      case None => baseMessage
+    }
+  }
+}
+
+case class FailedDecodeAttempt(
```
Baccata: let's make this a Throwable with an implemented [getMessage](https://github.com/disneystreaming/smithy4s/pull/1570/files#diff-932de23e3329370510b6ff51fe738be5afca381f0c2ca081c0a3de894f38ba96R31-R33). Then in `RawErrorResponse`, you can have `def getCause(): Throwable = failedDecodeAttempt.orNull`

Baccata: Also @ghostbuster91, @GaryAghedo, let's take care of this requirement here: https://github.com/disneystreaming/smithy4s/pull/1575#issuecomment-2296559749

modules/tests/src/smithy4s/tests/PizzaClientSpec.scala
```diff
@@ -112,27 +115,59 @@ abstract class PizzaClientSpec extends IOSuite {
...
+            path = smithy4s.codecs.PayloadPath(List()),
+            expected = "object",
+            message =
+              """Expected JSON object, offset: 0x00000000, buf:
+                |+----------+-------------------------------------------------+------------------+
+                ||          |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f | 0123456789abcdef |
+                |+----------+-------------------------------------------------+------------------+
+                || 00000000 | 6d 61 6c 66 6f 72 6d 65 64 20 62 6f 64 79       | malformed body   |
+                |+----------+-------------------------------------------------+------------------+""".stripMargin
+          )
```
Baccata: I really hate that this is part of the assertion. Let's avoid it : you could add an alternative to the definition of [clientTestForError](https://github.com/disneystreaming/smithy4s/blob/f3ca6635751fe85d29fec21059204f7917b5d28a/modules/tests/src/smithy4s/tests/PizzaClientSpec.scala#L219-L244) to abstract [this comparison away](https://github.com/disneystreaming/smithy4s/blob/f3ca6635751fe85d29fec21059204f7917b5d28a/modules/tests/src/smithy4s/tests/PizzaClientSpec.scala#L238), by doing : 

```scala
  def clientAssertError[E]( // feel free to find a better name
      name: String,
      response: Response[IO],
      assert: E => Unit
  )(implicit loc: SourceLocation) 
```

and have the current `clientTestForError` delegate to that one. Then your test could avoid comparing the error messages, as this is something that comes from an upstream library and really not something that should be tested here.

modules/tests/src/smithy4s/tests/PizzaSpec.scala
```diff
@@ -463,6 +473,123 @@ abstract class PizzaSpec
...
+        path = smithy4s.codecs.PayloadPath(List()),
+        expected = "object",
+        message =
+          """Expected JSON object, offset: 0x00000000, buf:
+            |+----------+-------------------------------------------------+------------------+
+            ||          |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f | 0123456789abcdef |
+            |+----------+-------------------------------------------------+------------------+
+            || 00000000 | 6d 61 6c 66 6f 72 6d 65 64 20 62 6f 64 79       | malformed body   |
+            |+----------+-------------------------------------------------+------------------+""".stripMargin
+      )
```
Baccata: Similar comment to my one [here](https://github.com/disneystreaming/smithy4s/pull/1570/files#r1721788084)

Baccata:
I think it's shaping up well, but there's still a couple things to take care of though. Good work though !

Baccata (approved):
Good work ! Thank you for this !

@kubukoz, wanna do a review as well ?

modules/core/src/smithy4s/http/RawErrorResponse.scala
```diff
@@ -0,0 +1,56 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.http
+
+case class RawErrorResponse(
```
ghostbuster91: **suggestion:** should we make a trait for it and hide the fact that it is a case class to make potential evolution of that class safe & possible?

Baccata: that's a good point actually. It doesn't need to be a trait, but there's [a few things](https://docs.scala-lang.org/overviews/core/binary-compatibility-for-library-authors.html#changing-a-case-class-definition-in-a-backwards-compatible-manner) that should be done.

kubukoz (commented):
looks good to me, let's just make the case class more bincompat-friendly

modules/core/src/smithy4s/http/RawErrorResponse.scala
```diff
@@ -0,0 +1,59 @@
...
+    code: Int,
+    headers: Map[CaseInsensitive, Seq[String]],
+    body: String,
+    private val failedDecodeAttemptField: FailedDecodeAttempt
```
kubukoz: **issue:** @GaryAghedo I think the intention was to make the whole case class's inner structure private as the linked page suggests. The constructor/extractor will have to become `private`, we'd have `withXXX` methods (immutable setters) instead of `copy`, and we should also probably compile under `-Xsource:3` even on Scala 2 if we're not doing that yet.

modules/core/src/smithy4s/http/RawErrorResponse.scala
```diff
@@ -84,6 +84,18 @@ object RawErrorResponse {
   ): RawErrorResponse =
     new RawErrorResponse(code, headers, body, failedDecodeAttempt)
 
+  def unapply(response: RawErrorResponse): Option[
```
ghostbuster91: **question:** do we want to keep the ability to use the `RawErrorResponse` as an extractor even though it might not contain all fields? 

Imo better just to make it private as suggested in the guide.

modules/core/src/smithy4s/http/RawErrorResponse.scala
```diff
@@ -0,0 +1,121 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.http
+
+case class RawErrorResponse(
```
ghostbuster91: **issue:** primary constructor should be private
```scala
case class RawErrorResponse private(...
```

modules/core/src/smithy4s/http/RawErrorResponse.scala
ghostbuster91: > we should also probably compile under `-Xsource:3` even on Scala 2 if we're not doing that yet.

So currently we are not doing that. 

There was apparently an attempt to enable it in https://github.com/disneystreaming/smithy4s/pull/555/files#r1006283612 but it is unclear why it was left commented out. 

~Do you know what is the reason behind following statement from the bincompat guide:~
> ~in Scala 2, you have to add the compiler option `-Xsource:3`~

~?~

I made an issue to enable it https://github.com/disneystreaming/smithy4s/issues/1584

modules/core/src/smithy4s/http/RawErrorResponse.scala
Baccata: Agreed

GaryAghedo: The reason I made it non private because I got this error
```
private method unapply in object RawErrorResponse is never used
[error] [core]   private def unapply(response: RawErrorResponse): Option[
[error] [core]               ^
[error] [core] No warnings can be incurred under -Werror.
```

kubukoz: In that case, you can silence the warning: https://www.scala-lang.org/2021/01/12/configuring-and-suppressing-warnings.html#configuring-warnings

modules/core/src/smithy4s/http/RawErrorResponse.scala
```diff
@@ -0,0 +1,124 @@
...
+    private val codeField: Int,
+    private val headersField: Map[CaseInsensitive, Seq[String]],
+    private val bodyField: String,
+    private val failedDecodeAttemptField: FailedDecodeAttempt
+) extends Throwable {
+
+  def code: Int = codeField
+  def headers: Map[CaseInsensitive, Seq[String]] = headersField
+  def body: String = bodyField
+  def failedDecodeAttempt: FailedDecodeAttempt = failedDecodeAttemptField
```
ghostbuster91: **question:** do we have to keep fields private and expose those getters? why not un-private fields in the case class def?

GaryAghedo: Good question

modules/tests/src/smithy4s/tests/PizzaClientSpec.scala
```diff
@@ -111,29 +113,49 @@ abstract class PizzaClientSpec extends IOSuite {
...
+      expect(error.failedDecodeAttempt.isInstanceOf[DecodingFailure])
+    }
   )
 
-  private def unknownResponse(
-      code: Int,
-      headers: Map[String, String],
-      body: String
-  ): UnknownErrorResponse =
-    UnknownErrorResponse(
```
kubukoz, line 132: **issue:** `UnknownErrorResponse` is no longer used, we should remove it.

In addition, the docs in `modules/docs/markdown/03-protocols/04-simple-rest-json/03-client.md` need updating because they still refer to `UnknownErrorResponse` - they should now mention `RawErrorResponse`, but there might be other differences too, e.g. we could mention when a particular `FailedDecodeAttempt` shows up

kubukoz: and finally, we'll need a CHANGELOG entry for this PR 😅  it's part of the checklist in the PR description.

kubukoz: ok, dealt with those

PR #1573, opened 2024-08-08
Improve ADT validators' performance

### LINKED ISSUE #1319: AdtMemberTraitValidator is a bit slow
In a model with a lot of member shapes, `AdtMemberTraitValidator` is increasingly slow. This is a problem because validators run on every `didChange` notification in the LSP, which can slow down your editor experience a lot (e.g. formatting will have to wait for this as well, assuming the current implementation).

I see one obvious culprit: in `AdtValidatorCommon`, every time `getReferences` is called, it iterates over all member shapes of the model. This happens for each `@adt` and each `@adtMember` target.

Here's the relevant code:

```java
	private static List<Reference> getReferences(Model model, Shape adtMemberShape, Shape adtParent) {
		return model.getMemberShapes().stream().flatMap(memberShape -> {
			boolean doesMemberTargetAdtShape = memberShape.getTarget() == adtMemberShape.getId();
			boolean isMemberShapeInDesiredTarget = memberShape.getContainer().equals(adtParent.toShapeId());
			if (doesMemberTargetAdtShape) {
				return Stream.of(new Reference(isMemberShapeInDesiredTarget, memberShape.getContainer()));
			} else {
				return Stream.empty();
			}
		}).collect(Collectors.toList());
	}
```

Perhaps we can have a quick win by pre-computing a map of member shapes whose target passes certain criteria. At first sight I'm not sure how to formulate that criteria in an efficient way though, so this needs some more careful inspection.

For reference, the models I'm concerned about have roughly 25k member shapes, and that number isn't going down anytime soon.

At the moment, on my machine it takes ~1000-1200ms for a single run of `AdtMemberTraitValidator`. `AdtTraitValidator` is noticeably faster (~100ms). This is calculated with `nanoTime()`.

Some profiler output:

image

(yeah, it shows less than 600ms here, which is much better, but a suspicious score nevertheless)

kubukoz:
Closes #1319. Also actually prevents ADTs from being mixins, which shouldn't have been possible in the first place.

Tested on a model with 92k shapes (including 78k members) using 288 `@adtMember`s and 33 `@adt`s.

Before: 11+ seconds

image

The original implementation was spending most its time in the `getReferences` method:

image

which had linear time complexity on the number of member shapes in the model, which let's be honest - are the most common type of shape by far in any realistic model.

After: <200ms

image

In the new implementation, there's no sharing of code (the validation patterns are different, so I didn't think it's worth sharing), and we utilize things like `getShapesWithTrait` and selectors to get better perf. Probably linear in the number of shapes annotated with each of the ADT traits.

- [ ] Added documentation
- [x] Updated changelog


kubukoz:
So far, so good - `AdtTraitValidator` is all but gone from the flame graph.

image

Baseline: https://github.com/disneystreaming/smithy4s/issues/1319#issuecomment-2704541220

kubukoz:
Got the AdtMemberTraitValidator too.

image

modules/protocol-tests/test/src/smithy4s/api/validation/AdtMemberTraitValidatorSpec.scala
```diff
@@ -93,18 +93,71 @@ object AdtMemberTraitValidatorSpec extends FunSuite {
...
-          "test#MyUnion must have exactly one member targeting test#struct"
+          "This shape must be referenced by test#MyUnion because of its smithy4s.meta#adtMember trait"
         )
         .build()
     )
     expect(result == expected)
   }
 
-  test("return error when structure is targeted by multiple unions") {
+  test("return no error when there are duplicate non-adtMember members") {
```
kubukoz: **thought:** I'm low key thinking to rewrite all these tests to use the IDL. I think the shape builder DSL is too weird to be practical, and it's certainly not readable.

As an added advantage, the selectors we have on the traits would actually run if we used the model assembler instead of a builder.

Baccata: Yeah agreed

modules/protocol/src/smithy4s/meta/validation/AdtTraitValidator.java
```diff
@@ -21,41 +21,67 @@
...
- * structures that the union targets must NOT be used within any other union.
- *
- * Also checks that the structures targeted are not empty (they must have at
- * least one member).
+ * All the members of an ADT union must be structures.
+ * Also, each such structure can only be referenced once in the whole model (from said union).
  */
 public final class AdtTraitValidator extends AbstractValidator {
+  private final Selector adtTargetedMemberSelector = Selector.parse(
+    String.format(":test(> member > :in(:root([trait|%s] > member > structure)))", AdtTrait.ID.toString())
```
Baccata: Can you put a comment on what this selector does ? The selector syntax is quite obtuse to me

Baccata: Can you put a comment on what this selector does ? The selector syntax is quite obtuse to me

kubukoz: will do

kubukoz: will do

modules/protocol/resources/META-INF/smithy/smithy4s.meta.smithy
```diff
@@ -24,13 +24,13 @@ structure packedInputs {}
...
 /// Example usage: @adtMember(MyUnion)
 @trait(selector: "structure :not([trait|error])")
-@idRef(failWhenMissing: true, selector: "union")
+@idRef(failWhenMissing: true, selector: "union :not([trait|mixin])")
 string adtMember
 
 /// Implies that all members of the union are annotated with the `adtMember` trait.
 /// Further signals that the `sealed trait` for this adt will extend the traits
 /// defined by any mixins that are present on all of the adt members.
-@trait(selector: ":test(union :test(> member > structure), :not([trait|mixin]))")
```
kubukoz, line 33: **thought:** The old behavior allowed using `adt` on mixins. I feel like this wasn't intentional, but it's hard to say without a test.

modules/protocol-tests/test/src/smithy4s/api/validation/AdtMemberTraitValidatorSpec.scala
kubukoz: it's actually less code, nice

modules/protocol/resources/META-INF/smithy/smithy4s.meta.smithy
kubukoz: gonna assume it wasn't intentional

Baccata (approved):

PR #1574, opened 2024-08-09
add support for jsonUnknown trait

benoitlouy:
Implement support for `jsonUnknown` trait defined in https://github.com/disneystreaming/alloy/pull/180

### DIFF
--- a/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
+++ b/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
@@ -1436,8 +1436,10 @@ private[smithy4s] class SchemaVisitorJCodec(
       (field, jLabel, default)
     }
 
-  private def nonPayloadStruct[Z](
-      fields: LabelledFields[Z],
+  private def structRetainUnknownFields[Z](
+      allFields: LabelledFields[Z],
+      knownFields: LabelledFields[Z],
+      fieldsForUnknown: LabelledFields[Z],
       structHints: Hints
   )(
       const: Vector[Any] => Z,
@@ -1445,21 +1447,13 @@ private[smithy4s] class SchemaVisitorJCodec(
   ): JCodec[Z] =
     new JCodec[Z] {
 
-      private[this] val (fieldsForUnknown, knownFields) = fields.partition {
-        case (field, _, _) => isForJsonUnknown(field)
-      }
-
-      private[this] val keepUnknown = fieldsForUnknown.nonEmpty
-
-      private[this] val isLabelForUnknownFields: String => Boolean =
-        if (keepUnknown)
-          fieldsForUnknown.map { case (_, label, _) =>
-            label
-          }.toSet
-        else _ => false
+      private[this] val isLabelForUnknownFields =
+        fieldsForUnknown.map { case (_, label, _) =>
+          label
+        }.toSet
 
       private[this] val handlers =
-        new util.HashMap[String, Handler](fields.length << 1, 0.5f) {
+        new util.HashMap[String, Handler](knownFields.length << 1, 0.5f) {
           knownFields.foreach { case (field, jLabel, _) =>
             put(jLabel, fieldHandler(field))
           }
@@ -1480,8 +1474,7 @@ private[smithy4s] class SchemaVisitorJCodec(
           cursor: Cursor,
           in: JsonReader
       ): scala.collection.Map[String, Any] => Z = {
-        val unknownValues =
-          if (keepUnknown) ListBuffer[(String, Document)]() else null
+        val unknownValues = ListBuffer[(String, Document)]()
         val buffer = new util.HashMap[String, Any](handlers.size << 1, 0.5f)
         if (in.isNextToken('{')) {
           if (!in.isNextToken('}')) {
@@ -1490,10 +1483,8 @@ private[smithy4s] class SchemaVisitorJCodec(
               val key = in.readKeyAsString()
               val handler = handlers.get(key)
               if (handler eq null) {
-                if (keepUnknown) {
-                  val value = documentJCodec.decodeValue(cursor, in)
-                  unknownValues += (key -> value)
-                } else in.skip()
+                val value = documentJCodec.decodeValue(cursor, in)
+                unknownValues += (key -> value)
               } else handler(cursor, in, buffer)
               in.isNextToken(',')
             }) ()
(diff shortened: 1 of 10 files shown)


modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -1445,14 +1480,21 @@ private[smithy4s] class SchemaVisitorJCodec(
           cursor: Cursor,
           in: JsonReader
       ): scala.collection.Map[String, Any] => Z = {
+        val unknownValues =
+          if (keepUnknown) ListBuffer[(String, Document)]() else null
```
Baccata: I appreciate the effort to minimise allocations :+1: . I would personally have gone for distinct `JCodec` implementations based on whether the structure has a `@jsonUnknown` field or not, which does induce some copy/pasting, but completely shaves off some checks on the critical path.

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -1467,14 +1509,32 @@ private[smithy4s] class SchemaVisitorJCodec(
...
-                else default
-              } else value
+              if (isLabelForUnknownFields(jsonLabel)) {
+                if (unknownValue == null) {
+                  if (default == null) {
+                    if (f.isRequired) Map.empty
+                    else None
+                  } else default
+                } else {
+                  if (f.isRequired) unknownValue else Some(unknownValue)
```
Baccata: you cannot make the assumption that `!required` implies `scala.Option` : we're planning on relaxing our Schema model to allow for other ways of modelling optionality (to cater for `java.util.Optional`, for instance). 

Moreover, maybe the user applies other type of customisations to what ends up annotated with `jsonUnknown`. Maybe they have a bijection from Document to circe's Json in place, etc. 

So what you need to do instead is to compile the `Document.Decoder` associated to the field that is annotated with `@jsonUnkown`, in order to run the unknownValue (converted to `Document`) through it to get a value that'll be guaranteed to be of the correct type, without having to make any assumption.

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -1467,14 +1509,32 @@ private[smithy4s] class SchemaVisitorJCodec(
...
-              if (value == null) {
-                if (default == null)
-                  cursor.requiredFieldError(jsonLabel, jsonLabel)
-                else default
-              } else value
+              if (isLabelForUnknownFields(jsonLabel)) {
+                if (unknownValue == null) {
+                  if (default == null) {
+                    if (f.isRequired) Map.empty
+                    else None
```
Baccata: See my comment [here](https://github.com/disneystreaming/smithy4s/pull/1574/files#r1713392500)

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
benoitlouy: There are now 2 `JCodec` implementations to separately handle structs with a `jsonUnknown` member and ones without.

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
benoitlouy: This has been addressed and the decoding now uses what you described.

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -1415,7 +1436,122 @@ private[smithy4s] class SchemaVisitorJCodec(
...
+                  } else value
+
+                case Some(docDecoder) =>
+                  if (unknownValue == null) {
+                    if (default == null) {
+                      docDecoder
+                        .decode(Document.obj())
+                        .getOrElse(
+                          throw new RuntimeException(
+                            s"${cursor.getPath(Nil)} Failed translating a Document.DObject to the type targeted by ${f.label}."
```
Baccata, line 1529: use `in.decodeError` instead, to throw the exception

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -1415,7 +1436,122 @@ private[smithy4s] class SchemaVisitorJCodec(
...
+                          )
+                        )
+                    } else default
+                  } else {
+                    docDecoder
+                      .decode(unknownValue)
+                      .getOrElse(
+                        throw new RuntimeException(
+                          s"${cursor.getPath(Nil)} Failed translating a Document.DObject to the type targeted by ${f.label}."
```
Baccata, line 1538: [ditto](https://github.com/disneystreaming/smithy4s/pull/1574/files#r1714988906)

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -1384,27 +1388,44 @@ private[smithy4s] class SchemaVisitorJCodec(
...
+      case m: Map[_, _] =>
+        m.foreach { case (label: String, value: Document) =>
+          writeLabel(label, out)
+          documentJCodec.encodeValue(value, out)
+        }
+      case Some(m: Map[_, _]) =>
+        m.foreach { case (label: String, value: Document) =>
+          writeLabel(label, out)
+          documentJCodec.encodeValue(value, out)
+        }
```
Baccata: Sorry, I should have signposted this bit as well, but this is falling under the similar notion to what I was describing [here](https://github.com/disneystreaming/smithy4s/pull/1574#discussion_r1713392500) : 

1. You cannot make the assumption that a smithy map necessarily translates to a Scala map 
2. You cannot make the assumption that the absence of `required` translates to Scala Option 
3. You cannot even make the assumption that at this level, you're dealing with `Documents`

Therefore you need to pre-compile the `Document.Encoder` and encode the field value against it, and pattern-match against the produced document to check it's a DObject, in which case you run the exact same `foreach` you have there

Baccata:
This is shaping up well ! This code is quite tricky and it's a very valuable PR, thank you very much for taking this on 👍

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
benoitlouy: This makes sense, thanks for the directions.

Baccata (approved):
Giving a pre-emptive approval for this, for @lewisjkl to consider when the upstream work (alloy) has been merged and released

lewisjkl (approved):
This looks good to me as well, great work on this one!

Baccata (approved):

PR #1575, opened 2024-08-16
Add info about failing discrimintator to UnknownErrorResponse

ghostbuster91:

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/modules/core/src/smithy4s/http/HttpErrorSelector.scala
+++ b/modules/core/src/smithy4s/http/HttpErrorSelector.scala
@@ -41,8 +41,8 @@ object HttpErrorSelector {
   def apply[F[_]: Covariant, E](
       maybeErrorSchema: Option[ErrorSchema[E]],
       compiler: CachedSchemaCompiler[F]
-  ): HttpDiscriminator => Option[F[E]] = maybeErrorSchema match {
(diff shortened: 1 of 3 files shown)


Baccata:
In all frankness, if we're gonna target 0.19, I think Gary's code change to [remove the type-parametrisation of the discriminator](https://github.com/disneystreaming/smithy4s/pull/1570/files#diff-607170a944b8ad9642f00c60495d837707cff84b84fe718fd54d0c79633eaa9bR233) is better to take care of the requirement of "what did we attempt" without needing to amend other functions. 

I'd rather you built on top of that, and changed the [FailedDecodeAttempt](https://github.com/disneystreaming/smithy4s/pull/1570/files#diff-932de23e3329370510b6ff51fe738be5afca381f0c2ca081c0a3de894f38ba96R40-R43) model to something like : 

``` 
sealed trait FailedDecodeAttempt 
object FailedDecodeAttempt {
  case class UnrecognisedDiscriminator(discriminator : HttpDiscriminator) extends FailedDecodeAttempt 
  case class DecodingFailure(...) extends FailedDecodeAttempt // what's in the current `FailedDecodeAttempt`
}
```

Therefore, I'm closing this PR.

PR #1576, opened 2024-08-30
fix: Add missing nano part when encoding timestamp to document

ghostbuster91:

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
+++ b/modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
@@ -99,9 +99,10 @@ class DocumentEncoderSchemaVisitor(
       hints
         .get(TimestampFormat)
         .getOrElse(TimestampFormat.EPOCH_SECONDS) match {
-        case DATE_TIME     => ts => DString(ts.format(DATE_TIME))
-        case HTTP_DATE     => ts => DString(ts.format(HTTP_DATE))
-        case EPOCH_SECONDS => ts => DNumber(BigDecimal(ts.epochSecond))
+        case DATE_TIME => ts => DString(ts.format(DATE_TIME))
+        case HTTP_DATE => ts => DString(ts.format(HTTP_DATE))
+        case EPOCH_SECONDS =>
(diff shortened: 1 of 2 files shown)


modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
```diff
@@ -101,7 +101,7 @@ class DocumentEncoderSchemaVisitor(
         .getOrElse(TimestampFormat.EPOCH_SECONDS) match {
         case DATE_TIME     => ts => DString(ts.format(DATE_TIME))
         case HTTP_DATE     => ts => DString(ts.format(HTTP_DATE))
-        case EPOCH_SECONDS => ts => DNumber(BigDecimal(ts.epochSecond))
+        case EPOCH_SECONDS => ts => DNumber(BigDecimal(s"${ts.epochSecond}.${ts.nano}"))
```
ghostbuster91: **thought:** I am not really a fan of constructing bigDecimal from string in here but I couldn't think of anything better.

Baccata: `BigDecimal(ts.epochSecond + 10e-9 * ts.nano)` ?

ghostbuster91: I should've made myself clearer. I thought about this approach, but I didn't want to use because of potential precision loss issue. 

Consider following:
```scala
val now = Instant.now()
val epochSecond: Long = now.getEpochSecond
val nano: Int = 999_999_999 // almost 1 second

val result: Double = epochSecond + 1e-9 * nano
println(s"Result with potential precision loss: ${BigDecimal(result)}")

val accurateBigDecimal =
  BigDecimal(epochSecond) + (BigDecimal(nano) * BigDecimal(10).pow(-9))

println(s"Accurate result : ${accurateBigDecimal}")

```
Which prints:
```
Result with potential precision loss: 1725278847
Accurate result : 1725278846.999999999
```

This is still probably a better idea than going through string parsing :)

Baccata (approved):

PR #1580, opened 2024-09-06
Update sbt to 1.10

### LINKED ISSUE #1538: Chore: update sbt in the project and scripted tests
In the `build.properties` scattered around the codebase, update sbt.version to whatever the latest stable version is (at time of writing, 1.10.0).

kubukoz:
Closes #1538.

- [ ] Added documentation
- [ ] Updated changelog


kubukoz:
I've bisected the header problem down to https://github.com/sbt/sbt/releases/v1.9.7, I suspect it's this change: https://github.com/sbt/sbt/pull/7404

the current state is that `headerLicense` is unset in most projects.

Baccata:
Well let's either set it or disable the header plugin altogether on irrelevant projects

modules/dynamic/test/src-jvm/smithy4s/dynamic/ModelWrapper.scala
```diff
@@ -35,7 +35,7 @@ class ModelWrapper(val model: Model) {
   private final implicit class JavaStreamOps[A](
       stream: java.util.stream.Stream[A]
   ) {
-    def toList: List[A] = stream.collect(Collectors.toList()).asScala.toList
+    def asList: List[A] = stream.collect(Collectors.toList()).asScala.toList
```
kubukoz, line 38: ok so I don't fully understand why it started to be a problem _now_ but it is indeed a problem for me.

I was getting "method `exists` doesn't exist on `java.util.List`" type errors. This is because in newer JDK versions there's a `toList()` method on `Stream` - it's been there since Java 16.

Thing is, we were already building on Java 17. I'm using Java 21 now as a default, so Metals probably has that as well, but my sbt in this project is still the one from the Nix flake, using Java 17.

I guess this could be a change in behavior between sbt or scala patch versions - didn't investigate more, but I find it slightly spooky that we can use newer Java features even though we have `-release 8` in our scalac options.

Baccata (approved):

modules/codegen-plugin/src/sbt-test/codegen-plugin/wildcard-config/test
```diff
@@ -12,10 +12,12 @@
...
 
-> ++2.13.10 root/compile
+> ++2.13.15
+> root/compile
 $ exists target/scala-2.13/src_managed/main/smithy/generated-metadata.smithy
 $ exists target/scala-2.13/resource_managed/main/META-INF/smithy/generated-metadata.smithy
 
-> ++3.3.0 root/compile
+> ++3.3.0
+> root/compile
```
kubukoz, line 21: I have no idea what happened here.

modules/dynamic/test/src-jvm/smithy4s/dynamic/ModelWrapper.scala
kubukoz: I get it now. The `temurin-jre-bin-17` derivation from the Nix flake doesn't have `ct.sym` in it, so Scalac just uses all the available JDK methods instead of just the ones matching the target version. Strange choice but ok.

PR #1588, opened 2024-09-24
Generate `transform` as a final def for cats-tagless friendliness

### LINKED ISSUE #1587: Minor annoyance: can't derive cats-tagless typeclasses for `Alg[_[_]]`s
given the generated

```scala
type FooApi[F[_]] = smithy4s.kinds.FunctorAlgebra[FooApiGen, F]
```

and a call to e.g. `cats.tagless.FunctorK.derived[FooApi]`, you'll get this error:

```scala
no implicit values were found that match type cats.tagless.FunctorK[
  [G[_$7]] =>>
    smithy4s.Transformation.PartiallyApplied[
      foo.bar.FooApiGen[
        [I, E, O, SI, SO] =>> G[O]]
    ]
]
```

If the `transform` method is removed from the algebra, it's fine again.

Perhaps we should be generating an extension method (or nothing) instead of a member method? The implementation of `transform` is just `Transformation.of[FooApiApiGen[F]](this)`, after all.

kubukoz:
Closes #1587 

- [ ] Added documentation
- [x] Updated changelog

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/internals/ModelLoader.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/ModelLoader.scala
@@ -51,7 +51,7 @@ private[codegen] object ModelLoader {
     val modelsInJars = deps.flatMap { file =>
       Using.resource(
         // Note: On JDK13+, the second parameter is redundant.
-        FileSystems.newFileSystem(file.toPath(), null)
+        FileSystems.newFileSystem(file.toPath(), null: ClassLoader)
       ) { jarFS =>
         val p = jarFS.getPath("META-INF", "smithy", "manifest")
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,14 @@ When adding entries, please treat them as if they could end up in a release any 
 
 Thank you!
 
+# 0.18.25
+
+* Make the `transform` method in generated `*Gen` algebras final. This should make it possible to derive e.g. `FunctorK` instances in cats-tagless automatically (see [#1588](https://github.com/disneystreaming/smithy4s/pull/1588)).
+
 # 0.18.24
 
 * Adds missing nanoseconds in Document encoding of EPOCH_SECOND timestamps
-* Add support for `alloy#jsonUnknown`, allowing structures to capture unknown JSON fields in one of their members. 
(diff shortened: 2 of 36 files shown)


modules/codegen/src/smithy4s/codegen/internals/ModelLoader.scala
```diff
@@ -51,7 +51,7 @@ private[codegen] object ModelLoader {
     val modelsInJars = deps.flatMap { file =>
       Using.resource(
         // Note: On JDK13+, the second parameter is redundant.
-        FileSystems.newFileSystem(file.toPath(), null)
+        FileSystems.newFileSystem(file.toPath(), null: ClassLoader)
```
kubukoz, line 54: **thought:** this was causing trouble for my sbt (@ghostbuster91 also reported seeing this):

```
[error] [codegen2_12] /Users/kubukoz/projects/smithy4s/modules/codegen/src/smithy4s/codegen/internals/ModelLoader.scala:54:21: ambiguous reference to overloaded definition,
[error] [codegen2_12] both method newFileSystem in class FileSystems of type (x$1: java.nio.file.Path, x$2: java.util.Map[String, _])java.nio.file.FileSystem
[error] [codegen2_12] and  method newFileSystem in class FileSystems of type (x$1: java.nio.file.Path, x$2: ClassLoader)java.nio.file.FileSystem
[error] [codegen2_12] match argument types (java.nio.file.Path,Null)
[error] [codegen2_12]         FileSystems.newFileSystem(file.toPath(), null)
[error] [codegen2_12]                     ^
[error] [codegen2_12] one error found
[error] [codegen2_12] (codegen2_12 / Compile / compileIncremental) Compilation failed
```

I'm not sure why it's failing now, but I'm sure it was fine a couple months ago. Anyway, doesn't seem important enough to investigate.

Baccata (approved):

modules/codegen/src/smithy4s/codegen/internals/ModelLoader.scala
kubukoz: explanation for future generations: https://github.com/disneystreaming/smithy4s/pull/1580#discussion_r1817473402

tl;dr the "JRE only" distribution of temurin used in the Nix flake was missing the ct.sym file, meaning Scalac would see all JDK methods, ignoring its `-release` flag (the target JVM version) - so it was seeing JDK methods that it shouldn't have seen.

Issue #1532, opened 2024-05-19
AWS: Account API ListRegions not working (unknown host)


kubukoz:
```scala
package smithy4sdemo

import cats.effect.IO
import cats.effect.IOApp
import com.amazonaws.account.Account
import org.http4s.ember.client.EmberClientBuilder
import smithy4s.aws.AwsClient
import smithy4s.aws.AwsEnvironment
import smithy4s.aws.kernel.AwsRegion

object Main extends IOApp.Simple {

  def run: IO[Unit] = EmberClientBuilder.default[IO].build.use { c =>
    AwsEnvironment.default(c, AwsRegion.US_EAST_1).use { env =>
      AwsClient(Account, env).use { api =>
        api.listRegions().debug().void
      }
    }
  }

}
```

Version: 0.18.19

Build:

```scala
val root = project
  .in(file("."))
  .settings(
    scalaVersion := "3.4.2",
    libraryDependencies ++= Seq(
      "com.disneystreaming.smithy4s" %% "smithy4s-aws-http4s" % smithy4sVersion.value,
      "org.http4s" %% "http4s-ember-server" % "0.23.27",
      "org.http4s" %% "http4s-ember-client" % "0.23.27",
    ),
    fork := true,
    smithy4sAwsSpecs += AWS.account,
  )
  .enablePlugins(Smithy4sCodegenPlugin)
```

Exception:

```
[info] DEBUG: Errored: java.net.UnknownHostException: ListRegions.us-east-1.amazonaws.com: nodename nor servname provided, or not known
[error] java.net.UnknownHostException: ListRegions.us-east-1.amazonaws.com: nodename nor servname provided, or not known
[error]         at java.base/java.net.Inet6AddressImpl.lookupAllHostAddr(Native Method)
[error]         at java.base/java.net.InetAddress$PlatformNameService.lookupAllHostAddr(InetAddress.java:934)
[error]         at java.base/java.net.InetAddress.getAddressesFromNameService(InetAddress.java:1543)
[error]         at java.base/java.net.InetAddress$NameServiceAddresses.get(InetAddress.java:852)
[error]         at java.base/java.net.InetAddress.getAllByName0(InetAddress.java:1533)
[error]         at java.base/java.net.InetAddress.getAllByName(InetAddress.java:1385)
[error]         at java.base/java.net.InetAddress.getAllByName(InetAddress.java:1306)
[error]         at java.base/java.net.InetAddress.getByName(InetAddress.java:1256)
[...]

PR #1596, opened 2024-09-27
Quick attempt at resolving AWS signing problems

### LINKED ISSUE #1568: Sagemaker: invalid signature (Credential should be scoped to correct service)
To reproduce

([repo with `sbt run`-able setup](https://github.com/kubukoz/demos/tree/smithy4s-sagemaker-InvalidSignatureException))

1. Generate code for AWS sagemaker, for example the `ListEndpoints` operation
2. Create a client for that service and call it with any parameters
3. See the error, `{"__type":"InvalidSignatureException","message":"Credential should be scoped to correct service: 'sagemaker'. "}`:

```
[info] {"__type":"InvalidSignatureException","message":"Credential should be scoped to correct service: 'sagemaker'. "}
[error] smithy4s.http.UnknownErrorResponse: status 400, headers: HashMap(content-length -> List(112), x-amzn-requestid -> List(5da79a17-1241-4404-a539-2efdd85b8a85), connection -> List(close), date -> List(Mon, 15 Jul 2024 23:38:37 GMT), content-type -> List(application/x-amz-json-1.1)), body:
[error] {"__type":"InvalidSignatureException","message":"Credential should be scoped to correct service: 'sagemaker'. "}
[error]         at smithy4s.http.UnknownErrorResponse$.apply(UnknownErrorResponse.scala:19)
[error]         at smithy4s.http.HttpResponse$Decoder$$anon$4.decode$$anonfun$1$$anonfun$1(HttpResponse.scala:248)
[error]         at sequence @ smithy4s.interopcats.package$$anon$1.zipMapAll(package.scala:31)
[error]         at sequence @ smithy4s.interopcats.package$$anon$1.zipMapAll(package.scala:31)
[error]         at sequence @ smithy4s.interopcats.package$$anon$1.zipMapAll(package.scala:31)
[error]         at map @ smithy4s.interopcats.package$$anon$1.zipMapAll(package.scala:31)
[error]         at flatMap @ smithy4s.interopcats.package$$anon$1.flatMap(package.scala:33)
[error]         at flatMap @ smithy4s.interopcats.package$$anon$1.flatMap(package.scala:33)
[error]         at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
[error]         at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
[error]         at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
[error]         at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
[error]         at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
[error]         at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
[error]         at clear @ org.http4s.client.middleware.Retry$.retryLoop$1(Retry.scala:90)
[error]         at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
[error]         at rethrow$extension @ fs2.Compiler$Target.compile$$anonfun$1(Compiler.scala:157)
[error]         at run$ @ hello.Main$.run(Main.scala:11)
[error] Nonzero exit code returned from runner: 1
```

kubukoz:
This is more of a random attempt rather than a methodically applied implementation, but early testing has proven it to be successful.

Potentially closes #1568, #1532 (works on my machine but needs investigation into whether this is actually a more correct implementation). It would be amazing if there are protocol tests for this...

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/modules/aws-http4s/src/smithy4s/aws/AwsClient.scala
+++ b/modules/aws-http4s/src/smithy4s/aws/AwsClient.scala
@@ -71,7 +71,11 @@ object AwsClient {
 
       def baseRequest(endpoint: OperationSchema[_, _, _, _, _]): F[HttpRequest[Blob]] = {
         awsEnv.region.map { region =>
-          val endpointPrefix = awsService.endpointPrefix.getOrElse(endpoint.id.name)
+          val endpointPrefix =
+            awsService.endpointPrefix
+              .orElse(awsService.arnNamespace.map(_.value))
+              .getOrElse(endpoint.id.name)
+
           val baseUri = HttpUri(
             scheme = HttpUriScheme.Https,
             host = s"$endpointPrefix.$region.amazonaws.com",
--- a/modules/aws-http4s/src/smithy4s/aws/internals/AwsSigning.scala
+++ b/modules/aws-http4s/src/smithy4s/aws/internals/AwsSigning.scala
@@ -60,7 +60,10 @@ private[aws] object AwsSigning {
   ): Client[F] => Client[F] = {
     val endpointPrefix = serviceHints
       .get(_root_.aws.api.Service)
-      .flatMap(_.endpointPrefix)
+      .flatMap { awsService =>
+        awsService.endpointPrefix
+          .orElse(awsService.arnNamespace.map(_.value))
+      }
       .getOrElse(serviceId.name)
       .toLowerCase()
 


modules/aws-http4s/src/smithy4s/aws/AwsClient.scala
```diff
@@ -71,7 +71,11 @@ object AwsClient {
 
       def baseRequest(endpoint: OperationSchema[_, _, _, _, _]): F[HttpRequest[Blob]] = {
         awsEnv.region.map { region =>
-          val endpointPrefix = awsService.endpointPrefix.getOrElse(endpoint.id.name)
+          val endpointPrefix =
+            awsService.endpointPrefix
+              .orElse(awsService.arnNamespace.map(_.value))
```
kubukoz, line 76: **question:** should this be `sigv4` rather than `arnNamespace`?

should we even be looking at `endpointPrefix`? `sigv4` is supposed to match `arnNamespace` but smithy says it's a SHOULD (see documentation trait of the `sigv4` trait members).

For what it's worth, both Sagemaker and Location work if sigv4 is used instead, in both locations of this PR.

kubukoz: we should eventually just go through all services on https://docs.aws.amazon.com/general/latest/gr/aws-service-information.html one-by-one and validate that we compute the right urls...

Baccata: Yeah, it's a bit misleading because of the `endpointPrefix` value name, but really we should be using the `sigv4` value here.

kubukoz: ok I'll change to sigv4 then, and maybe run through the services to see if it's present on all of them (and what the fallback should be).

modules/aws-http4s/src/smithy4s/aws/AwsClient.scala
```diff
@@ -66,22 +66,47 @@ object AwsClient {
...
+      * https://smithy.io/2.0/aws/aws-core.html#arnnamespace
+      *
+      * This only covers the `{prefix}.{region}.amazonaws.com` pattern; FIPS,
+      * dual-stack, non-standard partitions (`amazonaws.com.cn`, GovCloud) and
+      * operation-level endpoints all require evaluating
+      * `smithy.rules#endpointRuleSet`, which we do not do.
+      */
+    private val hostPrefix: String =
+      awsService.endpointPrefix
+        .orElse(awsService.arnNamespace.map(_.value))
```
kubukoz: Worth flagging where each of these rules actually comes from, since only some of it is documented.

**Documented.** The two primary rules are stated in the trait definitions themselves: [`endpointPrefix`](https://smithy.io/2.0/aws/aws-core.html#endpointprefix) resolves as `{endpointPrefix}.{region}.{dnsSuffix}` and is authoritative for the host; [`aws.auth#sigv4`'s `name`](https://smithy.io/2.0/aws/aws-auth.html#aws-auth-sigv4-trait) is "the signature version 4 service signing name to use in the credential scope" and is authoritative for signing.

**Inferred.** The fallback chains are not. `endpointPrefix` has no documented default, so there is no spec text saying what to do when it is absent. `arnNamespace` is chosen here because it is constrained to `^[a-z0-9.\-]{1,63}$` and [defaults to the lowercased service shape name](https://smithy.io/2.0/aws/aws-core.html#arnnamespace), which makes it DNS-safe — and because for the Account API (#1532) it empirically produces the host that the model's own `smithy.rules#endpointTests` assert. That is a good signal, not a guarantee.

The authoritative algorithm is `smithy.rules#endpointRuleSet`, which smithy4s does not evaluate. So this is a best-effort reconstruction of it for the `{prefix}.{region}.amazonaws.com` case, and the fallbacks are where it is most likely to diverge for some service we have not looked at. #2006 proposes checking that systematically against `endpointTests`.

PR #1602, opened 2024-09-28
Avoid collisions in typed node namespaces, add missing uppercase in alt nodes, remove ShapeTag from ADT alt members

### LINKED ISSUE #1601: ADT unions used as traits have unescaped names
Similar to #1592.

```smithy
$version: "2"

namespace demo

@smithy4s.meta#adt
@trait
union class {
    u: Struct
}

structure Struct {}

@class(u: {

})
string MyShape
```

```scala
[error] ./demo/MyShape.scala:13:10
[error] an identifier expected, but 'class' found
[error]     demo.class.Struct(),
[error]          ^^^^^
```

This doesn't happen with normal unions.

kubukoz:
Closes #1601 and the issue found [in the comments](https://github.com/disneystreaming/smithy4s/issues/1601#issuecomment-2380834208).

- [ ] Added documentation
- [x] Updated changelog


sampleSpecs/reservednames.smithy
```diff
@@ -146,3 +151,17 @@ structure Scala3ReservedKeywords {
     export: String
     enum: String
 }
+
+@smithy4s.meta#adt
+@trait
```
kubukoz, line 156: **thought:** maybe we should just make this illegal? using trait and adt like that

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -1617,7 +1617,10 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
       line"${ref.show}.${altName.capitalize}Case.widen".write
 
     case AltTN(_, _, AltValueTN.ProductAltTN(alt)) =>
-      alt.runDefault.write
+      // The `widen` is necessary in Scala 3.
+      // Without it, there are two schemas (the struct and the union) that match the type
+      // for the conversion to Hints.Binding.
+      line"${alt.runDefault}.widen".write
```
Baccata, line 1622: Okay, I don't mind the .widen call to help the compiler, but I think the ADT member's schema should not be `implicit`, ever. Would you mind addressing it in this PR ? 

That's something I find regrettable with regards to traditional implicit derivation solutions in Scala 2 : if a case class is an ADT member, it shouldn't be accompanied with its own typeclass instances, as the semantics of a program change drastically based on whether the type is widened or not.

sampleSpecs/reservednames.smithy
Baccata: I don't think it should be illegal, the semantics of these two being completely orthogonal. I do however think that dynamic bindings will just solve all these little gotchas.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
kubukoz: I agree, will give it a shot

kubukoz: done, this was apparently some previously-missed case, as we do this in all the other cases (just not 0-field product structs in a union)

kubukoz: also removing the widen

kubukoz: I rushed too much, apparently the widen is still necessary?

kubukoz: Widen is back, for Scala 2 :D

Baccata (approved):

PR #1603, opened 2024-10-01
Closes #1539 kebab caseificiation

### LINKED ISSUE #1539: Fix kebab-caseificiation
See https://github.com/disneystreaming/smithy4s/blame/89e88c972174fa0a3b2853901a52a5a6a4662c6d/modules/decline/src/core/commons.scala#L31

Running this on a string that starts with a lowercase erases the first letter.

dwalend:

- [ ] Added documentation
- [x] Updated changelog

### DIFF
--- a/modules/decline/test/src/smithy4s/decline/core/commonsSpec.scala
+++ b/modules/decline/test/src/smithy4s/decline/core/commonsSpec.scala
@@ -5,10 +5,14 @@ import weaver._
 object commonsSpec extends FunSuite {
 
   test("toKebabCase base case") {
-    expect(commons.toKebabCase("StartsWithUpperCase") == "starts-with-upper-case")
+    expect(
+      commons.toKebabCase("StartsWithUpperCase") == "starts-with-upper-case"
+    )
   }
 
   test("toKebabCase starts with lower case") {
-    expect(commons.toKebabCase("startsWithLowerCase") == "starts-with-lower-case")
+    expect(
+      commons.toKebabCase("startsWithLowerCase") == "starts-with-lower-case"
+    )
   }
 }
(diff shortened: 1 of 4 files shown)


dwalend:
It doesn't look like github linked this PR back to https://github.com/disneystreaming/smithy4s/issues/1539 . What's the right way to make that happen?

kubukoz:
> It doesn't look like github linked this PR back to #1539 . What's the right way to make that happen?

I think you need a hash in the PR title. Ideally, you can say "Closes #1539" and it'll link them so that when the PR gets merged, the issue gets closed.

modules/decline/test/src/smithy4s/decline/core/commonsSpec.scala
```diff
@@ -0,0 +1,14 @@
+package smithy4s.decline.core
+
+import weaver._
+
+object commonsSpec extends FunSuite {
```
kubukoz: nitpick: in tests, we usually name objects starting with an uppercase letter... although the object being tested is `commons`, I would probably rename this to match other suites

kubukoz (commented):
Looks like the formatter is complaining, can you run `sbt scalafmt` and commit the results? Thanks!

modules/decline/test/src/smithy4s/decline/core/commonsSpec.scala
dwalend: And in a PR about camel-case no less.

dwalend:
> > It doesn't look like github linked this PR back to #1539 . What's the right way to make that happen?
> 
> I think you need a hash in the PR title. Ideally, you can say "Closes #1539" and it'll link them so that when the PR gets merged, the issue gets closed.

No luck. I wonder if I had to get it right the first time.

Baccata (approved):

Baccata:
Thank you !

dwalend:
> Thank you !

Glad to. Next let me see if I can pick up something with a little more smithy4s in it.

PR #1606, opened 2024-10-04
Lenient numeric decoding

Baccata:
This adds opt-in logic to allow decoding numeric values from JSON strings. 
This also exposes a new method on the SimpleRestJsonBuilder to configure the underlying JSON codecs, without having to duplicate the smart-builder methods every time we add new options. 

- [x] Added documentation
- [x] Updated changelog

### DIFF
--- a/modules/http4s/src/smithy4s/http4s/SimpleRestJsonBuilder.scala
+++ b/modules/http4s/src/smithy4s/http4s/SimpleRestJsonBuilder.scala
@@ -35,7 +35,7 @@ class SimpleRestJsonBuilder private (
       simpleRestJsonCodecs
     ) {
 
-  @deprecated(message = "Use builder methods instead", since = "0.18.25")
+  @deprecated(message = "Use .withXXX methods instead", since = "0.18.25")
   def this(
       maxArity: Int,
       explicitDefaultsEncoding: Boolean,
--- a/modules/http4s/src/smithy4s/http4s/SimpleRestJsonBuilder.scala
+++ b/modules/http4s/src/smithy4s/http4s/SimpleRestJsonBuilder.scala
@@ -17,7 +17,17 @@
 package smithy4s
 package http4s
 
-object SimpleRestJsonBuilder extends SimpleRestJsonBuilder(1024, false, true)
+import smithy4s.json.Json
+import smithy4s.json.JsonPayloadCodecCompiler
+
+object SimpleRestJsonBuilder
+    extends SimpleRestJsonBuilder(
+      new internals.SimpleRestJsonCodecs(
+        jsonCodecs = Json.payloadCodecs,
+        explicitDefaultsEncoding = false,
+        hostPrefixInjection = true
+      )
+    )
 
 class SimpleRestJsonBuilder private (
     simpleRestJsonCodecs: internals.SimpleRestJsonCodecs
@@ -25,6 +35,7 @@ class SimpleRestJsonBuilder private (
       simpleRestJsonCodecs
     ) {
 
+  @deprecated(message = "Use builder methods instead", since = "0.18.25")
   def this(
       maxArity: Int,
       explicitDefaultsEncoding: Boolean,
@@ -32,7 +43,12 @@ class SimpleRestJsonBuilder private (
   ) =
     this(
       new internals.SimpleRestJsonCodecs(
-        maxArity,
+        Json.payloadCodecs
+          .withJsoniterCodecCompiler(
+            Json.jsoniter
+              .withMaxArity(maxArity)
+              .withExplicitDefaultsEncoding(explicitDefaultsEncoding)
+          ),
         explicitDefaultsEncoding,
         hostPrefixInjection
       )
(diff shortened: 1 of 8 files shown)


modules/http4s/src/smithy4s/http4s/SimpleRestJsonBuilder.scala
```diff
@@ -17,47 +17,67 @@
...
+      )
+    )
 
 class SimpleRestJsonBuilder private (
     simpleRestJsonCodecs: internals.SimpleRestJsonCodecs
 ) extends SimpleProtocolBuilder[alloy.SimpleRestJson](
       simpleRestJsonCodecs
     ) {
 
+  @deprecated(message = "Use builder methods instead", since = "0.18.25")
```
kubukoz: **suggestion:** as we don't have an "actual" builder, maybe

```suggestion
  @deprecated(message = "Use .withXXX methods instead", since = "0.18.25")
```

modules/json/src/smithy4s/json/JsoniterCodecCompiler.scala
```diff
@@ -78,6 +78,12 @@ trait JsoniterCodecCompiler extends CachedSchemaCompiler[JsonCodec] {
     */
   def withLenientTaggedUnionDecoding: JsoniterCodecCompiler
 
+  /**
+    * Enables lenient decoding of numeric values, where numbers may be carried by JSON strings
+    * as well as JSON numbers.
+    */
+  def withLenientNumericDecoding: JsoniterCodecCompiler
```
kubukoz, line 85: **question:** shall we just call it lenient decoding, in case we add more cases of leniency? such as, case-insensitivity (which might not be doable if there are members with names like `foo` and `FOO`, but might be possible in many cases)

kubukoz: and we'd just mention number handling because it happens to be the only thing for now

modules/json/test/src/smithy4s/json/SchemaVisitorJCodecCustomisationTests.scala
```diff
@@ -178,4 +202,39 @@ class SchemaVisitorJCodecCustomisationTests extends FunSuite {
 
     expect(result.f.isNegInfinity)
   }
+
+  test("decoding JSON string as a Float") {
```
kubukoz, line 206: **question:** something I don't get is: this works because InfinitySupport is on. Why does InfinitySupport have to imply that any other "string number" is valid?

modules/json/src/smithy4s/json/JsoniterCodecCompiler.scala
Baccata: I'd rather have more granular options than a "one size fits all" option.

modules/json/test/src/smithy4s/json/SchemaVisitorJCodecCustomisationTests.scala
Baccata: that's a sacrifice I'm agreeing to, in order to avoid having to implement N permutations, as the implementations are not compositional. 

I think agreeing to be generally lenient when you want infinity enabled is an okay compromise, considering it applies a similar "let's see if the next character is a quote" piece of logic.

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -80,153 +84,194 @@ private[smithy4s] class SchemaVisitorJCodec(
...
-      new JCodec[Int] {
-        def expecting: String = "int"
+    private abstract class NumericJCodec[A] extends JCodec[A] {
+      def decodeJsonNumber(cursor: Cursor, in: JsonReader): A
+      // Allows numerics to be received as JSON Strings
+      def decodeJsonString(cursor: Cursor, in: JsonReader): A
 
-        def decodeValue(cursor: Cursor, in: JsonReader): Int = in.readInt()
+      final def decodeValue(cursor: Cursor, in: JsonReader): A =
+        if (allowJsonStringNumerics) {
```
lewisjkl, line 93: It's just one if statement, but I think this could be moved off the hot path?

lewisjkl (approved):
Just the one potential thing in the comment below, otherwise looks good 👍

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
Baccata: Good question. In this case, it's hard : the problem is that due to the imperative nature of jsoniter, the logic is not compositional. If we want to get checks on the hotpath at all cost, that in an ideal world we'd have one class per combination of flags 

1. disallowed strings but allowed infinity 
2. allowed strings and allowed infinity 
3. disallowed strings and disallowed infinity 
4. allowed strings but disallowed infinity 

Right now I'm accepting this boolean check as a compromise to avoid having to repeat inherently error-prone bits of logic in 4 different implementations. I'm not too happy about it but I'd rather live with it.

lewisjkl: I think I'm following, for some reason I thought that you could also do at the class level something like:

```scala
private val decodeV: (Cursor, JsonReader) => A = if (allowJsonStringNumerics) {
  (cursor, in) => // ...
} else {
  (cursor, in) => // ...
}

final def decodeValue(cursor: Cursor, in: JsonReader): A = this.decodeV(cursor, in)
```

Baccata: I mean, maybe you're seeing something I'm not seeing. I'd be delighted if you attempted and managed to avoid the boolean check on the critical path without inducing a lot of repetition in the code 😄

Baccata: also there's the question of : is a method dispatch faster than a boolean check in this context ? not sure

lewisjkl: Yeah good point, it probably isn't worth worrying about, was mostly just asking to clarify my understanding

PR #1607, opened 2024-10-07
Add convenience truncate method to Timestamp type

### LINKED ISSUE #1598: Decrease timestamp resolution
Long story short the change in https://github.com/disneystreaming/smithy4s/pull/1576 prompted me to review in details `timestamp` datatype and `timestampFormat` trait. 

According to [smithy documentation](https://smithy.io/2.0/spec/simple-types.html#timestamp):
> 2.12. timestamp
A timestamp represents an instant in time in the proleptic Gregorian calendar, independent of local times or timezones. Timestamps support an allowable date range between midnight January 1, 0001 CE to 23:59:59.999 on December 31, 9999 CE, with a temporal resolution of 1 millisecond. This resolution and range ensures broad support across programming languages and guarantees compatibility with [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339.html).
> 
> 2.12.1. Timestamp serialization and deserialization
> The serialization format of a timestamp is an implementation detail that is determined by a [protocol](https://smithy.io/2.0/spec/protocol-traits.html#protocoldefinition-trait) and or [timestampFormat trait](https://smithy.io/2.0/spec/protocol-traits.html#timestampformat-trait). The format of a timestamp MUST NOT have any effect on the types exposed by tooling to represent a timestamp value.
> Protocols and timestampFormat traits MAY support temporal resolutions other than 1 millisecond. For example, the http-date timestamp format supports only seconds and forbids fractional precision. Modelers need to be aware of these limitations when defining timestamps to avoid an unintended loss of precision.
> The use of timestamps outside the allowable range risk not interoperating correctly across Smithy implementations; deserializers that encounter timestamps outside the allowable range SHOULD fail to deserialize the value.

and [timestamp format docs](https://smithy.io/2.0/spec/protocol-traits.html#timestamp-formats)

For `date-time` and `epoch-seconds`:
>  Values that are more granular than millisecond precision SHOULD be truncated to fit millisecond precision.

For `http-date`:
> A deserializer that encounters an http-date timestamp with fractional precision SHOULD fail to deserialize the value (for example, an HTTP server SHOULD return a 400 status code).

Base on the above I propose following changes:
1. for `timestampFormat=http-date` the value should be truncated to second precision during the serialization
2. for `timestampFormat=date-time | epoch-seconds` the value should be truncated to millisecond precision during the serialization

ghostbuster91:
This closes #1598 

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/modules/core/src-jvm-native/smithy4s/Timestamp.scala
+++ b/modules/core/src-jvm-native/smithy4s/Timestamp.scala
@@ -39,6 +39,16 @@ case class Timestamp private (epochSecond: Long, nano: Int)
 
   def conciseDate: String = formatToString(2)
 
+  /**
+    * @return a copy of this timestamp with a milisecond resolution
+    */
+  def truncateNanos: Timestamp = copy(nano = (nano / 1000000) * 1000000)
+
+  /**
+    * @return a copy of this timestamp with a second resolution
+    */
+  def truncateMillis: Timestamp = copy(nano = 0)
+
   override def toString: String = format(TimestampFormat.DATE_TIME)
 
   private[this] def formatToString(internalFormat: Int): String = {
--- a/modules/core/src-jvm-native/smithy4s/Timestamp.scala
+++ b/modules/core/src-jvm-native/smithy4s/Timestamp.scala
@@ -40,14 +40,14 @@ case class Timestamp private (epochSecond: Long, nano: Int)
   def conciseDate: String = formatToString(2)
 
   /**
-    * @return a copy of this timestamp with a milisecond resolution
+    * @return a copy of this timestamp truncated to a miliseconds precision
     */
-  def truncateNanos: Timestamp = copy(nano = (nano / 1000000) * 1000000)
+  def truncateToMillis: Timestamp = copy(nano = (nano / 1000000) * 1000000)
 
   /**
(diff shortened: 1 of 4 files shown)


modules/core/src-jvm-native/smithy4s/Timestamp.scala
```diff
@@ -39,6 +39,16 @@ case class Timestamp private (epochSecond: Long, nano: Int)
 
   def conciseDate: String = formatToString(2)
 
+  /**
+    * @return a copy of this timestamp with a milisecond resolution
+    */
+  def truncateNanos: Timestamp = copy(nano = (nano / 1000000) * 1000000)
```
Baccata: I'm finding the naming pretty confusing. I think it should be `truncateToSeconds` and `truncateToMillis` to make it clearer. For instance `truncateToSeconds` would truncate anything more precise than seconds.

Baccata: Alternatively : `toSecondPrecision` and `toMillsecondPrecision` ?

ghostbuster91: renamed as `truncateToSeconds` etc.

Baccata (approved):

PR #1610, opened 2024-10-11
Protocol in update lsp config

dwalend:

- [ ] Added documentation
- [x] Updated changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ Thank you!
 * Fixes an issue in which union members targetting Unit would fail to compile when used as traits (see [#1600](https://github.com/disneystreaming/smithy4s/pull/1600)).
 * Make the `transform` method in generated `*Gen` algebras final. This should make it possible to derive e.g. `FunctorK` instances in cats-tagless automatically (see [#1588](https://github.com/disneystreaming/smithy4s/pull/1588)).
 * Fixes commons.toKebabCase() sometimes drops the first letter (see [#1603](https://github.com/disneystreaming/smithy4s/pull/1603)).
-* Addes `com.disneystreaming.smithy4s:smithy4s-protocol dependency` to the generation of `smithy-build.json` in the `smithy4sUpdateLspConfig` tasks of the codegen plugins (see [#1610](https://github.com/disneystreaming/smithy4s/pull/1610)).
+* Adds `com.disneystreaming.smithy4s:smithy4s-protocol dependency` to the generation of `smithy-build.json` in the `smithy4sUpdateLspConfig` tasks of the codegen plugins (see [#1610](https://github.com/disneystreaming/smithy4s/pull/1610)).
 
 # 0.18.24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Thank you!
 * Fixes an issue in which union members targetting Unit would fail to compile when used as traits (see [#1600](https://github.com/disneystreaming/smithy4s/pull/1600)).
 * Make the `transform` method in generated `*Gen` algebras final. This should make it possible to derive e.g. `FunctorK` instances in cats-tagless automatically (see [#1588](https://github.com/disneystreaming/smithy4s/pull/1588)).
 * Fixes commons.toKebabCase() sometimes drops the first letter (see [#1603](https://github.com/disneystreaming/smithy4s/pull/1603)).
+* Added com.disneystreaming.smithy4s:smithy4s-protocol dependency to mill LSP plugin (see [#1610](https://github.com/disneystreaming/smithy4s/pull/1610)).
 
 # 0.18.24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ Thank you!
 * Fixes an issue in which union members targetting Unit would fail to compile when used as traits (see [#1600](https://github.com/disneystreaming/smithy4s/pull/1600)).
 * Make the `transform` method in generated `*Gen` algebras final. This should make it possible to derive e.g. `FunctorK` instances in cats-tagless automatically (see [#1588](https://github.com/disneystreaming/smithy4s/pull/1588)).
 * Fixes commons.toKebabCase() sometimes drops the first letter (see [#1603](https://github.com/disneystreaming/smithy4s/pull/1603)).
-* Added com.disneystreaming.smithy4s:smithy4s-protocol dependency to mill LSP plugin (see [#1610](https://github.com/disneystreaming/smithy4s/pull/1610)).
+* Addes `com.disneystreaming.smithy4s:smithy4s-protocol dependency` to the generation of `smithy-build.json` in the `smithy4sUpdateLspConfig` tasks of the codegen plugins (see [#1610](https://github.com/disneystreaming/smithy4s/pull/1610)).
 
 # 0.18.24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Thank you!
 
 * Optimises the conversion of empty smithy4s.Blob to fs2.Stream, to avoid performance degradation in Ember (see [#1609](https://github.com/disneystreaming/smithy4s/pull/1609))
 * Adds utility types for working with endpoint handlers (see [#1612](https://github.com/disneystreaming/smithy4s/pull/1612))
-* * Adds `com.disneystreaming.smithy4s:smithy4s-protocol` dependency to the generation of `smithy-build.json` in the `smithy4sUpdateLSPConfig` tasks of the codegen plugins (see [#1610](https://github.com/disneystreaming/smithy4s/pull/1610)).
+* Adds `com.disneystreaming.smithy4s:smithy4s-protocol` dependency to the generation of `smithy-build.json` in the `smithy4sUpdateLSPConfig` tasks of the codegen plugins (see [#1610](https://github.com/disneystreaming/smithy4s/pull/1610)).
 
 # 0.18.25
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Thank you!
 * Fixes an issue in which union members targetting Unit would fail to compile when used as traits (see [#1600](https://github.com/disneystreaming/smithy4s/pull/1600)).
 * Make the `transform` method in generated `*Gen` algebras final. This should make it possible to derive e.g. `FunctorK` instances in cats-tagless automatically (see [#1588](https://github.com/disneystreaming/smithy4s/pull/1588)).
 * Fixes commons.toKebabCase() sometimes drops the first letter (see [#1603](https://github.com/disneystreaming/smithy4s/pull/1603)).
+* Added com.disneystreaming.smithy4s:smithy4s-protocol dependency to mill LSP plugin (see [#1610](https://github.com/disneystreaming/smithy4s/pull/1610)).
```
kubukoz: **issue:** it's not just Mill, the sbt plugin also benefits from this.

it's not exactly the LSP plugin either, just the codegen plugin. How about:

```suggestion
* Addes `com.disneystreaming.smithy4s:smithy4s-protocol dependency` to the generation of `smithy-build.json` in the `smithy4sUpdateLspConfig` tasks of the codegen plugins (see [#1610](https://github.com/disneystreaming/smithy4s/pull/1610)).
```

also, we'll need a new patch version - 0.18.25 is out.

modules/codegen-plugin/src/sbt-test/codegen-plugin/update-lsp-config/build.sbt
```diff
@@ -2,23 +2,43 @@ import sbt.io.IO
 val subproj = project
 
 val subproj2 = project.enablePlugins(Smithy4sCodegenPlugin)
-
+//            "com.disneystreaming.smithy4s:smithy4s-protocol:${smithy4sVersion.value}"
```
kubukoz: **issue:** leftover?

modules/codegen-plugin/src/sbt-test/codegen-plugin/update-lsp-config/build.sbt
```diff
@@ -2,23 +2,43 @@ import sbt.io.IO
...
 val root = project
   .in(file("."))
   .aggregate(subproj, subproj2)
   .settings(
     TaskKey[Unit]("checkSmithyBuild") := {
       val generated = IO.readLines(file(".") / "smithy-build.json")
-      val expected = IO.readLines(file(".") / "expected.json")
+      val expected = s"""{
```
kubukoz: **suggestion:** I see why you had to inline it (the `smithy4sVersion` variable part), but I'm not too happy about it.

What if we invented a simple templating language so that the JSON file could still be its own file, with a placeholder like

```json
"com.disneystreaming.smithy4s:smithy4s-protocol:${SMITHY4S_VERSION}"
```

and then we'd do a `.replace("${SMITHY4S_VERSION}", smithy4sVersion.value)` upon loading that string?

also, that file wasn't deleted in the PR so it probably still exists as a leftover.

modules/codegen/test/src/smithy4s/codegen/internals/SmithyBuildSpec.scala
```diff
@@ -201,4 +201,32 @@ final class SmithyBuildSpec extends munit.FunSuite {
     assertEquals(Set.empty[os.FilePath], actual.sources.toSet)
     assertEquals(Set.empty[SmithyBuildPlugin], actual.plugins.toSet)
   }
+
+  test("generate json with SmithyBuildJson.toJson") {
+    val actual = SmithyBuildJson.toJson(
```
kubukoz, line 206: **question:** why do we even have two? `SmithyBuildJson.toJson` and `SmithyBuild.writeJson`? @Baccata

modules/codegen/src/smithy4s/codegen/SmithyBuildJson.scala
```diff
@@ -25,6 +25,10 @@ import smithy4s.codegen.internals.SmithyBuildMavenRepository
 import scala.collection.immutable.ListSet
 
 private[codegen] object SmithyBuildJson {
+
+  val protocolDependency =
+    s"com.disneystreaming.smithy4s:smithy4s-protocol:${BuildInfo.version}"
```
kubukoz: **nitpick:** wouldn't hurt us to add `organization` and `protocolArtifactName` as buildinfo keys. Totally fine if you don't, though.

modules/codegen/test/src/smithy4s/codegen/internals/SmithyBuildSpec.scala
Baccata: I think we should remove the `SmithyBuild` one (it's probably a leftover from a merge mistake). 

I think the reason is that we used to rely on the smithy-build pojo, but it was not flexible enough for what we needed to do. So we went straight-to-json I think. Better look at the history though

modules/codegen-plugin/src/sbt-test/codegen-plugin/update-lsp-config/build.sbt
Baccata: I agree with Jakub.

CHANGELOG.md
kubukoz: unresolving because of the 0.18.25 thing, just so that we don't forget

modules/codegen-plugin/src/sbt-test/codegen-plugin/update-lsp-config/build.sbt
dwalend: What would you think of changing from this string-to-string comparison to a json-to-json comparison? 

The string-to-string business seems really brittle - even if the json gets loaded from a string in a file the json-to-json comparison would be less brittle. 

json-to-json has a lot less magic, but if string-to-string is the standard then I won't rock the boat.

modules/codegen/test/src/smithy4s/codegen/internals/SmithyBuildSpec.scala
dwalend: Should I try to take that on in this fix, or does it belong in its own PR? (That seems like it'll touch a lot more of the system.)

modules/codegen/src/smithy4s/codegen/SmithyBuildJson.scala
dwalend: Were smithy4s to need to support a build tool with a slightly different format (bleep ?) then it'd save copying and hacking a magic string.

modules/codegen-plugin/src/sbt-test/codegen-plugin/update-lsp-config/build.sbt
kubukoz: I agree, we can stick to JSON comparisons on this level. If we care about formatting, we can test that in unit tests. I don't think we care about formatting much, though.

modules/codegen/test/src/smithy4s/codegen/internals/SmithyBuildSpec.scala
Baccata: You can leave it aside for now, no worries 👍

modules/codegen/src/smithy4s/codegen/SmithyBuildJson.scala
Baccata: Feel free to add it in this PR, if you'd like

modules/codegen/test/src/smithy4s/codegen/internals/SmithyBuildSpec.scala
kubukoz: https://github.com/disneystreaming/smithy4s/issues/1613

CHANGELOG.md
dwalend: I've pushed a merge with the latest of series/0.18

modules/codegen-plugin/src/sbt-test/codegen-plugin/update-lsp-config/build.sbt
dwalend: What do you like for a json library? Circe?

kubukoz: yeah circe is fine here

build.sbt
```diff
@@ -407,7 +407,9 @@ lazy val codegen = projectMatrix
       "smithyOrg" -> Dependencies.Smithy.org,
       "smithyVersion" -> Dependencies.Smithy.smithyVersion,
       "alloyOrg" -> Dependencies.Alloy.org,
-      "alloyVersion" -> Dependencies.Alloy.alloyVersion
+      "alloyVersion" -> Dependencies.Alloy.alloyVersion,
+      "smithy4sOrg" -> "com.disneystreaming.smithy4s",
```
kubukoz:
```suggestion
      "smithy4sOrg" -> organization,
```

(haven't tested it)

modules/codegen-plugin/src/sbt-test/codegen-plugin/update-lsp-config/build.sbt
dwalend: I just pushed a text version... I'll scratch at it a bit today, but this PR has been open more than a week. It's time to wrap it up.

kubukoz: yeah 😅

build.sbt
dwalend: There was a little more to it than that. Fixed.

dwalend:
I think that's everything listed in the PR. Anything else? Otherwise it's ready to merge.

Baccata (approved):
@kubukoz, I'll wait for your review

build.sbt
kubukoz: thanksss

kubukoz: interesting, I thought it'd like a setting key

dwalend: That sure looked like it should work. Just slapping in `organization` doesn't work either. I suspect there's something clever to do with sbt's type system, but didn't find it.

My core complaint about sbt (and maven) is that it's not self-consistent enough to learn patterns and gain speed from experience.

kubukoz (approved):

kubukoz:
thanks!

PR #1614, opened 2024-10-23
Fix hello world, use inlined files instead of strings

keynmol:
The snippet on the landing page isn't actually valid Smithy because it used 2.0 features without specifying version, and because it doesn't have a namespace.

With these changes I can simply run the CLI on it.

Additionally, removed the multiline strings which were weird to edit, and instead used webpack's raw-loader to use actual files. Page looks the same

CleanShot 2024-10-23 at 09 50 46@2x

### DIFF
--- a/modules/website/package-lock.json
+++ b/modules/website/package-lock.json
@@ -1,0 +1,13107 @@
+{
+  "name": "smithy4s",
+  "version": "0.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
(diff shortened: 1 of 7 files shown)


kubukoz (approved):

modules/website/package-lock.json
kubukoz, line 1: oops, looks like we shouldn't have added this. It slipped under my radar (now Yarn complains that there's a `package-lock` whereas there should only be a yarn lockfile)

PR #1621, opened 2024-11-11
Fix package name of quickstart example

yoohaemin:
... so it aligns with the declared namespace


kubukoz (approved):

kubukoz (changes requested):
hmm I was too quick to approve, looks like it's failing. We'll probably have to change it somewhere else too

Baccata:
Better change the namespace in the smithy code

Baccata:
Fixed in https://github.com/disneystreaming/smithy4s/pull/1639

PR #1623, opened 2024-12-02
Document timestamp encoder precision fix

### LINKED ISSUE #1622: Timestamp inconsistent JSON representation when using Document
Timestamp with epoch-seconds is inconsistently represented currently when encoding directly to JSON vs using a Document as an intermediate layer.

Reproduction
```scala
//> using scala "3.3"

//> using dep "com.disneystreaming.smithy4s::smithy4s-json:0.18.25"
import smithy4s.json.Json
import smithy4s.Document
import smithy4s.ShapeId
import smithy4s.Timestamp
import smithy4s.schema.Schema

case class Foo(seconds: Timestamp)

object Foo {
  given Schema[Foo] = Schema.struct(
    Schema.timestamp.required[Foo]("seconds", _.seconds).addHints(smithy.api.TimestampFormat.EPOCH_SECONDS.widen),
    )(Foo.apply).withId(ShapeId("document", "Foo"))
}

val encoder = Document.Encoder.fromSchema(Schema[Foo])

val foo = Foo(Timestamp(1, 0))

println(Json.writeDocumentAsBlob(encoder.encode(foo)).toUTF8String)

println(Json.writeBlob(foo).toUTF8String)
```

prints

```
{"seconds":1.000000000}
{"seconds":1}
```

The fix should be rather simple, I can create a PR once this issue is acked

msosnicki:

fixes #1622

- [ ] Added documentation
- [x] Updated changelog

### DIFF
--- a/modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
+++ b/modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
@@ -105,11 +105,16 @@ class DocumentEncoderSchemaVisitor(
         case EPOCH_SECONDS =>
           ts =>
             DNumber(
-              BigDecimal(ts.epochSecond) + BigDecimal(
-                ts.nano * 1 / 1000000000.0
-              )
-                .underlying()
-                .stripTrailingZeros()
+              BigDecimal({
+                val es = java.math.BigDecimal.valueOf(ts.epochSecond)
+                if (ts.nano == 0) es
+                else
+                  es.add(
+                    java.math.BigDecimal
+                      .valueOf(ts.nano.toLong, 9)
+                      .stripTrailingZeros
+                  )
+              })
             )
       }
     case PDocument => from(identity)
--- a/modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
+++ b/modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
@@ -105,15 +105,11 @@ class DocumentEncoderSchemaVisitor(
         case EPOCH_SECONDS =>
           ts =>
             DNumber(
-              BigDecimal(
-                java.math.BigDecimal
-                  .valueOf(ts.epochSecond)
-                  .add(
-                    java.math.BigDecimal
-                      .valueOf(ts.nano.toLong, 9)
-                      .stripTrailingZeros
-                  )
+              BigDecimal(ts.epochSecond) + BigDecimal(
+                ts.nano * 1 / 1000000000.0
               )
+                .underlying()
+                .stripTrailingZeros()
             )
       }
     case PDocument => from(identity)
--- a/modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
+++ b/modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
@@ -104,10 +104,17 @@ class DocumentEncoderSchemaVisitor(
         case HTTP_DATE => ts => DString(ts.format(HTTP_DATE))
         case EPOCH_SECONDS =>
           ts =>
-            val bd = (BigDecimal(ts.epochSecond) + (BigDecimal(
-              ts.nano
-            ) * BigDecimal(10).pow(-9))).underlying.stripTrailingZeros
-            DNumber(bd)
+            DNumber(
+              BigDecimal(
+                java.math.BigDecimal
+                  .valueOf(ts.epochSecond)
+                  .add(
+                    java.math.BigDecimal
+                      .valueOf(ts.nano.toLong, 9)
+                      .stripTrailingZeros
+                  )
+              )
+            )
       }
     case PDocument => from(identity)
     case PFloat    => from(float => DNumber(BigDecimal(float.toDouble)))
--- a/modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
+++ b/modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
@@ -104,10 +104,10 @@ class DocumentEncoderSchemaVisitor(
         case HTTP_DATE => ts => DString(ts.format(HTTP_DATE))
         case EPOCH_SECONDS =>
           ts =>
-            val epochSecondsWithNanos =
-              BigDecimal(ts.epochSecond) + (BigDecimal(ts.nano) * BigDecimal(10)
-                .pow(-9))
-            DNumber(epochSecondsWithNanos)
+            val bd = (BigDecimal(ts.epochSecond) + (BigDecimal(
+              ts.nano
+            ) * BigDecimal(10).pow(-9))).underlying.stripTrailingZeros
+            DNumber(bd)
       }
     case PDocument => from(identity)
     case PFloat    => from(float => DNumber(BigDecimal(float.toDouble)))
--- a/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
+++ b/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
@@ -394,8 +394,17 @@ private[smithy4s] class SchemaVisitorJCodec(
       }
 
       def encodeValue(x: Timestamp, out: JsonWriter): Unit = {
-        val bd = BigDecimal(x.epochSecond) + (x.nano * BigDecimal(10).pow(-9))
-        out.writeVal(bd.underlying.stripTrailingZeros)
+        out.writeVal(
+          BigDecimal(
+            java.math.BigDecimal
+              .valueOf(x.epochSecond)
+              .add(
+                java.math.BigDecimal
+                  .valueOf(x.nano.toLong, 9)
+                  .stripTrailingZeros
+              )
+          )
+        )
       }
 
       def decodeKey(in: JsonReader): Timestamp = {
--- a/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
+++ b/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
@@ -395,7 +395,9 @@ private[smithy4s] class SchemaVisitorJCodec(
 
       def encodeValue(x: Timestamp, out: JsonWriter): Unit = {
         out.writeVal(
-          BigDecimal(x.epochSecond) + BigDecimal(x.nano * 1/ 1000000000.0).underlying().stripTrailingZeros()
+          BigDecimal(x.epochSecond) + BigDecimal(x.nano * 1 / 1000000000.0)
+            .underlying()
+            .stripTrailingZeros()
         )
       }
 
--- a/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
+++ b/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
@@ -394,11 +394,14 @@ private[smithy4s] class SchemaVisitorJCodec(
       }
 
       def encodeValue(x: Timestamp, out: JsonWriter): Unit = {
-        out.writeVal(
-          BigDecimal(x.epochSecond) + BigDecimal(x.nano * 1 / 1000000000.0)
-            .underlying()
-            .stripTrailingZeros()
-        )
+        out.writeVal(BigDecimal({
+          val es = java.math.BigDecimal.valueOf(x.epochSecond)
+          if (x.nano == 0) es
+          else
+            es.add(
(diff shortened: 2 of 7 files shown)


modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
```diff
@@ -104,10 +104,10 @@ class DocumentEncoderSchemaVisitor(
         case HTTP_DATE => ts => DString(ts.format(HTTP_DATE))
         case EPOCH_SECONDS =>
           ts =>
-            val epochSecondsWithNanos =
-              BigDecimal(ts.epochSecond) + (BigDecimal(ts.nano) * BigDecimal(10)
-                .pow(-9))
-            DNumber(epochSecondsWithNanos)
+            val bd = (BigDecimal(ts.epochSecond) + (BigDecimal(
+              ts.nano
+            ) * BigDecimal(10).pow(-9))).underlying.stripTrailingZeros
```
msosnicki: potential issue: this is different from the current behavior in the [regular encoders](https://github.com/disneystreaming/smithy4s/blob/e369dd40850690cb38efd385c1cb16ed3d09a192/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala#L396), because it strips all the trailing zeros, rather than setting the scale to 9 at all times. Based on this [comment](https://github.com/disneystreaming/smithy4s/issues/1622#issuecomment-2500383589) jsoniter will respect the scales set on the `BigDecimal`, so striping all trailing zeros seem like a more general solution, rather than doing that only if `nano == 0`. 

Please let me know what you think, it definitely has to be consistent across both visitors, so I will adjust once this discussion is resolved.

kubukoz:
@Baccata can you take a look?

modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
Baccata: I'm happy with stripping trailing zeros, as long as the adjustment on the jsoniter-based codecs is not too complicated.

msosnicki: Perhaps I missed something, but to me it looks like it's as straighforward as the `Document` fix.

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -394,11 +394,8 @@ private[smithy4s] class SchemaVisitorJCodec(
       }
 
       def encodeValue(x: Timestamp, out: JsonWriter): Unit = {
-        if (x.nano == 0) {
-          out.writeVal(x.epochSecond)
-        } else {
-          out.writeVal(BigDecimal(x.epochSecond) + x.nano / 1000000000.0)
-        }
+        val bd = BigDecimal(x.epochSecond) + (x.nano * BigDecimal(10).pow(-9))
```
ghostbuster91: **issue:** shouldn't it be `BigDecimal(x.epochSecond) + (BigDecimal(x.nano) * BigDecimal(10).pow(-9))`?

otherwise we might lose some information due to overflow before the conversion to `BigDecimal` happens.

msosnicki: The result of this multiplication is already a `BigDecimal`, so I think it is fine

Baccata:
I'm happy with the change. Did you want to do anything else on this PR ?

plokhotnyuk:
I stored the following script in `timestamp.scala` and ran it using `scala-cli --power timestamp.scala`:
```scala
//> using jmh
//> using jmhVersion 1.37
package bench

import java.util.concurrent.TimeUnit
import org.openjdk.jmh.annotations.*

@BenchmarkMode(Array(Mode.Throughput))
@OutputTimeUnit(TimeUnit.SECONDS)
@Warmup(iterations = 2, time = 1, timeUnit = TimeUnit.SECONDS)
@Measurement(iterations = 2, time = 1, timeUnit = TimeUnit.SECONDS)
@State(Scope.Thread)
@Fork(1)
class Timestamp {

  @Param(Array("123456789", "123456000", "123000000", "0"))
  var nano: Int = 123456789

  var epochSecond: Long = System.currentTimeMillis() / 1000

  @Benchmark
  def pow() = BigDecimal((BigDecimal(epochSecond) + (nano * BigDecimal(10).pow(-9))).underlying.stripTrailingZeros)

  @Benchmark
  def scale() = BigDecimal(java.math.BigDecimal.valueOf(epochSecond).add(java.math.BigDecimal.valueOf(nano.toLong, 9).stripTrailingZeros))
}
```
On `Intel® Core™ i7-11800H` I got the following result using JDK 21:
```text
Benchmark           (nano)   Mode  Cnt          Score   Error  Units
Timestamp.pow    123456789  thrpt    2     689083.298          ops/s
Timestamp.pow    123456000  thrpt    2     753617.071          ops/s
Timestamp.pow    123000000  thrpt    2     745549.753          ops/s
Timestamp.pow            0  thrpt    2     768658.620          ops/s
Timestamp.scale  123456789  thrpt    2  184984953.592          ops/s
Timestamp.scale  123456000  thrpt    2  100797574.077          ops/s
Timestamp.scale  123000000  thrpt    2   83752128.337          ops/s
Timestamp.scale          0  thrpt    2  204584419.459          ops/s
```

msosnicki:
@plokhotnyuk 

The version you posted unfortunately breaks the `bootstrappedJS/DocumentSpec` (possibly something that could be reported to Scala JS)

Below you can find the results for the version that passes all tests, it behaves slightly worse, but still much better than the `pow` alternative:

```scala
@Benchmark 
def scalaBd() = BigDecimal(epochSecond) + BigDecimal(nano * 1/1000000000.0).underlying().stripTrailingZeros()
```

```
Benchmark             (nano)   Mode  Cnt          Score   Error  Units
Timestamp.pow      123456789  thrpt    2     657035.513          ops/s
Timestamp.pow      123456000  thrpt    2     716802.325          ops/s
Timestamp.pow      123000000  thrpt    2     714326.181          ops/s
Timestamp.pow              0  thrpt    2     719915.018          ops/s
Timestamp.scalaBd  123456789  thrpt    2   10934885.665          ops/s
Timestamp.scalaBd  123456000  thrpt    2    9444356.670          ops/s
Timestamp.scalaBd  123000000  thrpt    2    9101303.611          ops/s
Timestamp.scalaBd          0  thrpt    2   51482186.648          ops/s
Timestamp.scale    123456789  thrpt    2  164914297.116          ops/s
Timestamp.scale    123456000  thrpt    2   84470614.413          ops/s
Timestamp.scale    123000000  thrpt    2   69055873.696          ops/s
Timestamp.scale            0  thrpt    2  184479155.075          ops/s
```
measured on JDK 21, AMD Ryzen 7 3700X.

EDIT: unfortunately this one breaks the same test for Scala Native

plokhotnyuk:
The workaround for Scala.js and Scala Native is using of yet more optimized version:
```scala
BigDecimal({
   val es = java.math.BigDecimal.valueOf(epochSecond)
   if (nano == 0) es
   else es.add(java.math.BigDecimal.valueOf(nano, 9).stripTrailingZeros)
})
```
Here are results of benchmarks:
```text
Benchmark              (nano)   Mode  Cnt          Score   Error  Units
Timestamp.pow       123456789  thrpt    2     748598.854          ops/s
Timestamp.pow       123456000  thrpt    2     672352.329          ops/s
Timestamp.pow       123000000  thrpt    2     749301.419          ops/s
Timestamp.pow               0  thrpt    2     695728.197          ops/s
Timestamp.scalaBd   123456789  thrpt    2   12535456.203          ops/s
Timestamp.scalaBd   123456000  thrpt    2   12591612.068          ops/s
Timestamp.scalaBd   123000000  thrpt    2   12442059.688          ops/s
Timestamp.scalaBd           0  thrpt    2   59858539.270          ops/s
Timestamp.scaleOpt  123456789  thrpt    2  181264531.330          ops/s
Timestamp.scaleOpt  123456000  thrpt    2   92364559.709          ops/s
Timestamp.scaleOpt  123000000  thrpt    2   78387325.699          ops/s
Timestamp.scaleOpt          0  thrpt    2  203003351.828          ops/s
```

msosnicki:
@plokhotnyuk thanks for help, all green now

plokhotnyuk:
@msosnicki I've provided [a serialization method for timestamps](https://github.com/plokhotnyuk/jsoniter-scala/commit/4f1f650bf459cfad221c6d57a4792620498ae0c3) in [the 2.32.0 release of jsoniter-scala](https://github.com/plokhotnyuk/jsoniter-scala/releases/tag/v2.32.0), so after version upgrade you can serialize timestamps without any allocations or redundant CPU overhead as here:
```scala
def encodeValue(x: Timestamp, out: JsonWriter): Unit = out.writeTimestampVal(x.epochSecond, x.nano)
```
Also, `DNumber` creation could be improved by adding support of negative `epochSecond` values:
```scala
BigDecimal({
   val es = java.math.BigDecimal.valueOf(epochSecond)
   if (nano == 0) es
   else es.add(java.math.BigDecimal.valueOf({
      if (epochSeconds < 0) -nano
      else nano
   }.toLong, 9).stripTrailingZeros)
})
```

plokhotnyuk:
@msosnicki Please read my previous comment that I updated recently, where I propose more correct and efficient solution.

msosnicki:
> ```scala
> BigDecimal({
>    val es = java.math.BigDecimal.valueOf(epochSecond)
>    if (nano == 0) es
>    else es.add(java.math.BigDecimal.valueOf({
>       if (epochSeconds < 0) -nano
>       else nano
>    }.toLong, 9).stripTrailingZeros)
> })
> ```

Will add a comment about the new jsoniter method. 

Quesion about the negative nano from  your example, where is it documented? The same value before 1970 could be expressed in two ways: with flipping the sign of nano like you did, or adding to a negative `epochSeconds`.

plokhotnyuk:
@msosnicki @Baccata I've found only following documentation about the timestamp serialization methods: https://smithy.io/2.0/spec/protocol-traits.html#timestampformat-trait

It requires truncation to milliseconds and do not mention any support of dates before the Linux epoch (Jan 1, 1970).
But from current sources of `smithy4s.Timestamp` I see that we support parsing and serialization up to nanoseconds (without truncation to milliseconds) and for dates before the Linux epoch.

Also, some method implementations of `smith4s.Timestamp` require fixes to support negative `epochSecond`. For example `def epochMilli: Long = epochSecond * 1000 + nano / 1000000` should be replaced by 
```scala
def epochMilli: Long = epochSecond * 1000 + {
  if (epochSecond < 0) -nano
  else nano
} / 1000000
```

msosnicki:
> @msosnicki @Baccata I've found only following documentation about the timestamp serialization methods: https://smithy.io/2.0/spec/protocol-traits.html#timestampformat-trait
> 
> It requires truncation to milliseconds and do not mention any support of dates before the Linux epoch (Jan 1, 1970). But from current sources of `smithy4s.Timestamp` I see that we support parsing and serialization up to nanoseconds (without truncation to milliseconds) and for dates before the Linux epoch.
> 
> Also, some method implementations of `smith4s.Timestamp` require fixes to support negative `epochSecond`. For example `def epochMilli: Long = epochSecond * 1000 + nano / 1000000` should be replaced by
> 
> ```scala
> def epochMilli: Long = epochSecond * 1000 + {
>   if (epochSecond < 0) -nano
>   else nano
> } / 1000000
> ```

Yeah, I think the logic needs to change in multiple places, to correctly handle values below epoch. Will look into that

EDIT: only changed the decoding side, nanos are expected to always be positive (as in `Timestamp` class, tests that I added were raising constructor exceptions before).
So during decode we're always rounding down for seconds part (also for negative), no need to flip sings of nano in other places. 
It is more consistent with other parts of the code, like `Timestamp` constructors. 
For example `1969-12-31T23:59:58.123Z` (`-1877` in millis) is expected to be expressed as `Timestamp(-2L, 123000000)`, rather than `Timestamp(-1, 877000000)`.

ghostbuster91 (approved):

ghostbuster91:
@Baccata are we good to go?

kubukoz (commented):
please add a CHANGELOG entry for the next patch version

msosnicki:
> please add a CHANGELOG entry for the next patch version

done

Baccata:
Apologies, I've been out the past few weeks. 

> It requires truncation to milliseconds and do not mention any support of dates before the Linux epoch (Jan 1, 1970).

We're taking this with a grain of salt, as this is a protocol-specific concern and our `Timestamp` construct is protocol agnostic. We may add an option to truncate to the encoding side, but we'll always decode with maximum precision.

Baccata (approved):

PR #1629, opened 2024-12-19
Fix missing sbt command in CI build

plokhotnyuk:
My fork of this repo started to fail CI build with a missing command error for "sbt" and this small patch fixes that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,7 @@ jobs:
         uses: coursier/setup-action@v1
         with:
           jvm: temurin:1.17
+          apps: sbt
 
       - name: Run tests
         run: |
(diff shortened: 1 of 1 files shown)


kubukoz:
That's what we get for using ubuntu-latest :) https://github.com/actions/runner-images/issues/10636#issuecomment-2547174131 https://github.com/disneystreaming/smithy4s/blob/series/0.18/.github/workflows/ci.yml#L46

.github/workflows/ci.yml
```diff
@@ -61,6 +61,7 @@ jobs:
         uses: coursier/setup-action@v1
         with:
           jvm: temurin:1.17
+          apps: sbt
```
kubukoz, line 64: issue: We have other jobs using sbt, which will need the same treatment. Can you add sbt in there as well?

kubukoz (changes requested):

kubukoz (approved):
Thanks!

PR #1634, opened 2025-01-02
Update jsoniter-scala to 2.32.0

plokhotnyuk:
Updates a year old version mostly for the performance reason, see https://github.com/disneystreaming/smithy4s/blob/7deb6c32d63a644c729f4d52cece0a399191c401/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala#L398

### DIFF
--- a/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
+++ b/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
@@ -394,8 +394,15 @@ private[smithy4s] class SchemaVisitorJCodec(
         Timestamp(epochSeconds, ((timestamp - epochSeconds) * 1000000000).toInt)
       }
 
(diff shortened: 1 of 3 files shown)


kubukoz (approved):

kubukoz:
can't decide whether this is CHANGELOG-worthy, so I'll leave it up to you

kubukoz (changes requested):
mmm looks like this isn't as straightforward as we thought

image

plokhotnyuk:
I will provide a new release of jsoniter-scala with a fix for negative timestamp and added support for writing timestamp as JSON keys

Also, it seems that Scala Native should be upgraded too.

kubukoz:
> Also, it seems that Scala Native should be upgraded too.

sigh...

we're going to be **massively** blocked by Cats Effect - for good reasons - https://github.com/typelevel/cats-effect/issues/4076

PR #1637, opened 2025-01-16
Fix nullable decoding

lewisjkl:
Creating as a draft because I think the "solution" I have here may not be the right way to go at this, but wanted to get some thoughts about this in general before continuing. The issue is that if you have, for example, a Smithy structure that looks like:

```smithy
structure Foo {
  @required
  @nullable
  str: String
}
```

Smithy4s will decode the JSON payload `{}` into `Foo(Nullable.Null)`. I would think this is incorrect since the field is required and has no default value (if it had a default value of null, then it'd be working as intended).

I worked on tracking the issue down, and figured out that it is because our `Nullable` schema is a Bijection to `Option`. So when the JCodec Schema Visitor goes looking for the default value, it returns `Some(Nullable.Null)` rather than `None` for this case. I (at least temporarily) fixed this by modifying the bijection method of the `OptionDefaultVisitor`, but I'm not sure whether or not this approach is sound.

---

In this PR, I also investigated the case of:

```smithy
structure Foo {
  @required
  // @nullable // tested both with and without nullable
  str: RefinedType = null
}

// would have refinement trait on it
string RefinedType
```

The behavior for this was not as expected, for example in the _non-nullable_ case `{}` would result in `Missing required field` error. This is incorrect since in this case it should give an empty string (assuming empty string is a valid part of the domain set of possible values for the refinement type).

---

- [x] Added alloy compliance tests (when simpleRestJson protocol behaviour is expanded/updated)
- [x] Updated dynamic module to match generated-code behaviour
- [x] Added documentation
- [x] Updated changelog

### DIFF
--- a/modules/bootstrapped/test/src/smithy4s/schema/DefaultValueSpec.scala
+++ b/modules/bootstrapped/test/src/smithy4s/schema/DefaultValueSpec.scala
@@ -133,7 +133,7 @@ final class DefaultValueSpec extends FunSuite {
   test("refined") {
     val b: Schema[Int] =
       Schema.int.refined(smithy.api.Range(None, Option(BigDecimal(1))))
-    testCaseOpt(b, None)
+    testCaseOpt(b, Some(0))
   }
 
   test("recursive") {
--- a/modules/core/src/smithy4s/schema/Schema.scala
+++ b/modules/core/src/smithy4s/schema/Schema.scala
@@ -354,6 +354,9 @@ object Schema {
   private object OptionDefaultVisitor extends SchemaVisitor.Default[Option] {
     def default[A] : Option[A] = None
     override def option[A](schema: Schema[A]) : Option[Option[A]] = Some(None)
+    override def biject[A, B](schema: Schema[A], bijection: Bijection[A, B]): Option[B] = {
+      if (schema.hints.has[alloy.Nullable]) None else this.apply(schema).map(bijection.to)
+    }
   }
 
   def operation(id: ShapeId): OperationSchema[Unit, Nothing, Unit, Nothing, Nothing] =
--- a/modules/core/src/smithy4s/schema/Schema.scala
+++ b/modules/core/src/smithy4s/schema/Schema.scala
@@ -354,8 +354,6 @@ object Schema {
   private object OptionDefaultVisitor extends SchemaVisitor.Default[Option] {
     def default[A] : Option[A] = None
     override def option[A](schema: Schema[A]) : Option[Option[A]] = Some(None)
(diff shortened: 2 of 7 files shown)


modules/core/src/smithy4s/schema/Schema.scala
```diff
@@ -354,8 +354,9 @@ object Schema {
   private object OptionDefaultVisitor extends SchemaVisitor.Default[Option] {
     def default[A] : Option[A] = None
     override def option[A](schema: Schema[A]) : Option[Option[A]] = Some(None)
-    override def biject[A, B](schema: Schema[A], bijection: Bijection[A, B]): Option[B] =
-      this.apply(schema).map(bijection.to)
+    override def biject[A, B](schema: Schema[A], bijection: Bijection[A, B]): Option[B] = {
+      if (schema.hints.has[alloy.Nullable]) None else this.apply(schema).map(bijection.to)
```
lewisjkl, line 358: `this.apply(schema)` will return `Some(None)`, which when doing `.map(bijection.to)` becomes `Some(Nullable.Null)`. Basically these concepts are not treated the same since Option.empty is the absence of value while Nullable.Null is the inclusion of an explicit Null.

modules/core/src/smithy4s/schema/Schema.scala
```diff
@@ -354,8 +354,9 @@ object Schema {
   private object OptionDefaultVisitor extends SchemaVisitor.Default[Option] {
```
lewisjkl, line 354: Note that I was not able to create any test cases that indicated an implementation for refinements would do anything here. Every case I could think of would still bypass calling the refinement case. The `DefaultValueSchemaVisitor` did need to be updated however.

modules/bootstrapped/test/src/smithy4s/schema/DefaultValueSpec.scala
```diff
@@ -133,7 +133,7 @@ final class DefaultValueSpec extends FunSuite {
   test("refined") {
     val b: Schema[Int] =
       Schema.int.refined(smithy.api.Range(None, Option(BigDecimal(1))))
-    testCaseOpt(b, None)
+    testCaseOpt(b, Some(0))
```
lewisjkl, line 136: Looks like this test case was actually asserting the wrong thing. In this case, a null default on a non-nullable field means that the default value is assumed to be `0`

lewisjkl:
Before merging this, I think it will be helpful if I add documentation that shows each different combination of required, nullable, and null default and what happens (probably some tests to show this too).

Baccata (approved):

PR #1644, opened 2025-02-05
Add service transformation utils

kubukoz:

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/modules/core/src/smithy4s/schema/OperationSchema.scala
+++ b/modules/core/src/smithy4s/schema/OperationSchema.scala
@@ -19,6 +19,8 @@ package smithy4s
 package schema
 
 import smithy4s.internals.InputOutput
+import smithy4s.kinds.PolyFunction5
+import smithy4s.kinds.PolyFunction
 
 final case class OperationSchema[I, E, O, SI, SO] private[smithy4s] (
     id: ShapeId,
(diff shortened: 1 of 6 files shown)


modules/core/src/smithy4s/schema/OperationSchema.scala
```diff
@@ -96,3 +98,55 @@ final case class OperationSchema[I, E, O, SI, SO] private[smithy4s] (
     copy(streamedOutput = streamedOutput.map(f))
 
 }
+
+object OperationSchema {
```
kubukoz, line 102: **thought:** if these are deemed overkill, I'm ok with removing them.

kubukoz: the rest are more high-value though, I'd keep them as-is

Baccata: Couldn't these be methods on the OperationSchema class instead of static functions ?

kubukoz: no, if we want them as polyfunctions (and we do, if we want nice things like what you see in the new test), they have to be outside of the class

Baccata: fair nuff

Baccata (approved):

PR #1649, opened 2025-02-11
Use patterns instead of namespace literals for codegen

### LINKED ISSUE #1586: Allow to exclude nested namespaces
When generating scala classes from smithy model we often need to exclude certain namespaces due to various reasons. 

For that we were using `--exclude-ns` option in smithy4s-cli. However, recently we did some small refactoring that moved some of the shapes from the top namespace that was excluded (e.g. `a.b`) into sub-namespaces (e.g. `a.b.c` and `a.b.d`). 

This cause issues in the generated code as those nested namespaces were not excluded. 

Would it be possible to allow excluding namespaces in a more convenient way? e.g. `--exclude-ns a.b.*` would exclude `.a.b` as well as any other namespace that starts with `a.b`

msosnicki:

closes #1586

- [ ] Added documentation
- [ ] Updated changelog


modules/codegen/src/smithy4s/codegen/CodegenArgs.scala
```diff
@@ -18,15 +18,16 @@ package smithy4s.codegen
...
 final case class CodegenArgs(
     specs: List[os.Path],
     output: os.Path,
     resourceOutput: os.Path,
     skip: Set[FileType],
     discoverModels: Boolean,
-    allowedNS: Option[Set[String]],
-    excludedNS: Option[Set[String]],
+    allowedNS: Option[Set[NamespacePattern]],
+    excludedNS: Option[Set[NamespacePattern]],
```
Baccata: this change is not source compatible and `CodegenArgs` is very much part of the public API. Let's leave it as `String` please

Baccata (dismissed):

modules/codegen/test/src/smithy4s/codegen/internals/NamespacePatternSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+      "a.b.*" -> "a.b",
+      "a.b.*" -> "a.B",
+      "a.b.*" -> "a.b.C",
+      "a.b.*" -> "b.a",
+      "a.b.*" -> "a.bb.c",
+      "a.b.*" -> "acb.d",
+      "a.b*" -> "acb",
+      "a.b*" -> "b.a.c",
+      "a.b*" -> "a.bC",
+      "a.b*" -> "a.b.C"
```
msosnicki: I think this should pass - looks like `C` is a valid namespace segment after all, thought it's just lowercase chars

msosnicki: Done

modules/codegen/src/smithy4s/codegen/CodegenArgs.scala
kubukoz: could've done an additional `apply` method with the new signature tbh

kubukoz: @Baccata can you elaborate?

kubukoz (approved):
Think we just need a CHANGELOG entry, wouldn't mind a codegen unit test too

kubukoz (changes requested):
missclicked

modules/codegen/src/smithy4s/codegen/CodegenArgs.scala
Baccata: Sorry, my thumbs-down meant : adding an `apply` method doesn't fix the problem because it solves only constructors, not accessors

modules/codegen/test/src/smithy4s/codegen/internals/CodegenImplSpec.scala
```diff
@@ -102,4 +102,107 @@ final class CodegenImplSpec extends munit.FunSuite {
...
+    )(expectedCodegenNamespaces =
+      Set("allowed.namespace", "allowed.namespace.nested")
+    )
+  }
+
+  private def namespaceFilterTest(
+      inputNamespaces: List[String],
+      allowedNamespaces: List[String] = Nil,
+      forbiddenNamespaces: List[String] = Nil
+  )(expectedCodegenNamespaces: Set[String]) = {
```
kubukoz, line 177: **nitpick:** this should propagate an implicit location, but it's easy to add when necessary

kubukoz (approved):

Baccata (approved):

PR #1651, opened 2025-02-19
Memoize default values in Schema, not Field

kubukoz:
Surely this won't break anything...

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/modules/core/src/smithy4s/schema/Field.scala
+++ b/modules/core/src/smithy4s/schema/Field.scala
@@ -41,7 +41,7 @@ final case class Field[S, A](
   @deprecated("use .schema instead", since = "0.18.0")
   final def instance: Schema[A] = schema
 
-  lazy val getDefaultValue: Option[A] =
+  def getDefaultValue: Option[A] =
     schema.getDefaultValue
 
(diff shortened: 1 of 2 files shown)


Baccata:
can you elaborate on the why ?

plokhotnyuk:
I would prefer to have `Option[() => A]` instead (kind of evaluator function without any memorization) because instantiation of default values can be side effecting or the result value could be just mutable and should not be shared.

kubukoz:
> can you elaborate on the why ?

@Baccata I think it could be surprising that default values would get memoized in one place but not in another. My assumption is that the current behavior may not have been intentional, but maybe that's mistaken



> I would prefer to have `Option[() => A]` instead (kind of evaluator function without any memorization) because instantiation of default values can be side effecting or the result value could be just mutable and should not be shared.

there should be no side effects in instantiation of defaults from generated code. The only way I see it happening would be if you use a [refinement](https://disneystreaming.github.io/smithy4s/docs/codegen/customisation/refinements/) to a mutable type, and that's probably already broken in several ways...

Baccata (approved):

Issue #1398, opened 2024-02-08
Support more flexible explicit defaults


kubukoz:
Currently, explicit defaults are "all or nothing". Some users (myself included) might want to skip nulls, or skip empty collections - it would be great to have this configurable, e.g. as a `Schema[_] => Boolean` function rather than just `Boolean`.

One problem I see with this is that anything currently implemented with a `SchemaVisitor` won't have the original `Schema[A]`, but rather its pieces that _may_ be composed into a Schema again.

kubukoz:
current thinking is having something like this

```scala
.withDefaultEncoding(
  f: [A] => Field[_, A] => A => Boolean
)
```

then you can e.g. check if the field has an optional schema, and based on that + the value decide to e.g. write `Some(defaultValue)` but not `None`.

Baccata:
Okay, let's try it.

kubukoz:
done in #1652 #1662 and released in 0.18.31.

PR #1652, opened 2025-02-20
feat: Add support for flexible defaults encoding

ghostbuster91:

- [ ] ~Added bootstrapped code + smoke tests (when the rendering logic is modified)~ N/A
- [ ] ~Added build-plugins integration tests (when reflection loading is required at codegen-time)~ N/A
- [ ] ~Added alloy compliance tests (when simpleRestJson protocol behaviour is expanded/updated)~ N/A
- [ ] ~Updated dynamic module to match generated-code behaviour~ N/A
- [x] Added documentation
- [x] Updated changelog

Refers to #1398


modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
```diff
@@ -0,0 +1,134 @@
...
+        field: Field[?, A]
+    ): FieldRenderPredicateCompiler.ShouldSkip[A] = {
+      if (field.isRequired) Function.const(false)
+      else compileOptional(field)
+
+    }
+
+    def compileOptional[A](
+        field: Field[?, A]
+    ): FieldRenderPredicateCompiler.ShouldSkip[A]
```
ghostbuster91: **thought:** I was tempted to cast `Field[?,A]` to `Field[?, Optional[A]]` in `OptionalPredicate` but that doesn't work for fields with a default value.

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
```diff
@@ -0,0 +1,134 @@
...
+        Some(collection => c.tag.isEmpty(collection.asInstanceOf[f[a]]))
+      case BijectionSchema(inner, _)  => asEmptyCollectionPredicate(inner)
+      case RefinementSchema(inner, _) => asEmptyCollectionPredicate(inner)
+      case OptionSchema(inner) =>
+        asEmptyCollectionPredicate(inner)
+          .map(predicate =>
+            wrappedCollection =>
+              wrappedCollection.asInstanceOf[Option[Any]].exists(predicate)
+          )
+      case LazySchema(_)           => None // ?
```
ghostbuster91: Not sure what to do about that `LazySchema`. Should I simply write:
```scala
case LazySchema(inner) => asEmptyCollectionPredicate(inner.value)
```
?

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
```diff
@@ -0,0 +1,134 @@
...
+      case _: MapSchema[k, v]      => None
```
ghostbuster91: **question:** should maps be also taken care of here?

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
```diff
@@ -0,0 +1,134 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package smithy4s.codecs
+
+import smithy4s.schema.Field
+import smithy4s.schema.Schema
+
+trait FieldRenderPredicateCompiler { self =>
```
ghostbuster91: **question:** should this be `sealed`?

modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
```diff
@@ -76,13 +77,28 @@ object DocumentEncoder {
...
+      fieldRenderPredicateCompiler =
+        FieldRenderPredicateCompiler.fromExplicitDefaults(
+          explicitDefaultsEncoding
+        )
+    )
+
   def this(cache: CompilationCache[DocumentEncoder]) =
     this(cache, explicitDefaultsEncoding = false)
 
+  protected val explicitDefaultsEncoding: Boolean =
```
ghostbuster91, line 99: **thought:** kept for binary compatibility reasons

ghostbuster91:
I am not happy about the implementation of `SkipIfEmptyCollection` so I am fine with dropping it at this stage to minimize the amount of changes, but wanted to showcase it in case you think that it is worth to have something like that.

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
Baccata: No, users should be able to create their own.

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
```diff
@@ -0,0 +1,135 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.codecs
+
+import smithy4s.schema.Field
+import smithy4s.schema.Schema
+
+trait FieldRenderPredicateCompiler { self =>
```
Baccata: Let's rename to `FieldSkip`

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
```diff
@@ -0,0 +1,135 @@
...
+  def compile[A](
```
Baccata: let's change the interface to 

```
def compile(label: String, schema: Schema[A]) : A => Boolean 

final def compileField[S, A](field: Field[S, A]) : A => Boolean  = compile(field.label, field.schema)
```

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
```diff
@@ -0,0 +1,135 @@
...
+      field: Field[?, A]
+  ): FieldRenderPredicateCompiler.ShouldSkip[A]
+
+  def combine(
```
Baccata: let's make two combinators : `&&` and `||`. The behaviour will be clearer to end-users

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
Baccata: Let's call it `SkipNonRequired` instead. The term "optional" is more related to our Scala-specific interpretation, which differs to the absence of `@required` (for instance when default values are involve, as you point out)

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
```diff
@@ -0,0 +1,135 @@
...
+        a => r1(a) || r2(a)
+      }
+    }
+}
+
+object FieldRenderPredicateCompiler {
+
+  type ShouldSkip[A] = A => Boolean
+
+  trait OptionalPredicate extends FieldRenderPredicateCompiler {
```
Baccata: Don't make this public.

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
Baccata: I _think_ so. You'll have to add a test to ensure it doesn't recurse infinitely

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
```diff
@@ -0,0 +1,135 @@
...
+    ): FieldRenderPredicateCompiler.ShouldSkip[A]
+
+  }
+
+  val NeverSkip = new FieldRenderPredicateCompiler {
+    def compile[A](field: Field[_, A]): ShouldSkip[A] = Function.const(false)
+  }
+
+  private def asEmptyCollectionPredicate[F[_]](
+      schema: Schema[_]
```
Baccata: avoid using existential types as parameters : make it : 

```
private def emptyCollectionPredicate[F[_], A](schema: Schema[A]) : Option[A => Boolean] 
```

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
Baccata: yes

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
```diff
@@ -0,0 +1,135 @@
...
+    new FieldRenderPredicateCompiler.OptionalPredicate {
+      def compileOptional[A](field: Field[?, A]): ShouldSkip[A] = { a =>
+        a == None
+      }
+    }
+
+  val SkipIfEmptyOrDefaultOptionals =
+    SkipIfEmptyOptionals combine SkipIfDefaultOptionals
+
+  def fromExplicitDefaults(
```
Baccata: remove this, just inline the code where it's used

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+  val SkipIfEmptyOptionals =
+    new FieldRenderPredicateCompiler.OptionalPredicate {
+      def compileOptional[A](field: Field[?, A]): ShouldSkip[A] = { a =>
+        a == None
+      }
+    }
+
+  val SkipIfEmptyOrDefaultOptionals =
```
Baccata: I'd rather we removed all the ones that are mere combinations, and let the users combine them themselves

modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
Baccata: add a `@deprecated` onto it.

Baccata:
A good start 👍

ghostbuster91:
@kubukoz wanna have a look?

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
ghostbuster91: If we do that then we lose access to `isRequired`, `getDefaultValue` and other useful methods defined on the `Field` class. These methods delegate to `Schema` so we can still recreate that but I think that it makes more sense to check `isRequired` on a `field` instance rather than on `schema`. 

wdyt?

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
ghostbuster91: I actually started with two combinators at the beginning but was not happy with how it looked.

`ShouldSkip[A]` determines whether an element should be skipped during rendering. When combining multiple predicates, `OR` makes sense because if any condition suggests skipping, the element should be skipped.

Using `AND` means an element is skipped only if *all* predicates agree. While this is logically valid, it's rarely (if ever) useful.

For some, this might be counterintuitive. Consider `SkipIfDefaultOptionals` and `SkipIfEmptyOptionals`:
- If an element is **empty but not default**, `SkipIfEmptyOptionals` returns `true`, but `SkipIfDefaultOptionals` returns `false`, so the element **won’t be skipped** when using `and` combinator.
- The same issue happens in reverse—if it’s **default but not empty**, it also won’t be skipped.

The expected behavior is that **if any predicate says "skip," the element should be skipped.** This aligns naturally with `OR`, while `AND` leads to incorrect rendering in many cases.

Also, `ShouldSkip` together with the `or` combinator and the `NeverSkip` element forms a Monoid. That’s why I renamed `or` to `combine` to align with Monoid terminology. 

Let me know what you think.

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
ghostbuster91: I agree in general but I also think that users should have an easy way to use the previous behavior and `SkipIfEmptyOrDefaultOptionals` represents the old `explicitDefaults=false`.

modules/core/src/smithy4s/codecs/FieldSkipCompiler.scala
```diff
@@ -0,0 +1,144 @@
...
+  ): Option[A => Boolean] = {
+    import Schema._
+    schema match {
+      case c: CollectionSchema[f, A @unchecked] =>
+        Some(collectionA => c.tag.isEmpty(collectionA.asInstanceOf[f[A]]))
+      case b: BijectionSchema[inner, A @unchecked] =>
+        asEmptyCollectionPredicate[F, inner](b.underlying).map(predicateInner =>
+          collectionA => predicateInner(b.bijection.from(collectionA))
+        )
+      case r: RefinementSchema[inner, A @unchecked] =>
```
ghostbuster91: todo: add test cases for these two

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
kubukoz: isn't `explicitDefaults=false` the default?

we should have something for the current `explicitDefaults=true` though

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
kubukoz: I would change the whole thing to mean "keep if true" rather than "skip if true". It'd also be more familiar when you think about it as of a `filter` operation (which is keep if true). Doing it the other way around feels like it'd get messy when you want to negate anything, because you'd be dealing with double negations - the same reason why we avoid writing code like `if !dontRender` and instead write `if render`.

so I'd change to keep-if-true, add a `!` operator, and keep `&&` / `||`. Haven't thought about monoids.

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
kubukoz: agree with Kasper, checking `isRequired` on a random schema you were given feels odd

modules/core/src/smithy4s/codecs/FieldSkipCompiler.scala
```diff
@@ -0,0 +1,144 @@
...
+  ): FieldSkipCompiler.ShouldSkip[A]
+
+  def combine(
+      other: FieldSkipCompiler
+  ): FieldSkipCompiler =
+    new FieldSkipCompiler {
+
+      def compile[A](
+          field: Field[?, A]
+      ): FieldSkipCompiler.ShouldSkip[A] = {
```
kubukoz: I wonder if encoding this as a `case class FieldSkipCombine(left: ..., right: ...)` wouldn't be a better choice, for debuggability (toString, debugger, profiler)

but maybe it's overkill

modules/core/src/smithy4s/codecs/FieldSkipCompiler.scala
```diff
@@ -0,0 +1,144 @@
...
+
+    final def compile[A](
+        field: Field[?, A]
+    ): FieldSkipCompiler.ShouldSkip[A] = {
+      if (field.isRequired) Function.const(false)
+      else compileOptional(field)
+
+    }
+
+    def compileOptional[A](
```
kubukoz: nitpick: I'm not too happy about this having `Optional` but not having an Option in the signature, but since it's private I guess it's fine

modules/core/src/smithy4s/codecs/FieldSkipCompiler.scala
```diff
@@ -0,0 +1,144 @@
...
+      a => a != None && field.isDefaultValue(a)
+    }
+  }
+
+  val SkipIfDefaultOptionals: FieldSkipCompiler = skipIfDefaultOptionals
+
+  private case object skipIfEmptyOptionals
+      extends FieldSkipCompiler.SkipNonRequired {
+    def compileOptional[A](field: Field[?, A]): ShouldSkip[A] = { a =>
+      a == None
```
kubukoz: this relates to my previous comment about optional, but I feel icky about comparing an `A` that we don't know anything about to a None.

And now I just remembered about the Dynamic module - that's one space in which `==` can not always be relied upon (e.g. when comparing structs, `==` would just compare arrays (`smithy4s.dynamic.internals.DynStruct`) which is not what you want). Thankfully `Option` is a thing in Dynamic, but in theory it might be replaceable with a null as an optimization...

so I'd at least make sure to have a couple tests for these which go through `DSI`.

modules/json/src/smithy4s/json/internals/JsoniterCodecCompilerImpl.scala
```diff
@@ -19,27 +19,40 @@ package internals
...
+    fieldSkipCompiler: FieldSkipCompiler
 ) extends CachedSchemaCompiler.Impl[JCodec]
     with JsoniterCodecCompiler {
 
   type Aux[A] = JCodec[A]
 
   def withMaxArity(max: Int): JsoniterCodecCompiler = copy(maxArity = max)
 
+  @deprecated(
+    message = "Use withFieldSkipCompiler instead",
```
kubukoz: let's tell the users what exact compiler to use for the current `.withExplicitDefaultsEncoding(true)`

modules/json/src/smithy4s/json/internals/JsoniterCodecCompilerImpl.scala
```diff
@@ -19,27 +19,40 @@ package internals
...
+    since = "0.18.30"
+  )
   def withExplicitDefaultsEncoding(
-      explicitDefaultsEncoding: Boolean
+      explicitNulls: Boolean
```
kubukoz: this should probably be renamed back, for source compatibility (and also is `explicitNulls` even a good name, if we're not changing the semantic?)

modules/core/src/smithy4s/codecs/FieldSkipCompiler.scala
```diff
@@ -0,0 +1,144 @@
...
+        )
+      case o: OptionSchema[inner] =>
+        asEmptyCollectionPredicate(o.underlying)
+          .map(predicateInner =>
+            collectionA =>
+              collectionA.asInstanceOf[Option[inner]].exists(predicateInner)
+          )
+      case _: MapSchema[k, v] =>
+        Some(collectionA => collectionA.asInstanceOf[Map[k, v]].isEmpty)
+      case LazySchema(suspend)     => asEmptyCollectionPredicate(suspend.value)
```
kubukoz: wouldn't this blow up? thought we can't actually call `.value` until after compilation. Makes me wonder if this is covered by tests

kubukoz: ok I think I understand now - it doesn't blow up because we don't further inspect the schemas of structs/unions, which would be sitting between one occurrence of a recursive list and another. So we never see the same schema twice in this visitor.

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
Baccata: Mmm, yeah okay. But at least lose the existential type and do 

```
def compile[S, A](field: Field[S, A]) 
```

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
ghostbuster91: > isn't `explicitDefaults=false` the default?

yes, it is. 

> we should have something for the current `explicitDefaults=true` though

that is just the current `NeverSkip`

modules/core/src/smithy4s/codecs/FieldSkipCompiler.scala
ghostbuster91: me neither

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
Baccata: Yeah I agree with @kubukoz here, it's probably a lot more intuitive to use `keep if true`

Also regarding the thought regarding monoids : I'd rather not have it. When it comes to boolean-like constructs, I'd rather not have to think about whether we're talking about the additive or multiplicative monoid, and see explicit `&&` and `||` in combinators.

modules/json/src/smithy4s/json/internals/JsoniterCodecCompilerImpl.scala
ghostbuster91: not sure why it got renamed in the first place. Probably some metals quirk.

modules/core/src/smithy4s/codecs/FieldRenderPredicateCompiler.scala
ghostbuster91: ok good point with the `filter` operation. Should we now rename it to `FieldRenderCompiler`?

Baccata: I'd rename it to `FieldFilter`. I'd drop the `Compiler` part, since it's already expressed by the `def compile`.

modules/core/src/smithy4s/codecs/FieldSkipCompiler.scala
```diff
@@ -0,0 +1,141 @@
...
+    def compileOptional[A](field: Field[?, A]): ShouldRender[A] = {
+      asNonEmptyCollectionPredicate(field.schema) match {
+        case None             => Function.const(true)
+        case Some(isNonEmpty) => isNonEmpty
+      }
+    }
+  }
+
+  val SkipIfEmptyOptionalCollection: FieldSkipCompiler =
+    skipIfEmptyOptionalCollection
```
ghostbuster91: **thought:** Since I am not sure if this is a common pattern, I thought I would add a bit of explanation behind it.
Previously, this was just an anonymous instance assigned to a `val`. However, if you try to print it, it will show something like `FieldSkipCompiler$0afe1` — not very developer-friendly.

At first, I converted it to a `case object`, but that didn't work out due to `FieldSkipCompiler.SkipNonRequired` being private. That's why the `case object` is private, but there is a public `val` with an explicitly assigned type.

modules/core/src/smithy4s/codecs/FieldSkipCompiler.scala
ghostbuster91: `DSI` is used in `AwsComplianceSuite` so I think that it is covered already. Though I am not opposed to adding more tests for it, just not sure where to put them.

modules/core/src/smithy4s/codecs/FieldSkipCompiler.scala
ghostbuster91: I will add a comment about it since that is definitely not obvious at the first glance.

modules/core/src/smithy4s/codecs/FieldFilter.scala
```diff
@@ -0,0 +1,176 @@
...
+      // do not use field.hasDefaultValue as it returns always true for options
+      if (field.schema.getDefault.isDefined) { a =>
+        !field.isDefaultValue(a)
+      } else {
+        Function.const(true)
+      }
+    }
+  }
+
+  val SkipDefaultOptionValues: FieldFilter = skipDefaultOptionValues
```
Baccata: Rename to `SkipNonRequiredDefaultValues`

Baccata: Also why make a private case object and a value ?

modules/core/src/smithy4s/codecs/FieldFilter.scala
```diff
@@ -0,0 +1,176 @@
...
+
+    final def compile[S, A](
+        field: Field[S, A]
+    ): FieldFilter.Predicate[A] = {
+      if (field.isRequired) Function.const(true)
+      else compileOptional(field)
+
+    }
+
+    def compileOptional[S, A](
```
Baccata: rename to `compileNonRequired`

modules/core/src/smithy4s/codecs/FieldFilter.scala
```diff
@@ -0,0 +1,176 @@
...
+        field: Field[S, A]
+    ): FieldFilter.Predicate[A]
+
+  }
+
+  case object EncodeAll extends FieldFilter {
+    def compile[S, A](field: Field[S, A]): Predicate[A] = Function.const(true)
+  }
+
+  private def asNonEmptyCollectionPredicate[F[_], A](
```
Baccata: don't think the `F[_]` is needed

modules/core/src/smithy4s/codecs/FieldFilter.scala
```diff
@@ -0,0 +1,176 @@
...
+        case None             => Function.const(true)
+        case Some(isNonEmpty) => isNonEmpty
+      }
+    }
+  }
+
+  val SkipIfEmptyOptionalCollection: FieldFilter =
+    skipIfEmptyOptionalCollection
+
+  case object SkipIfEmptyCollection extends FieldFilter {
```
Baccata: Rename to `SkipEmptyCollections`

modules/core/src/smithy4s/codecs/FieldFilter.scala
```diff
@@ -0,0 +1,176 @@
...
+    def compileOptional[S, A](field: Field[S, A]): Predicate[A] = {
+      asNonEmptyCollectionPredicate(field.schema) match {
+        case None             => Function.const(true)
+        case Some(isNonEmpty) => isNonEmpty
+      }
+    }
+  }
+
+  val SkipIfEmptyOptionalCollection: FieldFilter =
+    skipIfEmptyOptionalCollection
```
Baccata: no need for a case object AND a val, is there?  Just make the case object public.

modules/core/src/smithy4s/codecs/FieldFilter.scala
```diff
@@ -0,0 +1,176 @@
...
+        // so we never see the same schema twice in this visitor.
+        asNonEmptyCollectionPredicate(suspend.value)
+      case _: EnumerationSchema[_] => None
+      case _: StructSchema[_]      => None
+      case _: UnionSchema[_]       => None
+      case _: PrimitiveSchema[_]   => None
+    }
+  }
+
+  private case object skipIfEmptyOptionalCollection
```
Baccata: rename to `SkipEmptyNonRequiredCollections`

modules/core/src/smithy4s/codecs/FieldFilter.scala
```diff
@@ -0,0 +1,176 @@
...
+
+  private case object skipUnsetOptions extends FieldFilter.SkipNonRequired {
+    def compileOptional[S, A](field: Field[S, A]): Predicate[A] = { a =>
+      a != None
+    }
+  }
+
+  val SkipUnsetOptions: FieldFilter = skipUnsetOptions
+
+  object SkipUnsetAndDefaultOptionValues extends FieldFilter {
```
Baccata: Change this to `val default = SkipUnsetOptions && SkipNonRequiredDefaultValues`

modules/docs/markdown/03-protocols/04-simple-rest-json/01-overview.md
```diff
@@ -86,9 +86,20 @@ See the section about [unions](../../04-codegen/02-unions.md) for a detailed des
 * This limit can be changed by setting the maxArity `smithy4s.http4s.SimpleRestJsonBuilder.withMaxArity(.)` to the desired value.
 * an example can be seen in the [client example](03-client.md)
 
-## Explicit Null Encoding
+## Field Filtering and Explicit Null Encoding
```
Baccata, line 89: Some documentation also ought to be added to the serialisation section, after this PR : https://github.com/disneystreaming/smithy4s/pull/1653

modules/core/src/smithy4s/codecs/FieldFilter.scala
ghostbuster91: https://github.com/disneystreaming/smithy4s/pull/1652#discussion_r1967425666

modules/core/src/smithy4s/codecs/FieldFilter.scala
ghostbuster91: we cannot since `skipIfEmptyOptionalCollection` is of type `FieldFilter.SkipNotRequired` which is private

modules/core/src/smithy4s/codecs/FieldFilter.scala
ghostbuster91: good catch, removed.

modules/core/src/smithy4s/codecs/FieldFilter.scala
Baccata: oh ... gotcha 👍

modules/docs/markdown/03-protocols/04-simple-rest-json/01-overview.md
kubukoz: @ghostbuster91 the other PR has been merged, please adapt

ghostbuster91: done :+1:

kubukoz (approved):
@baccata please have a final look too

modules/core/src/smithy4s/codecs/FieldFilter.scala
Baccata: The rename I've asked ("SkipNonRequiredDefaultValues") for still has to be applied

modules/core/src/smithy4s/codecs/FieldFilter.scala
```diff
@@ -0,0 +1,173 @@
...
+ *     https://disneystreaming.github.io/TOST-1.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.codecs
```
Baccata: should be moved to `smithy4s.schema` I think

modules/bootstrapped/test/src/smithy4s/DocumentSpec.scala
```diff
@@ -1146,6 +1196,195 @@ class DocumentSpec() extends FunSuite {
...
+        .withFieldFilter(strategy)
+        .fromSchema(structSchema)
+        .encode(MyStruct(Some(List.empty)))
+
+      expect.same(result, expected)
+    }
+
+  }
+
+  test("Recursive document structures shoould not blow up FieldFilter") {
```
Baccata: typo ("shoould")

Baccata:
Good work. Just a couple nitpicks (and also this : https://github.com/disneystreaming/smithy4s/pull/1652#discussion_r1971927935)

modules/core/src/smithy4s/codecs/FieldFilter.scala
ghostbuster91: Sorry, I forgot about this one due to the other comment about private case objects. Applied.

modules/core/src/smithy4s/codecs/FieldFilter.scala
ghostbuster91: done

Baccata (approved):

kubukoz:
@ghostbuster91 fyi if you said "Closes #1398" merging this would've closed the ticket, worth remembering in the future

PR #1656, opened 2025-02-24
Force rendering package object when a validated newtype is present

### LINKED ISSUE #1655: A lone validated newtype breaks compilation
Long story short, a validated newtype doesn't cause rendering of the package object (unlike normal newtypes and services), which breaks compilation due to the lack  of the type alias.

```smithy
$version: "2"

metadata smithy4sRenderValidatedNewtypes = true

namespace demo

@length(min: 0)
list MyNewtype {
    member: String
}
```

Image

Tested on 0.18.29.

As a workaround, one can put an unvalidated newtype (e.g. `string Foo`) into the namespace and that'll create a proper package object with both aliases.

kubukoz:
Closes #1655 

- [ ] Added documentation
- [x] Updated changelog


modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -288,24 +284,31 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
-    if (parts.size > 1) {
-      lines(
-        line"package ${parts.dropRight(1).mkString(".")}",
-        newline,
-        blk
-      )
-    } else blk
+    val packageContents = lines(serviceAliases, newline, typeAliases, newline)
+
+    packageContents.some
```
kubukoz, line 292: **thought:** that should prevent these things happening in the future 🤞

Baccata (approved):

PR #1662, opened 2025-03-19
Fix handling of non-Option optionals in field filtering

kubukoz:
By oversight in #1398, we kept the `_ != None` check, which makes certain usecases non-viable. Allowing bijections to Option seems like the right move.

- [ ] Added documentation
- [x] Updated changelog

### DIFF
--- a/modules/core/src/smithy4s/schema/FieldFilter.scala
+++ b/modules/core/src/smithy4s/schema/FieldFilter.scala
@@ -169,8 +169,11 @@ object FieldFilter {
 
     def apply[A](schema: Schema[A]): Predicate[A] = schema match {
       // nullables are technically never None, so we fall through
-      case OptionSchema(underlying) if !underlying.hints.has(Nullable) =>
-        _ == None
+      case OptionSchema(underlying) =>
+        if (underlying.hints.has(Nullable) && !underlying.isOption)
+          Function.const(false)
+        else
+          _ == None
 
       case BijectionSchema(underlying, bijection) =>
         this(underlying).compose(bijection.from)
--- a/modules/core/src/smithy4s/schema/FieldFilter.scala
+++ b/modules/core/src/smithy4s/schema/FieldFilter.scala
@@ -16,10 +16,12 @@
 
 package smithy4s.schema
 
-import smithy4s.Bijection
-import smithy4s.Lazy
-import smithy4s.Refinement
+import smithy4s.~>
 import alloy.Nullable
+import smithy4s.schema.Schema.OptionSchema
+import smithy4s.schema.Schema.BijectionSchema
+import smithy4s.schema.Schema.RefinementSchema
+import smithy4s.schema.Schema.LazySchema
 
 trait FieldFilter { self =>
   def compile[S, A](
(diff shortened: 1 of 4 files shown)


modules/json/test/src/smithy4s/json/JsonCodecApiTests.scala
```diff
@@ -82,26 +82,6 @@ class JsonCodecApiTests extends FunSuite {
     assertEquals(encoded, Blob("""{"a":null}"""))
   }
 
-  test(
```
kubukoz, line 85: This test was an exact duplicate of the one above.

ghostbuster91 (approved):

modules/core/src/smithy4s/schema/FieldFilter.scala
```diff
@@ -158,10 +163,45 @@ object FieldFilter {
...
-    def compileNonRequired[S, A](field: Field[S, A]): Predicate[A] = { a =>
-      a != None
+  private object IsNoneVisitor extends SchemaVisitor.Default[Predicate] {
+    def default[A]: Predicate[A] = Function.const(false)
+
+    override def biject[A, B](
+        schema: Schema[A],
+        bijection: Bijection[A, B]
+    ): Predicate[B] = {
+      if (schema.hints.has(Nullable)) {
```
Baccata: I'd rather this check was performed on the `def option` case. For consistency with the rest of the implementation that verify whether a field is nullable (https://github.com/disneystreaming/smithy4s/blob/series/0.18/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala#L1356-L1357)

modules/core/src/smithy4s/schema/FieldFilter.scala
```diff
@@ -158,10 +163,45 @@ object FieldFilter {
...
+        // nullables are technically never None
+        Function.const(false)
+      } else {
+        val inner = this(schema)
+        a => inner(bijection.from(a))
+      }
+    }
+
+    override def lazily[A](suspend: Lazy[Schema[A]]): Predicate[A] = {
+      val u = suspend.map(this(_))
```
Baccata: this needs to be a lazy val

modules/json/test/src/smithy4s/json/JsonCodecApiTests.scala
```diff
@@ -125,4 +105,38 @@ class JsonCodecApiTests extends FunSuite {
...
+        Json.jsoniter.withFieldFilter(FieldFilter.SkipUnsetOptions)
+      )
+
+      capi.encoders.fromSchema(schema).encode(input)
+    }
+
+    assertEquals(
+      go(None, schemaWithOption),
+      go(OptionalLike(None), schemaWithOptionEquivalent)
+    )
```
Baccata: Let's make this assertion consistent with the rest, as in change it to explicitly show a JSON payload. Whilst we're at it, let's also change it to cover exhaustively the changes in this fix, for instance : 

``` 
case class Options(option: Option[A], optional: Optional[A], refinedOptional: Optional[A], recursiveOption: Fix[Option])  
```

Baccata (approved):
Approving pre-emptively, assuming the necessary changes will be made before merging

modules/core/src/smithy4s/schema/FieldFilter.scala
kubukoz: You sure? It's not eagerly evaluating the lazy (hence the map)

Baccata: fair, but add a test to convince me :)

kubukoz: should be covered by the other comment, yeah

modules/json/test/src/smithy4s/json/JsonCodecApiTests.scala
kubukoz: done and done

PR #1666, opened 2025-03-25
Use the inputs when testing URI conversions in Http4sConversionSpec

bpholt:
Not a huge deal, but I discovered that `Http4sConversionSpec#http4sToSmithyAndBackUriTest` wasn't using its input in its test setup, so most of the tests were redundant. I believe this was the intended test behavior.

### DIFF
--- a/modules/http4s-kernel/test/src/smithy4s/http4s/kernel/Http4sConversionSpec.scala
+++ b/modules/http4s-kernel/test/src/smithy4s/http4s/kernel/Http4sConversionSpec.scala
@@ -39,7 +39,7 @@ object Http4sConversionSpec extends SimpleIOSuite {
   )
 
   http4sToSmithyAndBackUriTest(
-    uri"example.com",
+    uri"//example.com",
     uri"http://example.com/"
   )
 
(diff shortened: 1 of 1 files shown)


modules/http4s-kernel/test/src/smithy4s/http4s/kernel/Http4sConversionSpec.scala
```diff
@@ -39,7 +39,7 @@ object Http4sConversionSpec extends SimpleIOSuite {
   )
 
   http4sToSmithyAndBackUriTest(
-    uri"example.com",
+    uri"//example.com",
```
bpholt, line 42: Without the leading `//`, http4s parses this as the URI path, so the test fails because it was essentially trying to compare `http://localhost/example.com` to `http://example.com`.

kubukoz: i've had the "pleasure" of hitting this... yeah it's fun

kubukoz (approved):

kubukoz:
pretty sure the CLA isn't needed for a change like this

PR #1667, opened 2025-03-26
Improve handling of `@default(null)` and `@nullable`

### LINKED ISSUE #1581: Discrepancy of `null` handling when a default is present or not
Compare the following:

```smithy
structure Person {
  item: String = null
}

structure Person2 {
  item: String
}

structure Person3 {
  item: String = ""
}
```

Given a payload like `{"item": null}`, only `Person2` will be decoded successfully if you're using generated code.

Noteworthy:

- Dynamic will accept the input in all cases, but only `Person3` will have a field present with the default value. The others will simply not have that field.

Full repro: https://github.com/kubukoz/demos/tree/smithy4s-nulls-defaults (`sbt run` shows you all the results)

There's surely a bug here because Dynamic and Codegen don't work the same way, but which should it be? I'm leaning towards thinking that `= null` should be equivalent to nothing (see #1315), and I'm pretty sure all these cases should accept the input somehow.

ghostbuster91:

- [x] Updated dynamic module to match generated-code behaviour (it was the other way around this time)
- [ ] Added documentation
- [x] Updated changelog

This closes #1581


kubukoz:
dunno why now, but it looks like we got upgraded to the ubuntu 24.04 runners which don't have sbt, gonna have to install it manually now.

https://github.com/neandertech/langoustine/pull/199/files#r2008968475

modules/bootstrapped/test/src/smithy4s/DocumentSpec.scala
```diff
@@ -1003,15 +1003,16 @@ class DocumentSpec() extends FunSuite {
...
-    testFieldCombination(true, true, false)
-    testFieldCombination(true, false, true)
-    testFieldCombination(false, true, false)
+  for {
+    required <- List(true, false)
+    nullable <- List(true, false)
+    nullDefault <- List(true, false)
+  } yield {
+    test(
+      s"combinations of required, nullable, and null default ($required, $nullable, $nullDefault)"
```
ghostbuster91, line 1012: **note** this way it is clear when something fails which case is that.

Baccata:
@ghostbuster91, this ready for review ?

ghostbuster91:
> @ghostbuster91, this ready for review ? 

Yes

modules/bootstrapped/test/src/smithy4s/schema/DefaultValueSpec.scala
```diff
@@ -147,15 +147,16 @@ final class DefaultValueSpec extends FunSuite {
...
+
   private def testCaseOpt[A](schema: Schema[A], expect: Option[A])(implicit
       loc: Location
   ): Unit = {
     val sch = schema.addHints(smithy.api.Default(Document.DNull))
     val res = sch.getDefaultValue
     assertEquals(res, expect)
   }
-
-  private def testCase[A](schema: Schema[A], expect: A)(implicit
```
Baccata, line 158: Can we add tests with non-null documents while we're at it ? (Namely, `testCase[A](schema: Schema[A], document: Document, expect: A`)

Baccata:
Good work ! Just would like to take advantage of you making changes to this space to add tests to the `getDefault` function when there are actual non-null default values

modules/bootstrapped/test/src/smithy4s/schema/DefaultValueSpec.scala
ghostbuster91: sure thing :+1:

modules/bootstrapped/test/src/smithy4s/schema/DefaultValueSpec.scala
```diff
@@ -95,12 +140,11 @@ final class DefaultValueSpec extends FunSuite {
       Schema.int.optional[Foo]("y", _.y)
     )(Foo.apply)
     testCaseOpt(s, None)
-  }
-
-  test("union") {
```
ghostbuster91, line 100: **note** this case was invalid. `@default` trait cannot be applied on unions or their members

modules/bootstrapped/test/src/smithy4s/schema/DefaultValueSpec.scala
ghostbuster91: Done

modules/core/src/smithy4s/schema/Schema.scala
```diff
@@ -108,13 +108,12 @@ sealed trait Schema[A]{
...
-          case Nullable.Schema(_)  => 
-            Some(Nullable.Null.asInstanceOf[A])
-          case _ => None
-        }
-      case document => Document.Decoder.fromSchema(this).decode(document).toOption
+      case Document.DNull => NullableDefaultVisitor(this)
+      case document => 
+        println(document)
+        val res = Document.Decoder.fromSchema(this).decode(document)
+        println(res)
```
Baccata: tss tss tss

Baccata (approved):

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -1223,6 +1223,9 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     val defaultLine: Option[Line] = field.modifier match {
       // non-required fields with no default get a default of None
       case Field.Modifier(false, _, None) => Some(NameRef("scala.None").toLine)
+      // non-required, non-nullable fields with null default get a default of None
+      case Field.Modifier(false, false, Some(Field.Default(node, Some(_)))) if node == Node.nullNode =>
```
kubukoz, line 1227: **nitpick:** is it possible to do 

```suggestion
      case Field.Modifier(false, false, Some(Field.Default(Node.nullNode, Some(_)))) =>
```

?

CHANGELOG.md
```diff
@@ -33,6 +33,23 @@ Optics can still be rendered using metadata (global) and traits (selective), the
 
 The error type `smithy4s.http.UnknownErrorResponse` has been replaced with `smithy4s.http.RawErrorResponse`, which provides a more accurate description of an error response that failed to decode, including a full representation of the response code, headers, body and the discriminator if one was found.
 
+# ⚠️ Behavior Change: Handling of `@default(null)` and `@nullable`
```
kubukoz: **suggestion:** add a link to this pr

kubukoz (approved):

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
ghostbuster91: not really because `Node.nullNode` is a method that creates an instance of `NullNode`.
best I can do is:
```
case Field.Modifier(false, false, Some(Field.Default(_: NullNode, Some(_)))) =>
```

Then we should also change other places to make it consistent. Not sure if it is worth the noise. Better to make it a separate PR imo.

kubukoz: oh right, `nullNode` ain't a Scala `val`. ignore it then, we good.

kubukoz:
See you in the release 😭

kubukoz (commented):
forgor, the docs need updating - https://disneystreaming.github.io/smithy4s/docs/codegen/default-values#null-default

Issue #1663, opened 2025-03-20
Adjust URL-encoding rules


kubukoz:
We're relying on http4s to do our URLencoding for us, in both query and path parameters. However, it looks like this isn't following the smithy spec:

> Characters not defined as unreserved by [RFC 3986 section 2.3](https://datatracker.ietf.org/doc/html/rfc3986.html#section-2.3) MUST be [percent-encoded](https://datatracker.ietf.org/doc/html/rfc3986.html#section-2.1). That is, all characters except for alphanumerics and -._~.

https://smithy.io/2.0/spec/http-bindings.html#httplabel-serialization-rules

http4s skips escaping of more than that - e.g. `:`, which is most notably part of any AWS ARN, making it currently impossible to e.g. call Lambdas across accounts.

We should update this - do our own pathencoding, where we'll have control over which characters we escape.

Now, this can be an incompatible change that'll break some APIs if we just do it for everyone - would like to avoid that. I propose:

1. We do this for AWS protocols always
2. We enable it as opt-in in `SimpleRestJson`
3. Later - we switch to opt-out in `SimpleRestJson`, with a deprecated method
4. Even later - we remove the deprecated method

We could keep it forever, but since the spec says it's a MUST, I'd rather keep it consistent.

---

Bonus points - verify if our handling of queries is compliant.

bpholt:
Thanks for writing this up, @kubukoz!

bpholt:
FYI, I'm going to attempt picking this up. @kubukoz suggested in Discord:
> * the code we need to update is `toSmithy4sHttpUri` and `fromSmithy4sHttpUri` in `smithy4s.http4s.kernel` - they're duals of each other, will need to apply the same uri encoding consistently. Needs to be parameterized with a boolean or something equivalent, so we probably need an overload
> * you'll need to hardcode said boolean in `AwsClient`'s call to `fromSmithy4sHttpRequest` and `fromSmithy4sHttpRequest`
> * need to add a new builder method on `SimpleRestJson`, similar to `withMaxArity` and friends; sth like `.withCompliantLabelEncoding(bool)`, and said bool needs to be propagated all the way through the `SimpleRestJsonCodecs` to `toSmithy4sHttpRequest` / `fromSmithy4sHttpRequest` again. For the server _and_ client

Baccata:
@bpholt, FYI, I'm afraid I'm nowhere close to getting you the corporate CLA you've asked for :(

[...]

PR #1668, opened 2025-03-26
Correctly URI-encode HTTP Label values

bpholt:
I was a little surprised to find that I didn't need to modify `toSmithy4sHttpUri` to make this test pass, so I wanted to get some feedback before I get too much further into modifying `AwsClient` and `SimpleRestJson` as laid out in #1663.

Needs https://github.com/disneystreaming/alloy/pull/210

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -294,7 +294,11 @@ lazy val core = projectMatrix
       // https://github.com/disneystreaming/smithy4s/pull/1669
       ProblemFilters.exclude[IncompatibleMethTypeProblem](
         "smithy4s.http.HttpUnaryServerRouter.partialFunction"
-      )
+      ),
+      // originating in an Alloy update that removed ProtoCompactOffsetDateTime
+      ProblemFilters.exclude[MissingClassProblem]("alloy.proto.ProtoCompactOffsetDateTime"),
+      // originating in an Alloy update that removed ProtoCompactOffsetDateTime
+      ProblemFilters.exclude[MissingClassProblem]("alloy.proto.ProtoCompactOffsetDateTime$"),
     )
   )
   .jvmPlatform(allJvmScalaVersions, jvmDimSettings)
--- a/build.sbt
+++ b/build.sbt
@@ -286,6 +286,10 @@ lazy val core = projectMatrix
       ProblemFilters.exclude[DirectMissingMethodProblem](
         "smithy4s.http.HttpUnaryServerRouter#PartialFunctionRouter.this"
       ),
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        // this is a private case class so any problems are spurious
+        "smithy4s.http.HttpUnaryClientCodecs#HttpUnaryClientCodecsBuilderImpl.*"
+      ),
       // Breaking bin-compat to walk back ambiguous methods introduced in
       // https://github.com/disneystreaming/smithy4s/pull/1669
       ProblemFilters.exclude[IncompatibleMethTypeProblem](
--- a/modules/aws-http4s/src/smithy4s/aws/internals/AwsSigning.scala
+++ b/modules/aws-http4s/src/smithy4s/aws/internals/AwsSigning.scala
@@ -23,10 +23,7 @@ import cats._
 import cats.syntax.all._
 import fs2.Chunk
 import org.http4s._
-import org.http4s.Uri.Path.Segment
-import org.http4s.Uri.encode
 import org.http4s.client.Client
-import org.http4s.internal.CharPredicate.AlphaNum
 import org.typelevel.ci.CIString
 import smithy4s._
 import smithy4s.aws.kernel.AwsCrypto._
@@ -47,7 +44,6 @@ private[aws] object AwsSigning {
   val S3Control = "AWSS3ControlServiceV20180820"
 
   private val disableDoubleEncoding: Set[String] = Set(S3, S3Control)
-  private val dotSegments: Set[String] = Set("..", ".")
 
   def middleware[F[_]: Concurrent](
       awsEnvironment: AwsEnvironment[F]
@@ -134,43 +130,23 @@ private[aws] object AwsSigning {
           else {
             // Other services double encode and normalize
 
-            if (preparedRequest.uri.path.segments.exists(s => dotSegments.contains(s.encoded))) {
-              // if any of the segments are dot segments as described by RFC3986§3.3,
-              // defer to the Java path normalization algorithm, because that's what
-              // AWS expects. `org.http4s.Uri.removeDotSegments` sometimes removes too
-              // much; e.g. AWS expects a path of `foo/../..` to be normalized to `/..`;
-              // `removeDotSegments` returns `/` but `Uri#normalize` does the expected.
-              MonadThrow[F].catchNonFatal {
-                val uri = new URI(preparedRequest.uri.scheme.map(_.value).orNull, preparedRequest.uri.host.map(_.renderString).orNull, preparedRequest.uri.path.toAbsolute.renderString, preparedRequest.uri.fragment.orNull)
-                val almostNormalized = uri.normalize().getRawPath
+            MonadThrow[F].catchNonFatal {
+              val uri = new URI(
+                preparedRequest.uri.scheme.map(_.value).orNull,
+                preparedRequest.uri.host.map(_.renderString).orNull,
+                preparedRequest.uri.path.toAbsolute.renderString,
+                preparedRequest.uri.fragment.orNull
+              )
+              val almostNormalized = uri.normalize().getRawPath
 
-                val removeExtraneousTrailingSlash = almostNormalized.endsWith("/") &&
-                  !preparedRequest.uri.path.endsWithSlash &&
-                  almostNormalized.length > 1
-                if (removeExtraneousTrailingSlash)
-                  // normalization can add a final slash when we didn't have one
-                  // originally, so remove it if that's the case
-                  almostNormalized.substring(0, almostNormalized.length - 1)
-                else almostNormalized
-              }
-            } else {
-              // otherwise, let http4s handle normalization and encoding
-              val almostNormalized =
-                Uri
-                  .Path(
-                    preparedRequest.uri.path.toAbsolute.segments
-                      .map(_.encoded)
-                      .map(encode(_, toSkip = AlphaNum ++ "-_.~"))
-                      .map(Segment.encoded),
-                    absolute = true,
-                    endsWithSlash = preparedRequest.uri.path.toAbsolute.endsWithSlash
-                  )
-                  .normalize
-                  .toAbsolute
-
-              (if (preparedRequest.uri.path.endsWithSlash) almostNormalized.addEndsWithSlash
-               else almostNormalized).renderString
-                .pure[F]
+              val removeExtraneousTrailingSlash = almostNormalized.endsWith("/") &&
+                !preparedRequest.uri.path.endsWithSlash &&
+                almostNormalized.length > 1
+              if (removeExtraneousTrailingSlash)
+                // normalization can add a final slash when we didn't have one
+                // originally, so remove it if that's the case
+                almostNormalized.substring(0, almostNormalized.length - 1)
+              else almostNormalized
             }
           }
         }
--- a/modules/aws-http4s/src/smithy4s/aws/internals/AwsSigning.scala
+++ b/modules/aws-http4s/src/smithy4s/aws/internals/AwsSigning.scala
@@ -19,15 +19,19 @@ package internals
 
 import cats.effect.Concurrent
 import cats.effect.Resource
+import cats._
 import cats.syntax.all._
 import fs2.Chunk
 import org.http4s._
+import org.http4s.Uri.Path.Segment
+import org.http4s.Uri.encode
 import org.http4s.client.Client
+import org.http4s.internal.CharPredicate.AlphaNum
 import org.typelevel.ci.CIString
 import smithy4s._
 import smithy4s.aws.kernel.AwsCrypto._
 
-import java.net.URLEncoder
+import java.net.{URI, URLEncoder}
 import java.nio.charset.StandardCharsets
 
 /**
@@ -39,6 +43,11 @@ private[aws] object AwsSigning {
 
   // see https://raw.githubusercontent.com/awslabs/aws-sdk-kotlin/main/codegen/sdk/aws-models/s3.json
   val S3 = "AmazonS3"
+  // see https://raw.githubusercontent.com/awslabs/aws-sdk-kotlin/main/codegen/sdk/aws-models/s3-control.json
+  val S3Control = "AWSS3ControlServiceV20180820"
+
+  private val disableDoubleEncoding: Set[String] = Set(S3, S3Control)
+  private val dotSegments: Set[String] = Set("..", ".")
 
   def middleware[F[_]: Concurrent](
       awsEnvironment: AwsEnvironment[F]
@@ -94,7 +103,8 @@ private[aws] object AwsSigning {
 
     // S3 has special rules, in that it expects the X-Amz-Content-SHA256 to be set.
     val preSign: PreSigner[F] =
-      if (serviceName == S3) new PreSigner.S3InMemorySigned[F]
+      if (disableDoubleEncoding.contains(serviceName))
+        new PreSigner.S3InMemorySigned[F]
       else new PreSigner.Standard[F]
 
     val contentType = org.http4s.headers.`Content-Type`.headerInstance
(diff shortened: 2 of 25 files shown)


modules/aws-http4s/src/smithy4s/aws/internals/AwsSigning.scala
```diff
@@ -116,77 +126,124 @@ private[aws] object AwsSigning {
...
-              key.toString.toLowerCase + ":" + value.trim
+      preSign(request)
+        .mproduct[String] { case (_, preparedRequest) =>
+          if (disableDoubleEncoding.contains(serviceName))
+            // S3 and S3Control both disable double URL encoding and normalization.
+            preparedRequest.uri.path.toAbsolute.renderString.pure[F]
+          else {
+            // Other services double encode and normalize
+
+            if (preparedRequest.uri.path.segments.exists(s => dotSegments.contains(s.encoded))) {
```
bpholt: FYI, I think it would also be valid to simply use `java.net.URI#normalize` regardless of whether the path contains any dot segments. (The tests pass if I make that change.) I can switch to that if you'd prefer.

(That would also remove the reference to `org.http4s.internal.CharPredicate.AlphaNum` on line 163, which may be problematic since it appears to be an internal http4s API?)

bpholt:
@kubukoz @Baccata I think this is ready for review (CLA aside), at least for the portions that are specific to AWS. I think we agreed in Discord that the signing changes don't need to be configurable from Smithy traits, so I didn't make any changes in that direction. 

I wasn't sure if the changes to `fromSmithy4sHttpUri` and `fromSmithy4sHttpRequest` should be propagated out or not. Right now, I think `@simpleRestJson` would not use the new encoding rules (although it probably should), because `SimpleRestJsonCodecs#makeClientCodecs` doesn't set `encodeReservedCharacters` to true, but requests going to AWS will use the new rules since `AwsClient` does set that flag.

modules/aws-http4s/test/src-jvm/smithy4s/aws/internals/AwsSignatureTest.scala
```diff
@@ -89,12 +89,77 @@ object AwsSignatureTest extends SimpleIOSuite with Checkers {
...
+         |  - smithy4sRegion: $smithy4sRegion
+         |  - awsRequest: ${awsRequest.show}
+         |""".stripMargin
+    }
 
-    val genAwsRequest = for {
+    val genPathSegment: Gen[String] =
+      Gen
+        .frequency(
+          1 -> Gen.oneOf("..", "."),
```
bpholt, line 133: Path normalization should remove these when it's enabled, so these dot segments are sometimes added to the path to make sure that happens.

modules/aws-http4s/test/src-jvm/smithy4s/aws/internals/AwsSignatureTest.scala
```diff
@@ -89,12 +89,77 @@ object AwsSignatureTest extends SimpleIOSuite with Checkers {
...
+            .stringOf(
+              Gen.frequency(
+                9 -> Gen.oneOf(
+                  Gen.alphaNumChar,
+                  // this list from https://github.com/http4s/http4s/blob/ac5c5cc40f732df238ef38c8ef8f571b4e66bfe3/core/shared/src/main/scala/org/http4s/internal/UriCoding.scala#L26
+                  Gen.oneOf("-_.~".toIndexedSeq)
+                ),
+                1 -> Gen.oneOf(
+                  // this list from https://github.com/http4s/http4s/blob/ac5c5cc40f732df238ef38c8ef8f571b4e66bfe3/core/shared/src/main/scala/org/http4s/Uri.scala#L1069
+                  "!$&'()*+,;=:/?@".toIndexedSeq
```
bpholt, line 144: These are characters that should get encoded if they're present in a path segment and encoding is enabled.

modules/aws-http4s/test/src-jvm/smithy4s/aws/internals/AwsSignatureTest.scala
```diff
@@ -89,12 +89,77 @@ object AwsSignatureTest extends SimpleIOSuite with Checkers {
...
+            _,
+            toSkip = org.http4s.internal.CharPredicate.AlphaNum ++ "-_.~"
+          )
+        )
+
+    val genAwsRequest: Gen[SdkHttpFullRequest] = for {
       httpMethod <- Gen.oneOf(SdkHttpMethod.values().toList)
       host <- Gen.identifier
-      path <- Gen.listOfN(3, Gen.identifier).map(_.mkString("/"))
+      path <- Gen
```
bpholt, line 159: When I was working on this and trying to address some edge cases, I commented out this generator and used a variety of generators specific to the edge case I was working on. For example, things like `Gen.const("foo/../.")`, `Gen.const("this:might:be:an:aws:arn")`, etc. Unfortunately I didn't save the specifics, but they should be encoded in this generator.

bpholt:
Unfortunately I think this implementation escapes too much. After running the protocol compliance suite from Alloy and Smithy itself, I noticed that the spec actually says only values inserted into the URI [via the `httpLabel` trait](https://smithy.io/2.0/spec/http-bindings.html#id7) should actually be escaped in this way. Notably, literal values that were already in the URI should _not_ be escaped. 

(Or at least, that's how I'm now reading [the spec](https://smithy.io/2.0/spec/http-bindings.html#literal-character-sequences): "Patterns with no labels will match only requests containing the exact literal characters declared in the pattern", and then "Every label … MUST have the [`httpLabel` trait](https://smithy.io/2.0/spec/http-bindings.html#httplabel-trait)". Then, the serialization rules that describe escaping only apply to the `httpLabel` trait.)

I don't think we have enough information in `fromSmithy4sHttpUri` to make that distinction, so I'm taking a step back. My current thought is to try to change `smithy4s.http.internals.SchemaVisitorPathEncoder` and maybe `smithy4s.internals.SchemaVisitorPatternEncoder` to escape labelled values, using the escaping algorithm used by Smithy.

bpholt:
FWIW, I think we need to merge disneystreaming/alloy#210 before this will be fully tested by the Alloy compliance tests, but when I ran that locally before rebasing, it passed. (For some reason I can't get it to run now after pulling the latest updates; I get errors like "Unable to resolve trait `aws.protocols#ec2QueryName`" when executing `http4s / Test / managedResources` and `http4s / Test / envVars`. Maybe I'm not publishing Alloy correctly locally? I'm not very familiar with Mill.)

bpholt:
https://github.com/disneystreaming/smithy4s/blob/87897b8514c4e79074770197bd759f0b92971c8d/modules/aws-http4s/src/smithy4s/aws/AwsClient.scala#L91-L108

Other than `AWS_REST_JSON_1`, should these codec constructors use the HTTP label encoding changes I introduced, or should I leave the old behavior in place?

modules/aws-http4s/src/smithy4s/aws/internals/AwsSigning.scala
Baccata: yeah let's move :+1:, thank you for raising the concern

modules/core/src/smithy4s/internals/StructurePatternRefinementProvider.scala
```diff
@@ -67,7 +67,7 @@ object StructurePatternRefinementProvider {
           )
         )
     (input: A) => {
-      val result = encoder.encode(input)
+      val result = encoder.encode(input, false) // TODO should this encode?
```
Baccata: > // TODO should this encode? 

No need

modules/core/src/smithy4s/http/internals/PathEncode.scala
```diff
@@ -19,12 +19,14 @@ package smithy4s.http.internals
 import smithy4s.capability.Contravariant
 
 trait PathEncode[A] { self =>
-  def encode(a: A): List[String]
-  def encodeGreedy(a: A): List[String]
+  def encode(a: A, urlEncode: Boolean): List[String]
+  def encodeGreedy(a: A, urlEncode: Boolean): List[String]
```
Baccata: I'm not sure whether it makes sense to do it at this level to be honest. I'd rather it was a constructor parameter (in the `SchemaVisitorPathEncode` construct, for instance).

modules/core/src/smithy4s/http/internals/PathEncode.scala
```diff
@@ -55,4 +65,56 @@ object PathEncode {
...
+
+        case ((acc, Vector('%', '2')), 'F') if ignoreSlashes =>
+          (acc :+ '/') -> Vector.empty
+
+        case ((acc, Vector('%', '2')), 'F') if !ignoreSlashes =>
+          acc ++ Vector('%', '2', 'F') -> Vector.empty
+
+        case ((acc, lookback), c) =>
+          (acc ++ lookback :+ c) -> Vector.empty
+      }
```
Baccata: Would you mind rewriting it imperatively using a `Vector.builder` as an accumulator ? (Or even a StringBuilder)

modules/core/src/smithy4s/http/internals/SchemaVisitorPathEncoder.scala
```diff
@@ -78,7 +78,7 @@ object SchemaVisitorPathEncoder
       fields: Vector[Field[S, _]],
       make: IndexedSeq[Any] => S
   ): MaybePathEncode[S] = {
-    type Writer = S => List[String]
+    type Writer = (S, Boolean) => List[String]
```
Baccata: See https://github.com/disneystreaming/smithy4s/pull/1668/files#r2039441673

modules/http4s/test/src/smithy4s/http4s/SimpleRestJsonComplianceSuite.scala
```diff
@@ -95,7 +95,8 @@ object SimpleRestJsonComplianceSuite extends ProtocolComplianceSuite {
       val baseUri = uri"http://localhost"
       val suppliedHost =
         testHost.map(host => Uri.unsafeFromString(s"http://$host"))
-      SimpleRestJsonBuilder(service)
+      SimpleRestJsonBuilder
+        .apply(service)
```
Baccata: is that a syntactic preference or did something change in a source-incompatible manner ?

Baccata:
> Other than AWS_REST_JSON_1, should these codec constructors use the HTTP label encoding changes I introduced, or should I leave the old behavior in place? 

Let's leave it unchanged for now and revisit on a case-by-case basis (unless we find some evidence by analysis the spec that some protocols are inherently and unilateraly abiding by one rule)

Baccata:
A few comments, but great work overall !

modules/http4s/test/src/smithy4s/http4s/SimpleRestJsonComplianceSuite.scala
bpholt: I think I probably added it temporarily so I could more easily click through to the implementation, and then forgot to revert it. I'll switch it back.

modules/core/src/smithy4s/http/internals/PathEncode.scala
bpholt: I'm not sure how to implement it at the instance level and also [maintain the distinction between the old `path` and the new `encodedPath` methods in the implementation of `HttpEndpoint[I]`](https://github.com/Dwolla/smithy4s/blob/a0aad90f8b0b5b19e50417448b8aeb5467558a69/modules/core/src/smithy4s/http/HttpEndpoint.scala#L75-L76). 

I guess I could change `SchemaVisitorPathEncoder` from an object to a class, and then construct two `PathEncoder[I]`s in `HttpEndpoint.cast`, where `HttpEndpoint[I]`s are constructed, in case downstream code calls the old `path` method? Is that what you meant?

In addition to not disturbing the public API (most of this work has been to internal APIs, but `HttpEndpoint` is not marked as internal), we also need to maintain the distinction in behavior in order to continue the existing behavior in the AWS codecs other than `AWS_REST_JSON_1`, [as requested](https://github.com/disneystreaming/smithy4s/pull/1668#issuecomment-2796740816).

bpholt:
> > Other than AWS_REST_JSON_1, should these codec constructors use the HTTP label encoding changes I introduced, or should I leave the old behavior in place?
> 
> Let's leave it unchanged for now and revisit on a case-by-case basis (unless we find some evidence by analysis the spec that some protocols are inherently and unilateraly abiding by one rule)

FYI, it looks like `AWS_REST_XML` also requires encoding; without it, some of the compliance tests fail.

modules/core/src/smithy4s/http/internals/PathEncode.scala
bpholt: Addressed in 0c3df5ec7.

modules/core/src/smithy4s/http/internals/PathEncode.scala
bpholt: Addressed in fd284bd75.

Baccata:
Labelled as "DO NOT MERGE" while the CLA situation is being worked out

modules/core/src/smithy4s/http/internals/PathEncode.scala
Baccata: Thank you, that is indeed what I meant :+1:

modules/aws-http4s/test/src-jvm/smithy4s/aws/internals/AwsSignatureTest.scala
```diff
@@ -89,12 +89,77 @@ object AwsSignatureTest extends SimpleIOSuite with Checkers {
       awsRequest: SdkHttpFullRequest
   )
   object TestInput {
-    implicit val show: Show[TestInput] = Show.fromToString
+    implicit val showContentStreamProvider: Show[ContentStreamProvider] =
+      Show.show { csp =>
+        scala.io.Source.fromInputStream(csp.newStream()).mkString
+      }
+
+    implicit val showSdkHttpFullRequest: Show[SdkHttpFullRequest] = Show.show {
```
Baccata, line 97: this is much appreciated, thank you !

kubukoz:
ima put this back in draft instead of the scary label :P

Baccata (approved):
Happy with this in its current state. 

@bpholt has provided the signed corporate-CLA on behalf of Dwolla, which makes him exempt from having to sign the individual-CLA

@kubukoz, want to give it pass ?

modules/core/src/smithy4s/http/HttpEndpoint.scala
```diff
@@ -23,8 +23,14 @@ import smithy4s.http.internals.SchemaVisitorPathEncoder
 
 trait HttpEndpoint[I] {
   // Returns a list of path segments that should be appended to the base URL. These are not URL-encoded.
+  @deprecated(
+    "Path segments should be encoded in a specific way per the Smithy spec, so use encodedPath instead",
+    "v0.18.34"
```
kubukoz: it's gonna be .40 now, and maybe without the `v`

modules/core/src/smithy4s/http/HttpEndpoint.scala
```diff
@@ -61,15 +67,23 @@ object HttpEndpoint {
           )
         )
 
-      encoder <- SchemaVisitorPathEncoder(
+      labelEncodingEncoder <- new SchemaVisitorPathEncoder(true)(
```
kubukoz: suggestion: let's use a named param for the boolean

modules/core/src/smithy4s/http/HttpEndpoint.scala
kubukoz: also, should we actually deprecate it, if we keep using `path` in the case of `if rawHttpLabelValues`? I suppose the plan is to remove `rawHttpLabelValues` later and make `true` the default?

kubukoz (commented):
please add an entry in the changelog file, it'll be 0.18.40 at this point

modules/core/src/smithy4s/http/HttpEndpoint.scala
bpholt: Yes, I think that's the plan.

modules/core/src/smithy4s/http/HttpEndpoint.scala
```diff
@@ -24,8 +24,14 @@ import scala.annotation.tailrec
 
 trait HttpEndpoint[I] {
   // Returns a list of path segments that should be appended to the base URL. These are not URL-encoded.
+  @deprecated(
+    "Path segments should be encoded in a specific way per the Smithy spec, so use encodedPath instead",
+    "0.18.40"
+  )
   def path(input: I): List[String]
 
+  def encodedPath(input: I): List[String]
```
bpholt, line 33: I'm a little surprised this new method didn't trigger a MiMa error, but I didn't see anything that would have suppressed it, so I'm hoping it's ok.

build.sbt
```diff
@@ -286,11 +286,19 @@ lazy val core = projectMatrix
...
+        "smithy4s.http.HttpUnaryClientCodecs#HttpUnaryClientCodecsBuilderImpl.*"
+      ),
       // Breaking bin-compat to walk back ambiguous methods introduced in
       // https://github.com/disneystreaming/smithy4s/pull/1669
       ProblemFilters.exclude[IncompatibleMethTypeProblem](
         "smithy4s.http.HttpUnaryServerRouter.partialFunction"
-      )
+      ),
+      // originating in an Alloy update that removed ProtoCompactOffsetDateTime
+      ProblemFilters.exclude[MissingClassProblem]("alloy.proto.ProtoCompactOffsetDateTime"),
```
bpholt, line 302: These came in with the Alloy update. I'm not sure if filtering them out was the right call or not.

kubukoz: oh, I see now. The trait is no longer being generated...

looks good.

modules/core/src/smithy4s/http/HttpEndpoint.scala
kubukoz: I don't get it either, tbh. At least the hope is that nobody creates HttpEndpoint manually (as in, not via its `unapply` and `cast`)

btw. maybe we should seal this trait.

kubukoz: I tried a similar case and it emits `ReversedMissingMethod`. We actually have a filter for all of these (`ProblemFilters.exclude[ReversedMissingMethodProblem]("smithy4s.*")` in `project/MimaVersionPlugin.scala`) because it's a forward-compatibility problem.

personally I don't buy it, it's a binary risk like any other. What saves us is that nobody else should be implementing `HttpEndpoint`, as per the comment above.

project/Dependencies.scala
```diff
@@ -31,7 +31,7 @@ object Dependencies {
 
   val Alloy = new {
     val org = "com.disneystreaming.alloy"
-    val alloyVersion = "0.3.26"
+    val alloyVersion = "0.3.28-1-4a6c61"
```
kubukoz: **issue:** just noting to ourselves to use a stable number before we merge

modules/core/src/smithy4s/http/HttpEndpoint.scala
bpholt: Yep, makes sense. I missed `MimaVersionPlugin`.

> personally I don't buy it, it's a binary risk like any other. What saves us is that nobody else should be implementing HttpEndpoint, as per the comment above.

Are you suggesting I change something so that it wouldn't emit a `ReversedMissingMethod` error if they weren't being suppressed? I'm not sure what the best way to do that is; none of the options I can think of are very good. (e.g. use `???`, or delegate to `path`)

kubukoz: I wasn't suggesting any particular action in this case, just a more general wondering...ment for the other maintainers (@Baccata I'm looking at you) :)

in this case I think we're pretty safe because in practice this is "kinda sealed"

bpholt: Ok. I made the changes necessary to seal `HttpEndpoint`, but I'm happy to revert them if that's not the path we want to take.

kubukoz: oh I wasn't suggesting to actually seal it yet.

I would probably keep it open - the reasoning is, if someone has a custom implementation, their experience will be:

1. runtime linkage error
2. they add the missing implementation, whatever it is
3. no more error

whereas if we seal it now, they may be unable to build with the new smithy4s at all.

Not super realistic, but who knows. I'd leave this kind of change for 0.19

bpholt: Ok, works for me. I unsealed it.

kubukoz (approved):
just waiting for https://github.com/disneystreaming/alloy/actions/runs/16455931297 now so that we can bump alloy here

project/Dependencies.scala
bpholt: ✅

kubukoz (approved):

PR #1669, opened 2025-03-31
Encode error before applying middleware

zakolenko:

- [x] Added documentation
- [x] Updated changelog


modules/core/src/smithy4s/http/HttpUnaryServerRouter.scala
```diff
@@ -37,7 +37,8 @@ object HttpUnaryServerRouter {
       endpointMiddleware: Endpoint.Middleware[Request => F[Response]],
       getMethod: Request => HttpMethod,
       getUri: Request => HttpUri,
-      addDecodedPathParams: (Request, PathParams) => Request
+      addDecodedPathParams: (Request, PathParams) => Request,
+      encodeErrorsBeforeMiddleware: Boolean
```
kubukoz, line 54: **issue:** we'd need to support this in a binary compatible way, but it's a good start

modules/core/src/smithy4s/http/HttpUnaryServerRouter.scala
```diff
@@ -31,6 +31,28 @@ object HttpUnaryServerRouter {
...
+      makeServerCodecs,
+      endpointMiddleware,
+      getMethod,
+      getUri,
+      addDecodedPathParams,
+      encodeErrorsBeforeMiddleware = false
+    )
+  }
+
+  def v2[Alg[_[_, _, _, _, _]], F[_], Request, Response](
```
zakolenko: I’m not a fan of this approach either, but I wasn’t able to overload `apply` above - likely due to the second argument list

If you know a better way to maintain binary compatibility or can suggest a more suitable method name, I’d really appreciate it

Baccata: The only reason for the currying is the path-dependant `Impl` type (on the `impl` parameter). Therefore, you could simply add the boolean parameter on the first parameter set, which would allow to overload to `apply`

modules/core/src/smithy4s/server/UnaryServerEndpoint.scala
```diff
@@ -29,6 +29,16 @@ object UnaryServerEndpoint {
...
+  )(implicit F: MonadThrowLike[F]): Request => F[Response] = {
+    apply(interpreter, endpoint, codecs, middleware, encodeErrorsBeforeMiddleware = false)
+  }
+
+  def apply[F[_], Op[_, _, _, _, _], Request, Response, I, E, O, SI, SO](
+      interpreter: FunctorInterpreter[Op, F],
+      endpoint: Endpoint[Op, I, E, O, SI, SO],
+      codecs: UnaryServerCodecs[F, Request, Response, I, E, O],
+      middleware: (Request => F[Response]) => (Request => F[Response]),
+      encodeErrorsBeforeMiddleware: Boolean = false
```
Baccata: let's not add a default value : the overload already plays this role and the default value will make it harder to add more parameters down the line, should the need arise

modules/http4s/src/smithy4s/http4s/SimpleProtocolBuilder.scala
```diff
@@ -178,12 +199,30 @@ abstract class SimpleProtocolBuilder[P](
...
 
     def middleware(
         mid: ServerEndpointMiddleware[F]
     ): RouterBuilder[Alg, F] =
-      new RouterBuilder[Alg, F](service, impl, errorTransformation, mid)
+      middleware(mid, encodeErrorsBeforeMiddleware = false)
+
+    def middleware(
+        mid: ServerEndpointMiddleware[F],
+        encodeErrorsBeforeMiddleware: Boolean
```
Baccata: let's make a dedicate method to set this up as opposed to an additional parameter on the middleware method.

modules/http4s/src/smithy4s/http4s/SimpleProtocolBuilder.scala
```diff
@@ -178,12 +199,35 @@ abstract class SimpleProtocolBuilder[P](
...
-      new RouterBuilder[Alg, F](service, impl, errorTransformation, mid)
+      new RouterBuilder[Alg, F](
+        service,
+        impl,
+        errorTransformation,
+        mid,
+        encodeErrorsBeforeMiddleware = false
+      )
+
+    def middlewareWithEncodedErrors(
```
Baccata: Sorry if was I meant wasn't clear : what I asked for you to add : 

```scala
def encodeErrorsBeforeMiddleware(value: Boolean) : RouterBuilder[Alg, F]
```

the `middleware` method should not change the value of the boolean

zakolenko: @Baccata , no, it’s me who got it wrong
Thank you! will do as you proposed

modules/http4s/src/smithy4s/http4s/SimpleProtocolBuilder.scala
```diff
@@ -156,7 +171,13 @@ abstract class SimpleProtocolBuilder[P](
     def mapErrors(
         fe: PartialFunction[Throwable, Throwable]
     ): RouterBuilder[Alg, F] =
-      new RouterBuilder(service, impl, fe andThen (e => F.pure(e)), middleware)
+      new RouterBuilder(
```
kubukoz: **nitpick:** we could have a private `copy` method with default parameters everywhere, to avoid changing all these calls in the builder methods every time we add something. Would you mind adding that?

zakolenko: sure, thank you! - [done](https://github.com/disneystreaming/smithy4s/pull/1669/commits/79442d2e74b21753327e034c9aacaf86b8615e6f)

modules/http4s/src/smithy4s/http4s/SimpleProtocolBuilder.scala
zakolenko: [done](https://github.com/disneystreaming/smithy4s/pull/1669/commits/9491398f83ecf141a931b52aabb25ca90efebaa2)

Baccata (approved):
@kubukoz, I'll wait for your approval to merge

Baccata:
@zakolenko, actually, would you mind adding a line in the Changelog at the root of the project ?

modules/http4s/src/smithy4s/http4s/SimpleProtocolBuilder.scala
```diff
@@ -131,11 +131,26 @@ abstract class SimpleProtocolBuilder[P](
...
+      encodeErrorsBeforeMiddleware: Boolean
   )(implicit
       F: Concurrent[F]
   ) {
 
+    def this(
+        service: smithy4s.Service[Alg],
+        impl: FunctorAlgebra[Alg, F],
+        errorTransformation: PartialFunction[Throwable, F[Throwable]],
+        middleware: ServerEndpointMiddleware[F]
```
kubukoz: The constructor of `RouterBuilder` is package-private, I think we don't need this auxiliary one. Just have to add a MiMa exclusion to the build definition if it shouts at you.

modules/http4s/src/smithy4s/http4s/SimpleProtocolBuilder.scala
```diff
@@ -178,12 +193,25 @@ abstract class SimpleProtocolBuilder[P](
...
+
+    /**
+      * Configures whether errors that are in the smithy spec should be encoded into the HTTP response
+      * before the middleware is applied
+      *
+      * - If `true`, errors defined in the smithy spec will be encoded into the
+      *   response before middleware is applied. Non-Smithy errors will remain unencoded
+      *   and will be visible from [[middleware]].
+      * - If `false`, all errors (both smithy and non-smithy) will pass through the
+      *   middleware unencoded.
```
kubukoz, line 199: nitpick: might wanna mention in the Scaladoc that the default is false

kubukoz (approved):
just two final nitpicks

modules/http4s/src/smithy4s/http4s/SimpleProtocolBuilder.scala
zakolenko: Thank you, missed this

done - https://github.com/disneystreaming/smithy4s/pull/1669/commits/e542761625ab0302f0ef0debbb2600a6e1ea54ee

`PartialFunctionRouter` and `KleisliRouter` also defined as `private` so I've added a MiMa exclusion for them as well [here](https://github.com/disneystreaming/smithy4s/pull/1669/commits/f1239f0c61dedfec9873812ef8bcd061d8fc7757)

PR #1673, opened 2025-04-01
Support transitive mixins in `@adt`

### LINKED ISSUE #1312: Should transitive mixins be taken into account in `@adt`?
If you use `@adt` and all of the members apply a mixin directly, you get a sealed trait that also extends that mixin:

```smithy
$version: "2"

namespace input

use smithy4s.meta#adt

@adt
union Items {
  s1: Item1
  s2: Item2
}

@mixin
structure HasStuff {
  stuff: String
}

structure Item1 with [HasStuff] {}
structure Item2 with [HasStuff] {}
```

generates an `Items` that starts with:

```scala
sealed trait Items extends HasStuff with scala.Product with scala.Serializable {
```

However, if you insert a mixin between ≥1 of the members and the desired mixin:

```diff
$version: "2"

namespace input

use smithy4s.meta#adt

@adt
union Items {
  s1: Item1
  s2: Item2
}

@mixin
structure HasStuff {
  stuff: String
}

+ @mixin
+ structure HasHasStuff with [HasStuff] {}

- structure Item1 with [HasStuff] {}
+ structure Item1 with [HasHasStuff] {}
structure Item2 with [HasStuff] {}
```

then you don't see `HasStuff` in the trait definition:

```scala
sealed trait Items extends scala.Product with scala.Serializable {
```

Should this be supported by "flattening" the mixin dependencies, or should the models be updated to avoid transitive mixins if we want to use `@adt` as designed?

kubukoz:
Closes #1312.

- [x] Added documentation
- [x] Updated changelog


modules/bootstrapped/src/generated/smithy4s/example/AdtUnionWithTransitiveAndDirectMixins.scala
```diff
@@ -0,0 +1,91 @@
...
+
+import smithy4s.Hints
+import smithy4s.Schema
+import smithy4s.ShapeId
+import smithy4s.ShapeTag
+import smithy4s.schema.Schema.long
+import smithy4s.schema.Schema.struct
+import smithy4s.schema.Schema.union
+
+sealed trait AdtUnionWithTransitiveAndDirectMixins extends AdtMixinOne with TransitiveMixin with scala.Product with scala.Serializable { self =>
```
kubukoz, line 11: This is an interesting one: in both members, `AdtMixinOne` can be reached both directly AND transitively via `TransitiveMixin`. The current implementation adds them both here, which _should_ be OK, although it would be enough for the union to just specify `TransitiveMixin`.

I think this is OK, we're doing something slightly redundant for the benefit of having a simpler implementation in `SmithyToIR`.

Baccata (approved):

PR #1677, opened 2025-04-02
Open unions via `alloy#jsonUnknown`

### LINKED ISSUE #1585: JSON: Support open unions
Currently, we have a discrepancy between:

- Smithy's documentation stating that union shapes should be considered open, and
- Smithy4s's decision to prefer strict type safety and exhaustive pattern matching, as per Scala best practices.

We already support open enums, and the maintainers are in agreement that open unions should be supported in _some_ way, as an opt-in mechanism.

The current thinking is that we should extend the [`alloy#jsonUnknown` trait in Alloy](https://github.com/disneystreaming/alloy/blob/main/docs/serialisation/json.md#unknown-fields) to work on unions.

The desired semantics are as follows:

- A union member can be marked with `jsonUnknown`. This is what we call the catch-all case from now on.
  - [x] only one member can be marked with the trait (~~`structurallyExclusive` should do the job here~~ - it doesn't, but this is added via a custom java validator)
  - [x] ❓ ~~To be confirmed: only Document or any document shape?~~ Any document shape.
- The deserialization process will wrap the **entire document at the cursor** in the catch-all case if it doesn't find a matching tag (in tagged unions), discriminator (in discriminated unions) or if neither member matches (in untagged unions)
- Serialization of the catch-all case will write the entire document back.
- This will be supported in JSON and Document (de)serialization only, as the trait s protocol-specific.

This would be a runtime-only change, meaning that codegen isn't going to be aware of the trait (only the interpreters/codecs would)

If I missed something, please let me know.

kubukoz:
Closes #1585.

Alloy counterpart: https://github.com/disneystreaming/alloy/pull/208

Major TODOs:

- [x] Document support
  - [x] tagged unions
  - [x] discriminated unions
- [x] JSON support
  - [x] tagged unions
  - [x] discriminated unions
- [x] Documentation
- [x] alloy's own checklist in https://github.com/disneystreaming/alloy/pull/208
- [x] Changelog
- [x] Release the Alloy change and wait for this to turn green

CI is failing because snapshots are not working in the sbt/mill plugin. It'd be a bit tedious at the moment, because each scripted test / mill test would need to register a Sonatype Snapshots resolver. However, you can see that e.g. JS on 2.13 is passing - which includes the new protocol tests, yay.

- [x] Added alloy compliance tests (when simpleRestJson protocol behaviour is expanded/updated)
- [x] Added documentation
- [x] Updated changelog


modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -1040,27 +1064,39 @@ private[smithy4s] class SchemaVisitorJCodec(
   )(dispatch: Alt.Dispatcher[U]): JCodec[U] =
     new TaggedUnionJCodec[U](alternatives)(dispatch) {
 
-      def decodeValue(cursor: Cursor, in: JsonReader): U =
+      def decodeValue(cursor: Cursor, in: JsonReader): U = {
```
Baccata, line 1061: I'd rather we didn't change the existing implementation but rather created a new one specifically for the case where `JsonUnknown` is present

Baccata: Although, maybe not ...

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -1040,27 +1064,39 @@ private[smithy4s] class SchemaVisitorJCodec(
...
+              cursor.pop()
+              if (in.isNextToken('}')) result
+              else {
+                in.rollbackToken()
+                in.decodeError(s"Expected no other field after $key")
+              }
+            } else {
+              // ! - I'm not sure about this
+              in.rollbackToMark()
+              unknownTagHandler(cursor, in)
```
Baccata, line 1085: I think you need to check for non-nullity of `unknownTagHandler` here

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -1040,27 +1064,39 @@ private[smithy4s] class SchemaVisitorJCodec(
   )(dispatch: Alt.Dispatcher[U]): JCodec[U] =
     new TaggedUnionJCodec[U](alternatives)(dispatch) {
 
-      def decodeValue(cursor: Cursor, in: JsonReader): U =
+      def decodeValue(cursor: Cursor, in: JsonReader): U = {
+        // ! - I'm not sure about this
+        in.setMark()
```
Baccata, line 1062: :+1:

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -1040,27 +1064,39 @@ private[smithy4s] class SchemaVisitorJCodec(
...
+              val result = handler(cursor, in)
+              cursor.pop()
+              if (in.isNextToken('}')) result
+              else {
+                in.rollbackToken()
+                in.decodeError(s"Expected no other field after $key")
+              }
+            } else {
+              // ! - I'm not sure about this
+              in.rollbackToMark()
```
Baccata, line 1084: makes sense to me

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -1193,7 +1210,8 @@ private[smithy4s] class SchemaVisitorJCodec(
             in.rollbackToMark()
             in.rollbackToken()
             cursor.push(key)
-            val handler = handlerMap.get(key)
+            var handler = handlerMap.get(key)
+            if (handler eq null) handler = unknownTagHandler
```
Baccata, line 1211: 👍

modules/core/src/smithy4s/internals/DocumentDecoderSchemaVisitor.scala
```diff
@@ -545,22 +537,45 @@ class DocumentDecoderSchemaVisitor(
...
+          label -> decoder
         }
-        jsonLabel(alt) -> encoder
-      }.toMap
+        .toMap
 
     hints match {
       case Discriminated.hint(discriminated) =>
-        discriminatedUnion(discriminated, decoders)
+        discriminatedUnion(discriminated, decoders, handleUnknownTag)
```
Baccata, line 574: 👍

modules/core/src/smithy4s/internals/DocumentDecoderSchemaVisitor.scala
```diff
@@ -545,22 +537,45 @@ class DocumentDecoderSchemaVisitor(
...
       case Untagged.hint(_) =>
         untaggedUnion(decoders)
       case _ =>
-        taggedUnion(decoders)
+        taggedUnion(decoders, handleUnknownTag)
```
Baccata, line 578: 👍

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
kubukoz: not necessarily, if it's null then `in.discriminatorValueError(key)` kicks in, in the previous conditional

kubukoz: I'll rewrite these two ifs to just one, if that helps readability - will compare

Baccata: right yeah okay I see

kubukoz: turned into a single `if` with 3 branches 65e916b1d3aeca5cef56c5f603dba48961656cf0

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
kubukoz: sticking to a shared impl for the time being... Especially in the context of tagged unions I don't expect there to be a lot of overhead, the unhappy path has one extra `else if` but that should be it?

modules/docs/resources/markdown/04-codegen/02-unions.md
```diff
@@ -149,6 +149,41 @@ are  encoded as such
...
 
+#### Open unions
+
+By default, `union` shapes in Smithy4s are considered closed.
+
+This means that any tags/discriminators that are not listed in the union's Smithy definition, will not be accepted as valid values in deserialization. As such, adding/removing values to such a union definition is a breaking change.
+
+This behavior is at odds with the current [specification of Smithy](https://smithy.io/2.0/guides/evolving-models.html#updating-unions) as of April 2025:
+
+> Unions in Smithy are considered "open"; it is a backward-compatible change to add new members to a union
```
Baccata: I don't like this way of introducing it : this bit of specification is borderline nonsensical. 

What this PR solves is a **PROTOCOL-SPECIFIC** problem, not a problem with how we treat the generated code. As a matter of fact, talking about backward/forward compatibility is very much a protocol-specific notion. 

I think this bit of documentation that talks about how unions are serialised should generally go to a JSON specific section of the docs, but that's a problem for another time. What I'm saying is that you shouldn't tie this PR back to what AWS states about union backward compat.

modules/core/src/smithy4s/Document.scala
```diff
@@ -63,6 +63,8 @@ sealed trait Document extends Product with Serializable {
       value.map { case (k, v) => k + "=" + v.show }.mkString("{", ", ", "}")
   }
 
+  // Nests this document under the given key.
```
Baccata: Let's turn that into a proper bit of scaladoc

Baccata:
I'm happy with it, just some nitpicks about the documentation

modules/docs/resources/markdown/04-codegen/02-unions.md
kubukoz: > you shouldn't tie this PR back to what AWS states about union backward compat.

I still think it's worth a mention/link in codegen docs somewhere. To us it's a protocol-specific concern, but the spec is talking about model evolution in general, i.e. doesn't quantify which protocols it applies to. Someone with past Smithy exposure will be surprised to see we handle them differently by default

Baccata: Sure, but then how does AWS address this notion of open-unions in their own code-generators ? Cause that's what we should point out when we talk about this

kubukoz: smithy-java generates an `$Unknown` member by default:

image

modules/core/src/smithy4s/Document.scala
kubukoz: c243c7a1f748ae493bbec70d73f66751205d000c

modules/docs/resources/markdown/04-codegen/02-unions.md
Baccata: what's interesting is what's in the body of `$UnknownMember`, I wish you had screenshotted that 😅 . 

I'm gonna go with the assumption that they capture the JSON payload (like you do in this PR). The problem is that "open-unions" only make sense if code-generators are protocol specific. 

As long as you add that the wording of this bit of doc (ie open-unions don't make sense in protocol-agnostic solutions), I'm fine with keeping these docs here until we move them elsewhere. 

It means changing the `by default, `union` shapes in Smithy4s are considered closed.`, because "open" and "closed" are nonsensical in protocol-agnostic solutions.

kubukoz: since you asked...

image

image

it looks like they support decoding unknown members, but without keeping the actual structure that was decoded - just the member name. And you can't re-encode it back.

kubukoz: by the way, the above encoding in smithy-java sounds protocol-specific too, even though _that_ codegen is supposed to be agnostic - I suppose in some protocols you won't even have a member name to provide (say, if there was a way to encode open unions in protobuf)

kubukoz: tried to make it less controversial with ab5e6f9d7bf628e7e98e6228fcb7119443fe1cb6

Baccata (approved):

Baccata:
Squash it please :)

PR #1678, opened 2025-04-03
Sort hints before rendering, take A

### LINKED ISSUE #1660: Non-deterministic hint rendering
Something I noticed while working on #1573 - when I generate bootstrap code locally, I see some file changes. I'd previously encountered this in another codegen, and it's most likely caused by https://github.com/smithy-lang/smithy/issues/2492.

We should probably just sort traits by shapeId before rendering.

kubukoz:
Closes #1660 

This solution is more complex, preferring to keep the existing `Hint` structure as-is and adding ShapeId on top. See #1679 for take B

- [ ] Added documentation
- [ ] Updated changelog


kubukoz:
alright, this is wayyyyy more complex than the other solution, now that I realized we only render native hints anyway.

Issue #1628, opened 2024-12-18
Direct access to structure fields


kubukoz:
It would be "cool" (and probably useful in a bunch of ways) to have access to a particular `Field` instance, knowing the member name.

For example:

```smithy
structure Hello {
  @required foo: String = "woop"
}
```

```scala
println(Hello.fields.foo.getDefaultValue: Option[String])
```

This would sort of mirror how you can already access the individual `alt`s of a generated union shape.

meldmy:
We need this feature in our project to avoid duplicating default values in the code base having a smithy with default values

kubukoz:
@Baccata, thoughts on having this? POC in #1683.

Baccata:
Can either of you elaborate with a real usecase behind this ask, as in one that can't be easily solved via schema visitors, etc ? Like, why do you need the default value out of the context of serialisation, for instance ?

meldmy:
For example, we have content with a default maxSongs value defined for a page. This value is also included as a parameter in our domain events, and the UI app allows for increasing/decreasing this value. When we generate the event as a producer, we need to know how many items to calculate if the user hasn’t overridden the default.

In this case, our producer application would need to hardcode the default maxSongs value in its own codebase, rather than relying on the latest Smithy contract and its constraints. This duplication introduces potential risk and inconsistency

kubukoz:
@meldmy maybe this is a bit of an XY problem - the root of what you're trying to do is having a way to use the default value of the parameter, which is already present in the code, and maybe even has a default value in the constructor:

```smithy
structure Foo {
  @required requiredParam: String
  @required requiredWithDefault: String = "welp"
}
```

```scala
case class Foo(requiredParam: String, requiredWithDefault: String = "welp")
```

Would it be good enough for you to have an additional constructor that you can call, which will take `None` as "use the default if present"? In other words, what if we generated these additionally:

```scala
//object Foo {
def applyWithDefaults(requiredParam: String, requiredWithDefault: Option[String] = None): Foo =
  Foo(requiredParam, requiredWithDefault.getOrElse("welp"))
```

Then, when you want to use the default value you'd pass `None`, and otherwise you'd pass whatever value you intended to pass.

PR #1683, opened 2025-04-04
POC: direct access to fields

kubukoz:
POC for #1628.

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
@@ -705,7 +705,7 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
           case Field.TypeModification.None if modifier.required     => line"$tpe"
           case Field.TypeModification.None                          => line"$tpe"
           case Field.TypeModification.Option                        => line"$option[$tpe]"
-          case Field.TypeModification.Nullable if modifier.required => line"$tpe"
+          case Field.TypeModification.Nullable if modifier.required => line"Nullable[$tpe]"
           case Field.TypeModification.Nullable                      => line"Nullable[$tpe]"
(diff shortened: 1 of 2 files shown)


modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -782,6 +794,13 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
           line"${schemaImplicit}val schema: $Schema_[${product.nameRef}] = $constant_(${product.nameRef}()).withId(id).addHints(hints)"
         },
         renderTypeclasses(product.hints, product.nameRef),
+        newline,
+        // todo: skip if no fields
+        block(line"object fields") {
+          renderedFields.map { case (name, tpe, text) =>
+            line"val $name: $Field_[${product.nameRef}, $tpe] = $text"
```
kubukoz, line 801: **issue:** first culprit

image

that is indeed a bit unfortunate... and fixing the imports will probably be pretty tricky

kubukoz: that seems to be the only thing in `bootstrapped` itself, 47 of those to be precise (other primitives are also involved)

kubukoz: Core is happy:

image

although I'm starting to become concerned about class size limits on bigger structs.

Baccata: Yeah, not a fan of having this, without a proper usecase. 

Regarding parity with alts, I'd actually would rather the alts were rendered directly within the expression of the union schemas (literally going the other direction).

If we're gonna proceed : could you kindly run some metrics on, say, the EC2 spec (which is huge), to compute the difference in compilation time, and in the size of the jar before and after this change ? 

Alternatively, we could make that rendering an opt-in

PR #1691, opened 2025-04-11
Don't use -release flag

kubukoz:
Workaround/solution for https://github.com/scalameta/metals/issues/7359

Also, `-release` was deprecated in Scala 2.12 because:

> `-target is deprecated: Scala 2.12 cannot emit valid class files for targets newer than 8 (this is possible with Scala 2.13). Use -release to compile against a specific platform API version.`

The `-java-output-version` flags work on both 2.13 and 3.x:

image

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/project/Smithy4sBuildPlugin.scala
+++ b/project/Smithy4sBuildPlugin.scala
@@ -245,9 +245,9 @@ object Smithy4sBuildPlugin extends AutoPlugin {
   }
 
   def targetScalacOptions(scalaVersion: String) =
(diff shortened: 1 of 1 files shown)


kubukoz:
hmm no, this doesn't actually help the Metals case. But still, worth doing.

kubukoz:
or not. Im gonna close this for now, will revive when/if metals catches up on the bug

PR #1695, opened 2025-04-14
Publish Documentation in JAR

lewisjkl:
The reason I split the mdoc stuff into a separate sub-project was to avoid all of the dependencies in the published doc jar. Let me know if you all are cool with this or would prefer an alternative approach.

- [ ] Added documentation
- [ ] Updated changelog


build.sbt
```diff
@@ -81,10 +82,10 @@ lazy val allModules = Seq(
   complianceTests
 ).flatMap(_.projectRefs)
 
-lazy val docs =
+lazy val docsValidation =
```
kubukoz: **nitpick:** I'd look for a different name, like `docsForMdoc`, as it's more than validation

or maybe `docsRendering`, something of the like

kubukoz (approved):
i'm cool with it, just a local naming thingy

build.sbt
lewisjkl: Yeah fair point, will update

PR #1702, opened 2025-04-17
Fix #1699

### LINKED ISSUE #1699: Mixin not rendered even though the fields match in type
Occurs in 0.18.33.

Related: #425, #1510.

```smithy
$version: "2"

namespace hello

@mixin
structure MyMixin {
    @required
    description: String
}

structure StructWithDefault with [MyMixin] {
    @required description: String = "oh noes"
}

```

Generates:

```scala
final case class StructWithDefault(description: String = "oh noes")

...

trait MyMixin {
  def description: String
}
```

whereas `StructWithDefault extends MyMixin` is expected.

I believe this is because the `doFieldsMatch` check is too restrictive, and considers the presence/absence of a default value significant, even in situations when it doesn't affect the type of the field:

https://github.com/disneystreaming/smithy4s/blob/ad2b7c2e13c529bf92308c412cc6f2f8fd726380/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala#L212-L218

so unless there are good reasons not to, we should relax that constraint and only check if the relevant part (optionality/nullity).

Maybe it's just a matter of comparing the `.typeMod`?

kubukoz:
Closes #1699

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Thank you!
 
 # 0.18.34
 
+* codegen: Consider a mixin field "matching in type" even if it differs by default values in [#1702](https://github.com/disneystreaming/smithy4s/pull/1702).
 * codegen-cli: Ensure the command returns a failing exit code if command line arguments aren't valid in [#1694](https://github.com/disneystreaming/smithy4s/pull/1694).
 
 # 0.18.33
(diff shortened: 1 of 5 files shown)


kubukoz:
actually, I'll do this in #1701 - there's a dependency in a test I'm about to push.

PR #1707, opened 2025-04-29
Isolation via null parent ClassLoader

kubukoz:

- [ ] Added documentation
- [ ] Updated changelog


modules/codegen-cli/src/smithy4s/codegen/cli/Main.scala
```diff
@@ -36,27 +39,55 @@ object Main {
...
-                  "--dependencies to pull external JARs or use --local-jars to use",
-                  "JARs located on your file system."
-                ).mkString(" ")
+            Console.err.println("fork value: " + args.fork)
+            if (args.fork) {
+              val cp = resolveDependencies(
+                s"com.disneystreaming.smithy4s:smithy4s-codegen-cli_2.13:${BuildInfo.version}" ::
+                  args.dependencies,
+                args.localJars,
+                args.repositories
```
ghostbuster91, line 56: Very clever :+1: 

There is one downside though:
This is not ideal as coursier's resolution will just pick the latest versions preventing users from using anything older than what codegen was compiled against. 

But maybe this is good enough trade-off given its low complexity? :thinking:

kubukoz: imo this is actually good 😅 if you used older things you might be risking that the old version is not going to work with the codegen anyway, i.e. it might not be compliant with more recent requirements in the smithy model etc.

PR #1709, opened 2025-05-01
More predictable handling of trait/validator classes model loader

### LINKED ISSUE #336: Custom library-defined traits don't work with the codegen plugin
We might be doing something wrong, but it seems that custom library-defined traits are not properly supported by the codegen plugin.

**Our setup:**
- we have a library which defines some custom traits and registers the `TraitService` SPIs for them as well as some validators
- a service project is using the smithy4s codegen plugin & it depends on the above library as well as some JARs with smithy files

**The issue:**
When the model is loaded in the service project, it can't see the `TraitService` SPIs for our custom traits and it falls back to the dynamic model for those. Validators are picked-up fine, but they can't see the traits in the model since they rely on the runtime classes and won't work correctly with the dynamic model. Thus, we get validation errors everywhere. 

**The cause:**
The `ModelLoader` is only passing the classloader that can see external dependencies to the `discoverModels` call of the assembler and not to the factory method for the assembler itself:
```
Model
  .assembler() // this is using current classloader (which can only see the plugin's classpath)
  .discoverModels(upstreamClassLoader)
```
https://github.com/disneystreaming/smithy4s/blob/5ce1106e363bf1bec78e196791cbb26da6d803de/modules/codegen/src/smithy4s/codegen/ModelLoader.scala#L58
This way, it can't see any `TraitService` SPIs contributed by external dependencies, but it can still discover & load the smithy files contained in them.

**Our attempts to fix this:**
If we just pass the `upstreamClassloader` to the `Model.assembler` call, it loads the `TraitService` interface again from the `smithy-model` jar through this classloader and you end up in a situation where the SPI lookup fails with `"TraitService is not a TraitService"` type of error. We ran into similar kinds of issues when adding the current classloader as a parent of `upstreamClassLoader` and attempting to filter the external classpath.

It works if we abandon the staged model loading approach altogether and load everything from a unified classpath:
```
val validatorClassLoader = maybeDeps match {
  case Some(deps) => new URLClassLoader(deps, currentClassLoader)
  case None       => currentClassLoader
}

val modelAssembler =
  Model
    .assembler(validatorClassLoader)
    .discoverModels(validatorClassLoader) // discovering models from both internal & external classpath in a single pass
```
But that would surely break other things.

kubukoz:
Workaround / fix for https://github.com/smithy-lang/smithy/issues/2596. Even fixes #336.

tl;dr without passing the classloader to `assembler()`, we can run into strange cases where the validator gets loaded and the trait services don't - meaning the validators won't behave like they're supposed to.

- [ ] Added documentation
- [x] Updated changelog

### DIFF
--- a/modules/codegen-plugin/src/sbt-test/codegen-plugin/transformation-trait-classes/test
+++ b/modules/codegen-plugin/src/sbt-test/codegen-plugin/transformation-trait-classes/test
@@ -1,0 +1,4 @@
+# check if the app runs successfully.
+# if it doesn't compile, it could be because the transformation wasn't applied.
+# if the transformation throws, it's because of a bug like #336.
+> run
--- a/modules/codegen/src/smithy4s/codegen/internals/ModelLoader.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/ModelLoader.scala
@@ -76,7 +76,7 @@ private[codegen] object ModelLoader {
 
     // Loading the upstream model
     val upstreamModel = Model
-      .assembler()
+      .assembler(currentClassLoader)
       // disabling cache to support snapshot-driven experimentation
       .putProperty(ModelAssembler.DISABLE_JAR_CACHE, true)
       .addClasspathModels(currentClassLoader, discoverModels)
--- a/modules/codegen/src/smithy4s/codegen/internals/ModelLoader.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/ModelLoader.scala
@@ -74,9 +74,14 @@ private[codegen] object ModelLoader {
       }
     }
 
+    val validatorClassLoader = locally {
+      val jarUrls = deps.map(_.toURI().toURL()).toArray
+      new URLClassLoader(jarUrls, currentClassLoader)
+    }
+
     // Loading the upstream model
     val upstreamModel = Model
-      .assembler(currentClassLoader)
+      .assembler(validatorClassLoader)
       // disabling cache to support snapshot-driven experimentation
       .putProperty(ModelAssembler.DISABLE_JAR_CACHE, true)
       .addClasspathModels(currentClassLoader, discoverModels)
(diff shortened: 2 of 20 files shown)


modules/codegen/src/smithy4s/codegen/internals/ModelLoader.scala
```diff
@@ -74,9 +74,14 @@ private[codegen] object ModelLoader {
...
 
+    val validatorClassLoader = locally {
+      val jarUrls = deps.map(_.toURI().toURL()).toArray
+      new URLClassLoader(jarUrls, currentClassLoader)
+    }
+
     // Loading the upstream model
     val upstreamModel = Model
-      .assembler()
+      .assembler(validatorClassLoader)
```
kubukoz, line 84: This should make the classes (validators, trait services) visible to the assembler - but notably not making the models visible.

kubukoz:
ok I think the easiest way to test this will be to use a custom (non-alloy) trait class in a plugin/CLI test - as proven by https://github.com/polyvariant/smithy4s-bsp/pull/10 this finally allows us to use those properly.

the other thing would still be nice to test though, I'll need to think of a library that we can use as the example (or make a plugin test like that somehow)

modules/codegen-plugin/src/sbt-test/codegen-plugin/transformation-trait-classes/test
```diff
@@ -0,0 +1,4 @@
+# check if the app runs successfully.
+# if it doesn't compile, it could be because the transformation wasn't applied.
```
kubukoz, line 2: **thought:** it is WAY too easy to misconfigure a transformation, hence this extra care

kubukoz: BTW we had no other tests that involved user-provided transformations/validators, apparently. There's no `software.amazon.smithy.build.ProjectionTransformer` or `software.amazon.smithy.model.validation.Validator` in the test resources. I think these tests suffice, but maybe we should have a more minimal one for both too?

modules/codegen/src/smithy4s/codegen/internals/ModelLoader.scala
kubukoz: notably this doesn't solve the problems that #1705 and #1707 are attempting to solve, but it's a step in the right direction and doesn't clash with them

Baccata (approved):

modules/codegen-plugin/src/sbt-test/codegen-plugin/transformation-trait-classes/test
Baccata: Maybe indeed

PR #1717, opened 2025-05-18
cross-compile codegen module against scala3

ghostbuster91:

- [ ] Added documentation
- [ ] Updated changelog


modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -1644,60 +1644,76 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
+          })
+        }.asInstanceOf[T => Line]
+      case Primitive.Nothing => _ => sys.error("cannot happen") // this case can't happen
     }
 
   private def renderStringLiteral(raw: String): Line = {
-    import scala.reflect.runtime.universe._
-    val str = Literal(Constant(raw))
-      .toString()
+    // Escape the string as a Scala literal (surround with quotes, escape special characters)
```
kubukoz, line 1877: homestly I think we should just use scalameta for small things like these (rendering constants)

Baccata: fine, if you want to pull it I don't mind. But don't get your hopes up about rewriting the codegen in scalameta

kubukoz: yeah I'm not

ghostbuster91: I tried it in the past and while having the snippets to compile at the same time you compile your code is mind-blowing there was a lot of hassle sometimes to get them right.

Baccata: yeah, the ancestor to smithy4s was actually written with scalameta. Lots of friction.

kubukoz: hints:

- `libraryDependencies += ("org.scalameta" %% "scalameta" % "4.11.0").cross(CrossVersion.for3Use2_13)`
- `scala.meta.Lit.String(s).printSyntaxFor(scala.meta.dialects.Scala3)`

that's how I use it in https://github.com/kubukoz/smithy-playground/blob/c34be16f534795041295f4379a6ca8040297c4b1/modules/parser-gen/src/main/scala/playground/parsergen/ParserGen.scala#L355C25-L355C92 anyway

kubukoz:
we'll need this for #1790. @ghostbuster91 how do you feel about finishing this up?

ghostbuster91:
Sounds like a good case. I will try to do this this week.

kubukoz:
I believe this is done because of sbt2 / mill1

PR #1719, opened 2025-05-19
add mapping between HttpUri and java.net.URI

yisraelU:
Goal is to simply to have this logic in one place, given the many http libs on the JVM do have some interop or direct dependency on the URI data type. 

Binary compatibility breaks though due to adding the companion object for HttpUri which results in the companion object not extending AbstractFunctionN. To remedy this , I've added it in manually.

### DIFF
--- a/modules/core/src/smithy4s/http/HttpUri.scala
+++ b/modules/core/src/smithy4s/http/HttpUri.scala
@@ -69,7 +69,7 @@ object HttpUri
       path: IndexedSeq[String],
       queryParams: Map[String, Seq[String]],
       pathParams: Option[Map[String, String]]
-  ): HttpUri = HttpUri(scheme, host, port, path, queryParams, pathParams)
+  ): HttpUri = new HttpUri(scheme, host, port, path, queryParams, pathParams)
 
   def fromURI(uri: URI): HttpUri = {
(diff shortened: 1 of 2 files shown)


kubukoz (commented):
I'm not sure about whether this should be in core, or a separate module. Let's see what CI says about non-JVM platforms (I don't know how well URI is supported in JS/Native)

Baccata:
What's failing exactly now ?

modules/core/src/smithy4s/http/HttpUri.scala
```diff
@@ -30,4 +32,69 @@ final case class HttpUri(
...
+      HttpUri
+    ] {
+  def apply(
+      scheme: HttpUriScheme,
+      host: String,
+      port: Option[Int],
+      path: IndexedSeq[String],
+      queryParams: Map[String, Seq[String]],
+      pathParams: Option[Map[String, String]]
+  ): HttpUri = HttpUri(scheme, host, port, path, queryParams, pathParams)
```
kubukoz: You gotta use "new" here, otherwise it's an infinitely recursive call

Baccata (approved):

Baccata:
If you could also port this PR to target the `series/0.19` branch, that'd be great

PR #1721, opened 2025-05-21
(0.19) Add mapping between java.net.URI and HttpUri

yisraelU:
same  PR as #1719 targeting 0.19

### DIFF
--- a/modules/core/src/smithy4s/http/HttpUri.scala
+++ b/modules/core/src/smithy4s/http/HttpUri.scala
@@ -46,13 +46,10 @@ final case class HttpUri private (
 
   def userInfo: Option[String] = origin.flatMap(_.authority.userInfo)
 
-  /**
-   * Returns true if this is a relative URI (no authority)
-   * */
   def scheme: Option[HttpUriScheme] = origin.flatMap(_.scheme)
 
   /**
-   * Returns true if this is a relative URI (no scheme or authority)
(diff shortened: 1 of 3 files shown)


yisraelU:
Ok , I see that `host` parameter was changed to be `Option[String]` [here](https://github.com/disneystreaming/smithy4s/commit/92ab639b05c7f534ee094a609fd37dd3508ba0c9) hence the compilation errors.
Before I proceed , I'd like to clarify the model of HttpUri when Host is not provided
Is the intention to model a relative uri , or is a default meant to be assumed i.e. localhost
If its the former
This is a bit out of sync with the spec for a relative uri, as scheme is mandatory which does not exist in a relative uri. 
In addition one can add a port but leave out the host which is also invalid.

kubukoz:
It's supposed to be a relative uri... and I think you're right, it should be modeled in a way that doesn't let you pass a scheme/port without a host. Would you like to make that change?

yisraelU:
@kubukoz sure , I'll open a separate PR

yisraelU:
waiting on #1725

kubukoz:
Think this is good to update now :) 

pushed a commit with a changelog cleanup for 0.19. Once this is merged we can sync branches in #1731

yisraelU:
synced with 0.19

yisraelU:
@kubukoz ready for merge

kubukoz (approved):

PR #1722, opened 2025-05-22
Use TopDownIndex to gather service operations from resources

kubukoz:
Because why reinvent the wheel?

- [ ] Added documentation
- [ ] Updated changelog


kubukoz:
AFAIK the order of properties like service operations doesn't matter, so we don't need to be very concerned with this being possibly different from the Dynamic behavior.

Baccata (approved):

Issue #1584, opened 2024-09-12
Enable -Xsource:3 


ghostbuster91:
This should be enabled for both scala 2.12 and 2.13 builds. 
It ensures that the code behaves in the same way across all language versions. 

This fixes a bug in scala 2.12 and 2.13 where `copy` method is still accessible even though the primary constructor has been marked as `private`. 

Above is binary incompatible change so I guess this goes into 0.19 series.

ghostbuster91:
I will happily volunteer for this.

kubukoz:
I would suggest checking `-Xsource:3-cross`, apparently it's better suited for libs that are cross-built.

Baccata:
Go for it

ghostbuster91:
As usual, it was easier said than done. 

Turns out scala 2.12 does not support `Xsource:3-cross` atm. I asked if there are any plans to backport it [here](https://github.com/scala/scala/pull/10709#issuecomment-2396514517). 

Without this we cannot use these new features like the improved semantics for copy modifiers, meaning that if we want to be better at preserving binary compatibility we have to stop using case classes in public api. Enabling it only for 2.13 & 3 doesn't make sense imo.

We could enable `Xsource:3` that fails when a code with different semantics is detected but this isn't fully supported by scala 2.12 either as the following compiles without any errors/warnings:

```scala
//> using scala 2.12
//> using options -Xsource:3

case class Person private (age: Int, name: String)
```

Baccata:
>  we have to stop using case classes in public api 

I agree that ideally this should be a goal. We even had started on it, but lay-offs and reprioritisations forced us to park this effort : https://github.com/disneystreaming/smithy4s/pull/1391

lewisjkl:
Seeing as how `Xsource:3-cross` is not going to be supported in Scala 2.12 (per the issue linked above), I'm wondering if we should remove this issue from the 0.19 milestone. Any objections?

kubukoz:
Can we add it in 2.13 at least?

lewisjkl:
> Can we add it in 2.13 at least?

To be honest, I'm not really sure what the implications of that would be. I was just going off of the prior comment in this thread "Enabling it only for 2.13 & 3 doesn't make sense imo."

kubukoz:
I think the main benefit is that we can offer better binary compatibility guarantees on case classes, without a gigantic effort of rewriting those. `case class ... private(...)` is already a significant gain in those terms (both 2.13 and 3).

[...]

PR #1725, opened 2025-05-22
refactor httpuri to be rfc compliant

yisraelU:
This PR changes the model of HttpUri in a more strict fashion based off RFC 3986. In particular it does not allow construction of
   - a Uri with a scheme but no host
   - a Uri with a port without a host

While HttpUri still models both relative and absolute Uri's , any uri with either scheme or port must contain host.

For lack of better term I called the optional data type a HttpUriOrigin after [RFC 6454](https://www.rfc-editor.org/rfc/rfc6454.html) which seemed to contain the the most accurate description of `Scheme + Authority` 
Happy to change the naming to anything else

### DIFF
--- a/modules/core/src/smithy4s/http/HttpUri.scala
+++ b/modules/core/src/smithy4s/http/HttpUri.scala
@@ -16,10 +16,17 @@
 
 package smithy4s.http
 
+import scala.runtime.AbstractFunction4
+
+/**
+ * RFC 3986 compliant URI implementation.
+ * @param origin The origin component of the URI.
+ * @param path A sequence of URL-decoded URI path segments
+ * @param queryParams A map of query parameters where keys and values are URL-decoded
+ * @param pathParams Optional map of path parameters extracted during routing
+ */
 final case class HttpUri(
-    scheme: HttpUriScheme,
-    host: Option[String],
-    port: Option[Int],
+    origin: Option[HttpUriOrigin],
     /**
       * A sequence of URL-decoded URI segment.
       */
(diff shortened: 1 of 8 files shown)


yisraelU:
Hi @kubukoz , if you have some time for some advice here.

I have refactored the model a bit, and now the roundtrip to http4s uri tests fail .

The issue appears to me that the http4s uri model is actually incorrect from an rfc perspective (I did see quite a few issues raised around it being rfc compliant, and discussion of making a new uri model)

There are distinctions between a `URI` and an `HTTP-URI` , a `URI` for example has no requirement for an `Authority` i.e. `file:/...` however an `HTTP-URI` does require an `Authority` .
Essentially as the [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.1) puts it , 
> As such, the URI syntax is
   a federated and extensible naming system wherein each scheme's
   specification may further restrict the syntax and semantics of
   identifiers using that scheme.
As such the tests that assert `/` == `http:/` and such should be removed . 

I'd like to confirm this before I do so..

kubukoz:
Maybe instead of removing, I'd replace them with more direct tests that showcase what the encoded and roundtripped values are. If we're changing the behavior, let's at least keep it documented :)

modules/core/src/smithy4s/http/HttpUriAuthority.scala
```diff
@@ -0,0 +1,40 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.http
+
+final case class HttpUriAuthority(
```
kubukoz: **issue:** the constructor here should be private, and we should make sure the `unapply` / `apply` are private as well. Have a look at https://github.com/disneystreaming/smithy4s/issues/1584 too please

yisraelU: @kubukoz done.
On a side note do we want to continue with https://github.com/disneystreaming/smithy4s/pull/1391

kubukoz: I think so, but I don't know how easily it can be brought up-to-date

yisraelU:
@kubukoz ready for another review when you have time

modules/core/src/smithy4s/http/HttpUri.scala
```diff
@@ -30,4 +37,194 @@ final case class HttpUri(
...
+}
+
+object HttpUri
+    extends AbstractFunction4[
+      Option[HttpUriOrigin],
+      IndexedSeq[String],
+      Map[String, Seq[String]],
+      Option[Map[String, String]],
+      HttpUri
+    ] {
```
kubukoz: **issue:** I don't think this is necessary / useful, people should just us `apply` if they want to treat it as a function

kubukoz (approved):

kubukoz:
just one thing I'd change

modules/core/src/smithy4s/http/HttpUri.scala
Baccata: agreed

yisraelU: I added this because of [Mima](https://github.com/disneystreaming/smithy4s/actions/runs/15117877327/job/42493034033#step:9:210).
It seems when you define a companion object in 2.13 , Scala does not  generate the tupled and curried methods. 
see the [PR](https://github.com/disneystreaming/smithy4s/pull/1719) for 0.18

kubukoz: We're breaking bincompat in 0.19 already so it's fine to break mima on this branch

yisraelU: ok , ill remove then

yisraelU: done

yisraelU:
@kubukoz does CI need to be rerun, it has a status of failure right now, because of cancelled jobs

kubukoz:
image

why now... I understand the issue but I don't get why we're seeing it now and not on a previous PR to 0.19.

kubukoz:
ok, now I'm actually double confused because [this PR](https://github.com/disneystreaming/smithy4s/pull/1675) is already included in 0.19.

kubukoz:
ah ok, so the 2.13.10 version only shows up in one test, and I already fixed it on 0.18: https://github.com/disneystreaming/smithy4s/issues/1715

so I'd assume that has nothing to do with the cancellation. Let's retry.

and we're going to merge 0.18 into 0.19 soon after your PRs, as I'm waiting for them in  #1731.

PR #1733, opened 2025-05-27
fix uri encoding issue for conversion to java.net.URI , plus add tests

yisraelU:
the wrong constructor for java.net.URI was used [here](https://github.com/disneystreaming/smithy4s/blob/3ebf4b919963607c59ab6cd35208a057f4b7220f/modules/core/src/smithy4s/http/HttpUri.scala#L50) which takes in a raw string and does not apply encoding.
This PR fixes that and introduces some tests for it
Add filter for Native platform for url encoding roundtrip tests , as there is a bug in scala-native and was only patched in later versions

### DIFF
--- a/modules/bootstrapped/test/src/smithy4s/http/HttpUriSpec.scala
+++ b/modules/bootstrapped/test/src/smithy4s/http/HttpUriSpec.scala
@@ -39,4 +39,30 @@ final class HttpUriSpec extends FunSuite {
     val uri = httpUri.toURI
     assertEquals(httpUri, HttpUri.fromURI(uri))
   }
+
+  test("Roundtrip preserves uri encoding from java.net.URI") {
+    // This URI contains spaces, which should be encoded as %20
+    val uri = URI.create("http://example.com/foo%20bar?baz=qux%20quux")
+    val httpUri = HttpUri.fromURI(uri)
+    assertEquals(uri, httpUri.toURI)
+    assert(httpUri.path == IndexedSeq("foo bar"))
+    assert(httpUri.queryParams == Map("baz" -> List("qux quux")))
+  }
+  test("Roundtrip preserves uri encoding from HttpUri") {
+    // This HttpUri contains spaces, which should be encoded as %20, however HttpUri stores data in its decoded form
+    // so we expect the spaces to be present in the path and query parameters.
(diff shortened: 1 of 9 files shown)


modules/core/src/smithy4s/http/HttpUri.scala
```diff
@@ -96,4 +93,13 @@ object HttpUri
       .getOrElse(Map.empty)
     HttpUri(scheme, host, port, path, queryParams, None)
   }
+
+  def safeToInt(value: String): Option[Int] = {
```
lewisjkl: Is this ever used?

yisraelU:
Scala3Native test failing , looking into https://github.com/scala-native/scala-native/pull/4109

modules/core/src/smithy4s/http/HttpUri.scala
yisraelU: it was , i changed up the code , will remove

kubukoz:
Is that even there in 0.4?

yisraelU:
@kubukoz no , it's 5.x and on. I filtered out the tests

kubukoz:
ok, I see now. Please mention SN0.4/0.5 specifically in the string :)

Baccata (approved):

modules/bootstrapped/test/src/smithy4s/http/HttpUriSpec.scala
```diff
@@ -39,4 +40,45 @@ final class HttpUriSpec extends FunSuite {
...
+      assert(httpUri.path == IndexedSeq("foo bar"))
+      assert(httpUri.queryParams == Map("baz" -> List("qux quux")))
+    }
+  }
+
+  test("Roundtrip preserves uri encoding from HttpUri") {
+    if (Platform.isNative)
+      assume(
+        false,
+        "This test is not applicable for Scala Native, as we are on 4.x which has bugs in URI parsing and encoding"
```
kubukoz: **praise:** 
```suggestion
        "This test is not applicable for Scala Native, as we are on 0.4.x which has bugs in URI parsing and encoding"
```

kubukoz (approved):

lewisjkl (approved):

Issue #1567, opened 2024-07-15
Bad ambiguous URI resolution


Isammoc:
From [smithy 2.0 documentation](https://smithy.io/2.0/spec/http-bindings.html#specificity-routing), the server should route ambiguous URI in a specific order.

My current extract:

> a path with a non-label segment is considered more specific than one with a label segment in the same position

Reproduction:

1. Create a new project
    ```console
    $ sbt new disneystreaming/smithy4s.g8
    # default for all
    ```
2. Edit smithy definition 
    ```diff
    -  operations: [Hello]
    +  operations: [Hello, SpecialHello]
    ```
    and add new operation
    ```
    @readonly
    @http(method: "GET", uri: "/hello/me", code: 200)
    operation SpecialHello {
      output: Greeting
    }
    ```
3. Implement service (in `Main.scala`)
    ```scala
      override def specialHello(): IO[Greeting] = IO.pure(Greeting("Hello special me!"))
    ```
4. Start `sbt run`

**Current behavior**

The method for broader endpoint is run:
```console
$ curl http://localhost:9000/hello/me
{"message":"Hello me!"}
```

**Expected behavior**

More specific method is preferred:
```console
$ curl http://localhost:9000/hello/me
{"message":"Hello special me!"}
```

**Notes**

If this is from the same service, the order of operations seems to affect the affect the order of matching.

For different services, the order will be directly from the scala code that calls.

How can I help?

kubukoz:
We must have missed the fact that https://github.com/smithy-lang/smithy/issues/1029 has been resolved. We'll have to tackle this for sure...

kubukoz:
Had a quick read of the rules, at first sight it looks like we'll just have to sort the routes before combining them.

This code in [`HttpUnaryServerRouter.scala`](https://github.com/disneystreaming/smithy4s/blob/8d943ab5db5f0d72656d85928202955bc4e8d056/modules/core/src/smithy4s/http/HttpUnaryServerRouter.scala):

```scala
    private val httpEndpointHandlers: List[HttpEndpointHandler] =
      service.endpoints.toList
        .map { makeHttpEndpointHandler(_) }
        .collect { case Right(endpointWrapper) => endpointWrapper }

    private val perMethodEndpoint: Map[HttpMethod, List[HttpEndpointHandler]] =
      httpEndpointHandlers.groupBy(_.httpEndpoint.method)
```

(there are two occurrences of it)

[...]

PR #1743, opened 2025-06-04
Http4s check static query

Odomontois:

- [ ] Added documentation
- [x] Updated changelog


modules/bootstrapped/src/generated/smithy4s/example/PizzaAdminService.scala
```diff
@@ -118,6 +124,7 @@ object PizzaAdminServiceOperation {
     def getMenu(restaurant: String): GetMenu = GetMenu(GetMenuRequest(restaurant))
   }
   class Transformed[P[_, _, _, _, _], P1[_ ,_ ,_ ,_ ,_]](alg: PizzaAdminServiceGen[P], f: PolyFunction5[P, P1]) extends PizzaAdminServiceGen[P1] {
+    def checkQueryA(): P1[Unit, Nothing, CheckQueryOutput, Nothing, Nothing] = f[Unit, Nothing, CheckQueryOutput, Nothing, Nothing](alg.checkQueryA())
```
kubukoz: **thought:** it bothers me a bit that the ordering is... well, like this, but it seems to be a consequence of #1722. At some point maybe we'll sort these by the ordinal, it'd make more sense

sampleSpecs/pizza.smithy
```diff
@@ -401,3 +403,22 @@ structure HeadRequestOutput {
...
+    variant: String
+}
+
+@http(method: "GET", uri: "/query-check?variant=a", code: 200)
+@readonly
+operation CheckQueryA {
+    output : CheckQueryOutput
+}
+
+@http(method: "GET", uri: "/query-check?variant=b", code: 200)
```
kubukoz: **suggestion:** mind adding some multimap cases, such as multiple occurrences of `variant` and with multiple constant parameters?

modules/bootstrapped/src/generated/smithy4s/example/PizzaAdminService.scala
Odomontois: Yes, I would like to work on some form of topological sorting according to path/query generality.
But that would be much more work, so I've decided to start with a small PR and see how it'll go

kubukoz: I wasn't even suggesting to take http things into account here :) the code generator doesn't know about http and it should stay that way

modules/core/src/smithy4s/http/HttpUnaryServerRouter.scala
```diff
@@ -249,4 +244,10 @@ object HttpUnaryServerRouter {
 
   }
 
+  /** Checks if `larger` multimap is a subset of the `smaller` */
+  private def isSubset[K, V](larger: Map[K, Seq[V]], smaller: Map[K, Seq[V]]): Boolean =
```
yisraelU, line 304: Since this would be on the hot path, I think we should try to optimize this method as much as possible. 
right now per key its (m*n) , perhaps we should consider conversion to Set i.e.
``` 
val vSet = vs.toSet
vRequired.forall(vSet.contains)
```

kubukoz: I thought about it, but didn't think we should worry much about these maps being large. Regardless, if the implementation is simple enough we should do it.

Odomontois: Well, if the smaller multimap has `|Ks|` keys and each key has `|Vs|` values, and at the same time the larger multimap has `|Kl|` and each key has `|Vl|` values, the upper bound is something like
`C * |Ks| * |Vs| * |Vl|`
So it's not M* N, where M and N are the sizes of both maps, but it can grow if there are a lot of repeated query params, which could be fixed with a bit more sophisticated multimap datastructure instead of `Map[K, List[V]]`
But if we consider queries not having repeated params, it's more like O(N), where N is the size of query map in the `http.uri` field.

So, if we want to optimize it, we can try to (lazily) convert `Map[ParamName, List [ParamValue]]` from the request to `Map[ParamName, Set[ParamValue]]` we gain upper bound of `C * |Ks| * (|Vs| + |Vl|)`, which could be a performance win only when `|Vs| > 1`, i.e. there is a repeated query param in the `.smithy` file. 

In the typical case, when you don't have any required query parameters, ' .forall` will instantly return true, and the real cost is only the lambda allocation.

kubukoz: I'm not too concerned with this, like you said typical cases won't have that many repeated query params. Let's roll with the current implementation and not microoptimize ahead of time :)

modules/tests/src/smithy4s/tests/PizzaSpec.scala
```diff
@@ -463,6 +463,106 @@ abstract class PizzaSpec
       }
   }
 
+  routerTest("Respects Static Query Parameters") { (client, uri, log) =>
```
Odomontois, line 466: I've played with different variations of static query params, but it seems operations are shuffled during code generation, so it's impossible to guarantee any order., i.e., overlapping routes can't have a predictable outcome.

kubukoz: Can you add them nevertheless, and mark them as flaky with a mention of https://github.com/disneystreaming/smithy4s/issues/1567 ? That way, we'll know to fix them once the other thing is resolved

modules/core/src/smithy4s/http/HttpUnaryServerRouter.scala
Odomontois: Interestingly enough, smithy's own validation currently prohibits repeated params in query literals.

modules/tests/src/smithy4s/tests/PizzaSpec.scala
```diff
@@ -463,6 +463,142 @@ abstract class PizzaSpec
...
+      expect.same(codeXC, 200) &&
+      expect.same(
+        bodyXC,
+        response(
+          "variants" -> "c",
+          "staticVariants" -> "c",
+          "kinds" -> "x",
+          "staticKinds" -> "x"
+        )
+      )
```
kubukoz, line 520: **issue:** discarded expression. I'm not sure why we don't have a warning for this.

These assertions aren't actually being used

Baccata (approved):
@kubukoz do you want to give this another look ?

kubukoz (approved):

kubukoz:
just gonna update the changelog as we released .38 already

PR #1744, opened 2025-06-04
Add test case / quick fix for #1581

### LINKED ISSUE #1581: Discrepancy of `null` handling when a default is present or not

kubukoz:
Closes #1581

- [ ] Added documentation
- [ ] Updated changelog


kubukoz:
This is failing because nulls are only allowed in OptionSchema, whereas optional fields with defaults get a `Schema.field`, which is NOT an OptionSchema. Just a string schema + Default hint.

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -859,7 +859,7 @@ private[codegen] class SmithyToIR(
         case ShapeType.SHORT       => Node.from(0: Short)
         case ShapeType.FLOAT       => Node.from(0.0f)
         case ShapeType.BOOLEAN     => Node.from(false)
-        case ShapeType.BLOB        => Node.arrayNode()
+        case ShapeType.BLOB        => Node.from("")
```
kubukoz, line 864: **issue:** this appeared to be wrong from day 1, defaults of Blobs are written as strings, as validated by smithy itself. It was causing failures in the decoding of `seventeen`'s default.

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -1402,15 +1403,30 @@ private[smithy4s] class SchemaVisitorJCodec(
...
-      field: Field[Z, A]
+      field: Field[Z, A],
+      // nullable A
+      default: Any
   ): Handler = {
     val codec = apply(field.schema)
     val label = field.label
+
+    val decodeFn: (Cursor, JCodec[A], JsonReader) => A =
+      if (field.hints.has(Required) || default == null || field.hints.has(alloy.Nullable))
```
kubukoz: **issue:** I'm not sure I like this. It works, but I'm not convinced it's the right way due to my low familiarity with the JSON decoding impl. Thoughts welcome!

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
kubukoz: note that 0.19 completely changes this behavior and won't generate a default here at all, but we don't know when it'll drop so it's worth fixing now. #1667, #1696

lewisjkl:
I can see this going either way, but is it more expected as a caller of an API that an explicit `null` is assumed to be the default value when deserialized? I think I'd expect if I sent that, that it would actually be set to `null`, which in this case is not in fact possible

kubukoz:
> which in this case is not in fact possible

yeah, because fields with defaults don't have the Option in them. I think the behavior with using the default value is the only thing we can do, really.

In Dynamic, explicit nulls decode as the equivalent of None:

image

lewisjkl:
> > which in this case is not in fact possible
> 
> yeah, because fields with defaults don't have the Option in them. I think the behavior with using the default value is the only thing we can do, really.
> 
> In Dynamic, explicit nulls decode as the equivalent of None:
> 
> image

Yeah I mean that's fair enough. I guess ultimately it makes the most sense to have absence of value be treated exactly the same as an explicit null in the absence of the `nullable` trait. It just may be surprising to some users in this case when they PUT `{ "foo": null }` and then GET and receive `{ "foo": "test" }` (assuming "test" is the default there) 🤷‍♂️

kubukoz:
well, I know of users who were surprised that adding a default value made their previously-valid nulls invalid 😅 

I think it's the lesser evil to give them the default. Being less restrictive about what we accept, if you will...

Baccata (approved):

PR #1747, opened 2025-06-06
Expose multiple headers with the same name in responses

nviliunov-evolution-throwaway:

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/modules/http4s-kernel/src/smithy4s/http4s/kernel/package.scala
+++ b/modules/http4s-kernel/src/smithy4s/http4s/kernel/package.scala
@@ -91,8 +91,9 @@ package object kernel {
   def toSmithy4sHttpResponse[F[_]: Concurrent](res: Response[F]): F[Smithy4sHttpResponse[Blob]] =
     collectBytes(res.body).map { blob =>
       val headers = res.headers.headers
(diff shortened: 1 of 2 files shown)


modules/http4s-kernel/src/smithy4s/http4s/kernel/package.scala
```diff
@@ -91,8 +91,9 @@ package object kernel {
   def toSmithy4sHttpResponse[F[_]: Concurrent](res: Response[F]): F[Smithy4sHttpResponse[Blob]] =
     collectBytes(res.body).map { blob =>
       val headers = res.headers.headers
-        .map(h => CaseInsensitive(h.name.toString) -> Seq(h.value))
-        .toMap
+        .groupBy(h => CaseInsensitive(h.name.toString))
```
kubukoz, line 94: **question:** wait, we've been silently dropping repeated headers in responses **all this time**? wow.

kubukoz: at least this would only happen on clients...

however, `res.headers` should already have a MultiMap-like representation. Let's use that instead of our own grouping, if possisble.

PR #1756, opened 2025-06-19
Fix Hints methods to distinguish member and target hints

### LINKED ISSUE #1658: Strange things happening in `Hints`
We've found some suspicious activity in `Hints`:

- `ShapeTag#eq` (referential equality) usage in `get`: https://github.com/disneystreaming/smithy4s/blob/7cd62db35ac5911fbb455945154c9a178b6fd76a/modules/core/src/smithy4s/Hints.scala#L157 - there's no guarantee that two ShapeTags are the same instance, so we should probably stop doing that
- `filter` / `filterNot` lose information about whether the things were on the member or the target (`Hints.fromSeq` is used inside)
- `Hints#equals` is implemented by `toMap` equality, which also conflates member and target hints. Combined with the above it's a recipe for headache and loss of hair (ask @ghostbuster91)
- `Hints#toString` also follows `toMap` so it doesn't show you the difference if a hint jumps between groups.

I think we should fix all of these. I don't really know why we check for `eq` here:

https://github.com/disneystreaming/smithy4s/blob/7cd62db35ac5911fbb455945154c9a178b6fd76a/modules/core/src/smithy4s/Hints.scala#L157

Also, there might be other areas where we didn't really account for the member/target hint difference. `Hints` needs some careful review for that.

Jay-Lokhande:
This PR fixes #1658 by modifying that methods on `Hints` (such as `filter`, `equals`, and `toString`) preserve the distinction between `memberHints` and `targetHints`.

- [ ] Added documentation
- [x] Updated changelog

Closes Strange things happening in Hints #1658

### DIFF
--- a/modules/bootstrapped/test/src/smithy4s/HintsSpec.scala
+++ b/modules/bootstrapped/test/src/smithy4s/HintsSpec.scala
@@ -173,37 +173,60 @@ class HintsSpec() extends FunSuite {
     "Hints#filter and toString handle member and target bindings correctly"
   ) {
     import Document.syntax._
-    val staticMemberHint = HttpHeader("X-Member")
-    val staticTargetHint = HttpLabel()
-    val dynamicMemberHint = Hints.dynamic("smithy.api#jsonName" -> "foo")
-    val dynamicTargetHint = Hints.dynamic("smithy.api#documentation" -> "doc")
     val hints = Hints
-      .member(staticMemberHint)
-      .addMemberHints(dynamicMemberHint)
-      .addTargetHints(staticTargetHint)
-      .addTargetHints(dynamicTargetHint)
+      .member(HttpHeader("X-Member"))
+      .addMemberHints(Hints.dynamic("smithy.api#jsonName" -> "foo"))
+      .addTargetHints(HttpLabel())
+      .addTargetHints(Hints.dynamic("smithy.api#documentation" -> "doc"))
 
-    val expectedMemberStr =
-      s"${HttpHeader.id} -> $staticMemberHint, ${JsonName.id} -> ${Document
-        .obj(JsonName.id.show -> Document.DString("foo"))}"
-    val expectedTargetStr =
-      s"${HttpLabel.id} -> $staticTargetHint, ${Documentation.id} -> ${Document
-        .obj(Documentation.id.show -> Document.DString("doc"))}"
+    val memberHintsString =
+      """smithy.api#httpHeader -> X-Member, smithy.api#jsonName -> {smithy.api#jsonName="foo"}"""
+    val targetHintsString =
+      """smithy.api#httpLabel -> HttpLabel(), smithy.api#documentation -> {smithy.api#documentation="doc"}"""
     val expectedToString =
-      s"Hints(member=[$expectedMemberStr], target=[$expectedTargetStr])"
+      s"""Hints(member=[$memberHintsString], target=[$targetHintsString])"""
+
     expect.same(hints.toString, expectedToString)
+  }
+
+  test("Hints#filter preserves member and target hints for no-op filter") {
+    import Document.syntax._
+    val hints = Hints.empty
+      .addMemberHints(HttpHeader("X-Member"))
+      .addMemberHints(Hints.dynamic("smithy.api#jsonName" -> "foo"))
+      .addTargetHints(HttpLabel())
+      .addTargetHints(Hints.dynamic("smithy.api#documentation" -> "doc"))
 
     val filtered = hints.filter(_ => true)
+    expect.same(filtered.memberHints, hints.memberHints)
+    expect.same(filtered.targetHints, hints.targetHints)
     expect.same(filtered, hints)
+  }
 
-    expect.same(
-      filtered.memberHints,
-      Hints.member(staticMemberHint).addMemberHints(dynamicMemberHint)
-    )
-    expect.same(
-      filtered.targetHints,
-      Hints(staticTargetHint).addTargetHints(dynamicTargetHint)
-    )
+  test("Hints#filter returns empty hints when predicate is false") {
+    import Document.syntax._
+    val hints = Hints.empty
+      .addMemberHints(HttpHeader("X-Member"))
+      .addMemberHints(Hints.dynamic("smithy.api#jsonName" -> "foo"))
+      .addTargetHints(HttpLabel())
+      .addTargetHints(Hints.dynamic("smithy.api#documentation" -> "doc"))
+
+    val filtered = hints.filter(_ => false)
+    expect.same(filtered, Hints.empty)
+  }
+
+  test("Hints#filter selects specific hints by keyId") {
+    val hints = Hints.empty
+      .addMemberHints(HttpHeader("X-Member"))
+      .addTargetHints(HttpLabel())
+
+    val filtered = hints.filter(_.keyId == HttpLabel.id)
+    expect.same(filtered, Hints(HttpLabel()))
+  }
+
+  test("Hints#filter on empty hints returns empty hints") {
+    val filtered = Hints.empty.filter(_ => true)
+    expect.same(filtered, Hints.empty)
   }
 
   private def makeLazyHints(hints: => Hints): (Hints, () => Boolean) = {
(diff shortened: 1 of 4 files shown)


modules/core/src/smithy4s/Hints.scala
```diff
@@ -174,12 +184,19 @@ object Hints {
...
 
-    override def toString(): String =
-      s"Hints(${all.mkString(", ")})"
+    override def toString(): String = {
+      val memberStr =
+        memberHintsMap.map { case (k, v) => s"member:$k=$v" }.mkString(", ")
+      val targetStr =
+        targetHintsMap.map { case (k, v) => s"target:$k=$v" }.mkString(", ")
+      s"Hints(member=[$memberStr], target=[$targetStr])"
+    }
```
kubukoz: **suggestion:** 

```suggestion
    override def toString(): String = {
      val memberStr =
        memberHintsMap.map { case (k, v) => s"$k -> $v" }.mkString(", ")
      val targetStr =
        targetHintsMap.map { case (k, v) => s"$k -> $v" }.mkString(", ")
      s"Hints(member=[$memberStr], target=[$targetStr])"
    }
```

that way, we save some duplication by only saying `member` and `target` once, before each set of hints.

kubukoz:
general feedback:

- If you say in the description of the PR "Closes #1658", github will automatically recognize that, and when the PR is merged it will close the corresponding issue
- Follow the checklist in the PR description, e.g. here we should have a CHANGELOG entry and maybe a couple unit tests. The changelog entry should use the `core:` prefix as that's where the changes are.
- `HintsSpec` is the place to look at
- PRs and commits should generally be named using imperative verb form: https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form - e.g. "Make a distinction between member and target hints in methods on Hints"

last but not least, once the PR is ready to look at, you can undraft it :)

modules/core/src/smithy4s/Hints.scala
```diff
@@ -37,6 +37,9 @@ trait Hints {
   def memberHintsMap: Map[ShapeId, Hints.Binding]
   def targetHintsMap: Map[ShapeId, Hints.Binding]
 
+  def toMemberMap: Map[ShapeId, Hints.Binding] = memberHintsMap
+  def toTargetMap: Map[ShapeId, Hints.Binding] = targetHintsMap
```
kubukoz: **question:** what do we need these for?

Jay-Lokhande: sorry i forgot to remove this

Jay-Lokhande:
> general feedback:
> 
> * If you say in the description of the PR "Closes [Strange things happening in `Hints` #1658](https://github.com/disneystreaming/smithy4s/issues/1658)", github will automatically recognize that, and when the PR is merged it will close the corresponding issue
> * Follow the checklist in the PR description, e.g. here we should have a CHANGELOG entry and maybe a couple unit tests. The changelog entry should use the `core:` prefix as that's where the changes are.
> * `HintsSpec` is the place to look at
> * PRs and commits should generally be named using imperative verb form: https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form - e.g. "Make a distinction between member and target hints in methods on Hints"
> 
> last but not least, once the PR is ready to look at, you can undraft it :)

thank you for the suggestions

modules/bootstrapped/test/src/smithy4s/HintsSpec.scala
```diff
@@ -169,6 +169,43 @@ class HintsSpec() extends FunSuite {
     assertEquals(hints.get(Tags), Some(Tags(List("one", "two", "three"))))
   }
 
+  test(
+    "Hints#filter and toString handle static and dynamic bindings correctly"
```
kubukoz: **issue:** 
```suggestion
    "Hints#filter and toString handle member and target bindings correctly"
```

we check static vs dynamic too, but the root of the issue is member vs target

modules/bootstrapped/test/src/smithy4s/HintsSpec.scala
```diff
@@ -169,6 +169,43 @@ class HintsSpec() extends FunSuite {
...
+  ) {
+    import Document.syntax._
+    val staticMemberHint = HttpHeader("X-Member")
+    val staticTargetHint = HttpLabel()
+    val dynamicMemberHint = Hints.dynamic("smithy.api#jsonName" -> "foo")
+    val dynamicTargetHint = Hints.dynamic("smithy.api#documentation" -> "doc")
```
kubukoz: **suggestion:** I'd inline all these - it  helps see what's going on. Some duplication in tests can be useful sometimes.

For example:

```scala
  test(
    "Hints#filter and toString handle static and dynamic bindings correctly"
  ) {
    import Document.syntax._
    val hints = Hints.empty
      .addMemberHints(HttpHeader("X-Member"))
      .addMemberHints(Hints.dynamic("smithy.api#jsonName" -> "foo"))
      .addTargetHints(HttpLabel())
      .addTargetHints(Hints.dynamic("smithy.api#documentation" -> "doc"))

    val memberHintsString="""smithy.api#httpHeader -> HttpHeader("X-Member"), smithy.api#jsonName -> {smithy.api#jsonName -> "foo"}"""
    val targetHintsString ="""smithy.api#httpLabel -> HttpLabel(), smithy.api#documentation -> {smithy.api#documentation -> "doc"}"""

    val expectedToString =
      s"""Hints(member=[$memberHintsString], target=[$targetHintsString])"""

    expect.same(hints.toString, expectedToString)
  }
```

modules/bootstrapped/test/src/smithy4s/HintsSpec.scala
```diff
@@ -169,6 +169,43 @@ class HintsSpec() extends FunSuite {
...
+      s"${HttpHeader.id} -> $staticMemberHint, ${JsonName.id} -> ${Document
+        .obj(JsonName.id.show -> Document.DString("foo"))}"
+    val expectedTargetStr =
+      s"${HttpLabel.id} -> $staticTargetHint, ${Documentation.id} -> ${Document
+        .obj(Documentation.id.show -> Document.DString("doc"))}"
+    val expectedToString =
+      s"Hints(member=[$expectedMemberStr], target=[$expectedTargetStr])"
+    expect.same(hints.toString, expectedToString)
+
+    val filtered = hints.filter(_ => true)
```
kubukoz, line 219: **issue:** Filtering should have its own bunch of tests - let's keep this one focused on `toString`, and add some more (copy-pasting the `val hints` is fine)

modules/bootstrapped/test/src/smithy4s/HintsSpec.scala
```diff
@@ -169,6 +169,43 @@ class HintsSpec() extends FunSuite {
...
+    expect.same(filtered, hints)
+
+    expect.same(
```
kubukoz: **suggestion:** I would put a comment above this one, saying that we check the exact values of `memberHints` and `targetHints` so that we aren't fooled by the broken `equals` (in the past, `expect.same` would probably have passed even though `filtered` was collapsed to a single map, because `equals` also collapsed to one map before the comparson)

modules/bootstrapped/test/src/smithy4s/HintsSpec.scala
kubukoz: For `filter`, the first test can be with just `filter(_ => true)`, here are some more ideas:

- `filter(_ => false)` == `Hints.empty`
- `Hints(..., HttpLabel()).filter(_.keyId == HttpLabel().id) == Hints(HttpLabel())`
- `Hints.empty.filter(_ => true) == Hints.empty`

bonus: maybe something like this (pseudocode):

```scala
// it would be nice to generate these two with Scalacheck, but that'd be a tad more complex than what we're doing
val a = Hints.member(HttpLabel(), HttpHeader("c"))
val b = Hints.target(HttpQuery("a"), JsonName("b"))

expect.same((a ++ b).filter(a.contains), a)
expect.same((a ++ b).filter(b.contains), b)
expect.same((a ++ b).filter(v => a.contains(v) || b.contains(v)), a ++ b)
```

kubukoz (changes requested):
I have some proposals for improving the tests, but the implementation LGTM

Jay-Lokhande:
@kubukoz thank you for all the suggestions

kubukoz (approved):
Nice work, thank you! Added one more test for `filter` specifically and a comment on `ShapeTag` following the discussion from the ticket.

Will merge once this turns green.

PR #1759, opened 2025-06-20
Add fromStringOrUnknown and fromIntOrUnknown to Open Enums

### LINKED ISSUE #1626: Open enums: add `fromStringOrUnknown` convenience method
Currently, we only have `def fromString` and a `def $unknown`. I think it'd be useful to have a total, Option-less function like `def fromStringOrUnknown(s: String): TheEnum` generated in all open enums' companion objects.

Jay-Lokhande:
Add fromStringOrUnknown and fromIntOrUnknown to Open Enums
## PR Checklist 
- [x] Updated changelog

Closes #1626

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
@@ -1313,7 +1313,10 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
           lines(
             line"""final case class $$Unknown($paramName: $paramType) extends $name($stringValue, "$$Unknown", $intValue, Hints.empty)""",
             newline,
-            line"val $$unknown: $paramType => $name = $$Unknown(_)"
+            line"val $$unknown: $paramType => $name = $$Unknown(_)",
+            newline,
+            line"def fromStringOrUnknown(s: String): ${name.name} = fromString(s).getOrElse($$unknown(${if (isIntEnum) "s.toInt"
+            else "s"}))"
           )
         } else Lines.empty,
         newline,
--- a/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
@@ -1315,8 +1315,10 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
             newline,
             line"val $$unknown: $paramType => $name = $$Unknown(_)",
             newline,
-            line"def fromStringOrUnknown(s: String): ${name.name} = fromString(s).getOrElse($$unknown(${if (isIntEnum) "s.toInt"
-            else "s"}))"
+            if (isIntEnum)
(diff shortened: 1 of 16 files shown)


modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -1313,7 +1313,10 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
           lines(
             line"""final case class $$Unknown($paramName: $paramType) extends $name($stringValue, "$$Unknown", $intValue, Hints.empty)""",
             newline,
-            line"val $$unknown: $paramType => $name = $$Unknown(_)"
+            line"val $$unknown: $paramType => $name = $$Unknown(_)",
+            newline,
+            line"def fromStringOrUnknown(s: String): ${name.name} = fromString(s).getOrElse($$unknown(${if (isIntEnum) "s.toInt"
```
kubukoz: **issue:** oversight on my side - it should only be `fromStringOrUnknown` for string enums. Int enums should have `fromIntOrUnknown`.

Also, we don't have a `fromInt`, only a `fromOrdinal` - I think that's another bit of oversight, but we can fix that later - just use `fromOrdinal` for now.

so sth like

```scala
            line"def fromIntOrUnknown(i: Int): ${name.name} = fromOrdinal(i).getOrElse($$unknown(i))
```

in the case of int enums

Jay-Lokhande: i am thinking to use 
```scala
if (isIntEnum) 
....  fromOrdinal ....
else .... fromString ....
```

kubukoz: exactly

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -1313,7 +1313,12 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
           lines(
             line"""final case class $$Unknown($paramName: $paramType) extends $name($stringValue, "$$Unknown", $intValue, Hints.empty)""",
             newline,
-            line"val $$unknown: $paramType => $name = $$Unknown(_)"
+            line"val $$unknown: $paramType => $name = $$Unknown(_)",
+            newline,
+            if (isIntEnum)
+              line"def fromIntOrUnknown(i: Int): ${name.name} = fromOrdinal(i).getOrElse($$unknown(i))"
+            else
+              line"def fromStringOrUnknown(s: String): ${name.name} = fromString(s).getOrElse($$unknown(s))"
```
kubukoz: judging by previous lines, this should work too

```suggestion
              line"def fromIntOrUnknown(i: Int): $name = fromOrdinal(i).getOrElse($$unknown(i))"
            else
              line"def fromStringOrUnknown(s: String): $name = fromString(s).getOrElse($$unknown(s))"
```

kubukoz (changes requested):
looks good, let's just add a couple tests, and a changelog entry

kubukoz (approved):
Thanks :)

PR #1763, opened 2025-06-23
Sort service operations for rendering

kubukoz:
This should make adding new operations more predictable.

I was going back and forth between sorting based on location and by name, but model serializing affects locations so it'd become too unpredictable.

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
@@ -466,6 +466,8 @@ private[codegen] class SmithyToIR(
             .asScala
             .map(_.getId())
             .toList
+            // for stability
+            .sorted
 
         val operations = recursiveOperations(shape)
           .map(model.getShape(_).asScala)
--- a/modules/bootstrapped/src/generated/com/amazonaws/dynamodb/DynamoDB.scala
+++ b/modules/bootstrapped/src/generated/com/amazonaws/dynamodb/DynamoDB.scala
@@ -35,6 +35,8 @@ import smithy4s.schema.Schema.union
 trait DynamoDBGen[F[_, _, _, _, _]] {
   self =>
 
+  /** <p>Returns the regional endpoint information.</p> */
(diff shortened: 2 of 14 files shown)


modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -466,6 +466,8 @@ private[codegen] class SmithyToIR(
             .asScala
             .map(_.getId())
             .toList
+            // for stability
+            .sorted
```
Baccata, line 470: I'm honestly not sure about this : I think operation order ought to be considered semantically significant. Isn't there a way we could retain the original order ?

kubukoz: Smithy itself treats it as a Set. We don't have access to the syntax nodes either, so I don't think we have anything to sort by.

Next approximation would be the SourceLocation of the operation itself, but that does not have to match the order of operations you import. Plus this from the description:

> I was going back and forth between sorting based on location and by name, but model serializing affects locations so it'd become too unpredictable.

Baccata: https://github.com/smithy-lang/smithy/issues/591

Baccata: sorry, just leaving the link gives a passive-aggressive vibe ... my point is that the smithy people have agreed in the past that operations and resources should be ordered. If they've broken that again, we should point it out and have them fix it.

kubukoz: oh, in that case it's probably not the ordering of ops in the service itself, but the TopDownIndex. I'll follow up.

kubukoz: https://github.com/smithy-lang/smithy/issues/2699

ordering of ops in `getOperations` is fine. The index is borked.

kubukoz:
Closing, let's hope upstream fixes https://github.com/smithy-lang/smithy/issues/2699 soon, or else we can go back to manually constructing that list.

PR #1764, opened 2025-06-23
Add HTTP method and pattern to Scaladoc 

### LINKED ISSUE #728: Documentation: simpleRestJson add http method / path to scaladoc on service
Title says it all but basically: it's not the first time I work in a project where the specification is pulled from somewhere by the build tool. Some code is generated from it and I am implementing the service, and I run it to test it. But sometimes I forget what's the path to reach the specific operation.

Would adding something like the following be interesting:

```scala
// this is generated code
package thingy

// imports omitted

trait MyServiceGen[F[_, _, _, _, _]] {
  self =>

  /**
   * Method: GET
   * Path: /health_check
  */
  def getHealth(): F[Unit, MyServiceGen.GetHealthError, HealthResponse, Nothing, Nothing]

  def transform: Transformation.PartiallyApplied[MyServiceGen[F]] = Transformation.of[MyServiceGen[F]](this)
}
// ... rest is omitted
```

That would make it easier to figure out by using the IDE go-to-definition feature and ending up in the codegen where you find the info. You still don't have all the info, but sometimes that's all you need.

Jay-Lokhande:
## PR Checklist

- [x] Updated changelog

Closes #728

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Thank you!
 * core: Fix Hints methods to distinguish member and target hints (fixes [#1658](https://github.com/disneystreaming/smithy4s/issues/1658)) in [#1756](https://github.com/disneystreaming/smithy4s/pull/1756)
 * http4s: onError method added to the RouteBuilder in [#1755](https://github.com/disneystreaming/smithy4s/pull/1755)
 * codegen: Add fromStringOrUnknown and fromIntOrUnknown methods to Open Enums (fixes [#1626](https://github.com/disneystreaming/smithy4s/issues/1626) in [#1759](https://github.com/disneystreaming/smithy4s/pull/1759))
+* codegen: Add HTTP method and pattern to Scaladoc (fixes [#728](https://github.com/disneystreaming/smithy4s/issues/728)) in [#1764](https://github.com/disneystreaming/smithy4s/pull/1764)
 
 # 0.18.37
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,13 @@ Thank you!
 # 0.18.39
 
 * http4s: Partially fix [#1619](https://github.com/disneystreaming/smithy4s/issues/1619) by checking request query parameters against the static query parameters in [#1743](https://github.com/disneystreaming/smithy4s/pull/1743)
+* codegen: Add HTTP method and pattern to Scaladoc (fixes [#728](https://github.com/disneystreaming/smithy4s/issues/728)) in [#1764](https://github.com/disneystreaming/smithy4s/pull/1764)
 
 # 0.18.38
 
 * core: Fix Hints methods to distinguish member and target hints (fixes [#1658](https://github.com/disneystreaming/smithy4s/issues/1658)) in [#1756](https://github.com/disneystreaming/smithy4s/pull/1756)
 * http4s: onError method added to the RouteBuilder in [#1755](https://github.com/disneystreaming/smithy4s/pull/1755)
 * codegen: Add fromStringOrUnknown and fromIntOrUnknown methods to Open Enums (fixes [#1626](https://github.com/disneystreaming/smithy4s/issues/1626) in [#1759](https://github.com/disneystreaming/smithy4s/pull/1759))
-* codegen: Add HTTP method and pattern to Scaladoc (fixes [#728](https://github.com/disneystreaming/smithy4s/issues/728)) in [#1764](https://github.com/disneystreaming/smithy4s/pull/1764)
 
 # 0.18.37
 
--- a/modules/bootstrapped/src/generated/com/amazonaws/dynamodb/DynamoDB.scala
+++ b/modules/bootstrapped/src/generated/com/amazonaws/dynamodb/DynamoDB.scala
@@ -38,6 +38,7 @@ trait DynamoDBGen[F[_, _, _, _, _]] {
   /** <p>Returns an array of table names associated with the current account and endpoint. The output
     *       from <code>ListTables</code> is paginated, with each page returning a maximum of 100 table
     *       names.</p>
+    * 
     * @param ExclusiveStartTableName
     *   <p>The first table name that this operation will evaluate. Use the value that was returned for
     *           <code>LastEvaluatedTableName</code> in a previous operation, so that you can obtain the next page
--- a/modules/bootstrapped/src/generated/com/amazonaws/dynamodb/DynamoDB.scala
+++ b/modules/bootstrapped/src/generated/com/amazonaws/dynamodb/DynamoDB.scala
@@ -39,7 +39,6 @@ trait DynamoDBGen[F[_, _, _, _, _]] {
     *       from <code>ListTables</code> is paginated, with each page returning a maximum of 100 table
     *       names.</p>
     * 
-    * 
     * @param ExclusiveStartTableName
     *   <p>The first table name that this operation will evaluate. Use the value that was returned for
     *           <code>LastEvaluatedTableName</code> in a previous operation, so that you can obtain the next page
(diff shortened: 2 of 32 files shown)


modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -235,7 +235,8 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
 
   private def documentationAnnotation(
       hints: List[Hint],
-      skipMemberDocs: Boolean = false
+      skipMemberDocs: Boolean = false,
+      httpInfo: Option[(String, String)] = None
```
kubukoz: **suggestion:** you already have `hints` here, could use that instead of passing an extra parameter

modules/bootstrapped/src/generated/com/amazonaws/dynamodb/DynamoDB.scala
```diff
@@ -38,6 +38,7 @@ trait DynamoDBGen[F[_, _, _, _, _]] {
   /** <p>Returns an array of table names associated with the current account and endpoint. The output
     *       from <code>ListTables</code> is paginated, with each page returning a maximum of 100 table
     *       names.</p>
+    *
```
kubukoz, line 41: **thought:** so, these are new and kinda unrelated, but I think these newlines were intended already (see the `maybeNewline` part of the Renderer). They just happened to be ignored in the past.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -244,27 +244,46 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
+              doc.memberDocLines.flatMap { case (memberName, text) =>
+                s"@param $memberName" :: text
+                  .map(literalReplacements)
+                  .map("  " + _)
+              }.toList
+
+          List(shapeDocs, memberDocs)
+        }
+
+    val httpDocs: List[String] = hints
```
kubukoz: **thought:** I'm wondering if we went the right way about this. Maybe instead of accessing the Http hint here, we should make the http docs (as well as any other docs) part of the `Documentation` hint - just like the doc hint already contains shape and member docs.

This would mean:

1. Modifying https://github.com/disneystreaming/smithy4s/blob/0c3269259ad886a2d9e3545bfb1ecf7c5cd5b12f/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala#L985-L1026 to support `HttpTrait`
2. Removing the other case of `HttpTrait` handling in `SmithyToIR`
3. Adding some kind of field to `Hint.Documentation` like `protocolSpecific: List[List[String]]` which would allow inclusion of protocol-specific documentation sections
4. Rendering `protocolSpecific` here in a similar manner to how shape/member docs were handled before.

That way, we're staying in Smithy land for the protocol-specific parts, and we keep the renderer protocol-agnostic.
What do you think @lewisjkl?

kubukoz: in addition, work done in the `SmithyToIR` converter is easier to unit test.

Jay-Lokhande: I got the points, I would like to give another try on it

lewisjkl: Yeah I agree with you Jakub, let's keep it protocol agnostic wrt the HTTP trait and go with the field in Hint.Documentation 👍

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -258,12 +258,17 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
-
+        val protocolDocs: List[String] =
+          doc.protocolSpecificLines.flatten.map(literalReplacements)
         val maybeNewline =
-          if (shapeDocs.nonEmpty && memberDocs.nonEmpty) List("", "") else Nil
-        val allDocs = shapeDocs ++ maybeNewline ++ memberDocs
+          if (
+            (shapeDocs.nonEmpty || protocolDocs.nonEmpty) && memberDocs.nonEmpty
+          )
+            List("", "")
```
kubukoz: **issue:** let's remove one

```suggestion
            List("")
```

kubukoz: also, I think the `intercalate` trick from my commit worked a bit better, because it'd also introduce such a newline between protocol and shape docs too:

```scla
val allDocs = List(
  protocolDocs,
  shapeDocs,
  memberDocs
).filterNot(_.isEmpty).intercalate(List(""))
```

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -989,6 +989,13 @@ private[codegen] class SmithyToIR(
       .getTrait(classOf[DocumentationTrait])
       .asScala
       .foldMap(doc => split(doc.getValue()))
+    val httpDocs = shape
+      .getTrait(classOf[HttpTrait])
+      .asScala
+      .map { http =>
+        List(s"Method: ${http.getMethod}", s"Pattern: ${http.getUri.toString}")
```
kubukoz: **suggestion:** 
```suggestion
        List(s"HTTP ${http.getMethod} ${http.getUri.toString}"
```

it looks close to what you'd see in an actual HTTP request:

```
POST /foo/bar HTTP/1.1
```

kubukoz:
The failed tests need updating - the implementation looks good, it's just the expectations should change

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Thank you!
 * core: Fix Hints methods to distinguish member and target hints (fixes [#1658](https://github.com/disneystreaming/smithy4s/issues/1658)) in [#1756](https://github.com/disneystreaming/smithy4s/pull/1756)
 * http4s: onError method added to the RouteBuilder in [#1755](https://github.com/disneystreaming/smithy4s/pull/1755)
 * codegen: Add fromStringOrUnknown and fromIntOrUnknown methods to Open Enums (fixes [#1626](https://github.com/disneystreaming/smithy4s/issues/1626) in [#1759](https://github.com/disneystreaming/smithy4s/pull/1759))
+* codegen: Add HTTP method and pattern to Scaladoc (fixes [#728](https://github.com/disneystreaming/smithy4s/issues/728)) in [#1764](https://github.com/disneystreaming/smithy4s/pull/1764)
```
kubukoz: **issue:** this needs to be moved to a 0.18.39 section

modules/codegen/test/src/smithy4s/codegen/internals/ScaladocSpec.scala
```diff
@@ -122,4 +124,35 @@ final class ScaladocSpec extends munit.FunSuite {
...
+        |
+        |service Service {
+        |  operations: [GetHealth]
+        |}
+        |
+        |/// Health check operation
+        |@http(method: "GET", uri: "/health_check")
+        |operation GetHealth {
+        |  input: Unit
+        |  output: Unit
```
kubukoz: **nitpick:** these are implied, you can just skip input/output
```suggestion
```

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -258,12 +258,15 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
                 .map(literalReplacements)
                 .map("  " + _)
             }.toList
+        val protocolDocs: List[String] =
+          doc.protocolSpecificLines.flatten.map(literalReplacements)
+
+        val allDocs = List(protocolDocs, shapeDocs, memberDocs)
```
kubukoz: **issue:** shapeDocs should go first, I'd swap it so

```suggestion
        val allDocs = List(shapeDocs, protocolDocs, memberDocs)
```

kubukoz (approved):

PR #1765, opened 2025-06-23
bump alloy and smithy translate

yisraelU:


project/Dependencies.scala
```diff
@@ -31,7 +31,7 @@ object Dependencies {
 
   val Alloy = new {
     val org = "com.disneystreaming.alloy"
```
kubukoz: Gotta use the new org

project/Dependencies.scala
```diff
@@ -40,7 +40,7 @@ object Dependencies {
 
   val Smithytranslate = new {
     val org = "com.disneystreaming.smithy"
```
kubukoz: **issue:** :)

lewisjkl (approved):

PR #1766, opened 2025-06-24
fix: proper ordering on `perMethodEndpoint`

### LINKED ISSUE #1567: Bad ambiguous URI resolution

### LINKED ISSUE #1753: `TopDownIndex`  to gather service operations prevent workaround on ambiguous URI resolution
I wanted to follow up on this issue: [Bad ambiguous URI resolution #1567](https://github.com/disneystreaming/smithy4s/issues/1567)

    Note: If this is from the same service, the order of operations seems to affect the order of matching.

In our case, we were relying on the order of operations as a workaround to handle ambiguous URI resolution.

It looks like the ordering logic have changed in this PR: [Use TopDownIndex to gather service operations from resources #1722](https://github.com/disneystreaming/smithy4s/pull/1722), specifically in this change: [SmithyToIR.scala#L461](https://github.com/disneystreaming/smithy4s/pull/1722/files#diff-3d87f30f2b03db7c8e012d2c726b1e1a0f1227fe70d2e5c24166a9046c4428a5R461) affecting latest release `v0.18.37`.

```scala
def recursiveOperations(
    service: ServiceShape
): List[ShapeId] =
  TopDownIndex
    .of(model)
    .getContainedOperations(service)
    .asScala
    .map(_.getId())
    .toList
```

Would it be possible to revisit this, either by restoring the previous order or exploring a more deterministic solution to the URI ambiguity?

_Thank you for your time and all the great work on the project!_

tpetillot:
Closes #1753, Closes #1567

**Motivation:**
Static segment doesn't match before label ones.

**Modification:**
Add sorting on `perMethodEdnpoint` where `List[HttpEndpointHandler]` is sorted based on path segments. Following https://smithy.io/2.0/spec/http-bindings.html#specificity-routing

**Result:**
`/hello/me` would be match before `/hello/{name}`

- [ ] Added documentation
- [ ] Updated changelog


sampleSpecs/routing.smithy
```diff
@@ -0,0 +1,61 @@
+$version: "2"
+
+namespace smithy4s.example
```
kubukoz: **suggestion:** use a different namespace so that it doesn't end up generating in `bootstrapped` (it has namespace-based filtering). I think `tests` is a good place for it, so your build.sbt change makes sense.

After you make this change, you'll need to do `bootstrapped/clean` + `bootstrapped/managedSources` to make sure the extra files are removed.

modules/compliance-tests/test/src/HttpEndpointOrderingSpec.scala
```diff
@@ -0,0 +1,79 @@
...
+        val shuffleOrder = Random.shuffle(expectedOrder)
+
+        expect(shuffleOrder.sorted == expectedOrder)
+    }
+
+    pureTest("A[x] and B[x] are both literals then continue") {
+        val a = HttpEndpointDummy(path = List(PathSegment.static("abc")))
+        val b = HttpEndpointDummy(path = List(PathSegment.static("bcd")))
+
+        expect(HttpEndpoint.ordering.compare(a, b) == 0)
```
kubukoz: There's a problem with returning 0 - in certain situations, e.g. when you put endpoints in a `SortedSet`, if the elements' comparison returns zero, they'll be considered as duplicates.

We might want to implement `PartialOrdering` instead. As far as I can tell, `/abc` and `/bcd` should not be considered equivalent

modules/core/src/smithy4s/http/HttpUnaryServerRouter.scala
```diff
@@ -180,9 +180,12 @@ object HttpUnaryServerRouter {
         .map { makeHttpEndpointHandler(_) }
         .collect { case Right(endpointWrapper) => endpointWrapper }
 
-    private val perMethodEndpoint: Map[HttpMethod, List[HttpEndpointHandler]] =
-      httpEndpointHandlers.groupBy(_.httpEndpoint.method)
 
+    private val perMethodEndpoint: Map[HttpMethod, List[HttpEndpointHandler]] = {
+      httpEndpointHandlers.groupBy(_.httpEndpoint.method).map { case (mathod, handlers) =>
```
kubukoz: **issue:** typo: mathod -> method

modules/core/src/smithy4s/http/HttpEndpoint.scala
```diff
@@ -40,6 +41,23 @@ trait HttpEndpoint[I] {
 
 object HttpEndpoint {
 
+  private implicit val pathSegmentOrdering: Ordering[PathSegment] = Ordering.by[PathSegment, Int] {
+    case _: PathSegment.StaticSegment => 0
+    case _: PathSegment.LabelSegment => 1
+    case _: PathSegment.GreedySegment => 2
+  }
+
+  @nowarn("cat=deprecation")
```
kubukoz: **suggestion:** I'd rather avoid using deprecated methods. We could order by `endpoint.path.toIndexedSeq` instead

modules/tests/src/smithy4s/tests/RoutingSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+import org.http4s.dsl.Http4sDsl
+import weaver._
+import org.http4s.client.Client
+import org.http4s.Uri
+import cats.Show
+import org.http4s.Request
+import smithy4s.example.RoutingService
+
+
+abstract class RoutingSpec
```
kubukoz: **issue:** I think we're gonna want to implement this using compliance tests in Alloy instead. Something of this sort:

https://github.com/disneystreaming/alloy/blob/aea37789a1700a47ce8854f5355a34e8a15fc025/modules/protocol-tests/resources/META-INF/smithy/AddMenuItem.smithy

That makes them more reusable across languages, and more readable too (in some ways).

If you have trouble with that, I can help set you up (would rather this got resolved before a merge than in a follow-up, so that we don't release the intermediate state and then break compatibility in the `tests` module by removing)

kubukoz (changes requested):
thanks, this is looking very good already. A few suggestions/issues to fix before merge.

modules/core/src/smithy4s/http/HttpEndpoint.scala
tpetillot: Not sure how `toIndexedSeq` can help compare two `List[PathSegment]`? It doesn't seem there's a default implicit for `Ordering[IndexedSeq[A]]`

modules/tests/src/smithy4s/tests/RoutingSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+
+  routerTest("/{abc+}/def") {
+    (client, uri) =>
+      for {
+        res <- client.send[String](GET((uri / "abc" / "def" / "def")))
+      } yield {
+        val (code, body) = res
+        expect.same(code, 200) && expect(body == "\"greedyAbcDef\"")
```
tpetillot: @kubukoz any idea what’s causing this test to fail? Possibly a greedy matching issue?

modules/core/src/smithy4s/http/HttpEndpoint.scala
tpetillot: Found it: `import scala.math.Ordering.Implicits._`

modules/tests/src/smithy4s/tests/RoutingSpec.scala
tpetillot: > A greedy label is a label suffixed with the + qualifier that can be used to match more than one path segment. At most, one greedy label may exist in any path pattern, and if present, it **SHOULD** be the last label in the pattern. Greedy labels MUST be bound to a string shape.

modules/tests/src/smithy4s/tests/RoutingSpec.scala
tpetillot: I wasn't familiar with Alloy compliance testing, but I took a look at the [Smithy docs](https://smithy.io/2.0/additional-specs/http-protocol-compliance-tests.html#httpresponsetests-trait) and tried playing with `sampleSpec/test.smithy`. It didn’t seem to fail as expected, so I’m not entirely sure how it’s meant to integrate.

It seems like an upcoming direction, but since this PR is focused on fixing a bug, would it be ok to handle the test restructuring separately?

modules/core/src/smithy4s/http/HttpEndpoint.scala
```diff
@@ -80,4 +80,48 @@ object HttpEndpoint {
...
+   * The following algorithm is used to compare two paths
+   *
+   * Given two ambiguous URI patterns A and B with segments [A0, …, An] and [B0, …, Bm] with query string literals [AQ0, …, AQp] and [BQ0, …, BQq] (with both p
+   * and q possibly zero, i.e., without query string literals), the following steps are taken to compare them, for each index x from 0 to min(n, m)
+   *
+   * If A[x] and B[x] are both literals then continue (the literal values have to be equal otherwise the patterns are not ambiguous) If A[x] is a literal and
+   * B[x] is a label then A is more specific than B, If A[x] is a non-greedy label and B[x] is a greedy label then A is more specific than B If n > m then A is
+   * more specific than B If p > q then A is more specific than B
+   */
+  def lt(left: HttpEndpoint[_], right: HttpEndpoint[_]): Boolean = {
```
tpetillot: Http endpoint "lesser than" implementation

modules/core/src/smithy4s/http/HttpUnaryServerRouter.scala
```diff
@@ -180,9 +180,13 @@ object HttpUnaryServerRouter {
...
-    private val perMethodEndpoint: Map[HttpMethod, List[HttpEndpointHandler]] =
-      httpEndpointHandlers.groupBy(_.httpEndpoint.method)
-
+    private val perMethodEndpoint: Map[HttpMethod, List[HttpEndpointHandler]] = {
+      httpEndpointHandlers.groupBy(_.httpEndpoint.method).map { case (method, handlers) =>
+        method -> handlers.sortWith { case (x, y) =>
+          HttpEndpoint.lt(x.httpEndpoint, y.httpEndpoint)
+        }
+      }
+    }
```
tpetillot: Sorting

modules/compliance-tests/test/src/HttpEndpointSpec.scala
```diff
@@ -0,0 +1,146 @@
...
+  * (with both p and q possibly zero, i.e., without query string literals), the following steps are taken to compare them, 
+  * for each index x from 0 to min(n, m)
+  * 
+  * If A[x] and B[x] are both literals then continue (the literal values have to be equal otherwise the patterns are not ambiguous)
+  * If A[x] is a literal and B[x] is a label then A is more specific than B,
+  * If A[x] is a non-greedy label and B[x] is a greedy label then A is more specific than B
+  * If n > m then A is more specific than B
+  * If p > q then A is more specific than B
+  */
+object HttpEndpointSpec extends SimpleIOSuite {
```
tpetillot: Unit test

modules/tests/src/smithy4s/tests/RoutingSpec.scala
```diff
@@ -0,0 +1,96 @@
...
+import org.http4s.client.dsl.Http4sClientDsl
+import org.http4s.dsl.Http4sDsl
+import weaver._
+import org.http4s.client.Client
+import org.http4s.Uri
+import cats.Show
+import org.http4s.Request
+import smithy4s.routing.RoutingService
+
+abstract class RoutingSpec
```
tpetillot: Integration test

modules/tests/src/smithy4s/tests/RoutingSpec.scala
kubukoz: I'd say yes, but the `tests` module is published as a Maven artifact, so it's part of the public API - I'd rather resolve this before we release it.

Let me think about alternatives...

modules/core/src/smithy4s/http/HttpUnaryServerRouter.scala
kubukoz: TIL `sortBy` doesn't work with `PartialOrdering`

kubukoz: ah, you didn't use `PartialOrdering` after all - that also explains it

kubukoz:
I'll try to give this more attention tomorrow, busy day 🫠

kubukoz:
I managed to move the routing tests to Alloy.

Running on smithy4s 0.18:

```
[error] [http4s] - alloy.test.routing#AbcLabel(server|Request): RoutingAbcLabel 1ms
[error] [http4s]   unexpected error when running test Wrong endpoint called:  (modules/tests/src/smithy4s/tests/ProtocolComplianceSuite.scala:158)
[error] [http4s]    java.lang.Throwable: Wrong endpoint called
[error] [http4s] 
[error] [http4s]     [INFO] 19:26:27 [ProtocolComplianceSuite.scala:175] Test the AbcLabel operation with a label
[error] [http4s] 
[error] [http4s] - alloy.test.routing#AbcXyz(server|Request): RoutingAbcXyz 1ms
[error] [http4s]   unexpected error when running test Wrong endpoint called:  (modules/tests/src/smithy4s/tests/ProtocolComplianceSuite.scala:158)
[error] [http4s]    java.lang.Throwable: Wrong endpoint called
[error] [http4s] 
[error] [http4s]     [INFO] 19:26:27 [ProtocolComplianceSuite.scala:175] Test the AbcXyz operation
[error] [http4s] 
[error] [http4s] Failed: Total 337, Failed 2, Errors 0, Passed 335, Ignored 60, Canceled 33
[error] [http4s] Failed tests:
[error] [http4s]        smithy4s.http4s.SimpleRestJsonComplianceSuite
[error] [http4s] (http4s / Test / testOnly) sbt.TestsFailedException: Tests unsuccessful
```

Running on this branch:

```
[info] [http4s] Passed: Total 337, Failed 0, Errors 0, Passed 337, Ignored 60, Canceled 33
```

Notably, the tests were already passing if I just copied the operations from this PR verbatim, because ordering of ops accidentally matched the one we expected. I added an extra operation and renamed the greedy one to force some conflicts to happen no matter what order is used.

Will create an arlib PR soon and update this one 👍

kubukoz:
Alloy PR is merged, now we just need a release.

kubukoz:
this can go in after #1771

tpetillot:
@kubukoz Nice, I now see where you're going with generated specs ! 👏 Let me know if there's anything I can do to help get this merged.

kubukoz:
thanks, hopefully this is it :)

modules/compliance-tests/test/src/HttpEndpointSpec.scala
kubukoz: moving this to the `core` module (`compliance-tests` is a more specialised module)

kubukoz (approved):

kubukoz:
Thanks @tpetillot! It's great to see ambiguous URIs finally fixed :)

tpetillot:
Really appreciated your help and follow-up on this! 🎊

tpetillot:
Any idea regarding a shipping date for the fix @kubukoz ?

kubukoz:
I was waiting for #1769, we should be good to go now.

slandelle:
@kubukoz Gentle ping.
Can we expect a release with this regression fix, or will we have to wait for 0.19?
Asking because we had to pin scala-steward updates for smithy4s and we're eager to lift the pin.
Regards

kubukoz:
hey @slandelle, this is set to release in 0.18.39, we've just been having trouble with the new Sonatype Central uploads. I too hope we can get this delivered quickly.

re: steward, you can replace the pin with a retraction:

```hocon
updates.retracted = [
  {
    reason = "Ignore version 0.18.38 as it contains a problematic version of alloy",
    doc = "https://github.com/disneystreaming/alloy/pull/237",
    artifacts = [
      { groupId = "com.disneystreaming.smithy4s", version = { exact = "0.18.38" } }
    ]
  }
]
```

edit: I made a PR to make it part of the default config: https://github.com/scala-steward-org/scala-steward/pull/3674

kubukoz:
@slandelle it's out :)

slandelle:
@kubukoz Thanks a lot! And thanks for the scala-steward tips ;)

PR #1774, opened 2025-07-07
0.18 into 0.19, 2025.08

kubukoz:

- [ ] Added documentation
- [ ] Updated changelog


a-morales:
Went ahead and made this compile. I looked at the compiler error, but then saw this PR https://github.com/disneystreaming/smithy4s/pull/1667 So I think my changes are correct.

kubukoz:
looks about right! I'll give this some attention this week.

lewisjkl (approved):
Looks good, nothing unexpected to me 👍

lewisjkl:
I will keep going on this on Monday.. I was trying to get the latest merged in, but it looks like there's a bit more work I need to do to make 0.19 happy with bincompat friendly mode

kubukoz:
ah, bincompat mode... This one is on me:

https://github.com/disneystreaming/smithy4s/blob/8b2169c8f695975667e862484c28d489307da532/modules/codegen/test/src/smithy4s/codegen/BincompatCodegenIntegrationSpec.scala#L729-L732

Probably the easiest way to solve this:

1. Extract a separate module for these "integration" tests, so that they can depend on the locally-published snapshot of core instead of a stable version. The publishing of the snapshot can happen in a similar manner to how we do publishLocal for scripted tests.
2. Use `BuildInfo.version` instead of `latest.stable`

Step 1 eliminates the circular dependency the comment talks about (since this is in the tests of codegen, and codegen is used to generate core)

lewisjkl:
Hmm idk what is going on here now. Tests are passing for me locally, but seems to have an issue in CI. @a-morales or @kubukoz any ideas?

kubukoz:
Race condition on scala-cli starting the build server? I've seen it fail to do that, sporadically, even without concurrency. I'll give it a re-run.

build.sbt
```diff
@@ -482,6 +486,43 @@ lazy val codegen = projectMatrix
...
+    buildInfoKeys := Seq[BuildInfoKey](
+      version,
+      scalaBinaryVersion,
+      "smithyOrg" -> Dependencies.Smithy.org,
+      "smithyVersion" -> Dependencies.Smithy.smithyVersion,
+      "alloyOrg" -> Dependencies.Alloy.org,
+      "alloyVersion" -> Dependencies.Alloy.alloyVersion,
+      "smithy4sOrg" -> organization.value,
+      "protocolArtifactName" -> "smithy4s-protocol"
+    ),
```
kubukoz: question: can we use the buildinfo from `codegen` instead? IIRC it has everything but I might be missing some nuance.

a-morales:
So I can't seem to figure why the test is timing out even though we pass a timeout that is longer than what is shown in the error that is thrown. Any ideas what is going on @kubukoz

kubukoz:
The 30 second timeout is something internal to scala-cli/blooprifle. I would try the following:

- run a dummy `scala-cli compile` command on some arbitrary temporary directory, in the CI steps, before we run any tests
- add multiple `-v` flags to the scala-cli invocations so that we get more diagnostic info

kubukoz:
@a-morales it's a race condition. I have a fix, just testing it locally now

kubukoz:
basically, the core publishing wasn't happening before the test, but parallel to it. This isn't a problem when we build an "aggregate publishLocal" like in the case of the build tool plugins, but it's a problem here because there actually is an implicit dependency.

How I knew: after you added the temp directory run, I saw in the error message that there was a "not found" on the core module, so I inspected how we actually publish it...

somewhere in the meantime I ran the test locally, and noticed that it was failing while the `core3` module was still compiling. That confirmed the suspicion of the race condition.

kubukoz:
man, finally

a-morales (approved):

kubukoz:
I'll try to squash the fixup commits (as we're not squashing this full PR), bear with me

kubukoz:
actually, no, even a no-op rebase tries to have me resolving conflicts...

let's just have a bit of dirty history on main :P

PR #1775, opened 2025-07-09
Fix publishing

a-morales:
Fixes the error that showed up when trying to publish a non-snapshot build to Sonatype. The `ci-release` plugin runs the [`clean` command by default](https://github.com/sbt/sbt-ci-release/blob/main/plugin/src/main/scala/com/geirsson/CiReleasePlugin.scala#L208) and we add the generated files from the bootstrapped module into the list of files to clean https://github.com/disneystreaming/smithy4s/blob/series/0.18/build.sbt#L962-L965. 

Since files are removed, the git repo is dirty and the version get's updated mid-release. I figured if we are at the release stage we don't need to do a clean. 

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -138,6 +138,8 @@ jobs:
       - name: Publish ${{ github.ref }}
         run: sbt 'show version; ci-release'
         env:
+          # cleaning will removed generated code from bootstrapped module which will
+          # cause the repo to be dirty and change the version to a -SNAPSHOT version
           CI_CLEAN: ""
           PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
           PGP_SECRET: ${{ secrets.PGP_SECRET }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -138,6 +138,7 @@ jobs:
       - name: Publish ${{ github.ref }}
         run: sbt 'show version; ci-release'
         env:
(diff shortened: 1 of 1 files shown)


a-morales:
Kicked off a snapshot release to make sure this work as the snapshot workflow also uses `CI_CLEAN`. Looking at the snapshot publish from the last PR, the version listed under `sbt show version` was `0.18.38-38-e0d4391c-SNAPSHOT`, but the version published was `0.18.38-38-e0d4391c-20250707-2012-SNAPSHOT` so the git repo got dirty by the clean.

yisraelU:
@a-morales snapshot should be fixed too by your change

kubukoz:
wow, what a find @a-morales!

.github/workflows/ci.yml
```diff
@@ -138,6 +138,7 @@ jobs:
       - name: Publish ${{ github.ref }}
         run: sbt 'show version; ci-release'
         env:
+          CI_CLEAN: ""
```
kubukoz, line 143: **nitpick:** I would've suggested to keep a comment here about the reason it has to be there, but looking up this PR in git history shouldn't be much more difficult than that, so we'll know why it's there.

kubukoz (approved):

.github/workflows/ci.yml
a-morales: I'll add a comment.

PR #1776, opened 2025-07-11
POC: Generate bincompat-friendly core module

kubukoz:

- [ ] Added documentation
- [ ] Updated changelog


modules/core/src-jvm-native/smithy4s/Timestamp.scala
```diff
@@ -29,6 +29,7 @@ case class Timestamp private (epochSecond: Long, nano: Int)
     diff > 0 || diff == 0 && nano > other.nano
   }
 
+  @annotation.nowarn
```
kubukoz, line 32: **thought:** maybe we should make exceptions for "reasonably final" shapes like TimestampFormat... but then again, why would we assume it's reasonably final?

kubukoz:
as expected, this fails mima checks - the switch to bincompat-friendly itself is breaking, of course.

kubukoz:
this is in 0.19!

PR #1781, opened 2025-07-21
Remove deprecations for 0.19

### LINKED ISSUE #1330: 0.19: Remove deprecations
Remove everything that was deprecated in 0.18 and before.

This will break existing usage, so it should target the `series/0.19` branch.

stewSquared:
Resolves #1330 

- [ ] ~Added unit-tests (for runtime code)~
- [ ] ~Added bootstrapped code + smoke tests (when the rendering logic is modified)~
- [ ] ~Added build-plugins integration tests (when reflection loading is required at codegen-time)~
- [ ] ~Added alloy compliance tests (when simpleRestJson protocol behaviour is expanded/updated)~
- [ ] ~Updated dynamic module to match generated-code behaviour~
- [ ] ~Added documentation~
- [ ] Updated changelog

I imagine updating the changelog is unnecessary if the deprecations were previously mentioned.

### DIFF
--- a/modules/bootstrapped/test/src/smithy4s/DocumentSpec.scala
+++ b/modules/bootstrapped/test/src/smithy4s/DocumentSpec.scala
@@ -618,7 +618,6 @@ class DocumentSpec() extends ScalaCheckSuite {
     "document encoder - default values overrides + explicit defaults encoding = false"
   ) {
     val result = Document.Encoder
-      .withFieldFilter(FieldFilter.Default)
       .fromSchema(DefaultNullsOperationOutput.schema)
       .encode(
         DefaultNullsOperationOutput(
@@ -656,7 +655,7 @@ class DocumentSpec() extends ScalaCheckSuite {
 
   test("Document encoder - timestamp defaults") {
     val result = Document.Encoder
-      .withExplicitDefaultsEncoding(false)
+      .withFieldFilter(FieldFilter.Default)
       .fromSchema(TimestampOperationInput.schema)
       .encode(TimestampOperationInput())
     expect.same(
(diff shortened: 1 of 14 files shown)


modules/core/src/smithy4s/http/Metadata.scala
```diff
@@ -267,16 +251,6 @@ object Metadata {
   ) extends CachedSchemaCompiler.DerivingImpl[Encoder]
       with EncoderCompiler {
 
-    @deprecated
```
stewSquared, line 270: Just noting this didn't have a message or version in case that warrants scrutiny.

modules/core/src/smithy4s/internals/DocumentEncoderSchemaVisitor.scala
```diff
@@ -96,10 +96,6 @@ class DocumentEncoderSchemaVisitor(
   def this(cache: CompilationCache[DocumentEncoder]) =
     this(cache, explicitDefaultsEncoding = false)
 
-  @deprecated
```
stewSquared, line 99: Just noting this didn't have a message or version in case that warrants scrutiny.

modules/bootstrapped/test/src/smithy4s/DocumentSpec.scala
```diff
@@ -656,7 +656,7 @@ class DocumentSpec() extends ScalaCheckSuite {
 
   test("Document encoder - timestamp defaults") {
     val result = Document.Encoder
-      .withExplicitDefaultsEncoding(false)
+      .withFieldFilter(FieldFilter.Default)
```
kubukoz, line 658: **nitpick:** since it's the default, we can just skip this line. Applies to both cases

kubukoz:
Normally I'd only get to such tasks once we're relatively close to that 0.19 release, but it doesn't hurt so let's go with it :)

kubukoz:
superseded by #1877

Issue #1713, opened 2025-05-08
Open unions: tagged unions keep the whole object even if extra keys are present


kubukoz:
To reproduce:

```smithy
$version: "2"

namespace demo

use alloy#jsonUnknown

union MyOpenUnion {
  s1: Unit
  s2: Unit
  @jsonUnknown other: Document
}

```

Read this:

```json
{"a": 42, "b": "oops"}
```

it'll all be persisted in the Document. I believe the behavior should be to fail decoding due to multiple root keys.

You can try it out in Smithy Transcoder: https://kubukoz.github.io/smithy-transcoder/#N4Igtg9gJgpgNgFRgDwC4gFwgCQDcYBOAzgJYQB2GABADogBMdN5z5AhmDEQA5sDGMKrEjNmAVyKC2cOBACeAYgBWRCgFVyAa3IQA7iwNjyZclQCycgPLcY5DSarBmVKkQCM1e6mev6n496mVAACKupaOvpUEKgAFoTUACIQfGKc5IEAviAANCAAZhAEYGzoGKBh5Jig+STwUABiJHCohJggiQCiDQCCagAyCCCZmXkk5NxiZSBOIGx01AAs9Dm0IABGC2sQENxEdNl5+MQmmG6ZQA

kubukoz:
#1782 has fixed this, extra keys are now forbidden:

<img width="1386" height="619" alt="Image" src="https://github.com/user-attachments/assets/dcbb6344-11ef-4225-a1fc-d4cc8be512c7" />

kubukoz:
closed in #1782

PR #1782, opened 2025-07-22
Open lenient unions json support

msosnicki:

- [ ] Added documentation
- [x] Updated changelog


msosnicki:
~~TODO: update changelog~~

Done

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -1094,7 +1135,6 @@ private[smithy4s] class SchemaVisitorJCodec(
       }
     }
 
-  // todo: open unions here too
```
kubukoz, line 1097: **thought:** heh 😂

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -993,25 +993,68 @@ private[smithy4s] class SchemaVisitorJCodec(
...
+        alt.inject(handleInner(cursor, reader))
+      }
+
+      def handleTransformDocument(cursor: Cursor, reader: JsonReader)(
+          transform: Document => Document
+      ): U = {
+        val u = handleInner(cursor, reader)
+        alt.inject(documentTransformer(u, transform))
       }
+      def handleWrapInObj(key: String)(cursor: Cursor, reader: JsonReader): U =
```
kubukoz: **suggestion:** uncurry? we always pass all the params at once

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -1081,9 +1123,8 @@ private[smithy4s] class SchemaVisitorJCodec(
               }
             }
             // @jsonUnknown path
-            else if (unknownTagHandler ne null) {
-              in.rollbackToMark()
-              unknownTagHandler(cursor, in)
+            else if (unknownHandler ne null) {
+              unknownHandler.handleWrapInObj(key)(cursor, in)
```
kubukoz: **question:** don't we need to make sure a `}` token follows this object?

I like the idea though, the document transform is quite brilliant and saves us having to deal with the markers 👍

I think it also somewhat fixes https://github.com/disneystreaming/smithy4s/issues/1713. If I'm not mistaken, the only remaining part would be that we should reject multiple keys in the non-lenient case (and not just ignore them).

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
msosnicki: Refactored a bit more ;)

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
msosnicki: Great catch! Added a test, it was indeed failing :scream: 

Yeah, the marker approach was falling short in the following case:

```smithy
union SampleOpenUnion {
    str: String

    u: Unit

    @jsonUnknown
    unknown: Document
}
```

```json
{
  "str": null,
  "u": null,
  "other": {"whatever": "here"}
}
```

should be decoded as

```scala
SampleOpenUnion.unknown(
   Document.obj(
      "other" -> Document.obj("whatever" -> Document.fromString("here"))
   )
)
```

It would be cumbersome to get this behavior and keeping the whole object is not correct

It fixes the issue you mentioned, dropped a test for it.

msosnicki:
In the last commit made attempt to unify the tagged decoding process between lenient & non lenient cause there are some subtle things with the marker position and everything that need to be handled for each case separately, and there are a lot of similarities. 
I can revert it as it's not affecting the tests or anything

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -993,29 +996,89 @@ private[smithy4s] class SchemaVisitorJCodec(
...
+      }.toMap
+
+    private val unknownAlt = alternatives.find(_.hints.has(JsonUnknown))
+
+    protected def getHandler(key: String) = handlerMap
+      .get(key)
+      .orElse(
+        unknownAlt.map(alt =>
+          if (isDiscriminated) UnionJCodec.AltHandler.create(alt)
+          else UnionJCodec.unknownAltHandler(alt, key)
```
msosnicki: ~~TODO: compiled per invocation~~
Fixed

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -993,29 +996,103 @@ private[smithy4s] class SchemaVisitorJCodec(
...
+  }
+
+  private object UnionJCodec {
+
+    private type DocumentTransformer[A] = (A, Document => Document) => A
+
+    protected abstract class AltHandler[U, A] {
+      def alt: Alt[U, A]
+      def handle(cursor: Cursor, reader: JsonReader): U
+      def handleVariant(cursor: Cursor, reader: JsonReader): A
```
kubukoz: minor suggestion: I didn't like that we had 3 abstract methods here, plus `handleVariant` in `Mapped` didn't use its `map` (because it's never actually used, but that's not obvious from the code).

How about this:

```diff
diff --git a/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala b/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
index af83ee274..66b53944c 100644
--- a/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
+++ b/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
@@ -1021,13 +1021,15 @@ private[smithy4s] class SchemaVisitorJCodec(
     private type DocumentTransformer[A] = (A, Document => Document) => A
 
     protected abstract class AltHandler[U, A] {
-      def alt: Alt[U, A]
-      def handle(cursor: Cursor, reader: JsonReader): U
-      def handleVariant(cursor: Cursor, reader: JsonReader): A
+      def handle(cursor: Cursor, reader: JsonReader): U =
+        inject(handleVariant(cursor, reader))
+
+      protected def inject(a: A): U
+      protected def handleVariant(cursor: Cursor, reader: JsonReader): A
     }
 
     private object AltHandler {
-      def create[U, A](alt: Alt[U, A]): AltHandler[U, A] = new Impl(alt)
+      def create[U, A](alt: Alt[U, A]): AltHandler[U, A] = new FromAlt(alt)
 
       def openUnionUnknown[U, A](
           alt: Alt[U, A]
@@ -1041,16 +1043,13 @@ private[smithy4s] class SchemaVisitorJCodec(
           )
       }
 
-      private final class Impl[U, A](_alt: Alt[U, A]) extends AltHandler[U, A] {
+      private final class FromAlt[U, A](alt: Alt[U, A])
+          extends AltHandler[U, A] {
 
         private val codec = self.apply(alt.schema)
 
-        def alt: Alt[U, A] = _alt
-
-        def handle(cursor: Cursor, reader: JsonReader): U =
-          alt.inject(handleVariant(cursor, reader))
-
-        def handleVariant(cursor: Cursor, reader: JsonReader): A =
+        protected def inject(a: A): U = alt.inject(a)
+        protected def handleVariant(cursor: Cursor, reader: JsonReader): A =
           cursor.decode(codec, reader)
       }
 
@@ -1058,11 +1057,10 @@ private[smithy4s] class SchemaVisitorJCodec(
           underlying: AltHandler[U, A],
           map: A => A
       ) extends AltHandler[U, A] {
-        def alt: Alt[U, A] = underlying.alt
-        def handle(cursor: Cursor, reader: JsonReader): U =
-          alt.inject(map(underlying.handleVariant(cursor, reader)))
+        protected def inject(a: A): U = underlying.inject(a)
+
         def handleVariant(cursor: Cursor, reader: JsonReader): A =
-          underlying.handleVariant(cursor, reader)
+          map(underlying.handleVariant(cursor, reader))
       }
 
     }

```

kubukoz (approved):

kubukoz:
oh and add that changelog entry please, then we're good to merge

kubukoz:
> In the last commit made attempt to unify the tagged decoding process between lenient & non lenient cause there are some subtle things with the marker position and everything that need to be handled for each case separately, and there are a lot of similarities. I can revert it as it's not affecting the tests or anything

I think we'd need to benchmark to see if there's a difference. One would hope that the JVM is smart enough to optimize away the `if(isLenient)` checks, but without evidence it's just a bet

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
msosnicki: Yeah, exposing single method makes it cleaner, done

msosnicki:
> > In the last commit made attempt to unify the tagged decoding process between lenient & non lenient cause there are some subtle things with the marker position and everything that need to be handled for each case separately, and there are a lot of similarities. I can revert it as it's not affecting the tests or anything
> 
> I think we'd need to benchmark to see if there's a difference. One would hope that the JVM is smart enough to optimize away the `if(isLenient)` checks, but without evidence it's just a bet

0.18.39
```
[info] JsonBenchmark.lenientObj            thrpt    2   2934527.574          ops/s
[info] JsonBenchmark.lenientStr            thrpt    2   6828315.368          ops/s
[info] JsonBenchmark.lenientStrDecode      thrpt    2   5300187.016          ops/s
[info] JsonBenchmark.lenientStructDecode   thrpt    2   2532945.676          ops/s
[info] JsonBenchmark.regularObj            thrpt    2   3104166.638          ops/s
[info] JsonBenchmark.regularStr            thrpt    2   8329803.649          ops/s
[info] JsonBenchmark.regularStrDecode      thrpt    2  10487466.535          ops/s
[info] JsonBenchmark.regularStructDecode   thrpt    2   2712922.118          ops/s
[info] JsonBenchmark.regularUnknownDecode  thrpt    2   5392311.633          ops/s
```


This PR
```
[info] JsonBenchmark.lenientObj            thrpt    2   3023579.251          ops/s
[info] JsonBenchmark.lenientStr            thrpt    2   8635117.370          ops/s
[info] JsonBenchmark.lenientStrDecode      thrpt    2   6620934.813          ops/s
[info] JsonBenchmark.lenientStructDecode   thrpt    2   2941288.476          ops/s
[info] JsonBenchmark.lenientUnknownDecode  thrpt    2   3229369.633          ops/s
[info] JsonBenchmark.regularObj            thrpt    2   2894345.779          ops/s
[info] JsonBenchmark.regularStr            thrpt    2   8142303.405          ops/s
[info] JsonBenchmark.regularStrDecode      thrpt    2  11953159.338          ops/s
[info] JsonBenchmark.regularStructDecode   thrpt    2   3780413.114          ops/s
[info] JsonBenchmark.regularUnknownDecode  thrpt    2   5144994.675          ops/s
```

Results are more or less the same, the benchmark code can be found [here](https://gist.github.com/msosnicki/e8f6838f14dfd3ae553b7ce9d509a61f)

kubukoz:
nice! Can you expand the existing benchmarks in this repo in a follow-up? https://github.com/disneystreaming/smithy4s/tree/series/0.18/modules/benchmark/src/main/scala

kubukoz (approved):

PR #1786, opened 2025-07-25
WIP: Update primitives

a-morales:
This is dependent on https://github.com/disneystreaming/alloy/pull/239 and I'll make a PR in smithy-translate for stuff that is needed in the `protobuf` module, but this should be ready to review. Main thing in this PR is that I added `LocalDate`, `LocalTime` and `OffsetDateTime` implementations similar to `Timestamp` so that we don't have to bring in scala-java-time or something similar for ScalaJS support. 

Some open question I have:

- Should we move the new time related stuff to `smithy.time`?
- Is there a reason why there are two different implementations for Timestamp? One is for ScalaJS and the other is for jvm-native, main difference I saw between the two is some differences in how time and nanos are formatted and some JS specific functions to convert from/to a JS Date which can be moved to the `TimestampPlatform` and `TimestampCompanionPlatform` traits. 

Things still left to do:

- address some TODOs I left
- DRY up some of the repeated code in the time utilities. 

- [ ] Added documentation
- [ ] Updated changelog


a-morales:
I accidentally made this branch off the 0.18 branch instead of 0.19 and fixing the merge conflicts looked painful. I cherry picked my commits into another branch that is off the 0.19 branch and made another PR so closing this one in favor of the other which is here https://github.com/disneystreaming/smithy4s/pull/1787.

PR #1787, opened 2025-07-25
Add LocalDate, LocalTime, Duration and OffsetDateTime to primitives

a-morales:
This is dependent on https://github.com/disneystreaming/alloy/pull/239 and I'll make a PR in smithy-translate for stuff that is needed in the `protobuf` module, but this should be ready to review. Main thing in this PR is that I added `LocalDate`, `LocalTime` and `OffsetDateTime` implementations similar to `Timestamp` so that we don't have to bring in scala-java-time or something similar for ScalaJS support. 

Some open question I have:

- Should we move the new time related stuff to `smithy.time`?
- Is there a reason why there are two different implementations for Timestamp? One is for ScalaJS and the other is for jvm-native, main difference I saw between the two is some differences in how time and nanos are formatted and some JS specific functions to convert from/to a JS Date which can be moved to the `TimestampPlatform` and `TimestampCompanionPlatform` traits. 

Things still left to do:

- address some TODOs I left
- DRY up some of the repeated code in the time utilities. 
- The original draft PR https://github.com/disneystreaming/smithy4s/pull/1786 was against the 0.18 series branch. I think I cherry picked everything correctly but need to double check. 

- [ ] Added documentation
- [ ] Updated changelog


kubukoz:
> Should we move the new time related stuff to smithy.time?

I think `smithy4s.time` makes sense

> Is there a reason why there are two different implementations for Timestamp? 

I don't know of any - by what you're saying it looks like we could unify the two and just have platform-specific code for the conversions.

modules/cats/test/src/ShowVisitorSpec.scala
```diff
@@ -117,6 +119,42 @@ object ShowVisitorSpec extends FunSuite with CompatProvider {
...
+  }
+
+  test("localTime") {
+    val schema: Schema[LocalTime] = localtime
+    val foo = LocalTime(12, 24, 48, 0)
+    val showOutput = schemaVisitorShow(schema).show(foo)
+    expect.eql(showOutput, "12:24:48")
+  }
+
+  // TODO: For duration should show, show it as seconds.nanos format or the default `toString` for duration?
```
lewisjkl: Hmm since the show visitor is used more as a "inspect this thing" rather than creating a "machine-friendly encoding" like JSON, I think it is fine as is 🤷

modules/codegen/src/smithy4s/codegen/internals/CollisionAvoidance.scala
```diff
@@ -396,6 +396,10 @@ private[internals] object CollisionAvoidance {
     val timestamp_ = NameRef("smithy4s", "Timestamp")
     val document_ = NameRef("smithy4s", "Document")
     val uuid_ = NameRef("smithy4s", "UUID")
+    val localdate_ = NameRef("smithy4s", "LocalDate")
+    val localtime_ = NameRef("smithy4s", "LocalTime")
+    val duration_ = NameRef("java.time", "Duration")
+    val offsetdatetime_ = NameRef("java.time", "OffsetDateTime")
```
lewisjkl: I thought these were:

```suggestion
    val duration_ = NameRef("scala.concurrent.duration.", "Duration")
    val offsetdatetime_ = NameRef("smithy4s", "OffsetDateTime")
```

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -1667,6 +1671,14 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
       case Primitive.Timestamp =>
         ts => line"${NameRef("smithy4s", "Timestamp")}(${ts.getEpochSecond()}L, ${ts.getNano()})"
+      case Primitive.LocalDate =>
+        date => line"${NameRef("smithy4s", "LocalDate")}(${date.toEpochDay()})"
+      case Primitive.LocalTime =>
+        time => line"${NameRef("smithy4s", "LocalTime")}(${time.toSecondOfDay()}, ${time.getNano()})"
+      case Primitive.OffsetDateTime =>
+        time => line"""${NameRef("smithy4s", "OffsetDateTime")}(${time.toEpochSecond()}, ${time.getNano()}, scala.concurrent.duration.Duration(${time.getOffset().getTotalSeconds()}, "seconds"))"""
+      case Primitive.Duration =>
+        duration => line"${NameRef("scala.concurrent.duration", "Duration")}(${renderStringLiteral(duration.toString)})"
```
lewisjkl: I think the constants I put another comment on can be used instead of the one-off `NameRef`s here.

modules/core/src-js/smithy4s/LocalDateCompanionPlatform.scala
```diff
@@ -0,0 +1,33 @@
...
+ *  limitations under the License.
+ */
+
+package smithy4s
+
+import scalajs.js.Date
+
+private[smithy4s] trait LocalDateCompanionPlatform {
+
+  /** JVM platform only method */
```
lewisjkl:
```suggestion
  /** JS platform only method */
```
?

lewisjkl (commented):
Not done reviewing yet, will continue on tomorrow when I have some more time. This was a ton of work, good job getting through all of it!

modules/core/src/smithy4s/LocalDate.scala
```diff
@@ -0,0 +1,138 @@
...
+
+  val epoch = LocalDate(0)
+
+  def apply(epochDay: Long): LocalDate = new LocalDate(epochDay)
+
+  def apply(
+      year: Int, 
+      month: Int,
+      day: Int,
+  ): LocalDate = {
```
lewisjkl: What is the reason for throwing exceptions with invalid input rather than using an Either or Option? Perhaps we should make it clear this method is unsafe if we keep as is `unsafeApply` or something.

modules/core/src/smithy4s/LocalTime.scala
```diff
@@ -0,0 +1,148 @@
...
+    )
+    require(nano >= 0 && nano <= 999999999, "illegal nano")
+    new LocalTime(seconds, nano)
+  }
+
+  def apply(hour: Int, minute: Int, second: Int, nano: Int = 0): LocalTime = {
+    require(hour >= 0 && hour <=23, "illegal hour")
+    require(minute >= 0 && minute <=59, "illegal minute")
+    require(second >= 0 && second <=59, "illegal second")
+    require(nano >= 0 && nano <= 999999999, "illegal nano")
```
lewisjkl: Same as above with the unsafe methods

modules/core/src/smithy4s/OffsetDateTime.scala
```diff
@@ -0,0 +1,327 @@
...
+    require(month >= 1 && month <= 12, "illegal month")
+    require(
+      day >= 1 && (day <= 28 || day <= TimeUtil.maxDayForYearMonth(year, month)),
+      "illegal year, month, day combination"
+    )
+    require(hour >= 0 && hour <= 23, "illegal hour")
+    require(minute >= 0 && minute <= 59, "illegal minute")
+    require(second >= 0 && second <= 59, "illegal second")
+    require(nano >= 0 && nano <= 999999999, "illegal nano")
+    require(offset.toHours <= 18, "illegal offset")
```
lewisjkl: ditto

modules/core/src/smithy4s/TimeUtil.scala
```diff
@@ -0,0 +1,157 @@
...
+    else 28
+
+  def isLeap(year: Int): Boolean =
+    (year & 0x3) == 0 && { // (year % 100 != 0 || year % 400 == 0)
+      val cp = year * 1374389535L
+      val cc = year >> 31
+      ((cp ^ cc) & 0x1fc0000000L) != 0 || (((cp >> 37) - cc) & 0x3) == 0
+    }
+
+  // def parseOffset(s: String): Int = {
```
lewisjkl: Can delete?

modules/core/src/smithy4s/internals/DocumentDecoderSchemaVisitor.scala
```diff
@@ -164,6 +165,22 @@ class DocumentDecoderSchemaVisitor(
...
+      from("LocalTime") { case DString(string) =>
+        LocalTime.parseUnsafe(string)
+      }
+    case PDuration =>
+      from("Duration") { case DString(string) =>
+        Duration(string)
+      }
+    case POffsetDateTime => 
+      from("OffsetDateTime") { case DString(string) =>
+        OffsetDateTime.parseUnsafe(string)
```
lewisjkl: I think these should use `fromUnsafe` instead of `from` so that `PayloadError` will be thrown ultimately instead of random errors

modules/core/src/smithy4s/internals/DocumentKeyDecoder.scala
```diff
@@ -125,6 +129,20 @@ object DocumentKeyDecoder {
...
+          }
+
+          case PLocalTime => from(shortDesc) { case DString(string) =>
+            LocalTime.parseUnsafe(string)
+          }
+          case PDuration => from(shortDesc) { case DString(string) =>
+            Duration(string)
+          }
+          case POffsetDateTime => from(shortDesc) { case DString(string) =>
+            OffsetDateTime.parseUnsafe(string)
```
lewisjkl: Ditto with fromUnsafe

lewisjkl (commented):
Went through some more of it today and spotted a few more things. I know it is still WIP so feel free to ignore if any of the comments are not relevant (based on refactors/updates you are already doing)

modules/core/src/smithy4s/LocalDate.scala
a-morales: Just saw how it was done in the `Timestamp` and decided to keep it consistent with that, but I have nothing against making this return an `Either` or `Option`

 @kubukoz @Baccata have any opinions on this?

modules/core/src/smithy4s/TimeUtil.scala
a-morales: I think I may add a `smithy4s.time.Offset` type that can handle parsing an offset. The protobuf model for `OffsetDateTime` has 3 fields for `epochSeconds`, `nano` and `offset` where the offset is a string.

modules/core/src/smithy4s/LocalDate.scala
lewisjkl: Okay yeah I think keeping it consistent is fine and we can refactor all of them in one go if we decide that's the right path forward

modules/core/src/smithy4s/time/LocalTime.scala
```diff
@@ -0,0 +1,149 @@
...
+  def fromSeconds(seconds: Int): LocalTime = LocalTime(seconds, 0)
+
+  def parse(string: String): Option[LocalTime] = try {
+    Some(parseUnsafe(string))
+  } catch {
+    case NonFatal(_) => None
+  }
+
+  def parseUnsafe(s: String): LocalTime = {
+    println(s"parsing localTime $s")
```
lewisjkl:
```suggestion
```

modules/json/test/src/smithy4s/json/SchemaVisitorJCodecPropertyTests.scala
```diff
@@ -19,19 +19,19 @@ package json
 
 import cats.Show
 import com.github.plokhotnyuk.jsoniter_scala.core._
+import munit._
```
lewisjkl, line 22: Putting this comment on a random line, but I think we can add the new primitive types to https://github.com/disneystreaming/smithy4s/blob/f0b1e647e077f411438034601e12d45f6ab41999/modules/scalacheck/src/smithy4s/scalacheck/SchemaGenerator.scala#L49 and that will make it so these property tests take them into consideration

modules/json/test/src/smithy4s/json/SchemaVisitorJCodecTests.scala
```diff
@@ -1186,4 +1189,24 @@ class SchemaVisitorJCodecTests() extends FunSuite {
...
+  case class MyDuration(a: Duration)
+
+  object MyDuration {
+    implicit val schema: Schema[MyDuration] = {
+      val a = duration.required[MyDuration]("a", _.a)
+      struct(a)(MyDuration.apply)
+    }
+  }
+
+  test("Duration are correctly encoded/decoded from a BigDecimal") {
```
lewisjkl, line 1216: Can we also add test cases for the other new primitive types?

modules/xml/test/src/smithy4s/xml/XmlCodecSpec.scala
```diff
@@ -30,10 +30,11 @@ import smithy4s.Blob
 import smithy4s.Document
 import smithy4s.Hints
 import smithy4s.ShapeId
+import smithy4s.example.OpenEnumTest
+import smithy4s.example.OpenIntEnumTest
 import smithy4s.schema.Schema
 import smithy4s.schema.Schema._
 import weaver._
-import smithy4s.example.{OpenEnumTest, OpenIntEnumTest}
```
lewisjkl, line 36: Let's add test cases for the new primitives to this XmlCodecSpec too since the others are there already

smithy-build.json
```diff
@@ -4,7 +4,7 @@
   "maven": {
     "dependencies": [
       "software.amazon.smithy:smithy-protocol-test-traits:1.57.1",
-      "com.disneystreaming.alloy:alloy-core:0.3.26"
+      "com.disneystreaming.alloy:alloy-core:dev-SNAPSHOT"
```
lewisjkl: Let's put the latest version of alloy here now that it is published I believe

modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
```diff
@@ -18,17 +18,21 @@ package smithy4s
...
 
-import smithy4s.dynamic.model._
-
-import scala.collection.mutable.{Map => MMap}
-import smithy4s.schema.Schema._
-import smithy4s.internals.InputOutput
 import cats.Eval
 import cats.syntax.all._
-import smithy4s.schema.{Alt, EnumTag, EnumValue, Field}
+import smithy4s.dynamic.model._
```
lewisjkl, line 23: I think this `DynamicModelCompiler` will need to be updated to handle the new primitives similar to how it is handling UUID today. We'll want to be sure to add test cases around this too to make sure it is working properly (from a quick look I don't think we have any tests for UUID specifically here either so we can add that too). Anyway, lmk if you want to pair on this and I'll be happy to help

lewisjkl (commented):
This is awesome work! Just a few comments really and then I have a few ideas for what will be some good follow-up PRs to this one:

1. Let's enable scalafix checks in CI such that we can avoid the import diffs in the future (open for debate, lmk what you all think)
2. Update the [protocol definition](https://github.com/disneystreaming/alloy/blob/4824e9d4cd467e8a660b247f72a7ef37303f7c0a/modules/core/resources/META-INF/smithy/restjson.smithy#L9) in alloy so that these new format traits will be added to the hint mask for the SimpleRestJson protocol (to prevent them being filtered out) [see here](https://github.com/disneystreaming/smithy4s/blob/f0b1e647e077f411438034601e12d45f6ab41999/modules/http4s/src/smithy4s/http4s/internals/SimpleRestJsonCodecs.scala#L44)
3. It might be a good idea to add some [protocol tests to alloy](https://github.com/disneystreaming/alloy/tree/main/modules/protocol-tests) that capture the correct encoding formats for these new primitives. The protocol tests can serve as another way to sort of formalize the protocol and also we use them to generate tests in smithy4s that are more "end-to-end"

Just wanted to brain-dump that somewhere, but again let's not worry about it in this PR.

I'd like to get one more person to review this PR as well given how critical it is, maybe @yisraelU will volunteer as tribute?

yisraelU:
@lewisjkl I'll set aside time to review tomorrow

yisraelU:
Why are there so many 'ordering of import' changes ? it adds a lot of noise

a-morales:
> Why are there so many 'ordering of import' changes ? it adds a lot of noise

Sorry, apparently I have my editor config run a formatter on save :(. I'll implement @lewisjkl comment on enabling the CI check.

lewisjkl:
> > Why are there so many 'ordering of import' changes ? it adds a lot of noise
> 
> 
> 
> Sorry, apparently I have my editor config run a formatter on save :(. I'll implement @lewisjkl comment on enabling the CI check. 

Yeah this isn't the first time it has happened to one of us so I think it'd be good to just standardize on ScalaFix rules for imports and check in CI. Open to other opinions on this matter though

lewisjkl (approved):

yisraelU (approved):
would really like to see tests on those time-ops

PR #1791, opened 2025-07-29
Configure trigger generating Bijection for a newtype

Odomontois:
Since Bijection is a subtype of Function it works as an implicit conversion in scala 2, which confuses some users

- [ ] Added documentation
- [ ] Updated changelog


kubukoz:
Frankly I'd rather make Bijection just not be a subtype of Function. I see no reason why it should be usable as an implicit conversion

Odomontois:
> Frankly I'd rather make Bijection just not be a subtype of Function. I see no reason why it should be usable as an implicit conversion

Yes, but this would be too much of breaking change

kubukoz:
Dunno, it's the type of change I could see in 0.19 :)

Odomontois:
Ok, then.

Issue #1792, opened 2025-07-29
Default behavior of http clients changed affecting third party clients


yisraelU:
[This](https://github.com/disneystreaming/smithy4s/pull/1668) PR  introduced a change to how a specific group of characters are encoded client side, i.e. the colon character . This change is configured via a `boolean` field called `rawHttpLabelValues` and the default value is set to `false`. This means a change from existing behavior.
Additionally, an accompanying server side implementation for decoding was **not** provided, and therefore when utilizing a Smithy4s client and server together  with the default behavior will result in breakage when encountering one of those characters in the URI path

Per discussion with @kubukoz , suggested plan is.
- mark the recent release `0.18.40` as broken in the Scala Steward repo
-  subsequent pr  set default for http clients to be `true` , and aws clients to be false. Create release after merge to rectify release
- another pr adding server side decoding determined by the same parameter

kubukoz:
As we investigated with @yisraelU in the DMs, it appears that the http4s server actually always decodes `%3A` to a colon, so these paths are handled identically:

```
/foo:bar
/foo%3Abar
```

The decoding happens in https://github.com/disneystreaming/smithy4s/blob/2d5a17b4df45b04217d89f064c40351b1001306c/modules/http4s-kernel/src/smithy4s/http4s/kernel/package.scala#L75

so, for users of http4s servers, the client change might not change that much. Regardless, I believe the default behavior of the client should be restored, for the sake of systems not using http4s/smithy4s on the server.

yisraelU:
After discussion and a deeper dive , I now have a better understanding of the problem.
the encoding of the URI was shifted from an interpreter responsibility to a core smithy4s responsibility. This results in a doubly encoded path segment. Interpreters were expected to encode path segments until now are continuing to do so , while smithy4s is also providing encoding of the path

kubukoz:
As @yisraelU has found, the behavior with the "raw label encoding" flag set to true doesn't work either - it doesn't do ANY path encoding at all.

To sum up, 0.18.40 is totally broken and it's not just the defaults.

lewisjkl:
Can we consider this issue closed now?

yisraelU:
Yes I think so

PR #1793, opened 2025-07-29
Restore path encoding semantics

yisraelU:
I haven't started on the checklist below. 
This PR is a response to this issue https://github.com/disneystreaming/smithy4s/issues/1792

The goals are multifold
- rename `rawHttpLabelValues` to `!smithyPathEncoding` (for clarity: `smithyPathEncoding` means smithy4s does the smithy-compliant path encoding, whereas `!smithyPathEncoding` means we let http4s or other frameworks encode paths)
- set default of the smithyPathEncoding to false, which allows third party interpreters to maintain status quo
- set default of the Http4s interpreter to encode paths on its own, and provide implementations for both both true and false
- derive from the UnaryClientCodecs the `smithyPathEncoding` value for AWS clients as it is hardcoded per protocol

- [ ] Added documentation
- [x] Updated changelog

### DIFF
--- a/modules/core/src/smithy4s/http/HttpUnaryClientCodecs.scala
+++ b/modules/core/src/smithy4s/http/HttpUnaryClientCodecs.scala
@@ -17,16 +17,16 @@
 package smithy4s
 package http
 
-import smithy4s.capability.MonadThrowLike
 import smithy4s.client.UnaryClientCodecs
-import smithy4s.codecs.BlobDecoder
-import smithy4s.codecs.BlobEncoder
-import smithy4s.codecs.Decoder
-import smithy4s.codecs.PayloadError
+import smithy4s.codecs.{BlobEncoder, BlobDecoder}
+import smithy4s.codecs.{Decoder}
 import smithy4s.codecs.Writer
-import smithy4s.kinds.PolyFunction5
+import smithy4s.codecs.PayloadError
 import smithy4s.schema.CachedSchemaCompiler
 import smithy4s.schema.OperationSchema
+import smithy4s.capability.MonadThrowLike
+import smithy4s.kinds.PolyFunction5
+
 // scalafmt: { maxColumn = 120 }
 object HttpUnaryClientCodecs {
 
@@ -65,7 +65,7 @@ object HttpUnaryClientCodecs {
     def withResponseTransformation[Response0](f: Response0 => F[Response]): Builder[F, Request, Response0]
     def withHostPrefixInjection(enabled: Boolean): Builder[F, Request, Response]
     def withRawHttpLabelValues(enabled: Boolean): Builder[F, Request, Response]
-    def isRawHttpLabelValues: Boolean
+    def hasRawHttpLabelValues: Boolean
     def build(): UnaryClientCodecs.Make[F, Request, Response]
   }
 
@@ -119,7 +119,7 @@ object HttpUnaryClientCodecs {
     override def withRawHttpLabelValues(enabled: Boolean): Builder[F, Request, Response] =
       copy(rawHttpLabelValues = enabled)
 
-    def isRawHttpLabelValues: Boolean = rawHttpLabelValues
+    def hasRawHttpLabelValues: Boolean = rawHttpLabelValues
 
     def build(): UnaryClientCodecs.Make[F, Request, Response] = {
       val setBody: HttpRequest.Writer[Blob, Blob] = Writer.lift((req, blob) => req.copy(body = blob))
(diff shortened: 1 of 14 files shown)


modules/core/src/smithy4s/http/HttpUnaryClientCodecs.scala
```diff
@@ -118,6 +119,8 @@ object HttpUnaryClientCodecs {
     override def withRawHttpLabelValues(enabled: Boolean): Builder[F, Request, Response] =
       copy(rawHttpLabelValues = enabled)
 
+    def isRawHttpLabelValues: Boolean = rawHttpLabelValues
```
kubukoz: nitpick: `has` instead of `is`. Frankly I'd rename the whole concept to `urlEncodingHttpLabels` or something like that, but it's too late for now / should be done on 0.19

modules/http4s-kernel/src/smithy4s/http4s/kernel/package.scala
```diff
@@ -101,8 +101,12 @@ package object kernel {
       Smithy4sHttpResponse(res.status.code, headers, blob)
     }
 
-  def fromSmithy4sHttpUri(uri: Smithy4sHttpUri): Uri = {
-    val path = Uri.Path.Root.addSegments(uri.path.map(Uri.Path.Segment.encoded))
+  def fromSmithy4sHttpUri(uri: Smithy4sHttpUri, rawLabels: Boolean): Uri = {
+    val path = if (rawLabels) {
+      Uri.Path.Root.addSegments(uri.path.map(Uri.Path.Segment.encoded))
```
kubukoz: `rawLabels` == smithy4s doesn't encode them == http4s should encode them

so this conditional is inverted - in the case of `true` we should be calling `apply`. Otherwise, nobody encodes it

kubukoz: note that I keep getting confused with the booleans all the time due to "raw labels" being quite vague for me. Tests will prove if I got it right 😅 

but I'm judging by the pre-40 behavior, which was to use `apply`.

kubukoz: pushed a change that appears to fix the red aws-http4s tests by flipping this in c9f09bd716ad1d84ca30cfe73e15e8277cc59970

kubukoz: got lost in the force push 😓 now it's e062d98a8fe0ffd3b8eeb716439988da75031c54

yisraelU: the rawLabels name needs to go. everytime I need to remind myself what it means again

modules/http4s/src/smithy4s/http4s/SimpleRestJsonBuilder.scala
```diff
@@ -56,7 +56,7 @@ class SimpleRestJsonBuilder private (
           ),
         fieldFilter,
         hostPrefixInjection,
-        rawHttpLabelValues = false
+        rawHttpLabelValues = true
```
kubukoz: **thought:** it actually bothers me a lot that we have this boolean twice.

kubukoz: but I see no better way, so let's roll with it 👍

modules/http4s/test/src/smithy4s/http4s/SimpleRestJsonComplianceSuite.scala
```diff
@@ -95,7 +95,8 @@ object SimpleRestJsonComplianceSuite extends ProtocolComplianceSuite {
       val baseUri = uri"http://localhost"
       val suppliedHost =
         testHost.map(host => Uri.unsafeFromString(s"http://$host"))
-      SimpleRestJsonBuilder(service)
+      SimpleRestJsonBuilder
+        .withRawHttpLabelValues(false)(service)
```
lewisjkl: I was under the impression that this would default to true to retain the old behavior, but perhaps I'm turned around here 😅

yisraelU: this needs to be false to pass the restJson1 compliance tests 
This provides additional background https://github.com/disneystreaming/smithy4s/issues/1663

lewisjkl: Oh my bad, I missed this was in tests.

lewisjkl (approved):

kubukoz (approved):

PR #1794, opened 2025-07-30
Remove Bijection extends Function

Odomontois:
Which helps avoid accidental implicit conversion in scala 2

- [ ] Added documentation
- [X] Updated changelog

### DIFF
--- a/modules/core/src/smithy4s/Bijection.scala
+++ b/modules/core/src/smithy4s/Bijection.scala
@@ -16,7 +16,7 @@
 
 package smithy4s
 
-import scala.Predef.{identity => identity0, _}
+import scala.Predef.{identity => identity0}
 
 /**
   * A bijection is an association of two opposite functions A => B and B => A.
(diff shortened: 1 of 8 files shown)


modules/core/src/smithy4s/Bijection.scala
```diff
@@ -25,14 +25,16 @@ import scala.Predef.{identity => identity0, _}
...
-trait Bijection[A, B] extends Function[A, B] { outer =>
+trait Bijection[A, B] { outer =>
   def to(a: A): B
   def from(b: B): A
 
   final def apply(a: A): B = to(a)
 
   def swap: Bijection[B, A] = Bijection(from, to)
 
+  def toFunction: A => B = apply
```
kubukoz, line 36: nitpick: I'd get rid of this and just call `apply` or `to` when needed. Maybe even remove `apply` altogether...

Odomontois: Well, it's just nano-optimisation, since "Impl" already has a private `toFunction` field, we can avoid allocating lambdas in some cases.

kubukoz: alright, I'm convinced about `toFunction` then :P 

but `apply`, idk. It's not a major decision anyway, let's just leave it be.

kubukoz (approved):

kubukoz:
Please add a changelog entry and let's roll 👍

kubukoz:
Thanks!

ghostbuster91:
should the same be applied to `Surjection`?

PR #1795, opened 2025-08-01
add workflow to close stale issues

lewisjkl:
Raising this PR to be able to discuss the possibility of adding a workflow like this, I noticed we have some old PRs and I think in general it is nice to focus on PRs/Issues that have active interest/contributions.

Let me know your thoughts.

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,26 @@
+name: Close Inactive PRs/Issues
+on:
+  schedule:
+    - cron: "30 1 * * *" # At 01:30
+  workflow_dispatch:
+
+jobs:
+  close-issues:
+    timeout-minutes: 15
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v9
+        with:
+          debug-only: true # set to true if you want to test things out without any real impact to issues/PRs
+          ascending: true # start with oldest items
+          operations-per-run: 500
+          days-before-stale: 45
+          days-before-close: 7
+          stale-pr-label: "stale"
+          stale-issue-label: "stale"
+          stale-pr-message: "This PR is being marked as stale since it has had no activity for 45 days. Please comment below if you wish to prevent it from closing in 7 days."
+          close-pr-message: "This PR is being closed due to having no activity for 45 days plus no response from the previous comment. You can reopen the PR if you wish to continue work on it."
+          stale-issue-message: "This issue is being marked as stale since it has had no activity for 45 days. Please comment below if you wish to prevent it from closing in 7 days."
+          close-issue-message: "This issue is being closed due to having no activity for 45 days plus no response from the previous comment. You can reopen the issue if you wish to discuss further."


kubukoz:
Is there context for "why now"? In the meantime, here's my take:

I generally disagree with this idea. Issues being stale doesn't mean they're no longer there, it's just that whoever encountered them has found a decent workaround... or has given up.

Sometimes I open issues even when I'm not hitting them in real life, to serve as an acknowledgement of there being an issue, and maybe a good reproduction / hints for a solution. Prefer known issues to unknown/forgotten ones :) 

Having a lot of open issues doesn't necessarily make a project look buggy - in my opinion, it can show that enough people use it, and care, that they're willing to spend the time crafting a ticket.

As for tickets we legitimately don't care about, we can always close them ourselves due to being out of scope (or whatever reason makes sense in the context).

---

Happy to continue the conversation, maybe we can run it as a one-time thing, or schedule some time to review the older ones

lewisjkl:
> why now

Really just because I saw that we have PRs which are 2+ years old and I think in general that just adds clutter as there is virtually no hope for it being "merged" at this point and would need to be redone anyway with how much the codebase evolves over that time span. Plus if it is left unattended (no comments, no commits, etc) for 2+ years, are we ever going to get to it? Do we really need to?

> Issues being stale doesn't mean they're no longer there

With issues, I see your point more-so than with PRs. That being said, looking at the old issues (over 3 years old in some cases), I again question their relevance in 2025 and/or their need to ever be done if we've gone 3 years without worrying about it.

To me, that's why the workflow adds a comment saying "this will be closed in 7 days". That's an opportunity for us or whoever raised the issue to take a look at it and check, "is this still relevant? do I still care?" If no, then we do nothing and it is closed. If yes, we just add a comment (hopefully with some explanation as to why this is still relevant and yet has still not been addressed) to the Issue and now it won't be closed as stale. So all that to say, I actually think it is a way to get relevant issues floating to the top, and things that aren't relevant to get out of the way.

Worth noting, the timeframes are arbitrary. I'm fine to change the timeframes to whatever.. e.g. 1 year old receives a comment saying the issue will be closed in 30 days.

All of that being said, I'm not going to die on this hill.. so ultimately I will defer to the other maintainers, including yourself, to vote on / decide.

---

_(or we could just leave this PR open for the next 3 years and it can join the ranks <- me being a troll)_

kubukoz:
(I'll probably have more thoughts tomorrow / another day, it's late)

> 1 year old receives a comment saying the issue will be closed in 30 days.

that doesn't sound so bad. Also, I'd just go ahead and close the multi-year PRs straight away.

lewisjkl:
Yeah no worries, it is definitely not urgent at all and I appreciate the discussion in any case 🙏

yisraelU:
I agree with the automated workflow for PR's . It provides a reminder for the author and allows the author to respond and say "wait , I'm still working on it".
Issues on the other hand, we really need to set aside time as maintainers monthly, to go through any new issues , and analyze/prioritize them. We can use labels to convey a bit more information (and serve as a reminder for ourselves) , such as `not priority` ,`not affecting users` , `good idea` etc.. 
This would server to convey more info to the public at large, that 
1. these issues were read and thought about
2. here is what you can expect

kubukoz:
Some kind of triaging setup definitely sounds reasonable 👍 

e.g. an automation that takes new issues and tags them with `needs-triage` and reminds maintainers if there are any untriaged issues older than N days.

lewisjkl:
Yeah I agree with the above, I will look into this more when I get a chance to

PR #1806, opened 2025-08-11
Add hints to union members that are UnitCase

### LINKED ISSUE #1804: Hints are not rendered on Union members when they are targeting Unit
In the case of  a standard Union shape such as below, when a member targets the Unit shape, hints are not added to it.
In this particular case , notice the documentation hint is not added to the schema the represents the unit member.

Smithy:

```smithy
$version: "2"

namespace example

union Test {
    @jsonName("text")
    t:Text
    @documentation("no_text")
    noText: Unit
}

structure Text {
    @required
    body: String

}

```

Generated Code :

```scala

sealed trait Test extends scala.Product with scala.Serializable { self =>
  @inline final def widen: Test = this
  def $ordinal: Int

  object project {
    def t: Option[Text] = Test.TCase.alt.project.lift(self).map(_.t)
    def noText: Option[Test.NoTextCase.type] = NoTextCaseAlt.project.lift(self)
  }

  def accept[A](visitor: Test.Visitor[A]): A = this match {
    case value: Test.TCase => visitor.t(value.t)
    case value: Test.NoTextCase.type => visitor.noText(value)
  }
}
object Test extends ShapeTag.Companion[Test] {

  def t(t: Text): Test = TCase(t)
  /** no_text */
  def noText(): Test = Test.NoTextCase

  val id: ShapeId = ShapeId("example", "Test")

  final case class TCase(t: Text) extends Test { final def $ordinal: Int = 0 }
  /** no_text */
  case object NoTextCase extends Test { final def $ordinal: Int = 1 }
  private val NoTextCaseAlt = Schema.constant(Test.NoTextCase).oneOf[Test]("noText").addHints(hints)

  object TCase {
    val hints: Hints = Hints(
      smithy.api.JsonName("text"),
    ).lazily
    val schema: Schema[Test.TCase] = bijection(Text.schema.addHints(hints), Test.TCase(_), _.t)
    val alt = schema.oneOf[Test]("t")
  }

  trait Visitor[A] {
    def t(value: Text): A
    def noText(value: Test.NoTextCase.type): A
  }

  object Visitor {
      def default: A
      def t(value: Text): A = default
      def noText(value: Test.NoTextCase.type): A = default
    }
  }

  implicit val schema: Schema[Test] = union(
    Test.TCase.alt,
    NoTextCaseAlt,
  ){
    _.$ordinal
  }.withId(id).addHints(hints)
}
```

yisraelU:
PR to resolve #1804 
- adds an additional val in the Union companion object to hold the hints of the specific member and adds them to the `addHints` method call
- the adtmember smithy spec was used to test , as it already contained the use case of a union with a member targeting Unit.


yisraelU:
closing because i messed up the history

PR #1808, opened 2025-08-11
fix for hints in union->unit case

### LINKED ISSUE #1804: Hints are not rendered on Union members when they are targeting Unit

yisraelU:
closes #1804 
- adds an additional val in the Union companion object to hold the hints of the specific member and adds them to the `addHints` method call
- the adtmember smithy spec was used to test , as it already contained the use case of a union with a member targeting Unit. 
    -  This created the need to fix the AdtSmokeSpec

### DIFF
--- a/modules/bootstrapped/src/generated/smithy4s/example/OrderType.scala
+++ b/modules/bootstrapped/src/generated/smithy4s/example/OrderType.scala
@@ -66,7 +66,10 @@ object OrderType extends ShapeTag.Companion[OrderType] {
     val alt = schema.oneOf[OrderType]("inStore")
   }
   case object PreviewCase extends OrderType { final def $ordinal: Int = 2 }
-  private val PreviewCaseAlt = Schema.constant(OrderType.PreviewCase).oneOf[OrderType]("preview").addHints(hints)
+  val PreviewCaseHints: Hints = Hints(
+    smithy.api.JsonName("PREVIEW"),
+  ).lazily
+  private val PreviewCaseAlt = Schema.constant(OrderType.PreviewCase).oneOf[OrderType]("preview").addHints(PreviewCaseHints)
 
   object OnlineCase {
     val hints: Hints = Hints.empty
--- a/modules/bootstrapped/src/generated/smithy4s/example/OrderType.scala
+++ b/modules/bootstrapped/src/generated/smithy4s/example/OrderType.scala
@@ -65,11 +65,13 @@ object OrderType extends ShapeTag.Companion[OrderType] {
 
     val alt = schema.oneOf[OrderType]("inStore")
   }
-  case object PreviewCase extends OrderType { final def $ordinal: Int = 2 }
-  val PreviewCaseHints: Hints = Hints(
-    smithy.api.JsonName("PREVIEW"),
-  ).lazily
-  private val PreviewCaseAlt = Schema.constant(OrderType.PreviewCase).oneOf[OrderType]("preview").addHints(PreviewCaseHints)
+  case object PreviewCase extends OrderType {
+    final def $ordinal: Int = 2
+    val PreviewCaseHints: Hints = Hints(
+      smithy.api.JsonName("PREVIEW"),
+    ).lazily
+  }
+  private val PreviewCaseAlt = Schema.constant(OrderType.PreviewCase).oneOf[OrderType]("preview").addHints(PreviewCase.PreviewCaseHints)
 
   object OnlineCase {
     val hints: Hints = Hints.empty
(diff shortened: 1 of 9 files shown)


lewisjkl (approved):

lewisjkl:
I think there are some super edge cases where we could have naming collisions with the `val myAltHints` that we are creating now, but the same is true of the `Alt` suffix we are adding in the same spot, so I think probably not worth worrying about too much right now

yisraelU:
@lewisjkl perhaps the val , should be scoped inside the `case object ` ?

lewisjkl:
> @lewisjkl perhaps the val , should be scoped inside the `case object ` ?

Yeah that's a good idea actually, better safe than sorry even if the edge case is pretty unlikely

modules/bootstrapped/src/generated/smithy4s/example/OrderType.scala
```diff
@@ -65,8 +65,13 @@ object OrderType extends ShapeTag.Companion[OrderType] {
 
     val alt = schema.oneOf[OrderType]("inStore")
   }
-  case object PreviewCase extends OrderType { final def $ordinal: Int = 2 }
-  private val PreviewCaseAlt = Schema.constant(OrderType.PreviewCase).oneOf[OrderType]("preview").addHints(hints)
+  case object PreviewCase extends OrderType {
+    final def $ordinal: Int = 2
+    val PreviewCaseHints: Hints = Hints(
```
kubukoz: **suggestion:** rename to just hints? and let's move the alts inside as well? (with the name `alt` like the non-case-object cases)

lewisjkl: Yeah I agree, we can move the alts in as well since they are private it shouldn't break anything out in the wild

lewisjkl (approved):

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -1713,8 +1714,8 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     line"val tag: $EnumTag_[$parentType] = $EnumTag_.$tagStr"
   }
 
-  def renderHintsVal(hints: List[Hint]): Lines = {
-    val lhs = line"val hints: $Hints_"
+  def renderHintsVal(hints: List[Hint], prefix: Option[String] = None): Lines = {
```
kubukoz, line 1717: **nitpick:** I don't think `prefix` actually gets passed now

yisraelU: yeah , its not needed anymore , figured it still might be useful in the future..
I'll remove in future pr

PR #1810, opened 2025-08-13
[0.19] Render fields marked nullable as smithy4s.Nullable rather than Option

stewSquared:

- [x] ~Added build-plugins integration tests (when reflection loading is required at codegen-time)~
- [x] ~Added alloy compliance tests (when simpleRestJson protocol behaviour is expanded/updated)~
- [x] Added documentation
- [x] Updated changelog

### DIFF
--- a/modules/codegen/test/src/smithy4s/codegen/internals/RendererSpec.scala
+++ b/modules/codegen/test/src/smithy4s/codegen/internals/RendererSpec.scala
@@ -737,4 +737,44 @@ final class RendererSpec extends munit.ScalaCheckSuite {
       s"$definition does not contain Nullable[String]"
     )
   }
+
+  test("sparse collection types should be rendered as Nullable") {
+    val smithy =
+      """
+        |$version: "2.0"
+        |
+        |namespace smithy4s
+        |
+        |@sparse
+        |list SparseStringList {
+        |  member: String
+        |}
+        |
+        |@sparse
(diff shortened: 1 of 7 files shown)


modules/codegen/test/src/smithy4s/codegen/internals/RendererSpec.scala
```diff
@@ -711,4 +711,69 @@ final class RendererSpec extends munit.ScalaCheckSuite {
...
+        |  @required
+        |  nullableString: String
+        |}
+        |""".stripMargin
+
+    val contents = generateScalaCode(smithy).values
+    val definition = contents.find(_.contains("final case class NullableExample")).get
+
+    assert(
+      definition.contains("nullableString: Nullable[String]"),
```
stewSquared, line 736: So, this test already passes before my changes to Renderer.scala and ToLine.scala. I don't understand this yet, because I don't know where else they would found and rendered as Nullable.

The other test is for sparse list and map types, which can have nullable elements. Those were previously rendered as Option.

kubukoz:
suggestion: let's do it in 0.18 with a `@smithy4s.meta#asNullable` trait? And we can make it the default in 0.19.

Could even have a metadata setting like `smithy4sRenderDynamicHintBindings` from #1816 to drive this more globally without breaking backwards compatibility.

kubukoz:
Superseded by #1895

PR #1813, opened 2025-08-20
dynamic hint rendering

lewisjkl:
Changes from using static hint bindings in generated code to use dynamic ones instead. 

This behavior is configurable, but the default will be to render with dynamic bindings.

More rationale is added in the documentation added in this PR.

Todos:

- [ ] Update so configuration can be passed in via SBT plugin and CLI
- [ ] Update so smithy4s-core generated code uses static bindings ?
- [ ] Target 0.18 instead of 0.19 (assuming it defaults to static bindings) ?

- [ ] Added documentation
- [ ] Updated changelog


lewisjkl:
FYI I am working on a new PR here that will target 0.18 and default to static rendering mode

lewisjkl:
Closing in favor of #1816

Issue #1591, opened 2024-09-25
Feature request: Flag to Mark Generated Code as Package Private


bpholt:
Motivation: allowing libraries to generate smithy4s clients intended for use only within the library, avoiding binary compatibility issues caused by differences in the generated code over time.

Such a feature would remove the need for the Scalafix referenced in #1589.

Baccata:
I'm not necessarily opposed to it but you'd have to propose semantics (in this issue) and implement them yourself once @kubukoz  and I have agreed to it.

kubukoz:
I finally have a usecase so I want it 😅 

We usually configure this kind of thing via traits or metadata. I would use [Validated Newtypes](https://disneystreaming.github.io/smithy4s/docs/codegen/customisation/validated-newtypes) as inspiration, and implement both:

- apply the trait explicitly if you want fine-grained control and need only a small number of private shapes
- set metadata if you want a large blast radius.

I wouldn't apply the metadata globally, though.

```smithy
metadata smithy4sExtraModifiers = [
  {
    "selector": "[id|namespace=foo.bar] ~>"
    "modifiers": "private[foo]" // allow anything here, if it's broken it won't compile, and we'll futureproof ourselves by letting people add `open` or whatever they desire from day 1
  }
]
```

Selectors are flexible enough that you can choose any number of shapes, filter them by transitive relationships and so on.

---

Implementation-wise, I would have a transformation that adds the trait based on the metadata's selectors, then the codegen can just look for the trait (that's also pretty much [how the validated newtypes work](https://github.com/disneystreaming/smithy4s/blob/7cd62db35ac5911fbb455945154c9a178b6fd76a/modules/codegen/src/smithy4s/codegen/transformers/ValidatedNewtypesTransformer.scala)).

The big question I have is: do we want to retain that information in the model, for further codegen runs? Validated newtypes are retained in [CodegenRecord](https://github.com/disneystreaming/smithy4s/blob/7cd62db35ac5911fbb455945154c9a178b6fd76a/modules/codegen/src/smithy4s/codegen/CodegenRecord.scala#L28C5-L28C22), but that's because it was necessary to know what to generate on the use-site - we probably don't need that here.

kubukoz:
TBH I'd like selector-based metadata application in validated newtypes too. Maybe this could be the general pattern for configuring such things.

kubukoz:
[...]

PR #1816, opened 2025-08-21
Dynamic Hint Rendering 0.18

### LINKED ISSUE #1814: Render with Dynamic Hint Bindings
This may end up being included in 0.18 as opt-in behavior, but I'd like to include it in 0.19 as opt-out behavior. Meaning in 0.19 I want to change the default behavior to be rendering dynamic hint bindings rather than static ones. The primary motivation for this is that I have seen numerous occasions where changes in `smithy.api` or `alloy` traits leads to a situation where one must update smithy4s (regenerate the code in the core module so that things are up to date) in order to get things compiling again. This behavior is only seen when using smithy artifacts (published JARs containing smithy files) which depend on a different version of `smithy.api` or `alloy` than the core module of smithy4s does. These issues are very opaque and difficult for users to figure out without raising an issue.

Note that Smithy4s core generated code (alloy and smithy.api namespaces) do not need to use dynamic hint bindings, they can be static. The issue arises from the code generated in user-land having static hint bindings that are targeting code that was generated in smithy4s core.

---

Relates to #1813.

lewisjkl:

- [x] Added documentation
- [x] Updated changelog

relates to #1814


a-morales (approved):
Pretty nice solution using the smithy metadata to set the flags for rendering the code.

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -150,6 +150,10 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
         "Generated smithy files"
       )
 
+    val smithy4sRenderDynamicHintBindings = taskKey[Boolean](
```
kubukoz: **issue:** I don't think we should be going this way.

Historically, we've had some metadata-based configuration, e.g. `smithy4sRenderOptics`, which could be set in the build tool plugins. However, setting metadata directly in Smithy is very easy, so we've been phasing some of these out (e.g. `smithy4sRenderOptics` is actually gone from build tools in 0.19).

This is also something that wasn't done consistently, e.g. `smithy4sRenderOptics` has a build tool flag but `smithy4sErrorsAsScala3Unions` does not. I would suggest we consider this pattern deprecated and don't add any such keys even in 0.18.

modules/codegen-plugin/src/sbt-test/codegen-plugin/dynamic-hints/build.sbt
```diff
@@ -0,0 +1,21 @@
+import smithy4s.codegen.Smithy4sCodegenPlugin.autoImport.smithy4sOutputDir
+lazy val root = (project in file("."))
+  .enablePlugins(Smithy4sCodegenPlugin)
+  .settings(
+    scalaVersion := "2.13.16",
+    Compile / smithy4sRenderDynamicHintBindings := true,
+    libraryDependencies ++= Seq(
+      "com.disneystreaming.smithy4s" %% "smithy4s-core" % smithy4sVersion.value
+    ),
+    TaskKey[Unit]("check") := {
```
kubukoz: **nitpick:** I'd test this without invoking the build tools - just a test in `RendererSpec` should do fine, I think (this assumes we remove the sbt setting for this new flag)

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ When adding entries, please treat them as if they could end up in a release any
 
 Thank you!
 
+# 0.18.43
+
+* Add support for dynamic hint bindings in [#1816](https://github.com/disneystreaming/smithy4s/pull/1816)
```
kubukoz: **suggestion:** let's clarify


```suggestion
* Add support for generating dynamic hint bindings in [#1816](https://github.com/disneystreaming/smithy4s/pull/1816)
```

modules/codegen/test/src/smithy4s/codegen/internals/DynamicHintRenderingSpec.scala
```diff
@@ -0,0 +1,388 @@
...
+                        | @httpHeader("x-one")
+                        | one: String
+                        |
+                        | @timestampFormat("date-time")
+                        | two: Timestamp
+                        |}
+                        |
+                        |""".stripMargin
+
+    val scalaCode =
```
kubukoz: **suggestion:** let's avoid testing entire files here. Such tests are crazy hard to maintain if we don't use something like [`assertInlineSnapshot` from snapshot4s](https://siriusxm.github.io/snapshot4s/contributing/how-it-works#inline-snapshots).

In `RendererSpec` we've been doing tests based on substrings, I think that's an OK way to do it :) at least it prevents unrelated changes (e.g. those pertaining to imports)  from breaking unrelated tests.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -1717,15 +1731,30 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
+
+    rendered match {
       case Nil => lines(line"$lhs = $Hints_.empty")
       case args =>
-        line"$lhs = $Hints_".args(args).appendToLast(".lazily")
+        line"$lhs = $Hints_".args(args).appendToLast(getLazySuffix(hints))
     }
   }
 
+  // If all hints are any native bindings (static bindings)
```
kubukoz: **suggestion:** 
```suggestion
  // If there are any native hints (static bindings) to be generated,
```

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -123,6 +123,15 @@ private[codegen] class SmithyToIR(
       .flatMap(f => DefaultRenderMode.fromString(f.getValue))
       .getOrElse(DefaultRenderMode.Full)
 
+  private val renderDynamicHintBindings =
```
kubukoz: **suggestion:** let's make it controllable in a more fine-grained fashion, i.e. with a meta trait.

The reasoning is: metadata is global and cannot be overridden. If you publish a library with generated code, it may propagate to downstream codegen runs.

N.B. the Java trait implementations are now generated by default (i.e. if you don't add `@tags(["nocodegen"])`). This should alleviate some of the pains of creating a trait.

N.B.2 Such a trait could be a `structure {}`, that way we keep the option of making it more configurable in the future (e.g. `@dynamicHintBindings(except: [documentation])`)

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
kubukoz: Wrote down some other reasons to not do this: https://github.com/disneystreaming/smithy4s/issues/1537#issuecomment-3220087228

kubukoz (changes requested):

kubukoz:
I'm starting to lose track of all the metadata keys we have! If you don't mind, can we have some sort of package-private `Smithy4sMetadata` object that lists them all?

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
lewisjkl: I agree that metadata is less-ideal since it can't be overridden. What about a metadata entry that is basically an "allow list" for things that will be referenced using dynamic bindings? The default would be an empty list, but then we will have the flexibility to, internally, say that `alloy` and `smithy.api` (and some other internal namespaces) should only be referenced using dynamic hint bindings.

I don't love the trait-based approach mainly because it isn't something that I ever think needs to be applied on a shape-by-shape basis. I think the most granular you'd ever want to be is namespace-level. Or at least that is true of our internal use cases.

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
lewisjkl: Yeah tbh I didn't realize that we'd shifted away from this pattern, so I was trying to make things more uniform 😅 in any case, I'm happy to remove the build-plugin configurability

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
kubukoz: > because it isn't something that I ever think needs to be applied on a shape-by-shape basis

probably true

> the most granular you'd ever want to be is namespace-level

fair

---

I like an allow-list metadata more: you can even append to it later without conflicts (Smithy will just merge into a list). I would, however, I'd consider a yet another option - having the trait on individual shapes with the ability to apply it automatically on a larger scale, e.g.

https://github.com/disneystreaming/smithy4s/issues/1591#issuecomment-2910737483

this would let us gradually move towards "the only way to tweak codegen is via traits" with a generic solution for automating the repetitive work via selectors.

kubukoz: an additional axis of the problem is whether the trait should be applied on _the traits_ or _the shapes_. Or maybe we should support both...

lewisjkl: Yeah I suppose having some form of transformation out of the box that makes it easy to add the trait to all shapes in certain namespaces is another way that would accomplish the same thing.

> an additional axis of the problem is whether the trait should be applied on the traits or the shapes. Or maybe we should support both...

Yeah that is a tricky one, I _think_ it should be on the shapes rather than the traits for mainly one reason:

```smithy
@dynamicHintBindings
@since("1")
structure Test {
  @jsonName("aaa")
  a: String
}
```

This would mean that when `Test` is rendered, `since` and `jsonName` would be rendered with dynamic bindings. However, there is some ambiguity with:

```smithy
@dynamicHintBindings
@since("1")
@trait()
structure testTrait {
  @jsonName("aaa")
  a: String
}
```

Does this mean that when `testTrait` itself is rendered, it will use dynamic hint bindings, or does it mean when something uses `testTrait` as a hint then it will render it as a dynamic binding? Or both? I think it is more clear if the trait always means that the shape it is placed ON will use dynamic bindings to reference whatever hints it contains.

Idk if that makes sense, it is kind of tricky to explain it clearly as it is pretty meta 😅

CHANGELOG.md
yisraelU: I would add that this removes the static bindings too

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
kubukoz: re: the example you provided, I think the trait must have unambiguous behavior, so it'll either _always be applied on traits_ or it'll _always be applied on usages_. In other words we could have either:

- `@renderAsDynamic` (which must be placed on a `@trait` definition and makes its downstream usages rendered dynamically) or 
- `@hasDynamicBindings` (placed anywhere, makes the shape's upstream traits rendered dynamically)

dummy names, you get the idea. Probably don't need both.

If "all traits from core/alloy should be rendered dynamic" is the intention, I think `@renderAsDynamic` is the way to go:

```smithy
@trait(selector: "[trait|trait]")
structure renderAsDynamic {}
```

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -1717,15 +1731,30 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
 
+  // If there are any native hints (static bindings) to be generated,
+  // we need to use the lazily suffix to prevent compile time
+  // issues
+  private def getLazySuffix(hints: List[Hint]): String = {
+    if (
+      hints.forall {
+        case _: Hint.Native => false
+        case _              => true
+      }
```
kubukoz: **nitpick:** I'd flip it and write it as `if(hints.exists(isNative)) then ".lazily" else ""`, kind of shows the intention better and in line with the comment above

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -1692,6 +1692,14 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
       .run(true)
       ._2
 
+  private def renderHint(hint: Hint.DynamicBinding): Line = {
+    val ns = hint.shapeId.getNamespace
+    val name = hint.shapeId.getName
+    val sid = line"$ShapeId_(${renderStringLiteral(ns)}, ${renderStringLiteral(name)})"
+    val doc = renderNodeToLine(hint.data)
+    line"Hints.dynamic($sid, $doc)"
```
kubukoz, line 1700: **thought:** we do have an implicit conversion from tuples so this could be just `$sid -> $doc`, but I think `Hints.dynamic` may actually be better - shows the intention more clearly + fewer implicit conversions may be good for compile time.

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -1704,6 +1712,12 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     line"""val id: $ShapeId_ = $ShapeId_(${renderStringLiteral(ns)}, ${renderStringLiteral(name)})"""
   }
 
+  private def getRenderedHints(hints: List[Hint]): List[Line] =
+    hints.collect { case nt: Hint.Native => nt }.sortBy(_.shapeId).map(renderHint) ++ hints
```
kubukoz: **nitpick:** I would prefer to sort them together, i.e. not separate by static/dynamic. Not a strong opinion, but having a common ordering might make it easier to scan through a long list of mixed bindings

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -1283,6 +1283,18 @@ private[codegen] class SmithyToIR(
     )
   }
 
+  private def unfoldTraitNonConstraint(tr: Trait): Hint = {
```
kubukoz, line 1288: **thought:** might be worth a small comment here why non-constraint only:

```suggestion
  // We can only allow dynamic bindings for non-constraint traits, because constraints rely on types (static bindings) to find their refinement providers
  private def unfoldTraitNonConstraint(tr: Trait): Hint = {
```

modules/core/src/smithy4s/Hints.scala
```diff
@@ -255,10 +258,34 @@ object Hints {
         extends Binding {
       override def keyId: ShapeId = key.id
       override def toString: String = value.toString()
+
+      override def equals(obj: Any): Boolean = obj match {
```
kubukoz, line 262: **question:** why do we need statics to be equal to dynamics?

modules/core/src/smithy4s/Hints.scala
```diff
@@ -255,10 +258,34 @@ object Hints {
...
+        case sb: StaticBinding[_] =>
+          sb.key == this.key && sb.value == this.value
+        case db: DynamicBinding =>
+          this.toDynamicBinding == db
+        case _ => false
+      }
+
+      lazy val toDynamicBinding: DynamicBinding =
```
kubukoz: **nitpick:** I'd probably hide this for now (it's a public class)

kubukoz (approved):
I'd add something in `sampleSpecs` so that it generates a file in `bootstrapped/main` that we can check in (maybe just one with mixed bindings?) and we can roll 👍

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
lewisjkl: Yeah I was concerned about using the tuple + varargs constructor also because it expects String shapeIds. We could add another constructor that uses smithy4s.ShapeId, but I think it does get more messy, especially since you can have a mixture of dynamic and static bindings in the same place

modules/core/src/smithy4s/Hints.scala
lewisjkl: Initially I added it because tests were failing without it (when I was generating all dynamic bindings in bootstrapped). But then I thought since from the perspective of SchemaVisitors and in user-land, we want to more-or-less abstract the difference between dynamic and static hints, it makes sense. Like if we treat them the exact same way (e.g. they have the same exact meaning/impact on behavior) should they then be considered equal? I think yes, but ultimately I'm not entirely sure it matters...

kubukoz: they're equivalent, sure, but are they equal... I'm not sure! It'd be great If you can find out which tests were failing, then we can figure out why

lewisjkl: They were failing just because it was basically a test saying we are expecting A, B, C static hints, but the system under test was returning dynamic versions of A, B, and C. So yeah I guess there could be cases in schema visitors where people do equality checks on hints and it'd be surprising if they were not equal when they are equivalent, I think at least

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
lewisjkl: Made it so they are sorted together now, used an Either to track the full type (Hint.Native, Hint.DynamicBinding) since it makes sorting after the fact simpler and also makes calling their respective rendering methods simpler. The alternative would be some kind of method that matched on `Hint` (the supertype) and then delegated from there, I think that'd be kinda messy though. Open to other ideas as well, my brain is not functioning at 100% this late in the day to be honest 😅

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
lewisjkl: Rewrote using collectFirst, lmk if you like that better.

modules/core/src/smithy4s/Hints.scala
kubukoz: I'd defer to @Baccata, but from my POV I think it's ok to make them equal

kubukoz (approved):

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
kubukoz: I didn't see these comments! This looks fine :)

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
kubukoz: yep looks great

Issue #1331, opened 2023-12-18
0.19: Improvements to schema tags


kubukoz:
- `CollectionTag` should be unsealed: #1819 
- `Map` schemas should have a tag: #1818
- `Optional` schemas should have a tag #1826
- `Document` should be a tag? https://github.com/disneystreaming/smithy4s/issues/1659#issuecomment-2724935204

lewisjkl:
Should we consider this issue closed? I'm not sure I see enough value in DocumentTag for 0.19

kubukoz:
Yeah. Document would still be amazing but I don't have a pressing usecase so no issues with deferring until later/never

PR #1818, opened 2025-08-22
Add MapTag for MapSchema

a-morales:

This adds a `MapTag[_]` that is analogous to `CollectionTag[_]`. The only implementation within `Smithy4s` is `MapTag[Map]`, but this will allow for other tools to efficiently encode/decode Map like objects. 

Still need to make sure that existing tests pass, but this is good to at least get some feedback on

- [ ] Added documentation
- [ ] Updated changelog


modules/compliance-tests/src/smithy4s/compliancetests/internals/eq/EqSchemaVisitor.scala
```diff
@@ -51,14 +51,18 @@ object EqSchemaVisitor extends SchemaVisitor[Eq] { self =>
...
+      tag: MapTag[C],
       key: Schema[K],
       value: Schema[V]
-  ): Eq[Map[K, V]] = {
+  ): Eq[C[K, V]] = {
     implicit val valueEq: Eq[V] = self(value)
-    Eq[Map[K, V]]
+    tag match {
+      case MapTag.MapTag => Eq[Map[K, V]]
+      case _ => ???
```
kubukoz: **thought:** can probably be implemented with `iterator` without many issues, right?

modules/core/src/smithy4s/schema/MapTag.scala
```diff
@@ -0,0 +1,55 @@
...
+  def build[K, V](preserveOrder: Boolean)(put: (((K, V)) => Unit) => Unit): C[K, V]
+
+  def fromIterator[K, V](it: Iterator[(K, V)]): C[K, V] = build(put => it.foreach(put(_)))
+
+  def isEmpty[K, V](c: C[K, V]): Boolean
+  def empty[K, V]: C[K, V] = build(_ => ())
+}
+
+object MapTag {
+  case object MapTag extends MapTag[Map] {
```
kubukoz: **nitpick:** rename to `ScalaMapTag` or the like

modules/core/src/smithy4s/schema/MapTag.scala
```diff
@@ -0,0 +1,55 @@
...
+import scala.collection.immutable.ListMap
+
+trait MapTag[C[_, _]] {
+  def name: String
+
+  def iterator[K, V](c: C[K, V]): Iterator[(K, V)]
+  def build[K, V](put: (((K, V)) => Unit) => Unit): C[K, V]
+  def build[K, V](preserveOrder: Boolean)(put: (((K, V)) => Unit) => Unit): C[K, V]
+
+  def fromIterator[K, V](it: Iterator[(K, V)]): C[K, V] = build(put => it.foreach(put(_)))
```
kubukoz: **suggestion:** can we have a `fromScalaMap`? might be a less costly way to wrap / reuse something that already comes as a map, in some usages of MapTag. It can be implemented by default in a way similar to this.

kubukoz (commented):
keep going 👍

kubukoz:
ref: #1331

modules/core/src/smithy4s/schema/MapTag.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import scala.collection.immutable.ListMap
+
+trait MapTag[C[_, _]] {
+  def name: String
+
+  def iterator[K, V](c: C[K, V]): Iterator[(K, V)]
+  def toScalaMap[K, V](c: C[K, V]): Map[K, V] = iterator(c).toMap
+  def build[K, V](put: (((K, V)) => Unit) => Unit): C[K, V]
+  def build[K, V](preserveOrder: Boolean)(
```
yisraelU: not sure if preserve order belongs here as part of the `MapTag` typeclass. I don't think it has much to do with the capabilities that a Map interface is providing. I would have a `ScalaListMapTag` or whatever we want to call it as a different implementation entirely

a-morales: So I played around with removing the `preserveOrder` argument and going with a `ScalaListMapTag`, but the problem is that since the MapTag is paramatized on `C` it's hard to change that tag to an ordered one in the code. Seems that we would have to add a `@preserveOrder` smithy trait to denote that a Map should preserve the order of keys. 

We would have to update `SchemaVisitorJCodec` to not take in the `preserveMapOrder` boolean as an argument and instead rely on hints. We will also have to update the dynamic module meta model to have the `@preserveOrder` trait for String enums and Int enums since they get serialized to a map.

a-morales: @kubukoz @lewisjkl any thoughts on this?

lewisjkl: Yeah I think in the sense that the difference between a set and a list in smithy is denoted using `@uniqueItems` it would make sense that a similar "constraint" would exist at the protocol level for ordered maps, if we support it at all (which we kind of have to given the state of the dynamic module). As we talked about earlier, I think if ordering matters to a service in the request body, then the client ought to know that when it constructs said body. Similarly, if it matters to the client in a response, the service ought to take that into account. So it at least _feels_ like a protocol-level concern...

a-morales: I've created PR https://github.com/disneystreaming/smithy4s/pull/1820, which adds a `preserveKeyOrder` smithy trait and updates the codegen to pick the correct `MapTag` based if the trait is present or not.

modules/http4s-protocol-tests/src/smithy4s/tests/ProtocolComplianceSuite.scala
```diff
@@ -98,7 +98,8 @@ abstract class ProtocolComplianceSuite
   def genClientAndServerTests(
       impl: ReverseRouter[IO] with Router[IO],
       shapeIds: ShapeId*
-  )(dsi: DynamicSchemaIndex): List[ComplianceTest[IO]] =
+  )(dsi: DynamicSchemaIndex): List[ComplianceTest[IO]] = {
+    println(s"generating tests for $shapeIds")
```
yisraelU: left in ?

a-morales: Ahh yes, will remove

sampleSpecs/orderedMap.smithy
```diff
@@ -0,0 +1,11 @@
+$version: "1"
+
+namespace smithy4s.example
+
+// TODO: since @preserveKeyOrder changes the underlying type depending on scala version, the bootstrap module codegen task
+// should generate separate srcX_xx directories to render these differences
+// @alloy#preserveKeyOrder
+map OrderedMap {
```
lewisjkl, line 8:
```suggestion
@alloy#preserveKeyOrder
map OrderedMap {
```

We can put an issue in for the todo there I think

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -505,118 +506,121 @@ private[smithy4s] class SchemaVisitorJCodec(
...
-        } else if ((b >= '0' && b <= '9') || b == '-') {
-          in.rollbackToken()
-          new DNumber(in.readBigDecimal(null))
-        } else if (b == '[') {
-          new DArray({
-            if (in.isNextToken(']')) ArraySeq.empty[Document]
-            else
-              ArraySeq.unsafeWrapArray {
+        private val preserveKeyOrder =
+          preserveMapOrder || hints.has(PreserveKeyOrder)
```
lewisjkl: Left a comment on the other PR, but I'm wondering if we can get rid of `preserveMapOrder` in favor of using a transformation in the aws module to add the hint/trait rather than keeping this boolean around

a-morales: I spent some time figuring out the protocol tests and the solution was pretty simple. no need for transformers.

modules/cats/src/smithy4s/interopcats/SchemaVisitorShow.scala
```diff
@@ -99,15 +99,23 @@ final class SchemaVisitorShow(
...
+  override def map[C[_, _], K, V](
       shapeId: ShapeId,
       hints: Hints,
+      tag: MapTag[C],
       key: Schema[K],
       value: Schema[V]
-  ): Show[Map[K, V]] = {
+  ): Show[C[K, V]] = {
     implicit val showKey: Show[K] = self(key)
     implicit val showValue: Show[V] = self(value)
```
lewisjkl: I think these don't need to be implicit any longer

modules/compliance-tests/src/smithy4s/compliancetests/internals/eq/EqSchemaVisitor.scala
```diff
@@ -60,14 +60,24 @@ object EqSchemaVisitor extends SchemaVisitor[Eq] { self =>
...
-  override def map[K, V](
+  override def map[C[_, _], K, V](
       shapeId: ShapeId,
       hints: Hints,
+      tag: MapTag[C],
       key: Schema[K],
       value: Schema[V]
-  ): Eq[Map[K, V]] = {
+  ): Eq[C[K, V]] = {
     implicit val valueEq: Eq[V] = self(value)
```
lewisjkl: Also here, no need for an implicit here any longer I think

modules/compliance-tests/src/smithy4s/compliancetests/internals/eq/EqSchemaVisitor.scala
```diff
@@ -60,14 +60,24 @@ object EqSchemaVisitor extends SchemaVisitor[Eq] { self =>
...
-    Eq[Map[K, V]]
+    new Eq[C[K, V]] {
+      def eqv(x: C[K, V], y: C[K, V]): Boolean = {
+        tag.iterator(x).forall { case (key, xValue) =>
+          tag.get(y, key) match {
+            case Some(yValue) => valueEq.eqv(xValue, yValue)
+            case None         => false
+          }
```
lewisjkl: not a big deal either way, but could probably do `tag.get(y, key).exists(valueEq.eqv(xValue, _))`

modules/core/src-2.12/schema/MapTagCompanionPlatform.scala
```diff
@@ -0,0 +1,46 @@
...
+  type SeqMapType[K, V] = ListMap[K, V]
+
+  case object SeqMapTag extends MapTag[ListMap] {
+    override def name: String = "ListMap"
+
+    override def iterator[K, V](c: ListMap[K, V]): Iterator[(K, V)] =
+      c.iterator
+
+    override def build[K, V](
+        put: (((K, V)) => Unit) => Unit
```
lewisjkl:
```suggestion
        put: ((K, V) => Unit) => Unit
```

Does it need those extra parenthesis? Honest question really, not sure

modules/core/src-2.13/schema/MapTagCompanionPlatform.scala
```diff
@@ -0,0 +1,46 @@
...
+  type SeqMapType[K, V] = SeqMap[K, V]
+
+  case object SeqMapTag extends MapTag[SeqMap] {
+    override def name: String = "SeqMap"
+
+    override def iterator[K, V](c: SeqMap[K, V]): Iterator[(K, V)] =
+      c.iterator
+
+    override def build[K, V](
+        put: (((K, V)) => Unit) => Unit
```
lewisjkl: ditto

modules/core/src-3/schema/MapTagCompanionPlatform.scala
```diff
@@ -0,0 +1,46 @@
...
+  type SeqMapType[K, V] = SeqMap[K, V]
+
+  case object SeqMapTag extends MapTag[SeqMap] {
+    override def name: String = "SeqMap"
+
+    override def iterator[K, V](c: SeqMap[K, V]): Iterator[(K, V)] =
+      c.iterator
+
+    override def build[K, V](
+        put: (((K, V)) => Unit) => Unit
```
lewisjkl: ditto

lewisjkl (commented):
Mainly just a few potential nitpicks here and then the other comment I still have open above. Great work on all of this!

modules/core/src-2.12/schema/MapTagCompanionPlatform.scala
a-morales: With the extra parenthesis it's a function that takes in a single argument that is a tuple. Without the parenthesis, it's a function that takes two arguments. With that in mind though since the use case for this is for other libraries to generate their own map like thing I like the function taking two arguments instead of a single tuple for libraries that are not built in Scala. any opinions?

sampleSpecs/orderedMap.smithy
a-morales: Created https://github.com/disneystreaming/smithy4s/issues/1830 for this as well as a ticket in our internal tracker.

lewisjkl: So in addition to the above, can we also remove the TODO here and uncomment the trait now? Or maybe I am missing something

a-morales: So the trait can't be uncommented since the bootstrap module is cross-built across Scala 2.12, 2.13 and 3, and since the codegen uses scala 2.13 for the codegen the OrderedMap is a `SeqMap` which fails to compile for Scala 2.12.

lewisjkl (approved):

PR #1819, opened 2025-08-25
Update CollectionTag to be unsealed

a-morales:

This PR opens up the CollectionTag so that other collections not defined in smithy4s can be implemented and used. 

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Thank you!
 ## Unseal CollectionTag
 
 Make `CollectionTag` just a trait instead of a sealed trait to allow third party libraries to implement their own `CollectionTag` so that collections
-can be built efficiently.
+can be built efficiently. Existing and new usages of CollectionTags that start getting exhaustivity errors can handle unknown subtypes by using the CollectionTag instance's methods, such as `iterator` and `fromIterator`.
 
 ## Documentation fix
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ Thank you!
 
 # 0.19.0
 
+## Unseal CollectionTag
+
+Make `CollectionTag` just a trait instead of a sealed trait to allow third party libraries to implement their own `CollectionTag` so that collections
+can be built efficiently.
+
 ## Documentation fix
 
 Prevent documentation from being generated for case class when the field are not generated because they're annotated with `@streaming`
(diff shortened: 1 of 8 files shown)


a-morales:
ref: https://github.com/disneystreaming/smithy4s/issues/1331

kubukoz (changes requested):
Needs a CHANGELOG entry + formatting, other than lgtm

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -1000,10 +906,8 @@ private[smithy4s] class SchemaVisitorJCodec(
       member: Schema[A]
   ): JCodec[C[A]] = {
     tag match {
-      case CollectionTag.ListTag       => listImpl(member)
-      case CollectionTag.SetTag        => set(member)
-      case CollectionTag.VectorTag     => vector(member)
       case CollectionTag.IndexedSeqTag => indexedSeq(member)
```
lewisjkl, line 909: Is the implementation for IndexedSeq more optimized or is there another reason to use a custom implementation for it, but not the other collection types?

lewisjkl (commented):
Looks good, just one question from my side

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
a-morales: I kept it in since I saw that this implementation uses a different builder than the rest and based on the comment of the builder it seems to be for performance reasons. 

https://github.com/disneystreaming/smithy4s/blob/55cf50a6749ad38adec2140f5db7f7887703a237/modules/core/src/smithy4s/schema/CollectionTag.scala#L91-L97

lewisjkl: Ah I see, yeah fair enough then, I'm cool with that

lewisjkl (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ Thank you!
 
 # 0.19.0
 
+## Unseal CollectionTag
+
+Make `CollectionTag` just a trait instead of a sealed trait to allow third party libraries to implement their own `CollectionTag` so that collections
+can be built efficiently.
```
kubukoz: **suggestion:** 
```suggestion
can be built efficiently. Existing and new usages of CollectionTags that start getting exhaustivity errors can handle unknown subtypes by using the CollectionTag instance's methods, such as `iterator` and `fromIterator`.
```

kubukoz (approved):

PR #1820, opened 2025-08-27
change MapTag implementation based if @preserveKeyOrder is present or not

a-morales:

- [ ] Added documentation
- [ ] Updated changelog


modules/bootstrapped/src/generated/smithy4s/example/OrderedMap.scala
```diff
@@ -0,0 +1,19 @@
...
+import smithy4s.schema.Schema.bijection
+import smithy4s.schema.Schema.linkedHashMap
+import smithy4s.schema.Schema.string
+
+object OrderedMap extends Newtype[LinkedHashMap[String, String]] {
+  val id: ShapeId = ShapeId("smithy4s.example", "OrderedMap")
+  val hints: Hints = Hints(
+    alloy.PreserveKeyOrder(),
+  ).lazily
+  val underlyingSchema: Schema[LinkedHashMap[String, String]] = linkedHashMap(string, string).withId(id).addHints(hints)
```
kubukoz: **question:** can we make it be just a Map in the type signature?

kubukoz: sigh, mutable Map in there seems like something that'll cause trouble. I was hoping we could've wrapped it in an immutable wrapper, but apparently that's not the case.

Something like this could make sense:

```scala
  def wrapM[K, V](m: collection.mutable.Map[K, V]): collection.immutable.Map[K, V] =
    new collection.immutable.Map[K, V] {
      override def get(key: K): Option[V] = m.get(key)
      override def iterator: Iterator[(K, V)] = m.iterator
      override def removed(key: K): collection.immutable.Map[K, V] = m.toMap - key
      override def updated[V1 >: V](key: K, value: V1): collection.immutable.Map[K, V1] = m.toMap + (key -> value)
      override def size: Int = m.size
    }
```

Or maybe we use `scala.collection.Map` instead? That way, at least we don't expose the mutating methods, and we're one step closer to immutability.

CHANGELOG.md
```diff
@@ -7,10 +7,9 @@ Thank you!
 
 # 0.19.0
 
-## Unseal CollectionTag
+## Add support for `@alloy#preserveKeyOrder` trait
```
kubukoz, line 10: issue: should be separate entries

CHANGELOG.md
```diff
@@ -7,10 +7,9 @@ Thank you!
...
 
-Make `CollectionTag` just a trait instead of a sealed trait to allow third party libraries to implement their own `CollectionTag` so that collections
-can be built efficiently. Existing and new usages of CollectionTags that start getting exhaustivity errors can handle unknown subtypes by using the CollectionTag instance's methods, such as `iterator` and `fromIterator`.
+Any map types that are annotated with `@alloy#preserveKeyOrder` will be rendered as a `ListMap` instead of a `Map` so that key order is preserved.
```
kubukoz: just noting to update this once we finalize the design (it's a ListMap no more)

modules/bootstrapped/src/generated/smithy4s/example/OrderedMap.scala
a-morales: Unfortunately using `toMap` on a LinkHashMap to make it immutable destroys the key order which we want to preserve. If immutability is desired we can use `ListMap` which is immutable, but not performant. I'll give it a go to just using `scala.collection.Map` and see how that feels as a user.

a-morales:
So I played around with the following:

- just using `scala.collection.Map` as the type to hide the mutability methods on `LinkedHashMap` does not provide a good developer experience when using it.
- There is a `SeqMap` trait on scala 2.13+ to denote this types of maps where insertion order should be preserved, with the default implementation on the companion object using `VectorMap`. This would be great, but the `scala-collection-compat` library has an open issue to add support for the `SeqMap` trait https://github.com/scala/scala-collection-compat/issues/443
  - The work to add this to `scsala-collection-compat` would be non trivial

With this our options are:

- Stay with `ListMap` as the implementation to use, which is most likely the original reason why it was picked when `preserveKeyOrder` was added to the json visitor
- Drop scala 2.12 support so that we can use `SeqMap`
- (What I have committed now), change the implementation of the MapTag based on the scala version, using `ListMap` on scala 2.12 and `SeqMap` on scala 2.13+. This brought up issues in the `dynamic` module which has cross version support. The `dynamic` module uses codegen which is hardcoded to scala 2.13 so the bootstrap module doesn't compile for Scala 2.12. We can update the smithy4s codegen task to do cross version generation.

modules/bootstrapped/src/generated/smithy4s/example/OrderedMap.scala
```diff
@@ -0,0 +1,16 @@
...
+
+import smithy4s.Hints
+import smithy4s.Newtype
+import smithy4s.Schema
+import smithy4s.ShapeId
+import smithy4s.schema.Schema.bijection
+import smithy4s.schema.Schema.map
+import smithy4s.schema.Schema.string
+
+object OrderedMap extends Newtype[Map[String, String]] {
```
lewisjkl, line 11: Is this actually an ordered map? I don't see the preserve order trait, so I'm not clear on how that is tracked in this case

modules/core/src/smithy4s/schema/MapTag.scala
```diff
@@ -17,18 +17,14 @@
 package smithy4s
 package schema
 
-import scala.collection.immutable.ListMap
+// import scala.collection.immutable.TreeSeqMap
```
lewisjkl:
```suggestion
```

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -505,118 +507,121 @@ private[smithy4s] class SchemaVisitorJCodec(
         out.writeNonEscapedAsciiKey(x.toString)
     }
 
-    def document(maxArity: Int): JCodec[Document] = new JCodec[Document] {
-      import Document._
-      override def canBeKey: Boolean = false
+    // TODO: once preserveKeyOrder is removed this annotation can be remove
+    @scala.annotation.nowarn("cat=deprecation")
```
lewisjkl: Can it be removed now?

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -505,118 +507,121 @@ private[smithy4s] class SchemaVisitorJCodec(
...
+            })
+          } else if (b == '{') {
+            new DObject({
+              if (in.isNextToken('}')) Map.empty
+              else {
                 in.rollbackToken()
-                var arr = new Array[Document](4)
+                // We use the maxArity limit to mitigate DoS vulnerability in default Scala `Map` implementation: https://github.com/scala/bug/issues/11203
+                val obj =
+                  if (preserveMapOrder || hints.has(PreserveKeyOrder))
```
lewisjkl: Potential optimization, store the result of this boolean expression outside the level of this function so it doesn't need to be recomputed each time during decoding. `hints.has` should be `O(1)`, but 🤷

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -505,118 +507,121 @@ private[smithy4s] class SchemaVisitorJCodec(
...
+                  } else in.arrayEndOrCommaError()
+                }
+            })
+          } else if (b == '{') {
+            new DObject({
+              if (in.isNextToken('}')) Map.empty
+              else {
                 in.rollbackToken()
-                var arr = new Array[Document](4)
+                // We use the maxArity limit to mitigate DoS vulnerability in default Scala `Map` implementation: https://github.com/scala/bug/issues/11203
```
lewisjkl: I think this comment should move down a few lines now

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -628,7 +633,7 @@ private[smithy4s] class SchemaVisitorJCodec(
       case PBlob       => PrimitiveJCodecs.bytes
       case PBoolean    => PrimitiveJCodecs.boolean
       case PByte       => PrimitiveJCodecs.byte
-      case PDocument   => documentJCodec
+      case PDocument   => PrimitiveJCodecs.document(maxArity, hints)
```
lewisjkl, line 635: I think this was pulled out to the higher level for the same reason I posted above, just moving things off the hot-path wherever possible

sampleSpecs/orderedMap.smithy
```diff
@@ -0,0 +1,11 @@
+$version: "1"
+
+namespace smithy4s.example
+
+// TODO: since @preserveKeyOrder changes the underlying type depending on scala version, the bootstrap module codegen task
+// should generate separate srcX_xx directories to render these differences
+// @alloy#preserveKeyOrder
+map OrderedMap {
```
lewisjkl, line 8: Ah I see, for now we can just have the same as it currently is IMO and we could add an issue to enable this sort of thing for a limited number of the examples

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
a-morales: It can, I guess this is more relevant for the 0.18 branch in which the key order is set via the `preserveMapOrder` argument passed to the `SchemaVisitorJCodec`

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
a-morales: Yeah I saw that, but given the `preserveKeyOrder` trait is a hint we will have to pass it in since it's dependent on the case and it's not something that can be pulled out.

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
a-morales: So the protocol tests for `aws-http4s` depends on key ordering and since it's defined in the smithy repo protocol tests it's not something we can update with the `@preserveMapOrder` trait so the boolean argument needs to stay.

lewisjkl: We could potentially use a transformation to add the preserveMapOrder trait/hint and also add it to the hint mask. I'm not super familiar with the exact situation, but I think it'd be possible.

Issue #1817, opened 2025-08-22
Codegen produces non-existent `alloy.openapi.Summary` in hints


tpetillot:
**Generation result:**

```scala
withHints(alloy.openapi.Summary("Service summary"), smithy.api.Auth(Set(...)))
````

It looks like alloy.openapi.Summary is being generated from [DynamicModelCompiler.allHints](https://github.com/disneystreaming/smithy4s/blob/series/0.18/modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala#L143). 

However, this type doesn’t actually exist.
Are we missing something here, or is this an unintended generation issue?

The related class seems to be [SummaryTrait](https://github.com/disneystreaming/alloy/blob/4fcbad5d62c0a6018aaeb692240ff761400c576d/modules/core/src/alloy/openapi/SummaryTrait.java#L23)

**Comment by lewisjkl** [comment]
Hey there @tpetillot, thanks for raising an issue. I think we'll need a bit more context in order to help out here. So here are some assumptions I have about your situation, please correct them if I'm wrong at all:

- You are using smithy4s SBT plugin to generate code
- You are defining some smithy specs and in those you are using the `alloy.openapi#Summary` trait
- You are getting a compilation error telling you that `alloy.openapi.Summary` (the scala class) does not exist

Is that all correct?

You also mentioned the `DynamicModelCompiler`, are you using the smithy4s-dynamic module? This is less common, so I want to make sure I've got this right.

**Comment by kubukoz** [comment]
I think this is a version mismatch. Your version of Alloy is more recent than the one Smithy4s is built against.

Dynamic is probably a red herring (unrelated code)

**Comment by tpetillot** [comment]
Hello @lewisjkl and @kubukoz, thanks for your responses!

`DynamicModelCompiler` was just me guessing were the issue could come from, so must be unrelated.

On aligned version `0.3.28` of `alloy-core`, still have the following error:
```
object Summary is not a member of package alloy.openapi
       .withHints(alloy.openapi.Summary("test"), ...
```

kubukoz:
hmm we probably need to add alloy.openapi here, then:

https://github.com/disneystreaming/smithy4s/blob/0a44e7bddc3622e83c17f0512dcf435419b7efcd/build.sbt#L206-L212

In the meantime, you can do this:

- set `smithy4sAllowedNamespaces` to opt into generating `alloy.openapi`
- make sure to add any other namespaces you want to generate

[...]

PR #1822, opened 2025-08-29
fix: add 'alloy.openapi' to default allowed namespaces

tpetillot:
ref: https://github.com/disneystreaming/smithy4s/issues/1817

---

Note: i couldn't add alloy*, that was causing following error (among a lot of other ones)
```
[error] [core3] -- [E008] Not Found Error: /Users/tpetillot/workspace/repositories/disneystreaming/smithy4s/modules/core/src/smithy4s/schema/FieldFilter.scala:20:13 
[error] [core3] 20 |import alloy.Nullable
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ When adding entries, please treat them as if they could end up in a release any 
 
 Thank you!
 
+# 0.18.43
+
+* codegen: Allow `alloy.openapi.*` namespaces by default [#1822](https://github.com/disneystreaming/smithy4s/pull/1822)
+
 # 0.18.41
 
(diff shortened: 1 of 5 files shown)


kubukoz:
Thanks for the PR!

hmm, the pattern should've worked, according to [this pr](https://github.com/disneystreaming/smithy4s/pull/1649/files). Maybe it doesn't work if there's less than 2 segments in the namespace pattern 🤔 

Regardless, you'll need to run `bootstrapped/managedSources` in sbt locally (or just `compile`) and check in the generated files.

Also, please add a CHANGELOG entry for 0.18.43 (there's [a PR template](https://github.com/disneystreaming/smithy4s/blob/series/0.18/.github/PULL_REQUEST_TEMPLATE.md) with a checklist of things worth doing when submitting a PR)

kubukoz (approved):
Thanks :)

the entry for 0.18.42 is being added in another PR, btw

Issue #967, opened 2023-05-17
SUGGESTION: first class scalacheck generators 


caoilte:
It's pretty obvious to anyone who's been using smithy4s a few minutes that automatic scalacheck generators would be pretty neat. There seems to have been some work on this eg,

https://github.com/disneystreaming/smithy4s/tree/series/0.18/modules/scalacheck/src/smithy4s/scalacheck

but it's also been suggested by @kubukoz that ["it's not published, plus i don't think it handles validations"](https://discord.com/channels/1045676621761347615/1107711700515438604/1108067470692073482).

Regardless I don't think there is any documentation for current support or a roadmap for the remaining pieces that would make it first class so @kubukoz suggested I raise this issue to track.

kubukoz:
IIRC the main issue is the validations - generating things marked with [constraint traits](https://smithy.io/2.0/spec/constraint-traits.html#idref-trait) and [refinements](https://disneystreaming.github.io/smithy4s/docs/codegen/customisation/refinements/) would need to either:

- generate values that would pass the validation, OR
- generate values and let the generator filter out mismatching cases.

Option 2 is easier as it's a general solution that works for virtually all constraints. The huge problem is that it may filter out too many values and cause Scalacheck to fail because it won't be able to generate sufficient amounts of valid values.

We could special-case for the refinements we know about (length, range, regex - generating these is a tough task but might be possible, swagger UI does it albeit with some bugs), which is option 1... but I don't know how to do it for user-defined refinements: the generator would sort of have to be built into the schema, so that the SchemaVisitor that constructs generators would see it.

Baccata:
@kubukoz got it right : this is simply not something I'm willing to commit to providing a solution for, because the fact that we're letting users come with their own refinement rules (with custom constraints) makes it downright impossible to derive sensible scalacheck generators in a generic fashion. 

We do have a limited implementation but that we use for our own tests but it's certainly not something I want published as a consumable artifact. Feel free to copy/paste/adapt the code to fit your own usecase though. 

I also think that this is one of the few usecases that actually benefits from full-auto derivation with shapeless : 

* https://github.com/alexarchambault/scalacheck-shapeless

[...]

PR #1827, opened 2025-09-16
Include scalacheck in allModules

matejcerny:
The `scalacheck` module is currently not being published


Baccata:
This is not a mistake, we do not want to publish the scalacheck module. See context : 

https://github.com/disneystreaming/smithy4s/issues/967

kubukoz:
thanks @Baccata, I was going to reply but I couldn't find the ticket 🙏

PR #1834, opened 2025-09-24
add Cats instances for C[_] given CollectionTag[C]

bpholt:

- [ ] ~~Added bootstrapped code + smoke tests (when the rendering logic is modified)~~
- [ ] ~~Added build-plugins integration tests (when reflection loading is required at codegen-time)~~
- [ ] ~~Added alloy compliance tests (when simpleRestJson protocol behaviour is expanded/updated)~~
- [ ] ~~Updated dynamic module to match generated-code behaviour~~
- [ ] Added documentation
- [ ] Updated changelog

There have been a few times that I've wanted to interact with a `C[_] : CollectionTag` using e.g. `Functor[C]` or `Monad[C]`. I was able to build lawful instances so I thought this would be a useful contribution.

I'll update the changelog if/once there's consensus that this is otherwise ready to merge. Happy to add documentation, too, although I'm not sure exactly where.

### DIFF
--- a/modules/cats/src-3/smithy4s/interopcats/internal/NotGiven.scala
+++ b/modules/cats/src-3/smithy4s/interopcats/internal/NotGiven.scala
@@ -1,0 +1,22 @@
+/*
+ *  Copyright 2021-2025 Disney Streaming
+ *
+ *  Licensed under the Tomorrow Open Source Technology License, Version 1.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     https://disneystreaming.github.io/TOST-1.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.interopcats.internal
+
+type NotGiven[+A] = scala.util.NotGiven[A]
+
+object NotGiven:
+  export scala.util.NotGiven.given
--- a/modules/cats/src/smithy4s/interopcats/package.scala
+++ b/modules/cats/src/smithy4s/interopcats/package.scala
@@ -18,9 +18,11 @@ package smithy4s
 
 import scala.util.hashing.MurmurHash3
 import smithy4s.capability.MonadThrowLike
-import cats.MonadThrow
+import cats.{instances => _, _}
 import cats.syntax.all._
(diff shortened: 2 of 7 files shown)


modules/cats/src-3/smithy4s/interopcats/internal/NotGiven.scala
```diff
@@ -0,0 +1,22 @@
...
+ *     https://disneystreaming.github.io/TOST-1.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.interopcats.internal
```
bpholt, line 17: I put this in `.internal` so that it wouldn't be imported if someone did `import smithy4s.interopcats.*`. I'm happy to put it somewhere else if desired.

Baccata: If it's in `.internal`, it means that users should really not use it (yourself included), put it in `smithy.interopcats` instead.

modules/cats/src/smithy4s/interopcats/package.scala
```diff
@@ -60,4 +62,19 @@ package object interopcats {
     MurmurHash3.finalizeHash(hashResult, hashes.length)
   }
 
+  implicit def catsInstancesGivenCollectionTagForSet[C[_]](implicit
+      CT: CollectionTag[C],
+      @annotation.unused IsSet: C[Any] =:= Set[Any]
+  ): Traverse[C] with Monad[C] =
+    new instances.CatsInstancesForCollectionTag[C]
```
kubukoz, line 69: **question:** can you elaborate on why we have a separate instance for sets?

Cats normally doesn't even provide a functor for Set because of ordering problems and so on.

In addition, in a polymorphic context you won't see a value of `C[Any] =:= Set[Any]` (because the compiler can't know C is Set at that point), and it'd only be useful after you've matched specifically to `case SetTag`, at which point we might as well just provide a `Traverse[Set]`.

modules/cats/src/smithy4s/interopcats/package.scala
```diff
@@ -60,4 +62,19 @@ package object interopcats {
...
+
+  implicit def catsInstancesGivenCollectionTag[C[_]](implicit
```
kubukoz, line 71: **thought:** wonder if we can avoid the `NotGiven` check by putting this in a separate trait (also inherited by this object), playing with implicit priority like Cats does with its own instances. That'd be a nice way to get rid of the scala2 NotGiven.

PR #1836, opened 2025-10-01
add in scripted test for refinement usage

yisraelU:


a-morales (approved):

modules/codegen-plugin/src/sbt-test/codegen-plugin/refinements/test
```diff
@@ -0,0 +1 @@
+> compile
```
kubukoz: **suggestion:** I'd use `run` unless there's a good reason not to

PR #1848, opened 2025-10-20
config-based dynamic hint enable

lewisjkl:

- [x] Added documentation
- [x] Updated changelog

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
@@ -123,6 +123,27 @@ private[codegen] class SmithyToIR(
       .flatMap(f => DefaultRenderMode.fromString(f.getValue))
       .getOrElse(DefaultRenderMode.Full)
 
+  private case class NamespaceFragment(value: String) {
+    // Checks if `namespace` is "under" this fragment
+    // meaning is a sub-namespace of this fragment.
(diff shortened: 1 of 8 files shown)


modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -123,6 +123,27 @@ private[codegen] class SmithyToIR(
       .flatMap(f => DefaultRenderMode.fromString(f.getValue))
       .getOrElse(DefaultRenderMode.Full)
 
+  private case class NamespaceFragment(value: String) {
+    // Checks if `namespace` is "under" this fragment
+    // meaning is a sub-namespace of this fragment.
+    // This prevents things such as considering that "food" is
+    // under "foo"
+    def isNamespaceUnder(namespace: String): Boolean =
```
kubukoz: **suggestion:** instead of implicitly assuming that `smithy4sRenderDynamicHintNamespaces` should include the "prefix namespaces", maybe we should be consistent with the namespace regex in `NamespacePattern` from https://github.com/disneystreaming/smithy4s/pull/1649/files#diff-668170dc2ad0117ba5b65f2225be78b8dd128831e15f4deef08d8338d01334ddR280-R310

which is to say, maybe it should be `smithy4sRenderDynamicHintNamespacePatterns`

a-morales (approved):

Issue #1815, opened 2025-08-21
Pure Caching


lewisjkl:
Leaving this mostly as a placeholder for now with details to follow, but basically want a pure alternative to the current mutable-map-based CachedSchemaCompiler.

This will offer several advantages:
- Caching in the presence of transformations (currently cache keys are references, so transformations do not result in stable references)
- Injecting arbitrary/custom behavior into existing SchemaVisitors in user-land will be more possible due to recursion being abstracted to a different layer. Currently this sort of customization requires object-oriented "extending" of SchemaVisitor implementations.

PR #1857, opened 2025-11-13
Fix bug around Lazy caching and update Lazy model

a-morales:
While working on updating the Lazy model for #1815 I was writing a test to come up with a case where Lazy was being problematic. I found that schema visitors that implement the `CachedVisitor` which uses the `CompilationCache` along with using `transformTransitivelyK` on the schema results in the store of the `CompilationCache` to grow in proportion to how big a recursive structure is. Before my fix the size of the store in the test was the same order of magnitude as the size of the recursive structure (so if I created a cons list with a length of 1000 the store would have a size of 1000). With my fix the size of the store stays constant with the number of underlying schemas for a given schema. 

- [ ] Added documentation
- [ ] Updated changelog


modules/core/src/smithy4s/Lazy.scala
```diff
@@ -16,17 +16,24 @@
...
-final class Lazy[A](make: () => A) {
-  private[this] var thunk: () => A = make
-  lazy val value: A = {
-    val result = thunk()
-    thunk = null
-    result
-  }
-
-  def map[B](f: A => B): Lazy[B] = new Lazy(() => f(make()))
+sealed trait Lazy[A] {
```
a-morales: This change is not necessarily required for this PR.

modules/core/src/smithy4s/schema/Schema.scala
```diff
@@ -219,8 +221,14 @@ object Schema {
         underlying(u.copy(alternatives = u.alternatives.map(handleAlt(_))))
       case BijectionSchema(s, bijection) =>
         underlying(BijectionSchema(this(s), bijection))
-      case LazySchema(suspend) =>
-        underlying(LazySchema(suspend.map(this.apply)))
```
a-morales, line 228: This was the problematic line here and it's actually 2 bugs in the same line. 

both `underlying` and `this.apply` are the same so the transformation was essentially being applied twice (This is evident with the updated Lazy changes where applying a transform went from 
`LazySchema(Lazy.Root(_))` to `LazySchema(Lazy.Mapped(Lazy.Mapped(Lazy.Root(_), _), _))`. The other bug is that `this.apply` does not have a stable reference and it modifies the underlying schema which means that the store in the `CompilationCache` was getting a different transformed Schema which would not be in the store and thus be added and returned. Just adding a map that returns the same result for Lazy fixes the bug.

modules/core/src/smithy4s/Lazy.scala
Baccata: Does making it a `sealed abstract class` solve the MiMa issue ?

modules/core/src/smithy4s/schema/Schema.scala
Milyardo: From my [PR](https://github.com/disneystreaming/smithy4s/pull/1871/changes#diff-0641f3c258832869de66b9c14c6dd2810a17ac69f3e590bed1de43d77cf86203L230), this fix can be simplified to `LazySchema(Lazy(underlying(suspend.value)))`

Milyardo (changes requested):

modules/core/src/smithy4s/Lazy.scala
a-morales: Unfortunately it does not.

modules/core/src/smithy4s/schema/Schema.scala
a-morales: Trying your simplification causes some tests to fail. Need to investigate why.

Milyardo (approved):

yisraelU (approved):

PR #1864, opened 2025-12-25
[Core] Set Accept http header 

### LINKED ISSUE #1863: Smithy4s clients should set an appropriate Accept header
We've seen some issues making requests of specific server implementations when using smithy4s and `@simpleRestJson`, and I think it's because smithy4s isn't setting the `Accept` header when making requests. I was able to demonstrate this using this Smithy spec:
```smithy
$version: "2"
namespace echo.headers
use alloy#simpleRestJson

map RequestHeaders {
    key: String
    value: String
}

structure Response {
    @required
    value: RequestHeaders
}

@http(method: "GET", uri: "/", code: 200)
operation GetRoot {
    input: Unit,
    output: Response
}

@simpleRestJson
service EchoHeaders {
    operations: [GetRoot]
}
```
and demo app:
```scala
package echo.headers

import cats._
import cats.data.OptionT
import cats.effect.{Trace => _, _}
import cats.mtl.Local
import cats.syntax.all._
import com.comcast.ip4s._
import org.http4s.server.Server
import org.http4s.syntax.all._
import org.http4s.{Headers, HttpRoutes, Uri}

class EchoHeadersImpl[F[_] : Functor](requestHeaders: F[Headers]) extends EchoHeaders[F] {
  override def getRoot(): F[Response] =
    requestHeaders
      .map {
        _.headers
          .map { h =>
            h.name.toString -> h.value
          }
          .toMap
      }
      .map(Response(_))
}

class RequestHeaderProvidingMiddleware[F[_] : Monad](implicit L: Local[F, Headers]) extends (HttpRoutes[F] => HttpRoutes[F]) {
  override def apply(routes: HttpRoutes[F]): HttpRoutes[F] =
    HttpRoutes[F] { request =>
      Local[OptionT[F, *], Headers].scope {
        routes(request)
      }(request.headers)
    }
}

object EchoHeadersApp extends IOApp.Simple {
  private def server: Resource[IO, Server] =
    IO.local(Headers.empty)
      .toResource
      .flatMap { implicit l: Local[IO, Headers] =>
        SimpleRestJsonBuilder
          .routes(new EchoHeadersImpl(l.ask))
          .resource
          .map(new RequestHeaderProvidingMiddleware[IO])
      }
      .map(_.orNotFound)
      .flatMap {
        EmberServerBuilder
          .default[IO]
          .withHost(host"localhost")
          .withPort(port"0")
          .withHttpApp(_)
          .build
      }

  private def client(uri: Uri): Resource[IO, EchoHeaders[IO]] =
    EmberClientBuilder
      .default[IO]
      .build
      .flatMap {
        SimpleRestJsonBuilder(echo.headers.EchoHeaders)
          .client(_)
          .uri(uri)
          .resource
      }

  override def run: IO[Unit] =
    server
      .map(_.baseUri)
      .flatMap(client)
      .use(_.getRoot().flatMap(IO.println))
}
```

Running the app emits the following to the console:
```
15:09:46.674 [io-compute-1] INFO org.http4s.ember.server.EmberServerBuilderCompanionPlatform -- Ember-Server service bound to address: 127.0.0.1:49462
Response(Map(Host -> 127.0.0.1:49462, Date -> Tue, 16 Dec 2025 21:09:46 GMT, Connection -> keep-alive, User-Agent -> http4s-ember/0.23.33))
```
Note the absence of an `Accept` header, which I was expecting to be present and set to `Accept: application/json`.

This doesn't cause issues when using a smithy4s server, but some servers are stricter and refuse to handle requests without content negotiation via the `Accept` header.

Tested with Smithy4s 0.18.45.

yisraelU:
closes #1863 
- Adds an Accept header Writer that uses the Output schema to determine the MIME type for the header . Includes a default plus the ability to configure the header. 
Follows the same logic for the Content-Type header Writer
   - if rawStringsAndBlobs are set to true , use MediaTypeVisitor to derive the mediaType from the *Output* schema. This will ignore the `acceptMediaType` setting for the Codec. 
   - if not set , use the configured setting 
- sets restJson protocol to use `application/json` for the Http4s interpreter
- sets the appropriate accept header for the AWS protocols

- [x] Updated changelog


modules/http4s/src/smithy4s/http4s/internals/SimpleRestJsonCodecs.scala
```diff
@@ -138,6 +138,7 @@ private[http4s] class SimpleRestJsonCodecs(
       )
       .withBaseRequest(_ => baseRequest.pure[F])
       .withRequestMediaType("application/json")
+      .withAcceptMediaType("application/json")
```
yisraelU, line 141: sets default for simpleRestJson

modules/core/src/smithy4s/http/HttpUnaryClientCodecs.scala
```diff
@@ -210,8 +216,30 @@ object HttpUnaryClientCodecs {
               case None => inputEncoders.fromSchema(endpoint.input, inputEncoderCache)
             }
 
+          val acceptHeaderWriter: HttpRequest.Writer[Blob, I] = {
```
yisraelU, line 219: core logic starts here

lewisjkl (commented):
Not sure why CI is failing on that last commit, but this looks good to me. Once we get CI passing I think we are good to go

yisraelU:
@lewisjkl fixed.
Issue was, ironically, generated files are generated with they current year in the header . So rebasing on my other pr to update end year to 2026 resolved the issue

lewisjkl (approved):

sampleSpecs/acceptHeader.smithy
kubukoz, line 1: **thought:** IMO there was no need to add a `sampleSpec`, at least not one that'd get generated into files that are checked in. The change didn't make any codegen, after all.

Then again, we don't have such files (available in `bootstrapped/Test` but not checked in) so some more changes would've been required to get that done, feel free ignore this :)

kubukoz (commented):
good stuff!

PR #1868, opened 2026-01-09
[Core] - Content header writer refactor ,add content header test, and remove mediaTypeWriters

yisraelU:
This PR changes how the content-type header is added to an outbound request in the Client codecs.

Currently , the mediaTypeWriters is invoked only when the MetadataEncoders are set .
The default for the UnaryClientCodecs is to have the Metadata Encoders set to None . ( I assume this is because some protocols dont have metadata ) . I observed when creating a test for the Content-Type header , if I did not set the metadata encoder a content type header would not be added . This is a hidden dependency and is incorrect behavior . The source for this logic is [here](https://github.com/disneystreaming/smithy4s/blob/3a9eea61df19d145164b6a41974ed5b6c5775561/modules/core/src/smithy4s/http/HttpUnaryClientCodecs.scala#L189) , where if the metadata encoders field is `None` , we just return the mediaTypeWriters , which does not know how to decompose a schema and partition out the body part 

The logic for setting content-type header is ,
-  there must be content i.e. it should not be set if the body is empty 
- if rawStringsAndBlobs is set and no mediaType is set , content-type is set according to the default media-type , otherwise use the mediaType value
- default to the protocol configured by the requestedMediaType setting

- [ ] Updated changelog

### DIFF
--- a/modules/bootstrapped/test/src/smithy4s/http/ContentHeaderSpec.scala
+++ b/modules/bootstrapped/test/src/smithy4s/http/ContentHeaderSpec.scala
@@ -42,6 +42,7 @@ final class ContentHeaderSpec extends FunSuite {
   ], HttpResponse[Blob]] =
     HttpUnaryClientCodecs
       .builder[Id]
+      .withMetadataEncoders(Metadata.Encoder)
       .withBaseRequest(_ =>
         HttpRequest(
           HttpMethod.POST,
@@ -86,7 +87,7 @@ final class ContentHeaderSpec extends FunSuite {
   }
 
   test(
-    "Content-Type header uses default requestMediaType (text/plain) when no specific media type is set"
+    "Content-Type header is NOT set when body encoders are not configured (body is empty)"
   ) {
     val codecsMake = baseBuilder.build()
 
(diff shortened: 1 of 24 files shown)


modules/bootstrapped/test/src/smithy4s/http/ContentHeaderSpec.scala
```diff
@@ -0,0 +1,203 @@
...
+      Nothing,
+      Nothing
+    ](
+      ContentHeaderTestServiceOperation.DefaultContentHeader.schema
+    )
+    val request = codec.inputEncoder(DefaultContentHeaderInput("test data"))
+
+    val contentTypeHeader = extractContentTypeHeader(request)
+    // Note: The builder chain doesn't properly override requestMediaType, so it uses the default "text/plain"
+    // In production (http4s), the codec is configured directly with the desired requestMediaType
```
lewisjkl: Not sure I understand what is meant by this, can you elaborate?

lewisjkl (commented):
I think this looks good, just clarifying question on the one test

modules/bootstrapped/test/src/smithy4s/http/ContentHeaderSpec.scala
yisraelU: My apologies , this is leftover comments from a completely different attempt at solving the underlying  issue

yisraelU:
setting back to draft as I need to clarify some additional points

PR #1869, opened 2026-01-09
Update Dependency Versions

lewisjkl:
It seems scala-steward is not updating dependencies in our repo at the moment, I went ahead and updated some dependencies that have fallen behind. Will work on getting smithy-model all the way up to date, I think I need to update it in smithy-translate and alloy first so going to try that.

Note that this is not an exhaustive update of everything, I just manually went through and updated ones that seemed higher priority to keep up to date for the moment. We can be more exhaustive, especially with things like sbt plugins in another PR.

# Included Updates

- Scala 2.13.16 -> 2.13.18
- Smithy-translate 0.5.x -> 0.7.x
- Smithy (all libs, like smithy-model) 1.57.x -> 1.63.x (latest is 1.65, but having issues with it at the moment, going to look at this in another PR)
- Http4s 0.23.30 -> 0.23.33
- Alloy 0.3.33 -> 0.3.35
- Fs2 3.12.0 -> 3.12.2
- sbt-scalajs 1.18.2 -> 1.20.2
- sbt-scalafix 0.14.3 -> 0.14.5
- Kind Projector 0.13.3 -> 0.13.4

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -46,7 +46,8 @@ object Dependencies {
 
   val Cats = new {
     val core: Def.Initialize[ModuleID] =
-      Def.setting("org.typelevel" %%% "cats-core" % "2.13.0")
+      // must remain on 2.11 until we update scala-native version
+      Def.setting("org.typelevel" %%% "cats-core" % "2.11.0")
   }
 
   val Monocle = new {
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -10,7 +10,8 @@ object Dependencies {
 
   val Jsoniter = new {
     val org = "com.github.plokhotnyuk.jsoniter-scala"
-    val jsoniterScalaVersion = "2.38.8"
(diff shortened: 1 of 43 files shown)


project/Dependencies.scala
```diff
@@ -19,7 +20,7 @@ object Dependencies {
 
   val Smithy = new {
     val org = "software.amazon.smithy"
-    val smithyVersion = "1.57.1"
+    val smithyVersion = "1.63.0"
```
yisraelU, line 23: should we upgrade this to the latest version ?

yisraelU: nm , saw comments

yisraelU (approved):

project/Dependencies.scala
lewisjkl: Yeah working on getting to latest in alloy and smithy-translate and then will raise another PR here to update all of them

PR #1870, opened 2026-01-13
Query params refactor

yisraelU:
This PR is based off [this](https://github.com/disneystreaming/smithy4s/pull/1513) PR, focusing on remodeling the http query params to an RFC compliant model 

Adds additional tests , plus relies on existing tests to confirm conformance.

- the isSubset method was refactored to not be abstract as its coupled to http implementation and we specifically want to handle Strings
-

### DIFF
--- a/modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
+++ b/modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
@@ -17,46 +17,37 @@
 package smithy4s
 package compliancetests
 
-import cats.implicits._
 import org.http4s.Header
 import org.http4s.Headers
 import org.http4s.Uri
 import org.typelevel.ci.CIString
 
 import java.nio.charset.StandardCharsets
-import scala.collection.immutable.ListMap
 
 package object internals {
 
   private[compliancetests] def splitQuery(
       queryString: String
-  ): (String, String) = {
+  ): (String, Option[String]) = {
     queryString.split("=", 2) match {
       case Array(k, v) =>
         (
           k,
-          Uri.decode(
+          Some(Uri.decode(
             toDecode = v,
             charset = StandardCharsets.UTF_8,
             plusIsSpace = true
-          )
+          ))
         )
-      case Array(k) => (k, "")
+      case Array(k) => (k, None)
     }
   }
 
   private[compliancetests] def parseQueryParams(
       queryParams: Option[List[String]]
-  ): ListMap[String, List[String]] = {
-    queryParams.combineAll
+  ): Vector[(String, Option[String])] = {
+    queryParams.toVector.flatten
       .map(splitQuery)
-      .foldLeft[ListMap[String, List[String]]](ListMap.empty) {
-        case (acc, (k, v)) =>
-          acc.get(k) match {
-            case Some(value) => acc + (k -> (value :+ v))
-            case None        => acc + (k -> List(v))
-          }
-      }
   }
 
   private def escape(str: String): String = {
--- a/modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
+++ b/modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
@@ -33,11 +33,13 @@ package object internals {
       case Array(k, v) =>
         (
           k,
-          Some(Uri.decode(
-            toDecode = v,
-            charset = StandardCharsets.UTF_8,
-            plusIsSpace = true
-          ))
+          Some(
+            Uri.decode(
+              toDecode = v,
+              charset = StandardCharsets.UTF_8,
+              plusIsSpace = true
+            )
+          )
         )
       case Array(k) => (k, None)
     }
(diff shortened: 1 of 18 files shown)


modules/compliance-tests/src/smithy4s/compliancetests/internals/package.scala
```diff
@@ -17,46 +17,39 @@
...
+          Some(
+            Uri.decode(
+              toDecode = v,
+              charset = StandardCharsets.UTF_8,
+              plusIsSpace = true
+            )
           )
         )
-      case Array(k) => (k, "")
+      case Array(k) => (k, None)
```
yisraelU, line 44: this is because there was no "=" in the string , hence this was a "?foo" vs "?foo="

modules/core/src/smithy4s/http/HttpUnaryServerRouter.scala
```diff
@@ -305,10 +305,28 @@ object HttpUnaryServerRouter {
 
   }
 
-  /** Checks if `larger` multimap is a subset of the `smaller` */
-  private def isSubset[K, V](larger: Map[K, Seq[V]], smaller: Map[K, Seq[V]]): Boolean =
-    smaller.forall { case (k, vRequired) =>
-      larger.get(k).exists(vs => vRequired.forall(vs.contains))
+  /** Checks if `smaller` multimap is a subset of the `larger` */
+  private def isSubset(
```
yisraelU, line 309: changed type signature here , as we need to know more about the types to properly compare with empty string

a-morales (approved):

modules/core/src/smithy4s/http/HttpUri.scala
```diff
@@ -28,7 +28,7 @@ import java.nio.charset.StandardCharsets
 final case class HttpUri private (
     origin: Option[HttpUriOrigin],
     path: IndexedSeq[String],
-    queryParams: Map[String, Seq[String]],
+    queryParams: IndexedSeq[(String, Option[String])],
```
lewisjkl, line 31: What is the reason for representing this as an `IndexSeq` instead of a `Map` like you have below in `queryParamsAsMap`? `Map[String, Seq[Option[String]]]`

modules/core/src/smithy4s/http/Metadata.scala
```diff
@@ -120,20 +135,23 @@ case class Metadata(
     )
   }
 
-  def find(location: HttpBinding): Option[(String, List[String])] =
+  def find(
+      location: HttpBinding
+  ): Option[(Option[String], List[Option[String]])] =
```
lewisjkl: Why is the first member of the tuple optional? Does that not represent, for example, the name of the query parameter?

In any case we should probably add a comment to this method explaining exactly what it is doing and what the type signature means

modules/core/src/smithy4s/http/HttpUri.scala
yisraelU: IndexedSeq will preserve the ordering of the distinct keys too , thereby allowing for the exact reconstruction of the url string. The asMap function is just a helper for interfaces that require Maps

lewisjkl: I see, that makes sense

modules/core/src/smithy4s/http/Metadata.scala
yisraelU: The gist of this method is to use smithy http trait information to find values from the Metadata. It returns the values as a tuple of the first value aka the "head" and the "tail" if multiple values exist exists. For queries that would be using the key to find the first value and potentially more values if there are multiple instances of the same key within the query string. Because this PR changes the values to be Options , the "head" would be an option too

yisraelU: I'll add comments too

lewisjkl: I see, appreciate the explanation. Perhaps in this case having a class to represent the output would be worth it to make the intent totally clear (even the old implementation is not really obvious imo)

yisraelU: Good call! I've added a case class to model the return type and added some comments for additional clarification

yisraelU:
@lewisjkl addressed feedback

modules/core/src/smithy4s/http/Metadata.scala
```diff
@@ -144,6 +177,20 @@ object Metadata {
...
+   * @param head The first value. Can be None for valueless query parameters (e.g., ?foo),
+   *             or Some(value) for parameters with values (e.g., ?foo=bar).
+   *             Always Some for headers and path parameters.
+   * @param tail Additional values for multi-valued parameters. Each value is wrapped in Option
+   *             to support valueless query parameters in lists.
+   */
+  final case class BindingValues(
+      head: Option[String],
+      tail: List[Option[String]]
+  )
```
lewisjkl: For bincompat reasons perhaps we should make this just a `final class` since it is a public type

lewisjkl (approved):

Issue #1872, opened 2026-01-20
Expose refinement/validation evidence for ValidatedNewtype (analogous to Newtype#asBijection)


ghostbuster91:
**Motivation**

`smithy4s.Newtype` exposes a compile-time `Bijection[A, T]`, which makes it possible to generically lift schema compilers and codecs from a base type to the newtype:

```scala
given [A, T](using
  C: MyCustomCodec[A],
  B: Bijection[A, T]
): MyCustomCodec[T] =
  C.iso(B.to, B.from)
```

`ValidatedNewtype[A]` doesn't expose any compile-time evidence for `A => Either[String,T] + T => A`, so generic codec/schema lifting is not possible. Consider exposing a Prism/Refinement-like typeclass (partial isomorphism) for `ValidatedNewtype`.

**Use-case**

Consider a Smithy model like:
```smithy
@length(min: 1, max: 100)
@validatedNewtype
string Name
```

Which generates:
```scala
object Name extends ValidatedNewtype[String] {
  def apply(value: String): Either[String, Type] = ...
}
```

Assume we already have a codec for String:
```scala
given MyCustomCodec[String]
```

We would like to automatically derive:
```scala
given MyCustomCodec[Name]
```
in the same way as for Newtype, but with validation applied during decoding.

Conceptually, this requires only:
- `String => Either[String, Name]` (validation)
- `Name => String` (already available via .value)

However, because `ValidatedNewtype` does not exposes a Prism / Refinement / PartialIso-like typeclass
there is no way to express this transformation generically in user code.

**Why exposing a validator alone is not enough**

Exposing an implicit `String => Either[String, Name]` would already enable decoding for many codecs. However, a number of schema compilers and infrastructure codecs require more than just validation:
- a total mapping `Name => String` (already available via `.value`)
- optionally an unsafe constructor for value generation, default values, or round-tripping scenarios

A good conceptual middle ground already exists in smithy4s in the form of `Refinement[A, B]`, which models exactly this relationship: a partial transformation in one direction, and a total one in the other.

However, Refinement currently requires a reified `Constraint` and an associated `ShapeTag[Constraint]`. `ValidatedNewtype` does not have access to such constraint metadata, and in practice it cannot reconstruct it from the generated code.

This makes `Refinement` too heavy-weight to be reused directly for `ValidatedNewtype`.

[...]

PR #1873, opened 2026-01-20
Add asSurjection to ValidatedNewtype

ghostbuster91:
First attempt for #1872 

- [ ] Added documentation
- [x] Updated changelog

### DIFF
--- a/modules/core/src-3/ValidatedNewtype.scala
+++ b/modules/core/src-3/ValidatedNewtype.scala
@@ -30,6 +30,12 @@ abstract class ValidatedNewtype[A] extends AbstractNewtype[A] { self =>
 
   extension (orig: Type) def value: A = orig
 
+  implicit val asSurjection: Surjection[A, Type] =
+    new ValidatedNewtype.Make[A, Type] {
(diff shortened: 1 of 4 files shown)


modules/core/src-3/ValidatedNewtype.scala
```diff
@@ -38,5 +44,5 @@ abstract class ValidatedNewtype[A] extends AbstractNewtype[A] { self =>
 }
 
 object ValidatedNewtype {
-  private[smithy4s] trait Make[A, B] extends Bijection[A, B]
+  private[smithy4s] trait Make[A, B] extends Surjection[A, B]
```
kubukoz, line 47: **question:** was this Make even being used anywhere? What do we even need it for?

ghostbuster91: AFAIK it wasn't used anywhere, I changed it like that to match how it was done in `Newtype` but can remove it if you want.

https://github.com/disneystreaming/smithy4s/blob/series/0.18/modules/core/src-2/Newtype.scala#L43

Baccata (approved):

kubukoz (approved):

PR #1878, opened 2026-01-27
enable xsource 3 for scala 2.13

### LINKED ISSUE #1584: Enable -Xsource:3

lewisjkl:

- [ ] Added documentation
- [ ] Updated changelog

closes #1584

### DIFF
--- a/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
+++ b/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
@@ -1445,14 +1445,14 @@ private[smithy4s] class SchemaVisitorJCodec(
     val label = field.label
 
     val decodeFn: (Cursor, JCodec[A], JsonReader) => A = {
-      val allowExplicitNulls = !{
-          // required fields can't accept explicit nulls
-          field.hints.has(Required) ||
-          // if there was no default, we'd allow explicit nulls by virtue of having an OptionSchema
-          default == null ||
-          // nullables have separate handling in OptionSchema
-          field.hints.has(alloy.Nullable)
-        }
+      val allowExplicitNulls = ! {
+        // required fields can't accept explicit nulls
+        field.hints.has(Required) ||
+        // if there was no default, we'd allow explicit nulls by virtue of having an OptionSchema
+        default == null ||
+        // nullables have separate handling in OptionSchema
+        field.hints.has(alloy.Nullable)
+      }
 
       if (allowExplicitNulls)
         (cursor, codec, in) =>
--- a/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
+++ b/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
@@ -1445,8 +1445,7 @@ private[smithy4s] class SchemaVisitorJCodec(
     val label = field.label
 
     val decodeFn: (Cursor, JCodec[A], JsonReader) => A = {
-      val allowExplicitNulls =
-        ! {
+      val allowExplicitNulls = !{
           // required fields can't accept explicit nulls
           field.hints.has(Required) ||
           // if there was no default, we'd allow explicit nulls by virtue of having an OptionSchema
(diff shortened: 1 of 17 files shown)


modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -1445,15 +1445,14 @@ private[smithy4s] class SchemaVisitorJCodec(
...
-          // required fields can't accept explicit nulls
-          field.hints.has(Required) ||
-          // if there was no default, we'd allow explicit nulls by virtue of having an OptionSchema
-          default == null ||
-          // nullables have separate handling in OptionSchema
-          field.hints.has(alloy.Nullable)
-        }
+      val allowExplicitNulls = ! {
+        // required fields can't accept explicit nulls
+        field.hints.has(Required) ||
```
kubukoz, line 1450: I'm a bit puzzled, because `@nullable @required` is a thing and it does support explicit nulls...

kubukoz: proof (in 0.18): https://kubukoz.github.io/smithy-transcoder/#N4Igtg9gJgpgNgFRgDwC4gFwgCQDcYBOAzgJYQB2GABADogBMdN5z5AhmDEQA5sDGMKrEjNmRVAQCufVJIKCAyhOmoqwZlSoABNnDgQAngGJykvWwBGcGBu3yAjpJLyoVdp2pKCJcgHNbWg5OLlRsvjDUAJLkqDDhBMwAvswgADQgAGYQBGBs6BigAFZEFJigGSTwUABiJHCxBJggACIAotUAggCqADIIIImJ6T7ckvkg6uSadO42mG5mcKm2dGFz1AAsjOSJaSD4xGTkmACMiUA

lewisjkl: Hmm yeah that is interesting. I'll have to check the json tests to see if we ever have test cases that indicate one way or the other. Maybe this comment is misleading or it is handled in another code path. Just on my mobile so I'm not sure of the broader context atm.

In any case this PR doesn't change the behavior, but I'll look into it a bit tomorrow

kubukoz: yeah it's probably handled on a higher level, in an OptionSchema's code path (as it'll be wrapping around a primitive and have the same hints)

project/Smithy4sBuildPlugin.scala
```diff
@@ -294,11 +298,9 @@ object Smithy4sBuildPlugin extends AutoPlugin {
       else if (priorTo2_13(scalaVersion))
         filterScala2_12Options(commonCompilerOptions)
       else
-        commonCompilerOptions
-    // ++ Seq(
-    //   "-Xsource:3",
-    //   "-P:kind-projector:underscore-placeholders"
```
Milyardo, line 300: According to the migration [guide](https://docs.scala-lang.org/scala3/guides/migration/plugin-kind-projector.html) this flag is necessary for syntax that will cross compile between 2.13 and 3.

lewisjkl: Hmm if that's the case then how is it compiling successfully now?

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
lewisjkl: Yeah it is the case that it is handled at a higher level. The comments here are basically saying that required fields without the nullable trait can't have explicit nulls and then if it IS in fact annotated with nullable then that is handled at a higher level.

yisraelU (approved):

PR #1879, opened 2026-01-29
Validators for collections

### LINKED ISSUE #1843: Extend renderValidatedNewtypes to Support Collection Types
Smithy4s currently supports validated newtypes for constrained primitive shapes when `smithy4sRenderValidatedNewtypes` metadata key is set to `true`.

This proposal extends that mechanism to also include collection types (`list` and `map`).

## Motivation

The existing `renderValidatedNewtypes` feature ensures that constructing a constrained primitive newtype (e.g., a `NonEmptyString`) performs validation at the point of creation, returning either a valid instance or an error.
This helps prevent invalid values from being introduced.

However, collection types remain unvalidated at construction time, even when they have constraints such as `@length`.
This means users can create invalid in-memory objects (e.g., `List.empty`) that only fail later during the decoding on the other side of the wire.

Extending `renderValidatedNewtypes` to collections would ensure that all constrained shapes behave consistently and benefit from the same safety guarantees.

## Implementation details

Similarly to validated newtypes it should be possible to enable this feature on particular shapes by using `validateNewtype` trait or globally via smithy the metadata key.

This would require adding maps and lists to the set of valid targets of `validateNewtype` trait.

**Examples:**

```smithy
use smithy4s.meta#validateNewtype

@validateNewtype
@length(min: 5)
list Names {
  member: String
}
```

```smithy

metadata smithy4sRenderValidatedNewtypes = true

@length(min: 5)
list Names {
  member: String
}
```

msosnicki:
# Overview

Closes #1843 

- [ ] Added documentation
- [ ] Updated changelog


modules/core/src/smithy4s/Validator.scala
```diff
@@ -16,55 +16,378 @@
...
+      ): Collection[Col, Elem] =
+        new CollectionImpl(
+          getIterator,
+          mainBuilder = mainBuilder
+            .map(_.validating(constraint))
+            .orElse(Some(simple.validating(constraint))),
+          memberRefinements = memberRefinements
+        )
 
-    override def toSchema(a: Schema[A]): Schema[B] = {
```
msosnicki, line 62: This is not correct in the current implementation. See for example how `ValidatedString` changed - the underlyingType contains schema with all refinements attached to it. We also have `validator` with the same schema information duplicated. Then we have implicit val schema = validator.toSchema(underlyingSchema). 
So the resulting schema will have refinements applied twice.
I removed that duplication, but that means that the Validator.toSchema` is just applying the outermost bijection, which in my opinion defeats it's purpose.

modules/core/src/smithy4s/Validator.scala
```diff
@@ -16,55 +16,378 @@
...
+      var acc: Either[String, Unit] = Right(())
+      while (it.hasNext && acc.isRight) {
+        acc = validate(it.next())
+      }
+      acc
+    }
+
+  }
+
+  @deprecated
```
msosnicki, line 379: todo: add details for all deprecations + update changelog

modules/core/src/smithy4s/Validator.scala
```diff
@@ -16,55 +16,378 @@
 
 package smithy4s
 
-sealed trait Validator[A, B] {
+sealed trait Validator[A, B] { self =>
   def validate(value: A): Either[String, B]
 
+  // todo: deprecate
   def toSchema(a: Schema[A]): Schema[B]
```
msosnicki, line 23: I think this method is a bit problematic, especially the more complex scenario there is. For example, see the code generated for `ValidatedRefinedListConstrainedMember`.

The `underlyingType` contains the full schema besides bijection, and I haven't changed that. 

So doing the `schema = validator.toSchema(underlyingSchema)` where the validator contains all the schema bindings kind of duplicated. There was a bug present before  (see [here](https://github.com/disneystreaming/smithy4s/pull/1879/changes#r2742058921)) that was wrapping schemas in refinements multiple times and after fixing it the validator is just applying the outermost bijection.

modules/core/src/smithy4s/Validator.scala
```diff
@@ -16,55 +16,378 @@
 
 package smithy4s
 
-sealed trait Validator[A, B] {
+sealed trait Validator[A, B] { self =>
```
msosnicki, line 19: I tried not to change the current API too much, but given the issues highlighted in the other comments, I think that having a simpler interface, just:

```scala
trait Validator[A] {
  def validate(a: A): Either[String, A]
}
```

and having a schema visitor `Schema ~> Validator` that works for any type would be much better. at least at first glance. We would get rid of the duplication of schema bindings that are basically the same in the `Schema` and `Validator` members, and this would open up possibility to validate any type, not just the ones that the `@validatedNewtype` is restricted to. 
Not sure how it would fit the 0.18 though

Baccata: It's possible that by trying to unifying constraints with type refinements, we've made things awkward / more complicated they'd have to be. 

Maybe targeting 0.19 directly and trying to separate the concepts might be a good idea. It'd also be good to get the Disney and SXM contributors together in a call and decide and whiteboard it

modules/core/src/smithy4s/Validator.scala
```diff
@@ -16,55 +16,380 @@
...
 
 object Validator {
 
-  def of[A, B](bijection: Bijection[A, B]): ValidatorBuilder[A, B] =
-    new ValidatorBuilder[A, B](bijection)
-
-  final class ValidatorBuilder[A, B] private[smithy4s] (
-      bijection: Bijection[A, B]
-  ) {
+  trait Builder[A, B] { self =>
```
Baccata: I'm not too sure about this, as the exposure of public traits feels like a bincompat issue waiting to happen, because it's seemingly adding a lot of api surface. 

I'd rather you tried to have compositional functions 

```scala
sealed trait Validator[A, B]{ ...} 
object Validator {
   def list[A, B](validator: Validator[A, B]) : Validator[List[A], List[B]] = ??? 
   def map[K0, K1, V0, V1](key : Validator[K0, K1], value: Validator[V0, V1]) = ??? 
}
```

... but then, I think this is a little awkward too, and I think it'd be simplified by having a clearer delimitation between constraints (which may lead to the creation of validated newtypes) and type-refinements as they currently stand (which generally target types that are external to the metamode)

msosnicki: I made it package private for `smithy4s` and everything still works, so it might help a bit.

I started without a builder, just the methods directly on `Validator`, but the API was very awkward, using the type evidences all over the place. So it was even worse - it would impact the whole `Validator` interface, which is really public.

I could try with the approach you're suggesting here, just passing the fully constructed validators for collections. Whenever I have some time I can try that out.

modules/core/src/smithy4s/Validator.scala
ghostbuster91: > It's possible that by trying to unifying constraints with type refinements, we've made things awkward / more complicated they'd have to be.

I agree this topic should be revisited in 0.19. However, this PR doesn't change that much when it comes to design while it still fixes some issues.

PR #1885, opened 2026-02-06
Update codegen of numbers when using dynamic bindings

a-morales:

When using dynamic bindings during the codegen any numbers specified in hints gets interpreted as a floating point number. This was causing an issue in which we have a library doing a roundtrip test of going from code generated output back to a smithy model in which we encountered this validation error from Smithy https://github.com/smithy-lang/smithy/blob/main/smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java#L252

The fix here is when generating the dynamic hints if there's any number to do a check if it's a floating point number or not and render the right thing. I'll leave some comments here on the relevant changes

- [ ] Added documentation
- [ ] Updated changelog


modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -1929,7 +1929,11 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
       def nullNode(x: NullNode): Line =
         line"smithy4s.Document.nullDoc"
       def numberNode(x: NumberNode): Line =
-        line"smithy4s.Document.fromDouble(${x.getValue.doubleValue()}d)"
+        if (x.isFloatingPointNumber()) {
+          line"smithy4s.Document.fromDouble(${x.getValue.doubleValue()}d)"
+        } else {
+          line"smithy4s.Document.fromLong(${x.getValue.longValue()})"
+        }
```
a-morales, line 1936: This is the actual change

sampleSpecs/dynamic_hint_bindings.smithy
```diff
@@ -13,7 +13,7 @@ structure testDynamicBinding {
 }
 
 @testDynamicBinding(str: "test")
-@thisWillBeDynamic(test: 101)
+@thisWillBeDynamic(test: 101, test2: 123.456)
```
a-morales, line 16: Updated this spec to make sure that it renders correctly depending on what's in the spec

modules/bootstrapped/src/generated/smithy4s/example/ShouldHaveDynamicBinding.scala
```diff
@@ -15,14 +15,14 @@ object ShouldHaveDynamicBinding extends ShapeTag.Companion[ShouldHaveDynamicBind
   val hints: Hints = Hints(
     Hints.dynamic(ShapeId("smithy.api", "since"), smithy4s.Document.fromString("1")),
     Hints.dynamic(ShapeId("smithy4s.example", "testDynamicBinding"), smithy4s.Document.obj("str" -> smithy4s.Document.fromString("test"))),
-    Hints.dynamic(ShapeId("smithy4s.example.dynamic_traits", "thisWillBeDynamic"), smithy4s.Document.obj("test" -> smithy4s.Document.fromDouble(101.0d))),
+    Hints.dynamic(ShapeId("smithy4s.example.dynamic_traits", "thisWillBeDynamic"), smithy4s.Document.obj("test" -> smithy4s.Document.fromLong(101), "test2" -> smithy4s.Document.fromDouble(123.456d))),
```
a-morales, line 18: This is the codegen output of the smithy spec I modified

lewisjkl (approved):

PR #1886, opened 2026-02-06
series/0.18: Update codegen of numbers when using dynamic bindings 

a-morales:
This is the same as #1885 but back ported to the 0.18 branch. 

- [ ] Added documentation
- [ ] Updated changelog


lewisjkl (approved):

lewisjkl:
The tests that are failing are ones that I ignored on the 0.19 branch, you can do the same here until we get new artifacts published on the newer sonatype

modules/codegen/test/src/smithy4s/codegen/internals/ModelLoaderSpec.scala
```diff
@@ -81,7 +81,9 @@ class ModelLoaderSpec extends FunSuite {
     // nothing failed
   }
 
-  test("ModelLoader can load a dependency from s01 if it has a + in the name") {
+  test(
+    "ModelLoader can load a dependency from s01 if it has a + in the name".ignore
```
kubukoz, line 85: I need to republish that library some day 😅

PR #1892, opened 2026-02-19
Mill1x support

yisraelU:
In effort to target Mill 1.x  , we need to support Scala 3 in the codegen module which must also support Scala 2.12 .

A bunch of syntax changes were made to be compatible with Scala 3  .
VarArgs expansion is the only syntax that was not supported in Scala 2 , and therefore a config was added to silence the warning in the Scalac options
 
- [x] Updated changelog


modules/bootstrapped/src/generated/com/amazonaws/dynamodb/DynamoDB.scala
```diff
@@ -64,7 +64,7 @@ object DynamoDBGen extends Service.Mixin[DynamoDBGen, DynamoDBOperation] {
...
...ed NoSQL database service that provides fast and\n      predictable performance with seamless scalability. DynamoDB lets you offload the\n      administrative burdens of operating and scaling a distributed database, so that you don't have\n      to worry about hardware provisioning, setup and configuration, replication, software patching,\n      or cluster scaling.</p>\n\n         <p>With DynamoDB, you can create database tables that can store and retrieve any amount of\n      data, and serve any level of request traffic. You can scale up or scale down your tables'\n      throughput capacity without downtime or performance degradation, and use the AWS Management\n      Console to monitor resource utilization and performance metrics.</p>\n\n         <p>DynamoDB automatically spreads the data and traffic for your tables over a sufficient\n      number of servers to handle your throughput and storage requirements, while maintaining\n      consistent and fast performance. All of your data is stored on solid state disks (SSDs) and\n      automatically replicated across multiple Availability Zones in an AWS region, providing\n      built-in high availability and data durability. </p>")),
```
yisraelU, line 67: we used to escape a single quote , which was not necessary

modules/bootstrapped/src/generated/smithy4s/example/OrderType.scala
```diff
@@ -38,7 +38,7 @@ object OrderType extends ShapeTag.Companion[OrderType] {
   val id: ShapeId = ShapeId("smithy4s.example", "OrderType")
 
   val hints: Hints = Hints(
-    Hints.dynamic(ShapeId("smithy.api", "documentation"), smithy4s.Document.fromString("Our order types have different ways to identify a product\nExcept for preview orders, these don\'t have an ID")),
```
yisraelU, line 41: ditto

modules/bootstrapped/src/generated/smithy4s/example/OrderType.scala
```diff
@@ -51,7 +51,7 @@ object OrderType extends ShapeTag.Companion[OrderType] {
     val id: ShapeId = ShapeId("smithy4s.example", "InStoreOrder")
 
     val hints: Hints = Hints(
-      Hints.dynamic(ShapeId("smithy.api", "documentation"), smithy4s.Document.fromString("For an InStoreOrder a location ID isn\'t needed")),
```
yisraelU, line 54: ditto

modules/codegen/src-3/smithy4s/codegen/internal/MapTypeCompanionPlatform.scala
```diff
@@ -0,0 +1,23 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package smithy4s.codegen.internals
+
+import LineSegment.NameRef
+
+trait MapTypeCompanionPlatform {
+  val seqMapRef = NameRef("scala.collection.immutable.SeqMap")
```
yisraelU, line 22: same as scala 2 , just for cross compilation purposes

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -1949,10 +1951,25 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
     })
   }
 
+  private def escapeForStringLiteral(raw: String): String = {
```
yisraelU, line 1954: This replaced the Scala 2 version of storing source code string literals without using reflection.

project/Smithy4sBuildPlugin.scala
```diff
@@ -123,7 +178,8 @@ object Smithy4sBuildPlugin extends AutoPlugin {
 
     def millPlatforms(
         scalaVersion: String,
-        millVersions: Seq[String]
+        millVersions: Seq[String],
+        deps: ProjectMatrix*
```
yisraelU, line 182: Since sbt-projectmatrix uses exact Scala version matching for .dependsOn(), the Mill 1.x row (3.8.1) can't find a matching codegen row (3.3.6). 
To  work around this, millPlatforms now accepts project matrix dependencies explicitly and wires them per-row using depScalaVersion

lewisjkl (commented):
Looks great, I think we should also update the documentation mill example to show how to setup for mill 1.x now as well

modules/docs/resources/markdown/01-overview/03-installation.md
```diff
@@ -101,3 +103,39 @@ object example extends ScalaModule with Smithy4sModule {
...
+
+### Mill 1.x
+
+In your `build.mill`:
+
+```scala
+package build
+
+import mill._, mill.scalalib._
+import $ivy.`com.disneystreaming.smithy4s::smithy4s-mill-codegen-plugin_mill1:@VERSION@`
```
lewisjkl:
```suggestion
import $ivy.`com.disneystreaming.smithy4s::smithy4s-mill-codegen-plugin::@VERSION@`
```

I'd assume it will work like this now, but maybe not?

yisraelU: yes , you're right

modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -1881,42 +1881,44 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
...
             .getNano()}, scala.concurrent.duration.Duration(${time.getOffset().getTotalSeconds()}, "seconds"))"""
       case Primitive.Duration =>
-        duration => line"$duration_(${renderStringLiteral(duration.toString)})"
-      case Primitive.Document => renderNodeToLine(_)
-      case Primitive.Nothing  => v => (v: Nothing) // this case can't happen
-    }
+        (duration: java.time.Duration) => line"$duration_(${renderStringLiteral(duration.toString)})"
+      case Primitive.Document => (node: Node) => renderNodeToLine(node)
+      case Primitive.Nothing  => (_: Any) => sys.error("unreachable") // this case can't happen
+    }).asInstanceOf[T => Line]
```
rochala, line 1921: This could be simplified by migrating away from Aux pattern of Primitive to normal parameterized trait like:
```scala
private[internals] sealed trait Primitive[T]
private[internals] object Primitive {
  case object Unit extends Primitive[Unit]
  case object Blob extends Primitive[Array[Byte]]
```

Then Scala 3 typechecker will be satisfied with
```scala
  private def renderPrimitive[T](prim: Primitive[T]): T => Line =
    prim match {
      case Primitive.BigDecimal =>
        bd => line"scala.math.BigDecimal(${bd.toString})"
      case Primitive.BigInteger => bi => line"scala.math.BigInt(${bi.toString})"
      case Primitive.Unit       => _ => line"()"
      case Primitive.Double     => t => line"${t.toString}d"
      case Primitive.Float      => t => line"${t.toString}f"
      case Primitive.Long       => t => line"${t.toString}L"
      case Primitive.Int        => t => line"${t.toString}"
      case Primitive.Short      => t => line"${t.toString}"
      case Primitive.Bool       => t => line"${t.toString}"
      case Primitive.Uuid       => uuid => line"java.util.UUID.fromString(${renderStringLiteral(uuid.toString)})"
      case Primitive.String     => s => renderStringLiteral(s)
      case Primitive.Byte       => b => line"${b.toString}"
      case Primitive.Blob =>
        ba =>
          val blob = NameRef("smithy4s", "Blob")
          if (ba.isEmpty) line"$blob.empty"
          else
            line"$blob(Array[Byte](${ba.mkString(", ")}))"
      case Primitive.Timestamp =>
        ts => line"${NameRef("smithy4s", "Timestamp")}(${ts.getEpochSecond()}L, ${ts.getNano()})"
      case Primitive.Document => renderNodeToLine(_)
      case Primitive.Nothing  => v => (v: Nothing)
    }
```

But confirm me here, I may be wrong why Aux pattern was used here in the first place.

lewisjkl: Thanks for the input. Yeah it might be possible to move this direction, I'm not entirely sure. The reason it was done with Aux originally iirc has to do with the usage of recursion schemes in `SmithyToIr` https://github.com/disneystreaming/smithy4s/blob/ec25ae8f705ad4f9085411a8d496d193424bf880/modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala#L1400 and `TypedNode`. So I think it would be a larger refactor if we didn't have a GADT for primitive

lewisjkl (approved):
We can leave the PR open for the weekend in case anyone else wants to review, but I think it's good to go 🎉

rochala:
Hey, thanks for implementing support for mill 1.0. Are there any plans to backport this to 0.18 branch ? I'm happy to do the PR if you're willing to merge it.

PR #1893, opened 2026-02-23
Add type params to `struct` and `union` in codegen schemas

### LINKED ISSUE #1890: Generated code compilation is significantly slower in Scala 3
Hi 👋 I'm working on a codebase that generates code from a large number of smithy files which is cross-compiled to both Scala 2.13 and 3.3.

One "surprising" aspect is that the codegen modules takes ~10 seconds to compile in Scala 2.13 but over a minute in Scala 3 with the same number of source files.

I know this is likely an issue with the Scala 3 compiler, but I believe there is a way to significantly improve the compilation time of the generated code with a minor tweek to the codegen.

After profiling the compilation using `"-Yprofile-enabled", "-Yprofile-trace:/path/to/compiler.trace"`, it seems that a very significant amount of time is spent typechecking the `implicit val schema: Schema[Foo] = ???` values (average 60ms per val!!!).

Image

This seems to be an issue only for `struct` schemas in the companion object, but tbh I didn't look too much into it to see if it might be an issue with other ones as well (`union`, etc.)

```scala
implicit val schema: Schema[Foo] = struct(???)
```

What I found interesting is that the compilation time wasn't really affected by the number of arguments to `struct`, so I tried adding the type parameter explicitly (e.g., `struct[Foo](???)`) and that improved the compilation time for the `schema` val massively:

Image

I believe that adding the type parameter to the generated code should be a fairly trivial and safe change. I'd be happy to create a PR if you think agree with making this change and can point me to the right direction :)

kyri-petrou:
/fixes #1890

This PR adds type params for the partially applied schemas `struct` and `union` in generated code to improve compilation performance in Scala 3 (potentially for Scala 2 but I haven't checked). See #1890 for more info

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/modules/bootstrapped/src/generated/com/amazonaws/dynamodb/DescribeEndpointsResponse.scala
+++ b/modules/bootstrapped/src/generated/com/amazonaws/dynamodb/DescribeEndpointsResponse.scala
@@ -19,7 +19,7 @@ object DescribeEndpointsResponse extends ShapeTag.Companion[DescribeEndpointsRes
   // constructor using the original order from the spec
   private def make(endpoints: List[Endpoint]): DescribeEndpointsResponse = DescribeEndpointsResponse(endpoints)
 
-  implicit val schema: Schema[DescribeEndpointsResponse] = struct(
+  implicit val schema: Schema[DescribeEndpointsResponse] = struct[DescribeEndpointsResponse](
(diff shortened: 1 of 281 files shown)


kyri-petrou:
~~Agh seems that I need to make some more changes to the tests. I'll change the PR status to Draft until I fix those~~

OK tests are all passing now

yisraelU (approved):
Nice, Thank you

PR #1900, opened 2026-03-04
set up bincompat-friendly core codegen

### LINKED ISSUE #1777: Bincompat-friendly core
In 0.19, we should futureproof ourselves against changes in smithy.api / alloy that would normally make generated code binary incompatible - by using bincompat-friendly mode on them.

This means using the bincompat-friendly mode from #1500 / #1737 on smithy.api and alloy namespaces (anything else?).

lewisjkl:
Generate the smithy and alloy classes with bincompat friendly codegen enabled. Basically a port of #1776 with some changes to handle open enums like `TimestampFormat`. In these cases, I opted to throw exceptions off of the hot-path such that they'd be triggered at runtime, but during application start up. This is obviously not "pure" but personally I think that is a fine trade off given that it is very much an edge case. I don't think it makes sense to change / encode the possibility of this failure at a higher level which would add complexity with basically no benefit.

This also provides a metadata key for configuring namespaces that should be generated with bincompat friendly mode in addition to the existing trait-based approach.

Let me know what you think.

---

Closes #1777.

### DIFF
--- a/modules/core/src/smithy4s/time/Timestamp.scala
+++ b/modules/core/src/smithy4s/time/Timestamp.scala
@@ -31,11 +31,22 @@ case class Timestamp private (epochSecond: Long, nano: Int)
     diff > 0 || diff == 0 && nano > other.nano
   }
 
-  def format(format: TimestampFormat): String = format match {
-    case TimestampFormat.DATE_TIME     => formatToString(0)
-    case TimestampFormat.EPOCH_SECONDS => formatEpochSeconds
-    case TimestampFormat.HTTP_DATE     => formatToString(1)
+  /**
+   * Formats the Timestamp as a String using the given TimestampFormat
+   * 
+   * Returns an Option value since an unsupported TimestampFormat could be provided.
+   * 
+   * Supported formats are `DATE_TIME`, `EPOCH_SECONDS`, and `HTTP_DATE`
+   */
+  def format(format: TimestampFormat): Option[String] = format match {
+    case TimestampFormat.DATE_TIME     => Some(formatToString(0))
+    case TimestampFormat.EPOCH_SECONDS => Some(formatEpochSeconds)
+    case TimestampFormat.HTTP_DATE     => Some(formatToString(1))
+    case _                             => None
   }
+
+  def formatDateTime: String = formatToString(0)
+  def formatHttpDate: String = formatToString(1)
 
   def conciseDateTime: String = formatToString(3)
 
@@ -51,7 +62,9 @@ case class Timestamp private (epochSecond: Long, nano: Int)
     */
   def truncateToSeconds: Timestamp = copy(nano = 0)
 
-  override def toString: String = format(TimestampFormat.DATE_TIME)
+  override def toString: String = format(
+    TimestampFormat.DATE_TIME
+  ).get // we know `DATE_TIME` is a supported format
 
   private[this] def formatToString(internalFormat: Int): String = {
     val s = new java.lang.StringBuilder(32)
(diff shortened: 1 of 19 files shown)


modules/core/src/smithy4s/time/Timestamp.scala
```diff
@@ -31,12 +31,23 @@ case class Timestamp private (epochSecond: Long, nano: Int)
...
-    case TimestampFormat.EPOCH_SECONDS => formatEpochSeconds
-    case TimestampFormat.HTTP_DATE     => formatToString(1)
+  /**
+   * Formats the Timestamp as a String using the given TimestampFormat
+   * 
+   * Returns an Option value since an unsupported TimestampFormat could be provided.
+   * 
+   * Supported formats are `DATE_TIME`, `EPOCH_SECONDS`, and `HTTP_DATE`
+   */
+  def format(format: TimestampFormat): Option[String] = format match {
```
kubukoz, line 41: ooh controversial. But I'm ok with this

modules/docs/resources/markdown/04-codegen/01-customisation/16-binary-compatibility.md
```diff
@@ -88,6 +88,14 @@ structure Hello {
 
 This will change the structure of generated code so that it's possible to evolve in a bincompat safe way.
 
+An alternative way of configuring `bincompatFriendly` codegen is to configure the `smithy4sBinCompatNamespacePatterns` metadata key in your Smithy files.
+
+```smithy
+metadata smithy4sBinCompatNamespacePatterns = ["foo", "foo.*"]
+```
+
+The above metadata configuration will configure Smithy4s to generate `bincompatFriendly` for all shapes with ShapeIds starting with `foo#` or `foo.`.
```
kubukoz:
```suggestion
The above metadata configuration will configure Smithy4s to generate all shapes with ShapeIds starting with `foo#` or `foo.`, in bincompat-friendly mode, if possible.
```

since not all shapes even have a BCF mode. Like services.

modules/core/src/smithy4s/time/Timestamp.scala
lewisjkl: Yeah I'm kind of thinking it should be the responsibility of each codec to go to the correct format when it already has that information (which was already the case in quite a few spots as you'll see). When it doesn't have that information, it can choose to deal with the possibility of an unknown format

modules/docs/resources/markdown/04-codegen/01-customisation/16-binary-compatibility.md
kubukoz: actually, now I'm kinda worried that this will be a way to circumvent critical validation - especially for end users. Like, you can't use `@bincompatFriendly` on an adt union.

lewisjkl: Ah that's a good point I hadn't thought of. We could remove this option and then it will just force the use of a model transformer to add more namespaces without extra manual work. That way things will still be validated after the transformer..

kubukoz: yeah, that's how I've been doing it in https://github.com/polyvariant/smithy4s-bsp/blob/531ebb60261f62f766d6ce9d1680a12bc1cf887f/transformation/src/main/scala/MakeBincompatFriendly.scala - it tries to match only the shapes that are valid, though it's a bit duplicative in its guesses

lewisjkl: Nice, thanks for sharing 🙏

I'll update the PR to remove the metadata config tomorrow

a-morales (approved):

lewisjkl:
Going to merge this one since I think I've addressed all the feedback and we are still in milestone release land anyway. If you have any other thoughts on this one Jakub I'm happy to address in a follow up PR

kubukoz:
no it's alright 👍 sorry for the late response

PR #1903, opened 2026-03-06
work on compatibility with sbt 2.x

yisraelU:
The goal of this PR is to lay the foundations for **dual** support of both SBT 1 and 2 for our Smithy4s plugin.                                                                                                                                                                                                                                                                                                                                                                                                                          It cross-compiles the sbt codegen plugin for sbt 2 (2.0.0-RC9), alongside the existing sbt 1 build.                                                                                                                                                                                         

  Plugin cross-build:
  - codegenPlugin now builds two variants via sbt-projectmatrix: one for sbt 1 (Scala 2.12) and one for sbt 2 (Scala 3.8)
  - pluginCrossBuild / sbtVersion is set to 1.12.4 or 2.0.0-RC9 based on the Scala binary version
  - Both variants depend on sbt2-compat for API compatibility between sbt 1 and sbt 2
  - Conflicting transitive dependencies (scala-collection-compat_2.13, scala-xml_2.13) are excluded when building the Scala 3 variant. These exclusions are reflected in the published POM, so consumers should not need to add them manually. 

  Plugin source compatibility:
  - Smithy4sCodegenPlugin.scala compiles under both Scala 2.12 and Scala 3 using sbtcompat.PluginCompat for type conversions (File vs HashedVirtualFileRef, Settings type alias, Def.uncached)
  - Platform-specific code lives in src-2/ and src-3/ directories (Compat.scala)

  Scripted tests:
  - All 24 tests now pass under both sbt 1 and sbt 2
  - updated all file checks to use globs to handle both scala 2 and 3
  - update-lsp-config uses a ${SRC_MANAGED} placeholder in expected.json instead of a hardcoded path
  - transformation-trait-classes and validator-trait-classes use a cross-compatible packageBin conversion pattern for smithy4sAllDependenciesAsJars (sbt 1 returns File, sbt 2 returns HashedVirtualFileRef)

- [ ] Updated changelog


modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/project/plugins.sbt
```diff
@@ -7,3 +7,14 @@ sys.props.get("plugin.version") match {
          |Specify this property using the scriptedLaunchOpts -D.""".stripMargin
     )
 }
+
+// When running under sbt 2 (Scala 3), exclude Scala 2.13 variants that
```
yisraelU, line 11: This is repeated for all scripted tests , as is required for Scala 3

kubukoz:
Can we please target 0.18? We can forward-port to 0.19, but non-breaking changes should still be available to 0.18 users

modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/project/plugins.sbt
kubukoz: does this mean users will have to do that too?

modules/codegen-plugin/src/sbt-test/codegen-plugin/multimodule-aws/test
```diff
@@ -1,5 +1,4 @@
 # check if smithy4sCodegen works with libraries that were built with Smithy4s
 > show bar/smithy4sAllExternalDependencies
 > compile
-$ exists foo/target/scala-2.13/src_managed/main/smithy4s/foo/Lambda.scala
-$ absent bar/target/scala-2.13/src_managed/main/smithy4s/foo/Lambda.scala
```
kubukoz, line 5: Removing these makes me uneasy - these `exists`/`absent` tests are a bit low-level, clearly, but they're essential to some of these tests checking whether third-library-generated code is regenerated, or whether caching works as intended, etc.

I'd appreciate it if we can find a different strategy. If we can use globs here, like `$ exists foo/**/Lambda.scala`, that'd be great. If we can't, maybe we can have a custom task (in a shared plugin meant only for testing?) that does it in a version-independent way?

I wonder what the sbt maintainers would suggest

yisraelU: we cant use globs . SBT 2 will support globs but not SBT 1 . I'm in middle of investigating whta the maintainers are thinking cause this would be an issue for anyone that maintains sbt plugin tests

yisraelU:
> Can we please target 0.18? We can forward-port to 0.19, but non-breaking changes should still be available to 0.18 users

sure , I expected SBT 2 release to be after 0.19 release , hence I thought focus on 0.19 and perhaps backport to 0.18 , if necessary . But I can switch that up

modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/project/plugins.sbt
yisraelU: No , the issue is caused because we are using Sbt 1 to run tests for an Sbt plugin written in Scala 3.
This requires exclusion from the plugin classpath , not the dependencies for the plugin .  A user would be using Sbt 2 or the Scala 2.12 variant

modules/codegen-plugin/src/sbt-test/codegen-plugin/multimodule-aws/test
yisraelU: scratch that . I made a mistake will switch to globs

kubukoz: Amazing!

kubukoz:
@yisraelU I think we can publish with a milestone - we have some plugins at work based on smithy4s that might want to be ready for sbt2 too, so this would be useful

yisraelU:
@kubukoz That's fine with me. Is it ready as is to merge into 0.19?
Or did you have something else in mind

kubukoz:
I'll have another look later, only skimmed last time

modules/codegen-plugin/src/sbt-test/codegen-plugin/aws-newtype-flatten/project/plugins.sbt
kubukoz: I see, that's okay then 👍

kubukoz (approved):

PR #1821, opened 2025-08-29
Update Scala Native to 0.5.x

majk-p:
## TODOS
- [x] rebase on top of https://github.com/disneystreaming/smithy4s/pull/1823
- [x] http4s needs SN 0.5.x version (https://github.com/http4s/http4s/pull/7708)
- [x] [fs2-data](https://github.com/gnieh/fs2-data/) needs SN 0.5.x version (https://github.com/gnieh/fs2-data/pull/704) (for now xml module is disabled)
- [x] [weaver-test #189](https://github.com/typelevel/weaver-test/pull/189) needs publishing - published in https://github.com/typelevel/weaver-test/actions/runs/17407116140/job/49414856635
- [x] `CryptoPlatformCompat` needs adjusting to 0.5.x

PR #1916, opened 2026-04-13
Fix smithy4s-protobuf Scala.js dependency resolution

### LINKED ISSUE #1915: smithy4s-protobuf Scala.js artifact pulls in JVM-only dependencies since 0.18.51
In 0.18.51, the `smithy4s-protobuf` Scala.js artifact (`smithy4s-protobuf_sjs1_3`) pulls in JVM-only dependencies, causing Scala.js linker failures.

## Root cause

PR #1821 changed the non-JVM protobuf dependency in `build.sbt` from:

```scala
"com.thesamet.scalapb" %%% "protobuf-runtime-scala" % "0.8.14"
```

to:

```scala
"com.thesamet.scalapb" %% "scalapb-runtime" % scalapb.compiler.Version.scalapbVersion
```

The `%%%` → `%%` change means the Scala.js artifact now depends on the JVM variant of `scalapb-runtime`, which transitively pulls in `com.google.protobuf:protobuf-java` — a JVM-only library with no Scala.js equivalent.

## Comparison

**0.18.50** (working) — 9 transitive deps, all JS-compatible:
- `protobuf-runtime-scala_sjs1_3:0.8.14` (pure Scala, works on JS)

**0.18.51** (broken) — 13 transitive deps, several JVM-only:
- `com.google.protobuf:protobuf-java:3.19.6` (JVM only)
- `scalapb-runtime_3:0.11.17` (JVM, not `_sjs1_3`)
- `lenses_3:0.11.17` (JVM, not `_sjs1_3`)

## Error

The Scala.js linker fails with:

```
[error] Referring to non-existent class com.google.protobuf.CodedOutputStream
```

## Suggested fix

Use `%%%` instead of `%%` for the `scalapb-runtime` dependency in the non-JVM branch, or revert to `protobuf-runtime-scala` for JS/Native targets.

kubukoz:
## Summary

Use `%%%` instead of `%%` for `scalapb-runtime` in the non-JVM protobuf dependency branch, so that the Scala.js artifact (`scalapb-runtime_sjs1_3`) is resolved instead of the JVM-only one (`scalapb-runtime_3`), which transitively pulls in `protobuf-java` and breaks the Scala.js linker.

This was introduced in #1821 when the dependency was changed from `protobuf-runtime-scala` to `scalapb-runtime` but the `%%%` operator was accidentally changed to `%%`.

### Before (broken — 0.18.51)

The published POM for `smithy4s-protobuf_sjs1_3` depends on:
- `scalapb-runtime_3` (JVM only)
- `protobuf-java` (JVM only)
- `lenses_3` (JVM only)

### After (fixed)

The published POM depends on:
- `scalapb-runtime_sjs1_3` (Scala.js compatible)

Fixes #1915

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -832,7 +832,7 @@ lazy val protobuf = projectMatrix
         )
       else
         Seq(
-          "com.thesamet.scalapb" %% "scalapb-runtime" % scalapb.compiler.Version.scalapbVersion
+          "com.thesamet.scalapb" %%% "scalapb-runtime" % scalapb.compiler.Version.scalapbVersion
         )
     },
     Test / fork := virtualAxes.value.contains(VirtualAxis.jvm)


kubukoz:
something's up with `BincompatCodegenIntegrationSpec`...

PR #1920, opened 2026-04-19
Support proper smithy AWS models

### LINKED ISSUE #1752: Adjust to AWS now publishing proper Smithy models
https://aws.amazon.com/blogs/aws/introducing-aws-api-models-and-publicly-available-resources-for-aws-api-definitions/

This involves:

- deprecating https://github.com/disneystreaming/aws-sdk-smithy-specs/ (this can be done by maintainers later)
- moving its build-time transformations to something that we can run as part of smithy4s pre-codegen transformations
  - this would look somewhat similar to https://github.com/disneystreaming/smithy4s/blob/series/0.19/modules/codegen/src/smithy4s/codegen/transformers/AwsConstraintsRemover.scala, but be 
- documentation updates, build tool plugin updates - https://disneystreaming.github.io/smithy4s/docs/protocols/aws/aws/ , https://github.com/disneystreaming/smithy4s/blob/8d0ace7958ec325d881e2d4e85b3f7cd0d8222f8/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala and corresponding implementation code / Mill code

Make sure to test this end to end. Might not be a bad idea to have a separate repository just for some testing (and later maybe we'll do some publishing of compiled code as well, though that's always been controversial so we'll see)

majk-p:
Resolves https://github.com/disneystreaming/smithy4s/issues/1752

- [x] Added documentation
- [ ] Updated changelog


modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -160,11 +160,11 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
 
     val smithy4sAwsSpecsVersion =
       settingKey[String](
-        "Known version of the AWS specifications, produced by https://github.com/disneystreaming/aws-sdk-smithy-specs"
+        "Version of the AWS models BOM (software.amazon.api.models:all) used to discover service versions"
       )
 
     val smithy4sAwsSpecs =
-      settingKey[Seq[String]]("Aws modules to load")
+      settingKey[Seq[(String, String)]]("Aws modules to load")
```
kubukoz: issue: I'm afraid we'll need to make a new key for backward compat

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -160,6 +160,27 @@ private[codegen] class SmithyToIR(
   private val smithy4sBinCompatHintNamespacePatterns: Set[NamespacePattern] =
     smithy4sDefaultBinCompatHintNamespacePatterns
 
+  private val stdlibBincompatAddedShapes: Map[ShapeId, Hint.BincompatAdded] =
```
kubukoz, line 163: What do we need this for?

modules/protocol/resources/META-INF/smithy/smithy4s.meta.smithy
```diff
@@ -8,6 +8,13 @@ metadata suppressions = [
     }
 ]
 
+metadata smithy4sBincompatPreludeAdditions = [
```
kubukoz, line 11: suggestion: that seems like something we can replace with `apply cors$origins @bincompatAdded("0.19.1")` if needed. Is it a new field in smithy 1.69.0?

project/AwsBoilerplate.scala
```diff
@@ -88,4 +273,11 @@ object AwsBoilerplate {
     }
   }
 
+  private def jsonEscape(s: String): String =
```
kubukoz: can we use some json library in scope (sjsonnew, I guess...) instead? If smithy-model is in scope, we could use `Node.prettyPrintJson` and whatnot

project/AwsBoilerplate.scala
```diff
@@ -1,84 +1,269 @@
...
+    */
+  private def resolveServiceJar(
+      name: String,
+      version: String
+  ): Option[java.io.File] = {
+    try {
+      val jars = coursierapi.Fetch
+        .create()
+        .addDependencies(
+          coursierapi.Dependency.of("software.amazon.api.models", name, version)
```
kubukoz:
```suggestion
          coursierapi.Dependency.of(AwsModels.org, name, version)
```

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
majk-p: We bumped Smithy from 1.66.0 to 1.69.0, which added a new origins member to the `smithy.api#cors` prelude structure. Since we can't apply `@bincompatAdded` to shapes we don't own, we use model metadata to mark it. This prevents the new field from landing in the base constructor and breaking binary compat with code compiled against the previous Smithy version. I'm not sure this is the best way to do it, didn't have better ideas.

kubukoz: hmm okay, I think this might not be as bad as it sounded. "bincompatFriendly" is kind of added implicitly through metadata, so it makes sense that we'd add `bincompatAdded` through metadata too.

modules/protocol/resources/META-INF/smithy/smithy4s.meta.smithy
kubukoz: answering myself: no, because it requires `@bincompatFriendly` on a parent and we don't have one - since the prelude traits are bincompat-friendly by namespace

modules/protocol/resources/META-INF/smithy/smithy4s.meta.smithy
```diff
@@ -8,6 +8,13 @@ metadata suppressions = [
...
+    {
+        shape: "smithy.api#cors$origins"
+        version: "0.19.1"
```
kubukoz, line 14: I pondered for a second about whether this should be our version or the smithy-model version, but I think ours makes more sense. 👍

project/Dependencies.scala
```diff
@@ -168,11 +168,10 @@ object Dependencies {
...
   }
 
-  object AwsSpecSummary {
-    val org = "com.disneystreaming.smithy"
-    val name = "aws-spec-summary"
-    val awsSpecSummaryVersion = "2025.04.08"
-    val value = org % name % awsSpecSummaryVersion
+  object AwsModels {
+    val org = "software.amazon.api.models"
+    val bomVersion = "1.0.206"
```
majk-p: Newer version of this model are bugged and reference non-existing dependencies. I've raised that with AWS support.

modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala
```diff
@@ -160,11 +160,18 @@ object Smithy4sCodegenPlugin extends AutoPlugin {
...
     val smithy4sAwsSpecsVersion =
       settingKey[String](
-        "Known version of the AWS specifications, produced by https://github.com/disneystreaming/aws-sdk-smithy-specs"
+        "Version of the AWS models BOM (software.amazon.api.models:all) used to discover service versions"
       )
 
     val smithy4sAwsSpecs =
-      settingKey[Seq[String]]("Aws modules to load")
+      settingKey[Seq[String]](
+        "Aws modules to load (legacy, uses smithy4sAwsSpecsVersion)"
```
kubukoz, line 168: we can add `@deprecated` to additionally discourage use.

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -160,6 +160,27 @@ private[codegen] class SmithyToIR(
...
+      .getMetadata()
+      .asScala
+      .get("smithy4sBincompatPreludeAdditions")
+      .toSet
+      .flatMap((n: Node) => n.asArrayNode().asScala)
+      .flatMap(_.getElements().asScala)
+      .flatMap { (n: Node) =>
+        n.asObjectNode().asScala.flatMap { obj =>
+          for {
+            shapeIdStr <- obj.getStringMember("shape").asScala
```
kubukoz, line 174: suggestion: if one of those keys is missing, I'd rather fail loudly than ignore that

modules/mill-codegen-plugin/src-mill-1/smithy4s/codegen/mill/Smithy4sModule.scala
```diff
@@ -118,18 +118,26 @@ trait Smithy4sModule extends ScalaModule {
       }
   }
 
+  def smithy4sAwsSpecsVersion: T[String] = Task {
+    AWS.bomVersion
+  }
+
   def smithy4sAwsSpecs: T[Seq[String]] = Task {
```
kubukoz, line 125: same as in sbt, let's deprecate

modules/mill-codegen-plugin/src-mill-1/smithy4s/codegen/mill/Smithy4sModule.scala
```diff
@@ -118,18 +118,26 @@ trait Smithy4sModule extends ScalaModule {
       }
   }
 
+  def smithy4sAwsSpecsVersion: T[String] = Task {
```
kubukoz, line 121: do we need this here if it's already in `src-mill-shared`?

kubukoz (commented):
I'd like to see this in action somewhere, other than that lgtm. Maybe a scripted test in one of the build tools modules?

kubukoz:
snapshot on its way: https://github.com/disneystreaming/smithy4s/actions/runs/24889761074/job/72879428297

kubukoz (approved):

PR #1933, opened 2026-04-28
Render Maven Central in smithy-build.json when present in resolvers

### LINKED ISSUE #1631: `smithy4sUpdateLSPConfig` should render Maven Central if it's in `resolvers`
If you have any custom `resolvers` added in your build, then they'll be the only ones rendered into `smithy-build.json`. However, the default behavior for an empty resolver list is to use Maven Central (and probably m2 local), and using a different list removes that default.

It looks like our task deliberately doesn't write default resolvers, so as a result, adding a custom resolver means the LSP / smithy-cli won't use Maven Central anymore. I think we should be generating that repo, at least in situations where there are customer resolvers present.

kubukoz:
Previously, smithy4sUpdateLSPConfig (sbt) and LSP/updateConfig (mill) explicitly filtered out repo1.maven.org from the rendered repositories. This meant that adding a custom resolver removed the Maven Central fallback that smithy-cli would otherwise use, since populating the list disables the default.

Now Maven Central is rendered when it's present in the resolver list, and omitted when the user has deliberately removed it.

Fixes #1631

- [ ] Added documentation
- [ ] Updated changelog

### DIFF
--- a/modules/codegen-plugin/src/smithy4s/codegen/GenerateSmithyBuild.scala
+++ b/modules/codegen-plugin/src/smithy4s/codegen/GenerateSmithyBuild.scala
@@ -123,7 +123,7 @@ private[codegen] object GenerateSmithyBuild {
       .foldLeft(ListSet.empty[String])(_ + _)
 
   private val prepareResolvers: PartialFunction[Resolver, String] = {
-    case mr: MavenRepository if !mr.root.contains("repo1.maven.org") => mr.root
+    case mr: MavenRepository => mr.root
   }
 
   private def prepareInputDirs(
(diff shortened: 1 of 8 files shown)


modules/codegen-plugin/src/smithy4s/codegen/GenerateSmithyBuild.scala
```diff
@@ -123,7 +123,7 @@ private[codegen] object GenerateSmithyBuild {
       .foldLeft(ListSet.empty[String])(_ + _)
 
   private val prepareResolvers: PartialFunction[Resolver, String] = {
-    case mr: MavenRepository if !mr.root.contains("repo1.maven.org") => mr.root
+    case mr: MavenRepository => mr.root
```
kubukoz, line 126: hmm this doesn't quite work, apparently Maven Central is in `externalResolvers`. And that one is a Task...

PR #1934, opened 2026-04-28
Add `@unpackedOutput`

kubukoz:

- [x] Added documentation
- [x] Updated changelog


modules/codegen/src/smithy4s/codegen/internals/Renderer.scala
```diff
@@ -616,7 +616,12 @@ private[internals] class Renderer(compilationUnit: CompilationUnit) { self =>
         indent(
           line".withInput(${op.input.schemaRef})",
           Option(op.errors).filter(_.nonEmpty).as(line".withError(${opErrorDef}.errorSchema)"),
-          line".withOutput(${op.output.schemaRef})",
+          op.unpackedOutput match {
+            case Some(uo) =>
+              line".withOutput($Schema_.bijection(${op.output.schemaRef}, (_: ${op.output}).${uo.fieldName}, (x: ${op.renderEffectiveOutput}) => ${op.output}(${uo.fieldName} = x)))"
```
kubukoz:
```suggestion
              line".withOutput($Schema_.bijection(${op.output.schemaRef}, (_: ${op.output}).${uo.fieldName}, ${op.output}(_))"
```

modules/docs/resources/markdown/04-codegen/01-customisation/18-validated-newtypes.md
kubukoz, line 1: Renamed these two because their numbers were colliding with other pages

modules/protocol/src/smithy4s/meta/validation/UnpackedOutputTraitValidator.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * Operations marked with `@unpackedOutput` must have an output structure
+ * containing exactly one member, so that the generated method can return
+ * the inner member's type.
+ */
+public final class UnpackedOutputTraitValidator extends AbstractValidator {
```
lewisjkl, line 37: Would be good to have some tests for this I think, despite it being quite straightforward

lewisjkl (commented):
This is cool, could be nice in the future to have a built-in transformer that would add the trait wherever `httpPayload` trait is present

kubukoz:
It wouldn't really work, as long as there are more outputs. I think it's better to leave this explicit... even if we added it automatically for single field outputs, adding another field would be a more invasive change for the clients

kubukoz:
I'm wondering if it'd be acceptable / just as okay to have the trait live on the output member:

```
operation Foo {
  output := { @required @packedOutput result: String }
}
```

this would at least be easier to validate for us (just use the `structurallyExclusive` property on the trait), no java 😅 

it could still be applied externally with `apply FooOutput$result @packedOutput`.

Compare to now:

```
@packedOutput
operation Foo {
  output := { @required result: String }
}

//or
apply Foo @packedOutput
```

cc @bpholt

lewisjkl:
> It wouldn't really work, as long as there are more outputs. I think it's better to leave this explicit... even if we added it automatically for single field outputs, adding another field would be a more invasive change for the clients

Yeah to be clear, I meant an opt-in transformation, not one we'd apply automatically. But not a big deal either way.

> I'm wondering if it'd be acceptable / just as okay to have the trait live on the output member:

Yeah I don't have a strong opinion here. At least in our users' case though, they will usually use the apply syntax so having it shorter and not needing to reference the member is probably slightly preferred. Ultimately I think either way is fine though.

kubukoz:
~~@lewisjkl think this is good to go then. I'm fine with where the trait is right now.~~ never mind, I forgot there were pending comments

modules/protocol/src/smithy4s/meta/validation/UnpackedOutputTraitValidator.java
kubukoz: yeah, I'll add some on the next run...

kubukoz: added in 702d14f

lewisjkl (approved):

PR #1946, opened 2026-05-05
2026-05 batch merge + add missing changelogs

kubukoz:
because I'm not patient enough to do the merge-update-wait-merge-update-wait cycle 8 times.


kubukoz:
doing this the right way after all, starting with #1929

Issue #1948, opened 2026-05-05
Codegen produces smithy.api.String instead of java.lang.String when applying certain alloy traits


a-morales:
For some reason when applying `alloy#dateFormat` and `alloy#localTimeFormat` to structure members the codegen will set the type to `smithy.api.String` instead of `java.lang.String`. I have a reproduction with the sample specs here https://github.com/disneystreaming/smithy4s/compare/series/0.19...fix-codegen-bug-when-using-some-alloy-traits#diff-0404a245eb249ab9b2101d500c74059fcd49f0bec5a731555dd80e429d6a18a2R13

DenisNovac:
Hey there, does it affect `@uuidFormat` ? I am not sure since generated class uses the correct `java.util.UUID`. However, my code stopped to compile when i try to pass java `UUID` as `Newtype[UUID]`.

```
type mismatch;
[error]  found   : java.util.UUID
[error]  required: com.github.MyWrappedId
[error]     (which expands to)  com.github.MyWrappedId.Type
[error]           myId = entry.myId.value,  // here i pass .value which is java UUID from Scala newtype library
```

This code compiles fine in 0.18.50

kubukoz:
Have you tried 0.19.3?

DenisNovac:
I had that error in 0.19.0 and latest 0.19.4

kubukoz:
Can you prepare a full reproduction? i.e. the smithy code and the scala code you're trying. This can be a new ticket (and we can close this one)

DenisNovac:
Yup, here: https://github.com/DenisNovac/smithy-bug-demo
Gives same error:
```
[error]UserServiceImpl.scala:19:46: type mismatch;
[error]  found   : java.util.UUID
[error]  required: com.github.denisnovac.smithytest.UserId
[error]     (which expands to)  com.github.denisnovac.smithytest.UserId.Type
[error]           userId = returnedDomainUser.userId.value
[error]                                              ^
```

PR #1949, opened 2026-05-05
Fix codegen bug when structure member had format trait applied

a-morales:
Addresses #1948, will leave a comment to what was happening. 

- [ ] Added documentation
- [ ] Updated changelog


modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -855,14 +855,17 @@ private[codegen] class SmithyToIR(
         case shape if shape.getId() == durationShapeId =>
           Type.PrimitiveType(Primitive.Duration).some
         case T.durationSecondsFormat(_) =>
-          Type
-            .Alias(
-              x.namespace,
-              x.name,
```
a-morales, line 861: The bug was that when the trait was applied we created an alias in which it used the namespace/name of the member target, in this case `smithy.api#BigDecimal`, which correctly doesn't get codegen. The fix is to check the id and if it's part of the `smithy.api` namespace just return the Primitive type.

kubukoz (approved):

modules/codegen/src/smithy4s/codegen/internals/SmithyToIR.scala
```diff
@@ -888,32 +891,41 @@ private[codegen] class SmithyToIR(
...
+            Type.PrimitiveType(Primitive.Uuid).some
+          else
+            Type
+              .Alias(
+                x.namespace,
+                x.name,
+                Type.PrimitiveType(Primitive.Uuid),
+                isUnwrapped = false
+              )
+              .some
```
lewisjkl, line 904: Might not need UUID handling here, doesn't hurt anything though

lewisjkl (approved):

Issue #1345, opened 2024-01-06
scala FQNs may get mistaken for a relative namespace


kubukoz:
Consider this smithy:

```smithy
$version: "2"

namespace hello

long Long
```

It works fine. However, if you add another file:

```smithy
$version: "2"

namespace hello.scala

structure NotThatLong {}
```

then the first one's Scala no longer compiles:

```scala
[error] -- [E008] Not Found Error: /Users/kubukoz/projects/demos/target/scala-3.3.1/src_managed/main/scala/hello/package.scala:3:25 
[error] 3 |  type Long = hello.Long.Type
[error]   |              ^^^^^^^^^^^^^^^
[error]   |              type Type is not a member of object hello.Long.
[error]   |              Extension methods were tried, but the search failed with:
[error]   |
[error]   |                  Cyclic reference involving val schema
[error] -- [E008] Not Found Error: /Users/kubukoz/projects/demos/target/scala-3.3.1/src_managed/main/scala/hello/Long.scala:10:34 
[error] 10 |object Long extends Newtype[scala.Long] {
[error]    |                            ^^^^^^^^^^
[error]    |                            type Long is not a member of hello.scala
[error] -- [E008] Not Found Error: /Users/kubukoz/projects/demos/target/scala-3.3.1/src_managed/main/scala/hello/Long.scala:15:37 
[error] 15 |  val underlyingSchema: Schema[scala.Long] = long.withId(id).addHints(hints)
[error]    |                               ^^^^^^^^^^
[error]    |                               type Long is not a member of hello.scala
```

it actually gets easier to understand if both types are named `Long`:

```diff
- structure NotThatLong {}
+ structure Long {}
```

because then the message says:

```scala
[error] -- [E007] Type Mismatch Error: /Users/kubukoz/projects/demos/target/scala-3.3.1/src_managed/main/scala/hello/Long.scala:15:69 
[error] 15 |  val underlyingSchema: Schema[scala.Long] = long.withId(id).addHints(hints)
[error]    |                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]    |                     Found:    smithy4s.schema.Schema[Long]
[error]    |                     Required: smithy4s.schema.Schema[hello.scala.Long²]
[error]    |
[error]    |                     where:    Long  is a class in package scala
[error]    |                               Long² is a class in package hello.scala²
```

I think this would have to be solved by referring to `_root_.scala.Long`.

[...]

PR #1950, opened 2026-05-06
feat: Allow to customize generated package name

ghostbuster91:

- [x] Added documentation
- [x] Updated changelog


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ When adding entries, please treat them as if they could end up in a release any
 
 Thank you!
 
+# 0.18.51
```
kubukoz: way off, 0.18.52 is out already

kubukoz:
I think we should have a usage of this in `bootstrapped/`, checked-in.

I would also deprecate the old methods of enabling/disabling namespaces, but only when we forward-port this to 0.19.

modules/docs/resources/markdown/04-codegen/01-customisation/18-package-remapping.md
```diff
@@ -0,0 +1,83 @@
...
+- `com.example.*` - matches `com.example` followed by any additional segments
+- `com.example*` - like above, but also matches `com.example` itself
+
+```kotlin
+metadata smithy4sCodegen = {
+    excludedNamespaces: ["com.example.ignored", "com.internal.*"]
+}
+```
+
+## Cross-namespace references
```
denisrosca, line 67: Does this work across multiple jar dependencies? As you note, the smithy4s model loader strips the metadata from upstream modules.

So what happens in the this scenario:
```
- Module A defines `com.example.first` shapes and codegens with packagePrefix: "gen". Generated classes live at `gen.com.example.first.MyString`. Module A also writes `com.example.first` into the smithy4sGenerated manifest, so downstream modules will treat it as already-generated and skip it.
- Module B depends on A's jar and codegens its own `com.example.second` shapes that reference `com.example.first#MyString`.
```
How does module B know to use the remapped namespace? Won't this break at module B generation/compile time due to the module A's metadata being stripped?

kubukoz: I don't know if it already happens, but if we put a copy of the package mapping into smithy4sGenerated metadata, that part should propagate from transitives

ghostbuster91: The wording was bit imprecise. Here is what `ModelLoader` does:
```
    // Appending all metadata that is not Smithy4s-specific, as well as relevant
    // Smithy4s-related metadata, into the resulting model.
    upstreamModel.getMetadata().asScala.foreach {
      case (CodegenRecord.METADATA_KEY, _) => ()
      case (k, _) if k.startsWith("smithy4s") =>
        sanitisingModelBuilder.removeMetadataProperty(k)
      case _ => ()
    }
```
So the `smithy4sGenerated` key is preserved and now it also holds the mapping.

Having said that we weren't consuming rendered packages from the previous runs when determining the current mappings. This should be fixed.

kubukoz: sounds like some sbt tests are due

kubukoz:
let's rebase this on the latest 0.18, you'll get the CI gains

modules/docs/resources/markdown/04-codegen/01-customisation/18-package-remapping.md
```diff
@@ -0,0 +1,96 @@
...
+
+When package remapping is active, cross-namespace `Type.Ref` nodes inside declarations are remapped automatically. Generated import statements will use the remapped package name, so the compiled Scala code remains self-consistent.
```
kubukoz: issue: end users may not know what "`Type.Ref` nodes inside declarations" are, we should rephrase to just say "references to types from previous codegen runs" or something like that

modules/docs/resources/markdown/04-codegen/01-customisation/18-package-remapping.md
```diff
@@ -0,0 +1,96 @@
...
+This also works across module boundaries. When Smithy4s generates code for Module A with a `packagePrefix`, it writes the namespace-to-package mapping into the `smithy4sGenerated` tracking manifest:
+
+```smithy
+metadata smithy4sGenerated = [{
+  namespaces: ["com.example.api"],
+  renderedPackages: { "com.example.api": "gen.com.example.api" }
+}]
+```
+
+When Module B depends on Module A's jar and runs codegen, the `ModelLoader` discovers Module A's Smithy model (and its manifest) from the classpath. Smithy4s reads the `renderedPackages` entries and applies them as additional package mappings before rendering Module B's code. Any `Type.Ref` pointing at `com.example.api` in Module B's generated code will be resolved to `gen.com.example.api` — matching the actual location of Module A's compiled classes.
```
kubukoz: again, ModelLoader is too much detail for a user-oriented doc page, it's more like CONTRIBUTING.md material.

kubukoz (approved):

kubukoz:
want to see if the other maintainers have any concerns, but after this and #1965 we can do a new patch release.

Then we forward-port to 0.19, sounds good?

lewisjkl (commented):
Can you elaborate on the intended use cases where each of these configurations would be needed? I feel it can add a lot of complexity / make it harder to find generated code than it otherwise would be, and generally make it harder to troubleshoot for users posting issues. I think we'd need a pretty strong justification to support adding it and maybe something less open-ended could solve the intended use case.

ghostbuster91:
There are two use cases driving this:

1. Simple package renaming. Sometimes you just need to control the package name under which generated code lands, without being tied to the namespace structure of the Smithy files.

2. Multiple codegen variants from the same Smithy sources. We have a project where the same Smithy files are used to generate code in two different ways. There's a shared JVM library that contains one variant of the generated code (e.g. without type refinements applied), and a downstream module that depends on this library but needs to generate from the same Smithy files with a different codegen configuration (e.g. with type refinements applied), because it uses those models for a slightly different purpose.

Without the ability to remap the target package, the two generated outputs would collide. Remapping lets the downstream module land its variant in a distinct package so both can coexist.

kubukoz:
btw. package renaming is useful to avoid some conflicts, e.g. [smithy4s-bsp](https://github.com/polyvariant/smithy4s-bsp/blob/3979a92b5363177e692cbd28b85b3ee7f05c5595/transformation/src/main/scala/RenameScalaNamespace.scala#L31-L34).

> Multiple codegen variants from the same Smithy sources.

It can also be considered a form of shading. [sbt-shading](https://github.com/coursier/sbt-shading) has some problems that make it unsuitable for daily usage, IMO

lewisjkl:
Yeah I see the case for having a package/namespace prefix that the user could specify. That would handle the shading concern and the collision one. I'm less sure that a full rewrite is necessary or desirable. The full `foo.bar` => `baz.whatever` is more concerning to me from the perspective of debugging/helping users out. A namespace prefix is a lot easier to spot and I think it solves the problems you are listing above.

Maybe there are scenarios I'm not considering, but I just think the least powerful version of this we can go with that solves the problem is ideal. Fwiw, internally in another codegen project we do support namespace prefixes and it has proved useful but still simple. In that case, we actually make it sort of an all or nothing. Meaning all generated code from this invocation of the plugin will have a prefix or it won't. Basically it is a way of shading all the things you generated to avoid conflicts in downstream dependencies (or from upstream for that matter).

kubukoz:
I was wondering if we could keep just the prefixing part, but not globally (I can see why you might want to prefix e.g. just a "standard library" type of namespace, and not an API namespace, for example). However, that doesn't solve the bsp problem as I only wanted to rename certain namespaces that conflicted with other names (e.g. the `_root_.scala` package).

kubukoz:
In general I think detaching the generated packages from the Smithy namespaces is an ability worth having. Prefixing is a special case, but since we have to persist the mapping in the intermediate metadata... might as well go all the way

lewisjkl:
> that doesn't solve the bsp problem as I only wanted to rename certain namespaces that conflicted with other names (e.g. the _root_.scala package).

Can you elaborate on what you mean here? I want to make sure I am getting the full picture.

In the end, I don't feel strongly enough about this to prevent it from going through - feel free to merge and we will review more internally if we see any usage of the feature and see how it feels and works in practice.

kubukoz:
> > that doesn't solve the bsp problem as I only wanted to rename certain namespaces that conflicted with other names (e.g. the _root_.scala package).
> 
> Can you elaborate on what you mean here? I want to make sure I am getting the full picture.

I linked this earlier - in bsp, there's a [`bsp.scala` namespace](https://github.com/build-server-protocol/build-server-protocol/blob/1abbbb26a1d5ce93ea2267c28b24f0b05876e7f1/spec/src/main/resources/META-INF/smithy/bsp/extensions/scala.smithy), and when you generate code for it, other code using packages from `scala.` (such as `scala.Product` and many others) doesn't compile - #1345.

It's a pretty big change (fixing #1345), so instead I opted for a workaround - just renaming the `scala` part of the namespace, via [a transformation](https://github.com/polyvariant/smithy4s-bsp/blob/3979a92b5363177e692cbd28b85b3ee7f05c5595/transformation/src/main/scala/RenameScalaNamespace.scala#L31-L34). Same for `java`. And later on I realized the BSP traits are in `traits` which is too global, so I prefixed that with `bsp.` for consistency.

Doing this with transformations only works as long as the shapeIds aren't used for anything at runtime.

lewisjkl:
Makes sense, appreciate the explanation. Yeah like I said I'm good to merge this and then provide feedback if I notice any friction in the wild.

kubukoz:
https://github.com/disneystreaming/smithy4s/releases/tag/v0.18.53

PR #1952, opened 2026-05-06
CI: split compile/test/mima/checks/docs/plugins/release into parallel jobs

kubukoz:
Disclaimer: this is all AI-generated work, but it was directed by my own ideas on how to speed up the pipeline.

I do think we could de-boilerplate the YML a little bit, but the direction I'd like to steer in would be to use something like [sbt-github-actions](https://github.com/sbt/sbt-github-actions/) (or sbt-typelevel's version of that). We can tackle this separately. But then again, I think GHA YML was unmaintainable before the age of Claude, and it's actually the perfect candidate for vibe engineering.

I intend to backport this to 0.18, as we're still getting [contributions](#1950) there, and will probably continue to do so for a few more months.

---

## Summary

Restructures CI from a monolithic 6-cell test matrix into a graph of parallel, cache-reusing jobs. **Brings PR wall-clock from 34:22 (reference run [25463921864](https://github.com/disneystreaming/smithy4s/actions/runs/25463921864)) down to 16:48 ([run 25468673536](https://github.com/disneystreaming/smithy4s/actions/runs/25468673536)) — a 51% reduction.**

## Job graph

```
compile (matrix: 6 cells, produces /tmp/remote-cache artifacts)
   ├─> test          (matrix: 6 cells, pulls cache, runs test_*)
   ├─> mima          (matrix: 3 jvm cells, pulls cache)
   ├─> docs          (pulls 2.13/jvm cache; mdoc + website)
   ├─> sbt-plugin    (matrix: 24 cells — one per scripted test folder)
   ├─> mill-plugin   (matrix: 3 cells — one per mill version)
   └─> release       (matrix: 6 cells; runs only on tags/snapshots)
            └─> release-bundle  (tags only; merges per-shard sona-staging
                                 slices and runs sonaRelease once)

pre-existing in parallel (no compile dep):
checks       (scalafmt + scalafix on 2.13/jvm + headerCheck)

build-success-checkpoint  waits on all of the above.
```

## Key changes

**Compile/test split with shared cache**
- The `compile` matrix now produces `/tmp/remote-cache` artifacts containing Compile *and* Test classes plus Zinc analysis (`META-INF/inc_compile.zip`).
- `test` and `mima` jobs download the matching artifact and `pullRemoteCache_$BUILD_KEY` before doing their work — Zinc skips compile when source hasn't changed.
[...]

PR #1953, opened 2026-05-06
CI: pre-compile codegen chain (on top of #1952)

kubukoz:
**Stacked on top of #1952.**

Adds a `pre-compile` job that compiles `protocol`, `codegen`, and `codegen-cli` once before the matrix, pushing their cache. Every matrix cell pulls this cache before its own work, so the codegen-cli classpath used by `genSmithy` is satisfied without recompile. `test` and `mima` cells also pull the codegen chain explicitly because their cell-scoped `pullRemoteCache_$BUILD_KEY` aggregates don't include codegen-cli (which is only 2.13/jvm).

**Why:** today every cell rebuilds `codegen` (~38 sources, ~27s) from scratch even though every cell needs the same compiled artifact for source generation. With pre-compile, the work happens once. The bigger win is reduced peak memory pressure within each cell — fewer modules to schedule concurrently.

## Test plan

- [ ] `pre-compile` job runs and uploads `compilation-pre.zip`
- [ ] Compile cells download pre-compile cache and skip codegen-cli/codegen/protocol compile
- [ ] Test cells (especially js/native) skip recompile of codegen-cli when invoking `genSmithy`
- [ ] Total wall-clock vs. #1952 baseline

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,8 +34,52 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  pre-compile:
+    name: Pre-compile (codegen)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout current branch
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ env.CHECKOUT_REF }}
+
+      - name: Cache
+        uses: coursier/cache-action@v6
+        with:
+          extraKey: pre-compile
+
+      - name: Setup Java and Scala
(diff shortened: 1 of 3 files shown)


kubukoz:
Numbers from the CI runs show pre-compile adds ~1.5 min wall-clock (3 min for the pre-compile job + matrix cells starting after it) while saving ~30s on the slowest cell. Net negative as configured. The premise was that compounding wins across many cells would help, but the cells were already running in parallel — the saved per-cell time wasn't on the critical path. Closing.

PR #1954, opened 2026-05-06
CI: separate link step for js/native test cells (on top of #1952)

kubukoz:
**Stacked on top of #1952.**

Splits the js/native test cells into two steps: link first, then run. Linker work (Test/fastLinkJS for Scala.js, Test/nativeLink for Scala Native) is memory-heavy; today it runs concurrently with test execution, creating avoidable peak memory pressure on the runners.

Adds a `testLink_<scala>_<platform>` alias generated by the build plugin, only for js and native doublets. JVM cells skip the link step via an `if:` guard.

**Why:** the linker is single-threaded per module but multiple modules can link in parallel — competing for memory with concurrently-running tests for already-linked modules. By isolating linking, we let it use the full runner without contention, and the test step starts from cached linker output.

## Test plan

- [ ] Link step runs only for js/native cells
- [ ] Test step skips relink (linker output cached on disk)
- [ ] No OOMs on JS/Native cells
- [ ] Total wall-clock for js/native cells vs. #1952 baseline

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,8 +34,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build:
-    name: Test ${{matrix.scalaVersion}} (${{matrix.scalaPlatform}})
+  compile:
+    name: Compile ${{matrix.scalaVersion}} (${{matrix.scalaPlatform}})
     strategy:
       fail-fast: true
       matrix:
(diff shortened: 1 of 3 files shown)


kubukoz:
Splitting link from test added 2nd-sbt-boot overhead without preventing the memory contention it was meant to address — Test 3_0 (native) actually got slower (~16 min vs ~13 min on the base PR). Closing.

PR #1955, opened 2026-05-06
CI: scope scalafix check to 2.13/jvm only (on top of #1952)

kubukoz:
**Stacked on top of #1952.**

The `checks` job currently runs `scalafixAll --check`, which iterates every project regardless of axis. The build plugin (Smithy4sBuildPlugin.scala:680) already comments that scalafix is intended for 2.13/jvm projects only, and exposes scoped aliases `scalafix_2_13_jvm` / `scalafixTests_2_13_jvm` for exactly this case. Switch to those aliases.

## Test plan

- [ ] `checks` job runs scalafix only on 2.13/jvm projects (not 2.12, 3.0, js, native)
- [ ] No unexpected scalafix violations slip through (the rules currently in `.scalafix.conf` operate on 2.x semanticdb anyway)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,8 +34,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build:
-    name: Test ${{matrix.scalaVersion}} (${{matrix.scalaPlatform}})
+  compile:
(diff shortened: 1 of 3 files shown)


kubukoz:
Merged into #1952 (modernize-ci) locally and pushed.

PR #1956, opened 2026-05-06
CI: matrix scripted tests across folders (on top of #1952)

kubukoz:
**Stacked on top of #1952.**

The existing `sbt-plugin` job runs 24 scripted tests serially in ~18 min wall-clock. publishLocal of smithy4s modules is fast (driven by the remote cache wired in #1952), so the wall-clock is dominated by per-test sub-sbt invocations that each compile and run sample sources.

Splits the job into:
1. `sbt-plugin-discover` — lists `modules/codegen-plugin/src/sbt-test/codegen-plugin/` into a JSON array (job output).
2. `sbt-plugin` — matrix over that list, each cell runs `sbt 'scripted codegen-plugin/<folder>'`. `fail-fast: false` so a flake in one folder doesn't cancel the others.

Total wall-clock should drop to whatever the slowest single test takes, plus per-cell setup overhead.

## Test plan

- [ ] `sbt-plugin-discover` lists all 24 folders correctly
- [ ] All 24 cells succeed
- [ ] Wall-clock per cell vs. monolithic job
- [ ] Total billable runner-minutes (expect 2-3x increase since each cell pays setup overhead)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,8 +34,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build:
-    name: Test ${{matrix.scalaVersion}} (${{matrix.scalaPlatform}})
+  compile:
+    name: Compile ${{matrix.scalaVersion}} (${{matrix.scalaPlatform}})
     strategy:
       fail-fast: true
       matrix:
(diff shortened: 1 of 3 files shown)


kubukoz:
Merged into #1952 (modernize-ci) locally and pushed.

PR #1961, opened 2026-05-12
Add `@transient` annotation

xuwei-k:
edit by admin: https://www.scala-sbt.org/2.x/docs/en/reference/cached-task.html#per-task-key-opt-out


kubukoz:
👋 

Can you share more about what problem this solves?

lenguyenthanh:
probably this: https://github.com/sbt/sbt/pull/9032, I guess if we bump to 2.0.0-RC12 then we need to add `@transient` anyway.

kubukoz:
I think this should be documented in sbt more prominently...

edit: [it is](https://www.scala-sbt.org/2.x/docs/en/reference/cached-task.html#per-task-key-opt-out).

kubukoz (approved):

PR #1968, opened 2026-05-26
relax default ns filter

lewisjkl:
By default, smithy4s was filtering out all namespaces that started with  from codegen. This is problematic because it does not generate every  namespace inside of the core module. So if you want to use another module, there isn't a great way to do it. This relaxes the constraint to make it only prevent codegeneration on things that are already generated in core (which is smithy.api).

- [ ] Added documentation
- [x] Updated changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,16 @@ When adding entries, please treat them as if they could end up in a release any 
 
 Thank you!
 
+# 0.19.6
+
+- Relax default filter in codegen module to allow generating namespaces that start with `smithy.` as long as they are not `smithy.api` in [#1968](https://github.com/disneystreaming/smithy4s/pull/1968)
+
 # 0.19.5 
 
 - Port of [#1950](https://github.com/disneystreaming/smithy4s/pull/1950) to 0.19 series
   codegen: Add `smithy4sCodegen` metadata key for render-time package remapping. Supports `packagePrefix` (prepend a prefix to all generated packages), `packageMappings` (per-namespace overrides), and `excludedNamespaces` (skip namespaces from codegen). See [Package Remapping](https://disneystreaming.github.io/smithy4s/docs/codegen/customisation/package-remapping) for details.
 - codegen: Add `allowedNamespaces` field to the `smithy4sCodegen` Smithy metadata key — restricts codegen to the listed namespace patterns. When the build-tool `allowedNamespaces` / `excludedNamespaces` settings are also set, the two sources are unioned.
 - codegen: Deprecate the build-tool `allowedNamespaces` / `excludedNamespaces` settings (sbt `smithy4sAllowedNamespaces` / `smithy4sExcludedNamespaces`, the equivalent mill targets, and the CLI `--allowed-ns` / `--excluded-ns` flags) in favor of the `smithy4sCodegen` Smithy metadata key.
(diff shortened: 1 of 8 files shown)


yisraelU (approved):

modules/codegen/test/src/smithy4s/codegen/internals/CodegenImplSpec.scala
```diff
@@ -281,7 +281,10 @@ final class CodegenImplSpec extends munit.FunSuite {
       )
       .map { case (_, result) => result.namespace }
       .toSet
-    assertEquals(generatedNamespaces, expectedCodegenNamespaces)
+    assertEquals(
+      generatedNamespaces.filterNot(_.startsWith("smithy")),
```
kubukoz: should this also say `smithy.api`?

modules/codegen/test/src/smithy4s/codegen/internals/TestUtils.scala
```diff
@@ -53,6 +53,8 @@ object TestUtils {
       loc: Location
   ): Unit = {
     val scalaResults = generateScalaCode(smithySpec).values.toList
+      // filter out smithy dependencies for the sake of assertions
+      .filterNot(_.startsWith("package smithy"))
```
kubukoz, line 57: ditto

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Thank you!
   codegen: Add `smithy4sCodegen` metadata key for render-time package remapping. Supports `packagePrefix` (prepend a prefix to all generated packages), `packageMappings` (per-namespace overrides), and `excludedNamespaces` (skip namespaces from codegen). See [Package Remapping](https://disneystreaming.github.io/smithy4s/docs/codegen/customisation/package-remapping) for details.
 - codegen: Add `allowedNamespaces` field to the `smithy4sCodegen` Smithy metadata key — restricts codegen to the listed namespace patterns. When the build-tool `allowedNamespaces` / `excludedNamespaces` settings are also set, the two sources are unioned.
 - codegen: Deprecate the build-tool `allowedNamespaces` / `excludedNamespaces` settings (sbt `smithy4sAllowedNamespaces` / `smithy4sExcludedNamespaces`, the equivalent mill targets, and the CLI `--allowed-ns` / `--excluded-ns` flags) in favor of the `smithy4sCodegen` Smithy metadata key.
+- Relax default filter in codegen module to allow generating namespaces that start with `smithy.` as long as they are not `smithy.api` in [#1968](https://github.com/disneystreaming/smithy4s/pull/1968)
```
kubukoz: I'm actually cutting a 0.19.5 now, can you move this to the next patcH?

modules/codegen/test/src/smithy4s/codegen/internals/CodegenImplSpec.scala
lewisjkl: Well in this case the test is not looking for smithy namespace files and since the smithy.test dependencies are there in the class path it will generate them. So filtering them out shortens the assertion to just relevant parts

kubukoz: how about this, just to be specific: `!startsWith("smithy.api") & !startsWith("smithy.test")` - just to make sure other `smithy.*` namespaces don't pass through by accident

kubukoz: consider this a nitpick tho

kubukoz (approved):

kubukoz (approved):

PR #1970, opened 2026-05-27
Allow opting out of coursier's default repositories during codegen

### LINKED ISSUE #1969: Codegen always queries Maven Central via coursier defaults; expose option to disable
### Summary

`Smithy4sCodegenPlugin` resolves codegen dependencies through `coursier.Fetch(FileCache())`, which seeds the resolver list with coursier's [default repositories](https://github.com/coursier/coursier/blob/main/modules/coursier/shared/src/main/scala/coursier/Resolve.scala) — including Maven Central — *in addition to* whatever the user passes via sbt's `resolvers`. There is no way to opt out.

For users who run behind an internal repository proxy (e.g. Artifactory, Nexus) that mirrors Maven Central, this is undesirable: when both repos are queried, coursier may pick Maven Central for the actual artifact download, even though the proxy has the artifact and is authenticated.

### Why this matters now

Maven Central enforces per-IP rate limits (see [Sonatype's 429 FAQ](https://central.sonatype.org/faq/429-error/)) and we (and presumably others) are now seeing 429s from Maven Central during codegen. Example error:

```
[error] coursier.error.FetchError$DownloadingArtifacts: Error fetching artifacts:
[error] https://repo1.maven.org/maven2/software/amazon/smithy/smithy-model/1.66.0/smithy-model-1.66.0.jar: download error: Caught java.io.IOException (Server returned HTTP response code: 429 for URL: ...)
[error] (... / Compile / smithy4sCodegen) coursier.error.FetchError$DownloadingArtifacts: ...
```

This happens despite an internal mirror being configured via sbt `resolvers`, because Maven Central is also in the resolver list and wins the race for some artifacts.

### Reproducer

Any project using `Smithy4sCodegenPlugin` with `sbt.override.build.repos=true` pointing exclusively at an internal mirror — `smithy4sCodegen` will still hit `repo1.maven.org` because the codegen plugin's coursier instance ignores those overrides and uses defaults.

### Relevant code

[`Smithy4sCodegenPlugin.scala`](https://github.com/disneystreaming/smithy4s/blob/series/0.18/modules/codegen-plugin/src/smithy4s/codegen/Smithy4sCodegenPlugin.scala) reads sbt resolvers and passes URLs to `CodegenArgs.repositories`:

```scala
val res = (conf / resolvers).value.toList.collect { case m: MavenRepository => m.root }
```

[`ModelLoader.scala`](https://github.com/disneystreaming/smithy4s/blob/series/0.18/modules/codegen/src/smithy4s/codegen/internals/ModelLoader.scala) then constructs the fetch:

```scala
val fetch = Fetch(FileCache())
  .addRepositories(repos: _*)
  .addDependencies(deps: _*)
```

`Fetch(FileCache())` includes coursier's default repositories (Maven Central, ivy2Local, etc.). `addRepositories` *adds* to that list rather than replacing it, so there is no way for the user to disable Maven Central even when their sbt `resolvers` setting is exhaustive.

### Proposed fix

Expose a setting on `Smithy4sCodegenPlugin` (e.g. `smithy4sCodegenUseDefaultRepositories: SettingKey[Boolean]`, defaulting to `true` for back-compat) that, when set to `false`, configures `Fetch` to skip default repositories and use only the ones derived from sbt's `resolvers`. Equivalent of coursier CLI's `--no-default`.

For users behind a proxy, setting it to `false` would eliminate Maven Central from the codegen's resolver list entirely.

### Workaround

Users can override coursier's default repositories globally via `-Dcoursier.repositories=...` in `SBT_OPTS`, but that is a blunt instrument that affects every coursier consumer in the JVM, not just smithy4s codegen.

### Related

#1723 (sbt proxy repository overrides ignored — same root cause: codegen uses its own coursier `Fetch`)

kubukoz:
## Summary

Adds a way to disable coursier's default repository list (Maven Central, ivy2Local, etc.) during codegen dependency resolution, so users behind an internal repository proxy can avoid querying Maven Central entirely.

`CodegenArgs` / `DumpModelArgs` gain a new `allowDefaultRepositories: Boolean = true` field. When `false`, `Fetch` uses `withRepositories` (replace) instead of `addRepositories` (additive), so only the user-supplied repos are queried — equivalent to coursier CLI's `--no-default`.

Surfaced as:
- **sbt**: `smithy4sAllowDefaultRepositories: SettingKey[Boolean]`
- **Mill** (0.11 / 0.12 / 1.x): `def smithy4sAllowDefaultRepositories: T[Boolean]`
- **CLI**: `--no-default-repositories` on `generate` and `dump-model`

Defaults to `true` everywhere, so existing builds are unaffected.

Fixes #1969.

## Why

Maven Central enforces per-IP rate limits and we (and likely others) are hitting `429 Too Many Requests` during codegen. With this change, users behind a proxy that mirrors Central can keep codegen entirely off Maven Central.

## Test plan

- [x] `codegen` / `codegen-cli` / `codegen-plugin` (Scala 2.12 + 3) / `mill-codegen-plugin` (0.11, 0.12, 1.x) / `codegen-integration` all compile
- [x] `codegen-cli` `CommandParsingSpec` updated and passing — covers `--no-default-repositories` parsing for both subcommands and the `true` default when the flag is omitted
- [x] `codegen` test suite passes
- [x] Manual sandbox at `/tmp/smithy4s-1969-test/` (locally-published plugin): with `Compile / smithy4sAllowDefaultRepositories := false` and no resolvers configured, codegen fails with `ResolutionError$CantDownloadModule` for `smithy4s-protocol` — confirming Maven Central / ivy2Local are not consulted. With the default `true`, the same project resolves and generates as expected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/modules/codegen/src/smithy4s/codegen/internals/ModelLoader.scala
+++ b/modules/codegen/src/smithy4s/codegen/internals/ModelLoader.scala
@@ -43,13 +43,15 @@ private[codegen] object ModelLoader {
       repositories: List[String],
       transformers: List[String],
       discoverModels: Boolean,
-      localJars: List[os.Path]
+      localJars: List[os.Path],
+      allowDefaultRepositories: Boolean
   ): (ClassLoader, Model) = {
     val currentClassLoader = this.getClass().getClassLoader()
     val deps = resolveDependencies(
       dependencies :+ protocolDependency,
       localJars,
-      repositories
+      repositories,
+      allowDefaultRepositories
     )
 
     val modelsInJars = deps.flatMap { file =>
@@ -138,7 +140,8 @@ private[codegen] object ModelLoader {
   private def resolveDependencies(
       dependencies: List[String],
       localJars: List[os.Path],
-      repositories: List[String]
+      repositories: List[String],
+      allowDefaultRepositories: Boolean
   ): Seq[File] = {
     val maybeRepos = RepositoryParser.repositories(repositories).either
     val maybeDeps = DependencyParser
@@ -163,10 +166,11 @@ private[codegen] object ModelLoader {
     }
     val resolvedDeps: Seq[java.io.File] =
       if (deps.nonEmpty) {
-        val fetch = Fetch(FileCache())
-          .addRepositories(repos: _*)
-          .addDependencies(deps: _*)
-        fetch.run()
+        val baseFetch = Fetch(FileCache())
+        val withRepos =
+          if (allowDefaultRepositories) baseFetch.addRepositories(repos: _*)
+          else baseFetch.withRepositories(repos)
+        withRepos.addDependencies(deps: _*).run()
       } else {
         Seq.empty
       }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ When adding entries, please treat them as if they could end up in a release any 
 
 Thank you!
 
+# 0.19.7
+
+- codegen: Add a way to opt out of coursier's default repositories (e.g. Maven Central, ivy2Local) when resolving codegen dependencies. Sbt: set `smithy4sAllowDefaultRepositories := false`. Mill: override `def smithy4sAllowDefaultRepositories = false`. CLI: pass `--no-default-repositories` to `generate`/`dump-model`. Useful for users behind an internal repository proxy that mirrors Maven Central. Defaults to `true` for backwards compatibility. See [#1969](https://github.com/disneystreaming/smithy4s/issues/1969).
+
 # 0.19.6
 
 - Relax default filter in codegen module to allow generating namespaces that start with `smithy.` as long as they are not `smithy.api` in [#1968](https://github.com/disneystreaming/smithy4s/pull/1968)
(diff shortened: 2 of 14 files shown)


modules/codegen/src/smithy4s/codegen/internals/ModelLoader.scala
```diff
@@ -163,10 +166,11 @@ private[codegen] object ModelLoader {
     }
     val resolvedDeps: Seq[java.io.File] =
       if (deps.nonEmpty) {
-        val fetch = Fetch(FileCache())
-          .addRepositories(repos: _*)
-          .addDependencies(deps: _*)
-        fetch.run()
+        val baseFetch = Fetch(FileCache())
+        val withRepos =
+          if (allowDefaultRepositories) baseFetch.addRepositories(repos: _*)
```
kubukoz, line 171: I'm wondering... should we just prepend by default? It'd be a silent behavior change, which some folks may not like, e.g. if they add snapshot repos here. (they'd be checked first, which is usually a waste)

lewisjkl: Yeah I mean in hindsight it would have been better I think, but not sure we should change it now. I don't feel too strongly either way though

lewisjkl (approved):

PR #1977, opened 2026-07-06
switch over to use interfaces library to make life better

### LINKED ISSUE #1922: Move to coursier-interface
Migrate our use of Scala-cross-compiled coursier to coursier-interface, the Java library with shading. This should simplify the build for Scala 3 and also our usage of the coursier APIs.

ref: #1903 + [discord](https://discord.com/channels/1045676621761347615/1045676622491164724/1480941749496189070)

Make sure to use at least https://github.com/coursier/interface/releases/tag/v1.0.29-M4 .

yisraelU:
closes #1922 

- [x] Updated changelog

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -427,7 +427,7 @@ lazy val codegen = projectMatrix
       Dependencies.Circe.core.value,
       Dependencies.Circe.parser.value,
       Dependencies.Circe.generic.value,
-      "io.get-coursier" % "interface" % "1.0.29-M4",
+      Dependencies.CoursierInterface.core,
       Dependencies.Mima.core % Test
     ),
     libraryDependencies ++= {
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,4 @@
+import com.typesafe.tools.mima.core._
 import software.amazon.smithy.model.traits.TraitService
 import _root_.java.util.stream.Collectors
 import java.nio.file.Files
@@ -253,7 +254,15 @@ lazy val core = projectMatrix
     scalacOptions ++= Seq(
       "-Wconf:msg=value noInlineDocumentSupport in class ProtocolDefinition is deprecated:silent"
     ),
-    mimaBinaryIssueFilters ++= Seq.empty
+    // smithy-waiters 1.72.0 added a new `message` field to the `Acceptor` structure
+    // (see waiters.smithy upstream), which shifted the arity of the generated
+    // `smithy.waiters.Acceptor.apply` case class constructor. This is generated
+    // prelude code for the built-in waiters trait, not hand-authored API.
+    mimaBinaryIssueFilters ++= Seq(
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "smithy.waiters.Acceptor.apply"
+      )
+    )
   )
   .jvmPlatform(allJvmScalaVersions, jvmDimSettings)
   .jsPlatform(allJsScalaVersions, jsDimSettings)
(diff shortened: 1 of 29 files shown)


build.sbt
```diff
@@ -253,7 +254,15 @@ lazy val core = projectMatrix
...
-    mimaBinaryIssueFilters ++= Seq.empty
+    // smithy-waiters 1.72.0 added a new `message` field to the `Acceptor` structure
+    // (see waiters.smithy upstream), which shifted the arity of the generated
+    // `smithy.waiters.Acceptor.apply` case class constructor. This is generated
+    // prelude code for the built-in waiters trait, not hand-authored API.
+    mimaBinaryIssueFilters ++= Seq(
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "smithy.waiters.Acceptor.apply"
+      )
+    )
```
lewisjkl, line 265: I was just thinking that we could use the bincompat-friendly codegen from @kubukoz for this, but I don't think we can actually because the trait we'd need to use won't be able to be added via the smithy apply syntax.. I think you can't use apply syntax on anything inside of prelude.

Is there a way to do this that you know of @kubukoz ?

kubukoz: yes, @majk-p used it somewhere already iirc. let me dig it up

kubukoz: `smithy4sBincompatPreludeAdditions` - https://github.com/disneystreaming/smithy4s/pull/1920/changes#diff-20fd97b354b82750271e4a60e0954a9bb23e469524f1ea6fff646dd4288591bfR11

modules/codegen/src/smithy4s/codegen/internals/ModelLoader.scala
```diff
@@ -137,40 +140,55 @@ private[codegen] object ModelLoader {
...
+  }
+
+  // Builds the Fetch request without running it, so the repository/dependency
+  // wiring can be unit-tested without performing any network resolution.
+  private[internals] def buildFetch(
+      dependencies: List[Dependency],
+      repositories: List[MavenRepository],
+      allowDefaultRepositories: Boolean
+  ): Fetch = {
+    val baseFetch = Fetch.create()
```
kubukoz, line 170: I think we need to call `.withCache()` so that creds auto-load - similarly to the Scala version.

build.sbt
```diff
@@ -418,8 +427,7 @@ lazy val codegen = projectMatrix
       Dependencies.Circe.core.value,
       Dependencies.Circe.parser.value,
       Dependencies.Circe.generic.value,
-      ("io.get-coursier" %% "coursier" % "2.1.24")
-        .cross(CrossVersion.for3Use2_13),
+      "io.get-coursier" % "interface" % "1.0.29-M4",
```
kubukoz: pull out to `Dependencies`?

build.sbt
kubukoz: but I'm afraid `smithy.waiters` doesn't have `@bincompatFriendly` :( so adding it now will be even more breaking.

Baccata (approved):

Baccata:
PS : my approval is conditioned by the addressing of other people's comments, just stating that I'm okay with the direction

build.sbt
lewisjkl: Yeah that makes sense, then for now we will stick with the mima override

yisraelU:
Addressed feedback.

modules/codegen/src/smithy4s/codegen/internals/ModelLoader.scala
yisraelU: This happens automatically here in the ApiHelper.cache()  https://github.com/coursier/interface/blob/v1.0.29-M4/interface/src/main/scala/coursier/internal/api/ApiHelper.scala#L388-L392

lewisjkl (approved):

PR #1978, opened 2026-07-08
Support @structurePattern targeting unions

Baccata:
Depends on https://github.com/disneystreaming/alloy/pull/280

Extend SchemaVisitorPatternDecoder and SchemaVisitorPatternEncoder to handle union schemas using the magic {label} and {value} pattern identifiers, as specified by the upcoming alloy change (disneystreaming/alloy#280).

Also bump alloy to a dev-SNAPSHOT that relaxes the structurePattern validator to accept unions as targets, and exclude transitive alloy from smithytranslate to avoid version conflicts.

- [x] Added sample shape to check compilation of generated code.
- [x] Added documentation
- [x] Updated changelog
- [ ] Upgrade alloy once upstream PR is merged and a release is cut

### DIFF
--- a/modules/core/src/smithy4s/internals/SchemaVisitorPatternEncoder.scala
+++ b/modules/core/src/smithy4s/internals/SchemaVisitorPatternEncoder.scala
@@ -91,6 +91,57 @@ private[internals] final class SchemaVisitorPatternEncoder(
     }
   }
 
+  override def union[U](
+      shapeId: ShapeId,
+      hints: Hints,
+      alternatives: Vector[Alt[U, _]],
+      dispatch: Alt.Dispatcher[U]
+  ): MaybePathEncode[U] = {
+    type Writer = U => List[String]
+
+    def compileAlt[A](alt: Alt[U, A]): Option[A => List[String]] = {
+      self(alt.schema).map(_.encode)
+    }
+
+    val altEncoders: Option[Map[Int, Any => List[String]]] = {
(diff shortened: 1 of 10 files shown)


modules/core/src/smithy4s/internals/SchemaVisitorPatternEncoder.scala
```diff
@@ -91,6 +91,57 @@ private[internals] final class SchemaVisitorPatternEncoder(
...
+  ): MaybePathEncode[U] = {
+    type Writer = U => List[String]
+
+    def compileAlt[A](alt: Alt[U, A]): Option[A => List[String]] = {
+      self(alt.schema).map(_.encode)
+    }
+
+    val altEncoders: Option[Map[Int, Any => List[String]]] = {
+      val entries = alternatives.zipWithIndex.map { case (alt, idx) =>
+        compileAlt(alt).map(enc =>
```
kubukoz: I haven't looked very closely but can't we use `Alt.Dispatcher` to make this a little more type safe?

Baccata: Nah, unfortunately the `Option` layer prevents that. We could get around it by changing the way the visitor works so that it'd build a `A => String => String` or something, but ultimately it'd be a larger changeset for a minor benefit, I think :/

lewisjkl (approved):

PR #1981, opened 2026-07-10
BugFix : ClassCastException when encoding DArray backed by Object[]

Baccata:
The previous fix for #1158 only short-circuited on empty arrays. Non-empty ArraySeq instances backed by Object[] (produced by Iterator.map(...).toIndexedSeq in DocumentEncoderSchemaVisitor) still triggered the unsafe cast to Array[Document].

Guard the fast path with an isInstanceOf check on the underlying array, falling back to the safe foreach path when the runtime type doesn't match.

- [x] Updated changelog

### DIFF
--- a/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
+++ b/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
@@ -521,7 +521,8 @@ private[smithy4s] class SchemaVisitorJCodec(
               // DArray values built via Iterator.map(...).toIndexedSeq (e.g. in DocumentEncoderSchemaVisitor)
               // produce ArraySeq backed by Object[], making an unchecked cast crash at runtime.
               // See https://github.com/disneystreaming/smithy4s/issues/1158
-              case x: ArraySeq[_] if x.unsafeArray.isInstanceOf[Array[Document]] =>
+              case x: ArraySeq[_]
+                  if x.unsafeArray.isInstanceOf[Array[Document]] =>
                 val xs = x.unsafeArray.asInstanceOf[Array[Document]]
                 var i = 0
                 while (i < xs.length) {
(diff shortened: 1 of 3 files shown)


modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
```diff
@@ -517,18 +517,17 @@ private[smithy4s] class SchemaVisitorJCodec(
...
-                    encodeValue(xs(i), out)
-                    i += 1
-                  }
+              // Fast path: only safe when the backing array is actually Array[Document].
+              // DArray values built via Iterator.map(...).toIndexedSeq (e.g. in DocumentEncoderSchemaVisitor)
+              // produce ArraySeq backed by Object[], making an unchecked cast crash at runtime.
+              // See https://github.com/disneystreaming/smithy4s/issues/1158
+              case x: ArraySeq[_]
+                  if x.unsafeArray.isInstanceOf[Array[Document]] =>
+                val xs = x.unsafeArray.asInstanceOf[Array[Document]]
```
yisraelU, line 529: purely cosmetic , can collapse the cast check and cast 

```
  x.unsafeArray match {
    case xs: Array[Document] => ...
```

yisraelU (approved):

yisraelU:
for future reference , I dont think the bug is triggered by ` Iterator.map(...).toIndexedSeq ` . That will yield a Vector1 
```
  val viaIterator: IndexedSeq[String] =
    Iterator("a", "b").map(identity).toIndexedSeq
  println(viaIterator.getClass.getName) //  scala.collection.immutable.Vector1
  ```
I think more than likely an arrayseq is being created somewhere else and getting passed through

Baccata:
Good point, I've amended the hallucinated comment to say "niche case" whilst I try and understand the actual codepath leading to the issue. But in the meantime, the fix seems sound and does work.

Baccata:
Yeah it's as I thought : the issue specifically happens when proguard is involved (lambdas in particular).

modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala
lewisjkl: Is `unsafeArray` expensive at all? If so worth making this change, otherwise yeah doesn't really matter either way

Baccata: Nope, there's no computation it's just a field access : https://www.scala-lang.org/api/3.x/scala/collection/immutable/ArraySeq.

lewisjkl (approved):

PR #1987, opened 2026-07-16
Codegen 'eviction'

lewisjkl:
This is a fix for the issue that recently popped up in CI (without us even bumping any versions) due to the dynamo pom.xml file specifying a range and a new smithy-build version getting pulled transitively.

This change is effectively putting an `eviction` in place so that the child class loader does not duplicate what the parent class loader has already loaded.

See changelog entry for more detail.

---

Typical disclaimer: I used AI in the creation of this PR. AI helped to identify the root cause of the issue, @yisraelU helped to come up with the idea of "eviction" as a solution from there, and I had the idea of using `BuildInfo` to filter the class paths this way. From there it was a slam dunk for Claude to implement the fix as described. There you go.. everyone gets credit, even our good friend, Claude.

- [ ] Added documentation
- [x] Updated changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ When adding entries, please treat them as if they could end up in a release any 
 
 Thank you!
 
+# 0.19.11
+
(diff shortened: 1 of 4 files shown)


kubukoz (approved):
Pretty clever!

a-morales (approved):

yisraelU (approved):

PR #1991, opened 2026-07-28
Dynamic validation

msosnicki:

- [x] Added documentation
- [x] Updated changelog


modules/dynamic/src-jvm/DynamicSchemaValidation.scala
```diff
@@ -0,0 +1,148 @@
...
+              )
+            case PBlob =>
+              schema.reifyHint(RefinementProvider.blobLengthConstraint)
+            case PTimestamp | PDocument | PBoolean | PUUID | PLocalDate |
+                PLocalTime | PDuration | POffsetDateTime =>
+              schema
+          }
+        case e: EnumerationSchema[?] =>
+          enumSchema(e.asInstanceOf[EnumerationSchema[Enum[?]]])
+            .asInstanceOf[Schema[A]]
```
kubukoz: Casts are pretty risky in Dynamic land. Any way we can make this work without them?

modules/dynamic/src-jvm/DynamicSchemaValidation.scala
kubukoz, line 1: I don't think we need to limit ourselves to JVM here, this could work on all platforms.

modules/dynamic/src-jvm/DynamicSchemaIndexCompanionPlatform.scala
```diff
@@ -27,6 +27,20 @@ private[dynamic] trait DynamicSchemaIndexCompanionPlatform {
...
+  /**
+    * Loads a dynamic schema index model from a smithy model.
+    *
+    * @param performValidation when true, constraint traits (`@length`, `@range`,
+    * `@pattern` etc) are reified into Schema objects that get enforced upon
+    * decoding, instead of being kept as inert hints (which is the default dynamic
+    * behaviour), mirroring the fact that codegen-produced schemas enforce these
+    * constraints but dynamically-loaded do not).
+    */
+  def loadModel(
```
kubukoz, line 41: We also have a platform-agnostic variant that works on our own model of `Model` - let's add `applySchemaRefinements` to it too

modules/dynamic/src-jvm/DynamicSchemaIndexCompanionPlatform.scala
```diff
@@ -27,6 +27,20 @@ private[dynamic] trait DynamicSchemaIndexCompanionPlatform {
     */
   def loadModel(
       model: software.amazon.smithy.model.Model
+  ): DynamicSchemaIndex = loadModel(model, performValidation = false)
+
+  /**
+    * Loads a dynamic schema index model from a smithy model.
+    *
+    * @param performValidation when true, constraint traits (`@length`, `@range`,
```
kubukoz: My first thought seeing this name was that we'd be doing model validations rather than schema validations...

maybe `addSchemaValidators` or `applySchemaRefinements` would fit better?

modules/dynamic/src-jvm/DynamicSchemaValidation.scala
```diff
@@ -0,0 +1,148 @@
...
+  *
+  * This object provides a transformation for dynamically-loaded schemas, so that validation 
+  * hints are reintroduced at a `Schema` level.
+  */
+private[dynamic] object DynamicSchemaValidation {
+
+  def reifyConstraints(index: DynamicSchemaIndex): DynamicSchemaIndex =
+    new DynamicSchemaIndex {
+      def allServices: Iterable[DynamicSchemaIndex.ServiceWrapper] =
+        index.allServices
```
kubukoz, line 44: 1. I believe this isn't right: operation inputs/outputs/errors won't have their traits reified. UNLESS the default DSI implementation calls `getSchema(shapeId)` - in which case they would be reified - but we can't assume that about the `index` instance here, as it's an implementation detail and not something enforced by a contract.

To be safe, we could convert those services to builders and map each endpoint's input/output/errors, reifying the constraints right there. I think a simpler alternative to this would be embedding the reification process directly in the DynamicCompiler rather than in a "middleware" (here)

modules/dynamic/src-jvm/DynamicSchemaValidation.scala
```diff
@@ -0,0 +1,148 @@
...
+      def allSchemas: Iterable[Schema[_]] =
+        index.allSchemas.map(reifySchema(_))
+      def getSchema(shapeId: ShapeId): Option[Schema[_]] =
+        index.getSchema(shapeId).map(reifySchema(_))
```
kubukoz: I don't like that we map the schemas on read. Not only is it per-call overhead, but it can also produce different instances of schemas when you call it multiple times, which can break things like schema compilation caches.

modules/dynamic/src-jvm/DynamicSchemaValidation.scala
```diff
@@ -0,0 +1,148 @@
...
+      def metadata: Map[String, Document] = index.metadata
+    }
+
+  private def reifySchema[A](schema: Schema[A]): Schema[A] =
+    schema.transformTransitivelyK(ReifyConstraints)
+
+  private object ReifyConstraints extends (Schema ~> Schema) {
+
+    private def void[C, A](
```
kubukoz: I think at this point we should move this to `RefinementProvider`. It was always a bit of a hack, having it as a more first-class thing would feel better

modules/dynamic/src-jvm/DynamicSchemaValidation.scala
```diff
@@ -0,0 +1,148 @@
...
+
+    private implicit class SchemaOps[A](schema: Schema[A]) {
+      def reifyHint[B](rp: RefinementProvider[B, A, ?]): Schema[A] =
+        schema.hints.get(rp.tag).fold(schema)(schema.validated(_)(void(rp)))
+    }
+
+    private def collection[C[_], B](
+        schema: Schema.CollectionSchema[C, B]
+    ): Schema[C[B]] =
+      schema.tag match {
```
kubukoz: AFAIK that's not an exhaustive match (`CollectionTag` is open in 0.19). We should have another case for arbitrary tags.

modules/dynamic/src-jvm/DynamicSchemaValidation.scala
```diff
@@ -0,0 +1,148 @@
...
+        .reifyHint(RefinementProvider.lengthConstraint[B](_.toString.length))
+        .reifyHint(RefinementProvider.rangeConstraint[B, Int](_.ordinal()))
+        .reifyHint(RefinementProvider.patternConstraint[B](e => e.toString))
+
+    def apply[A](schema: Schema[A]): Schema[A] =
+      schema match {
+        case t @ PrimitiveSchema(_, _, tag) =>
+          tag match {
+            case PString =>
+              t.reifyHint(RefinementProvider.stringLengthConstraint)
```
kubukoz, line 93: also: `@idRef`

modules/dynamic/src-jvm/DynamicSchemaValidation.scala
```diff
@@ -0,0 +1,148 @@
...
+          }
+        case e: EnumerationSchema[?] =>
+          enumSchema(e.asInstanceOf[EnumerationSchema[Enum[?]]])
+            .asInstanceOf[Schema[A]]
+        case c @ CollectionSchema(_, _, _, _) => collection(c)
+        case m: MapSchema[c, k, v] =>
+          m.reifyHint(
+            RefinementProvider.lengthConstraint[c[k, v]](m.tag.iterator(_).size)
+          )
+        case other => other
```
kubukoz: I'd handle the remainder explicitly, [like in smithy-playground](https://github.com/kubukoz/smithy-playground/blob/a0e19e561b2da513a2005bb5165a82f06940beeb/modules/core/src/main/scala/playground/smithyutil/AddDynamicRefinements.scala#L96C6-L96C120)

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Thank you!
 # 0.19.11
 
 - codegen: Fix an `IllegalAccessError` (e.g. `class ...IncludeClosures cannot access its abstract superclass ...BackwardCompatHelper`) that could occur during `smithy4sCodegen` when a project dependency pulled a different version of a Smithy library (`smithy-build`, `smithy-model`, etc.) than the one bundled with the codegen plugin. The model-loading `URLClassLoader` used the plugin classloader as its parent, so a duplicate copy of a plugin-provided module on the child loader could split a package across two classloaders and break package-private access. The codegen now drops any dependency already provided by the parent classloader (tracked via the new `BuildInfo.codegenDependencies`) from the child classloader, so Smithy versions no longer need to be aligned between the plugin and the project's dependencies.
+- dynamic: Add `DynamicSchemaIndex.loadModel(model, performValidation: Boolean)` overload. When `performValidation = true`, constraint traits (`@length`, `@range`, `@pattern`, ...) are reified into schema refinements instead of being kept just as hints, matching the behaviour of codegen-produced schemas.
```
kubukoz: .11 is already out, we need a new section :)

msosnicki:
@kubukoz thanks for review, I think all your comments are now addressed

modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
```diff
@@ -87,6 +87,17 @@ private[dynamic] object Compiler {
      */
   protected[dynamic] def compile(
       model: Model
+  ): DynamicSchemaIndex = compile(model, applySchemaRefinements = false)
+
+  /**
+     * @param knownHints hints supported by the caller.
```
kubukoz: knownHints is long gone

kubukoz: do we need an overload if this is `protected[dynamic]`? I don't think it's callable by users

kubukoz: if mima flags this, we should silence it. Maybe just silence the whole `smithy4s.dynamic.internals` package.

kubukoz (commented):
pls resolve the conflicts, also got a few small comments and we can get this merged

modules/dynamic/src/smithy4s/dynamic/internals/DynamicModelCompiler.scala
msosnicki: Done, all green

kubukoz (approved):

PR #1998, opened 2026-08-21
Fix greedy label matching when followed by static segment

### LINKED ISSUE #1997: Greedy label followed by static segment never matches
A route pattern with a greedy label followed by a static segment, like `/base/{foo+}/end`, never matches any request and always returns a 404. In `matchPath` the `GreedySegment :: StaticSegment :: _` branch requires `greedyAcc.nonEmpty`, but `greedyAcc` is only populated in the branch that requires the greedy segment to be last, so the guard can never hold. The Smithy spec allows literal segments after a greedy label.

stasimus:
## Summary
- Routes like `/base/{foo+}/end` never matched, always 404ing
- `greedyAcc` was only ever populated when the greedy label was the last segment, so the `GreedySegment :: StaticSegment` branch's guard could never be satisfied
- Rewrote greedy matching to split on the fixed-length suffix after the greedy segment

Fixes #1997

- [ ] Added bootstrapped code + smoke tests (when the rendering logic is modified) — not applicable, no rendering logic touched
- [ ] Added build-plugins integration tests (when reflection loading is required at codegen-time) — not applicable
- [ ] Added alloy compliance tests (when simpleRestJson protocol behaviour is expanded/updated) — this is a bug fix to existing path-matching behavior, no alloy-side test added
- [ ] Updated dynamic module to match generated-code behaviour — not applicable, `matchPath` is shared runtime code used by both codegen and the dynamic module
- [ ] Added documentation — not applicable, internal routing fix
- [x] Updated changelog

### DIFF
--- a/modules/bootstrapped/test/src/smithy4s/http/MatchPathSpec.scala
+++ b/modules/bootstrapped/test/src/smithy4s/http/MatchPathSpec.scala
@@ -87,6 +87,36 @@ class MatchPathSpec() extends munit.FunSuite with munit.ScalaCheckSuite {
     expect.eql(result, Some(expected))
   }
 
+  test("Allows for greedy labels followed by a static segment") {
+    // /base/{foo*}/end
+    val path: List[PathSegment] =
+      List(
+        PathSegment.static("base"),
+        PathSegment.greedy("foo"),
+        PathSegment.static("end")
+      )
+    val expected = Map("foo" -> "a/b")
+    val result = doMatch(path)("base", "a", "b", "end")
+
+    expect.eql(result, Some(expected))
+  }
+
+  test(
+    "Greedy labels capture segments equal to the trailing static segment"
+  ) {
+    // /base/{foo*}/end
+    val path: List[PathSegment] =
+      List(
+        PathSegment.static("base"),
+        PathSegment.greedy("foo"),
+        PathSegment.static("end")
+      )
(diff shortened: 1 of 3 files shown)


lewisjkl (approved):
Looks good, no concerns from my end. Thanks for submitting this!

@kubukoz any concerns on your side?

modules/bootstrapped/test/src/smithy4s/http/MatchPathSpec.scala
```diff
@@ -87,6 +87,36 @@ class MatchPathSpec() extends munit.FunSuite with munit.ScalaCheckSuite {
     expect.eql(result, Some(expected))
   }
 
+  test("Allows for greedy labels followed by a static segment") {
```
kubukoz, line 90: Can we have a case where the request doesn't match the final static segment? e.g. call `base/a/b/not-end`.

kubukoz:
> Looks good, no concerns from my end. Thanks for submitting this!
> 
> @kubukoz any concerns on your side?

just one test case I'd like to see, other than that lgtm

modules/bootstrapped/test/src/smithy4s/http/MatchPathSpec.scala
stasimus: Added, thanks.

PR #2004, opened 2026-09-05
Compare range constraints exactly instead of via Double

### LINKED ISSUE #2003: Range validation loses precision by going through Double
`RangeConstraint` in `RefinementProvider` converts every input through `Numeric.toDouble` and then rebuilds a `BigDecimal` from that double before comparing it to the `@range` bounds. Anything that a double can't represent exactly gets compared against a rounded value, so the check gives the wrong answer for large longs, big integers and precise decimals.

Observed on `series/0.19` (fefe0414f) with `RefinementProvider.numericRangeConstraints[N].make(range)`:

| Constraint and input | Expected | Actual |
| --- | --- | --- |
| `Long`, max `9007199254740992`, input `9007199254740993` | Reject | Accept |
| `Long`, min `9007199254740993`, input `9007199254740993` | Accept | Reject, `Input must be >= 9007199254740993, but was 9007199254740992` |
| `Long`, min `Long.MinValue`, max `Long.MaxValue`, input `Long.MaxValue` | Accept | Reject, `but was 9.223372036854776E+18` |
| `BigInt`, min `0`, input `10^400` | Accept | Reject, `Numeric values must not be NaN or pos/neg infinity. Found Infinity` |
| `BigDecimal`, max `0.1`, input `0.10000000000000000000000000000000001` | Reject | Accept |

The `Long.MaxValue` case is probably the most likely to bite in practice: a `@range(min: ..., max: 9223372036854775807)` bound on a long lets the max value itself through on the model side but the generated validation rejects it. The `BigInt` case means any `@range` on a `BigInteger` shape rejects values above ~1.8e308 with a misleading NaN/infinity error.

Values that are already exact (`Int`, `Long`, `BigInt`, `BigDecimal`) should be converted to `BigDecimal` directly, with the double path only used for `Float`/`Double`.

stasimus:
## Summary
- `RangeConstraint` converted every value with `Numeric.toDouble` and rebuilt a `BigDecimal` from that, so anything a double can't hold exactly was compared against a rounded value: `Long` values above 2^53 slipped past or were wrongly rejected, `Long.MaxValue` was rejected against a `max: 9223372036854775807` bound, and any `BigInt` above ~1.8e308 failed with a NaN/infinity error
- The conversion is now picked once per constraint from the `Numeric` instance: `BigDecimal` and `BigInt` are used as is, integrals go through `toLong`, floats and doubles keep the old `toDouble` path with the NaN/infinity check
- Side effect: integral inputs no longer print a trailing `.0` in validation messages (`but was 11` instead of `but was 11.0`), JVM test expectations updated
- New `RangeConstraintSpec` covers the cases from the issue

Related codegen side (bounds rendered through `doubleValue()`) is in #2002.

Fixes #2003

- [ ] Added bootstrapped code + smoke tests (when the rendering logic is modified) - not applicable, runtime only
- [ ] Added build-plugins integration tests (when reflection loading is required at codegen-time) - not applicable
- [ ] Added alloy compliance tests (when simpleRestJson protocol behaviour is expanded/updated) - not applicable
- [ ] Updated dynamic module to match generated-code behaviour - not applicable, the dynamic module uses the same `RefinementProvider`
- [ ] Added documentation - not applicable
- [x] Updated changelog


kubukoz (approved):

kubukoz:
can you resolve the conflicts?

stasimus:
Merged, these were tests and CHANGELOG.md.

kubukoz (approved):

